PR #1, opened 2015-01-08
Set exitCode instead of calling exit() directly in VM configurations.

cbracken:
This avoids side-effects of terminating the VM, such as preventing coverage collection and other interaction with the VM observatory.


cbracken:
In testing against a few dozen packages, this turned up a fair number of hangs. In several cases, this revealed resource leaks in the code under test, in others, failed cleanup in the tests themselves. Unfortunately debugging these issues is pretty painful today. While I still thing this is probably the right change to make, until that changes, I'd rather hold off on this.

PR #3, opened 2015-02-05
Support document.readyState value: "complete"

nicolasgarnier:
Some browsers (like content_shell) don't use readyState = "loaded" but readyState = "complete" instead.

This causes Content_shell to hang if no tests are being ran because notifyDone() is never called in that case.


nex3:
:+1:, I'll merge this.

PR #58, opened 2015-04-08
Travis

kevmoo:


devoncarew:
On Travis, Chrome needs to be started with `--no-sandbox`. This may be why the tests are failing here...

https://github.com/google/webdriver.dart/blob/master/.travis.yml#L7

nex3:
As I've mentioned on other packages, I'm not a fan of moving packages piecemeal over to Travis while the official bots are still running. It supports fewer operating systems and fewer browsers than our bots, and the pull request benefits are negligible since we rarely get major pull requests.

Speaking of pull requests, please submit changes via Rietveld CLs.

kevmoo:
@nex3 I'm just using this to play on Travis – I can make that more clear in the title.

Sorry for the noise. No desire to land this CL as is.

Issue #61, opened 2015-04-09
Groups with spaces in their name always pass their tests


alan-knight:
OK, a few caveats - in the Dart repo, using test.py, and with useHtmlIndividualConfiguration. I suspect the fault is at least partially in test.py, but it might be easier to deal with in the test package. For an example, in a Dart repo checkout, add a file foo_test.dart, like

      import 'dart:html';
      import 'package:unittest/unittest.dart';
      import 'package:unittest/html_individual_config.dart';

      main() {
        useHtmlIndividualConfiguration();

        group('You have got to be kidding me', () {

          test("failure", () {
            expect(false, isTrue);
          });
        });

        group('it_works_if_you_have_no_spaces_in_the_group_name', () {

          test("also_failure", () {
            expect(false, isTrue);
           });
        });
      }

Run this from the top-level directory with a command-line like 

      tools/test.py -m release -c dart2js -r chrome -j 1 html/foo_test

I get one pass and one failure.

alan-knight:
The file goes in dart/tests/html/foo_test.dart

alan-knight:
Repo bug is at https://code.google.com/p/dart/issues/detail?id=23156

There's something more to it, because there are other files that have these characteristics that don't suffer from this problem, but I don't at this point know what the difference is.

nex3:
What output are you seeing? Where does it say it passes, in the `unittest` output or the test runner output? If you can't tell, just paste it here and I'll take a look.

alan-knight:
https://gist.github.com/alan-knight/31a2e15c086cef0bb271

zoechi:
Isn't this mentioned somewhere that spaces in test/group name isn't supported for html tests. I stumbled upon this when I worked on a testrunner last year but don't remember where.

whesse:
The unittest framework is not supporting spaces in the group name that is passed in in the url.  Both the urls
http://127.0.0.1:36966/root_build/generated_tests/dart2js-chrome/tests_html_foo_test/test.html?crossOriginPort=58705&group=You%20have%20got%20to%20be%20kidding%20me
and
http://127.0.0.1:36966/root_build/generated_tests/dart2js-chrome/tests_html_foo_test/test.html?crossOriginPort=58705&group=You+have+got+to+be+kidding+me

report "Pass" in the browser, in the actual web page displayed.  The latter, with + signs, is what test.py is using to run the test.

whesse:
[...]

PR #62, opened 2015-04-10
Enable group names containing spaces in HtmlIndividualConfiguration

whesse:
Fixes dart issue http://dartbug.com/23156 and test issue https://github.com/dart-lang/test/issues/61

### DIFF
--- a/lib/html_individual_config.dart
+++ b/lib/html_individual_config.dart
@@ -34,6 +34,8 @@ class HtmlIndividualConfiguration extends htmlconfig.HtmlConfiguration {
         }
 
         var testGroupName = groups.single.split('=')[1];
+        testGroupName =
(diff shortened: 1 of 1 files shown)


lib/html_individual_config.dart
```diff
@@ -34,6 +34,8 @@ class HtmlIndividualConfiguration extends htmlconfig.HtmlConfiguration {
         }
 
         var testGroupName = groups.single.split('=')[1];
+        testGroupName =
+            testGroupName.replaceAll('+', ' ').replaceAll('%20', ' ');
```
nex3: It seems like it would be much more robust to use `Uri.parse` and then accessing this via `Uri.queryParameters.`

kevmoo:
Could we add a test case for this, too?

nex3:
Closing this out, since the `0.11.x` branch is no longer supported.

PR #97, opened 2015-04-30
Add support for a WebDriver-based test runner

DrMarcII:


nex3:
I don't want to bake WebDriver support in to the core test package. It should really be a plugin that users can include or not as they desire.

PR #108, opened 2015-05-07
stack_trace 1.2.0 isn't compatible with test.

zoechi:
I tried `pub downgrade` and I got the error

 `Class 'LazyTrace' has no instance method 'foldFrames' with matching arguments.`

 After limiting stack_trace to '^1.2.1' it worked again.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -17,7 +17,7 @@ dependencies:
   shelf_static: '^0.2.0'
   shelf_web_socket: '^0.0.1'
   source_span: '^1.0.0'
-  stack_trace: '^1.2.0'
+  stack_trace: '^1.2.1'
   string_scanner: '^0.1.1'
   yaml: '>=0.9.0 <3.0.0'
 


nex3:
Did you mean to attach a different email to this commit?

zoechi:
This is a common issue with my account. I guess I renamed it once several years ago when I started using GitHub and GitHub states in the FAQs or somewhere that this doesn't work well in all cases but I didn't know about it back then. My previous contributions to Dart repos were accepted anyway.

nex3:
I believe it's you, I just didn't know how much you cared about making the commit sync up with your account.

pubspec.yaml
```diff
@@ -17,7 +17,7 @@ dependencies:
   shelf_static: '^0.2.0'
   shelf_web_socket: '^0.0.1'
   source_span: '^1.0.0'
-  stack_trace: '^1.2.0'
+  stack_trace: '^1.2.1'
```
nex3: Generally I don't worry that much about bugs that are fixed in patch versions; `1.2.0` isn't so much incompatible as it is completely broken. But it is nice to have `pub downgrade` work, so sure.

PR #116, opened 2015-05-17
Fix typo in sample YAML

eee-c:
The label should be `transformers`, not `transformer`.


eee-c:
I signed it (updated the existing one).

PR #270, opened 2015-06-10
Add BufferedStringSink to wrap print().

DrMarcII:
Change reporters to take a StrinkSink as a parameter.
Remove dep on 'dart:io' in compact.dart.


DrMarcII:
The goal of this is to make the interaction of the reporters with the surrounding environment more consistent by having StringSink's (such as stdout) injected into the reporters.

nex3:
Can you give me some context behind this? In general, the test runner stays away from dependency injection by design.

PR #277, opened 2015-06-18
Windows platform: Fix incorrect path separator in pub url.

DisDis:


DisDis:
I signed it!

nex3:
This is a good catch, but it's not quite the right way to fix it. Just replacing backslashes with forward slashes won't always create a valid URL that means the same thing as a path; you need extra logic (available in `p.toUri`) for that. I'll make another patch that fixes it.

PR #339, opened 2015-09-20
When using --pub-serve, only compile tests when browser.isJS is true

greglittlefield-wf:
### Ultimate problem
When the  `--pub-serve` option is used, JavaScript source maps are fetched for test suites, which causes the test files to be compiled via dart2js.

These source maps are fetched unnecessarily when run in Dartium and content shell, which causes  unnecessary dart2js compilation.

This compilation can take a while, and makes running browser tests in Dartium and content shell while using `--pub-serve` much more sluggish.

### Changes
* Only fetch the source maps for test suites if `browser.isJs` is true.
* Add test case with content shell that verifies that test suite load errors are still handled gracefully when `browser.isJs == false`.

### DIFF
--- a/lib/src/runner/browser/server.dart
+++ b/lib/src/runner/browser/server.dart
@@ -230,21 +230,21 @@ void main() {
       var suitePrefix = p.withoutExtension(
           p.relative(path, from: p.join(_root, 'test')));
 
-      var jsUrl;
+      var dartUrl;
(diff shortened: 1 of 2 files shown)


greglittlefield-wf:
I signed it! (My company, Workiva, signed it.)

nex3:
Thanks for the patch!

PR #342, opened 2015-09-25
new flag "fail-fast" - testing stops after the first error

d2m:
2 tests added:

```pub run test:test -N failFast test/runner/expanded_reporter_test.dart``` and
```pub run test:test -N failFast test/runner/compact_reporter_test.dart```

This is a first try - the idea is to cancel all subscriptions when an error is detected in the reporter. I am sure there are problems with async tests etc. What do you think? 

New command line option:
```pub run test:test -f```
should run all tests until the first error.

### DIFF
--- a/lib/src/runner.dart
+++ b/lib/src/runner.dart
@@ -63,7 +63,8 @@ class Runner {
         verboseTrace: config.verboseTrace,
         printPath: config.paths.length > 1 ||
             new Directory(config.paths.single).existsSync(),
-        printPlatform: config.platforms.length > 1);
+        printPlatform: config.platforms.length > 1,
+        failFast: config.failFast);
 
     return new Runner._(config, loader, engine, reporter);
   }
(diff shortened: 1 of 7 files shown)


nex3:
I'm worried that this flag will behave in inconsistent ways when multiple test suites are being run in parallel. The user may have an expectation that running with `--fail-fast` multiple times will fail on the same test each time, but there's no guarantee that will happen.

Even if we're okay with that, there's also a question of what the UI is for other tests that are running in parallel. We can't just exit the test runner immediately when the first failure happens—we have to give parallel tests a chance to run their tear-down logic and clean up. How do we surface that to the user?

d2m:
We could disable parallel test execution with ```--fail-fast```. I do that currently by passing only a single test file or create a file that imports and runs different test files. Canceling subscriptions should trigger tear-down logic and clean up.

nex3:
> We could disable parallel test execution with ```--fail-fast```.

It seems like it would be very confusing to users that telling their tests to fail quickly would actually make them run *slower*.

> Canceling subscriptions should trigger tear-down logic and clean up.

The issue with tear downs isn't the implementation, it's the UI.

d2m:
Looks like we have different use cases.

PR #350, opened 2015-10-12
[WIP] Tags (fixes #16)

### LINKED ISSUE #16: Support for tags
During development I won't care about a lot of tests, especially integration tests (which are often slow(er)). By using tags I could declare sets of unit tests I care about and are likely to break because of what I'm working on.
This would make it more convenient to run tests frequently.
Running all tests is more appropriate for checkin hooks and CI.

This feature could be combined with #6, #7, #14 

Examples:
- run only server tests (tag: `server`)
- run only fast server tests (tags: `server,unittest`) (as don't run `integration` tests)
- run all firefox tests (tags: `firefox`)
- could be combined with #14 (tags: `-skip`) (exclude `skip`)
- run tests for a specific feature (tags: `authentication`)

It should be supported to add tags at group and test level, where tags on a group apply also to tests within that group.

I just found an example where this is already done http://olivinelabs.com/busted/.

yjbanov:
This is not yet meant for merging. Just checking if it makes sense. Notable AIs:

- add tests
- switch from `List` to `Set`

### DIFF
--- a/lib/src/backend/metadata.dart
+++ b/lib/src/backend/metadata.dart
@@ -33,6 +33,9 @@ class Metadata {
   /// The reason the test or suite should be skipped, if given.
   final String skipReason;
 
+  /// A set of tags attached to a test
+  final List<String> tags;
+
   /// Platform-specific metadata.
   ///
   /// Each key identifies a platform, and its value identifies the specific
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -13,3 +13,6 @@ packages
 # Include when developing application packages.
 pubspec.lock
 .packages
+
+# Don't commit IntelliJ project directories
+.idea
```
kevmoo: Do a global ignore for these, yo

yjbanov: Done. How do you choose what should be ignored globally and what should be ignored locally?

kevmoo: Ignore things that are part of the core tool set (like files created by pub).

Not files specific to your tool set (like your IDE)

nex3:
Could we move this to [Rietveld](http://codereview.chromium.org)? I really dislike doing large reviews on GitHub.

yjbanov:
Done: https://codereview.chromium.org/1405633004

lib/src/backend/metadata.dart
```diff
@@ -88,27 +91,30 @@ class Metadata {
...
-          Map<PlatformSelector, Metadata> onPlatform})
+          Map<PlatformSelector, Metadata> onPlatform, List<String> tags})
       : testOn = testOn == null ? PlatformSelector.all : testOn,
         timeout = timeout == null ? const Timeout.factor(1) : timeout,
         skip = skip,
         onPlatform = onPlatform == null
             ? const {}
-            : new UnmodifiableMapView(onPlatform);
+            : new UnmodifiableMapView(onPlatform),
+        this.tags = tags ?? const <String>[];
```
zoechi: This limits the use with older Dart versions. 
I don't care myself, I just wondered if this is intentional

yjbanov: Yep. Will fix. BTW, the review moved to https://codereview.chromium.org/1405633004/

PR #363, opened 2015-11-22
Update sdk declaration in pubspec.yaml

jirkadanek:
This solves the problem that "<1.14.0" does not include the current dev release 1.14.0-dev.1.0

With the dev releases of 1.14 already out, this package should declare it is compatible with it.


kevmoo:
@nex3 has a patch out for this – just need to land it

https://codereview.chromium.org/1463023002/

Issue #372, opened 2015-12-09
Allow any object as message for test() and group()


munificent:
It would be useful to let you pass any object as the "message" for these and have test implicitly call `toString()` on them. This is consistent with `print()` and `StringBuffer.write()`. It's particularly useful for `group()` since it would let you do:

```dart
group(MyAwesomeClass, () {
  ...
});
```

And then have `MyAwesomeClass` be a validated identifier, allow go-to-definition, etc.

PR #379, opened 2016-01-06
Allow object to be passed in as a group / test description

jayudey-wf:
There was an issue created, https://github.com/dart-lang/test/issues/372, that requested the ability to pass in an object as the description for the group and test methods.  

```
group(MyAwesomeClass, () {
  ...
});
```

This change allows a user to do that and is accompanied with unit tests exercising the intended change.

### DIFF
--- a/lib/test.dart
+++ b/lib/test.dart
@@ -78,7 +78,8 @@ Declarer get _declarer {
 
 // TODO(nweiz): This and other top-level functions should throw exceptions if
 // they're called after the declarer has finished declaring.
-/// Creates a new test case with the given description and body.
+/// Creates a new test case with the given description (converted to a string)
+/// and body.
 ///
 /// The description will be added to the descriptions of any surrounding
 /// [group]s. If [testOn] is passed, it's parsed as a [platform selector][]; the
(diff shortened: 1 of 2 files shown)


jayudey-wf:
I believe Workiva already has an agreement on file.

@googlebot

lib/test.dart
```diff
@@ -171,10 +175,18 @@ void test(String description, body(), {String testOn, Timeout timeout, skip,
...
-        tags, Map<String, dynamic> onPlatform}) =>
-    _declarer.group(description, body,
-        testOn: testOn, timeout: timeout, skip: skip, tags: tags);
+void group(description, body(),
+    {String testOn,
+    Timeout timeout,
+    skip,
+    tags,
+    Map<String, dynamic> onPlatform}) =>
+  description == null
```
nex3: I don't want to allow description-less groups other than the implicit root group. If someone has a reason to group `setUp` behavior or whatever, they can be bothered to explain it in text.

lib/test.dart
```diff
@@ -118,9 +118,13 @@ Declarer get _declarer {
 ///
```
nex3: Earlier in this comment, change "the given description" to "the given description (converted to a string)". Same with the comment for `group()`.

nex3:
Done reviewing! Just a couple small changes and I'll merge it in.

jayudey-wf:
@nex3 comments addressed

nex3:
Thanks for the patch!

kevmoo:
:metal:

jayudey-wf:
thanks for letting me contribute!

Issue #12, opened 2015-02-19
Support pluggable reporters


nex3:
There should be a means for third parties to create reporters that are dynamically loaded and used by unittest.

This is particularly tricky because reporters are likely to be desired on a user-by-user rather than project-by-project basis, so there's some friction to just putting it into a project's pubspec.

Fox32:
I think for the first implementation it would be sufficient if I could specify a reporter via command line. 

The main use case I consider is using the unittest command without a special reporter locally while developing, but another configuration on the continuous integration server. There I might need the report in the JUnit format (like [this](https://github.com/renggli/dart-junit)) or one of the many other formats.

nex3:
I agree; we should definitely also support setting reporters from the command line.

nex3:
This is blocked on #48.

jayudey-wf:
this is needed in order to utilize saucelabs unit testing capabilities, https://docs.saucelabs.com/reference/rest-api/?_ga=1.245459780.1583316598.1430344943#start-js-unit-tests

nex3:
A reporter plugin isn't a good fit for Sauce Labs. A reporter just takes the results of tests being run and displays them to the user, it doesn't run the test itself. #49 or #99 might be a better fit.

butlermatt:
I'd love this functionality, along with #48. I had a previous project using unittest which I wrote my own custom configuration to create Dart Koans (See: https://github.com/butlermatt/dart_koans/blob/master/lib/src/config_koans.dart )

I tried to revive this project this evening with Test but it doesn't seem that same configurability is available with test.

PR #381, opened 2016-01-11
Adding reporter to create xunit test-reports

jayudey-wf:
Internally we use several services that expect xunit formatted test reports.  Currently this isn't available in the `Test` package.  

This change adds a reporter that produces xunit compatible output as well as testing to verify that the output created by the reporter is being properly formatted.

xunit standard used - https://svn.jenkins-ci.org/trunk/hudson/dtkit/dtkit-format/dtkit-junit-model/src/main/resources/com/thalesgroup/dtkit/junit/model/xsd/junit-7.xsd

### DIFF
--- a/lib/src/runner.dart
+++ b/lib/src/runner.dart
@@ -24,6 +24,7 @@ import 'runner/reporter.dart';
 import 'runner/reporter/compact.dart';
 import 'runner/reporter/expanded.dart';
 import 'runner/reporter/json.dart';
+import 'runner/reporter/xunit.dart';
 import 'runner/runner_suite.dart';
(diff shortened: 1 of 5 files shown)


jayudey-wf:
I signed it!

nex3:
It's really cool that you did this! But... I don't think we can take it in to the core test package :disappointed:. We really try hard to keep the set of built-in reporters small and very broadly useful.

This is a great thing to put in a reporter plugin. We don't have those implemented yet (#12), unfortunately, but they will be usable before too long and you'll be able to use pretty much all the code you wrote here for them.

PR #401, opened 2016-03-03
Use web_socket_channel rather than http_parser.

nex3:


lib/src/runner/browser/browser_manager.dart
```diff
@@ -6,7 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:async/async.dart';
-import 'package:http_parser/http_parser.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
```
kevmoo: sorting

test/runner/browser/code_server.dart
```diff
@@ -4,7 +4,7 @@
 
 import 'dart:async';
 
-import 'package:http_parser/http_parser.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
```
kevmoo: sorting

kevmoo:
some sorting, otherwise LGTM

PR #408, opened 2016-03-14
Default to using the system packageRoot.

mgiambalvo:
Currently, the default is to use 'pacakges/' in the cwd as the package
root. That means if you set a different packageRoot as a flag to the
dart vm, you also need to set it as a flag to package:test. To simplify
things, package:test can just use whatever the Platform packageRoot is
by default, so you only need to set it once.

### DIFF
--- a/lib/src/runner/configuration.dart
+++ b/lib/src/runner/configuration.dart
@@ -66,7 +66,7 @@ class Configuration {
   final bool _pauseAfterLoad;
 
   /// The package root for resolving "package:" URLs.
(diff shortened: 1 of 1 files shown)


nex3:
I don't think I can accept this as-is. `Platform.packageRoot` is not a reliable field; it's just the literal text of the `--package-root` flag that was passed to the original process, which means that it won't work if automatic package resolution is used or if the executable is spawned in an isolate with different package-resolution semantics than the root.

Can you give me some context on why you're making this change? In general, the `test` executable doesn't officially support being run in any way other than through `pub run test`.

PR #411, opened 2016-03-24
Fix all strong mode warnings in 0.11.x.

nex3:
I can't figure out how to get Rietveld to base this on the right branch, so here's a PR instead.


kevmoo:
LGTM

PR #419, opened 2016-04-08
Fix additional strong-mode warnings.

nex3:
Looks like we hadn't run the analyzer against all the top-level
libraries.


kevmoo:
LGTM

PR #446, opened 2016-08-01
Add solo_test, solo_group API

### LINKED ISSUE #127: Add ability to run solo tests
something like test(solo:true) even better a Test annotation that can be selected from the dart editor to select that only a single test should run

srawlins:
Fixes #127

This is a fairly simple implementation of `solo_test` and `solo_group`, restricted to the "backend" code (`lib/src/backend/declarer.dart` and `lib/src/backend/metadata.dart`).
#### WIP

This PR is functional, but I'm throwing it out there as a "request for comments." What do the authors think?

TODO:
- [ ] Write up some text in README.
- [ ] Write up some doc comments for `solo_test` and `solo_group`. These should be lengthy, with explanations of inheritance, and what to expect in output.
#### Notes
- The real API change is to add a `solo:` named arg to `test` and `group`, but the backward compatibility, and real beauty, comes in implementing the `package:unittest` methods: `solo_test` and `solo_group`. Those five characters, `solo_` are what people love so much about unittest's feature.
- The feature works by declaring all non-solo tests or groups as "skipped". So there is no new status in the reporting. For example:
  
  ``` text
  $ cat a.dart
  import 'lib/test.dart';
  
  void main() {
    test('foo', () { expect('a', equals('a')); });
    test('bar', () { expect('a', equals('a')); });
    solo_test('baz', () { expect('a', equals('a')); });
  }
  $ dart a.dart
  00:00 +0: foo
  00:00 +0 ~1: bar
  00:00 +0 ~2: baz
  00:00 +1 ~2: All tests passed!
  ```
  
  We see that it looks like one test ran, and two were skipped. This is the analogy of "solo".
- I don't think this feature needs a new status (something like "00:00 !1 ~2: All tests passed!"), but we could if we wanted.
- We could change the "All tests passed!" text to be different if any tests were skipped, but that would be a separate feature, unrelated to this PR.
- I did not change the return code if a solo test/group is found. We could if we wanted to, but again, this could be a separate feature (different return code if any tests are skipped), unrelated to this PR.
- We could add one line of text to the output, like "Solo test/group found! Skipping everything not solo." But I don't think it's required.
- I didn't do anything fancy to `solo` as a metadata attribute, which means it behaves like others (specifically, `skip`. This means it has the same "inheritance" rules as `skip` (tests and groups inherit value from containing group). For example:
  
  ``` text
  $ cat b.dart
  import 'lib/test.dart';
  
  void main() {
    group('group1', () {
      test('test A', () { expect('a', equals('a')); });
      test('test B', () { expect('a', equals('a')); });
    });
  
    solo_group('group2', () {
      test('test C', () { expect('a', equals('a')); });
      solo_test('test D', () { expect('a', equals('a')); });
    });
  }
  $ dart b.dart
  00:00 +0: group1 test A
  00:00 +0 ~1: group1 test B
  00:00 +0 ~2: group2 test C
  00:00 +1 ~2: group2 test D
  00:00 +2 ~2: All tests passed!
  ```
  
  `test A` and `test B` are skipped; simple enough. `test D` is run; no surprise there. `test C` is the only interesting case: it is run, even though it is a _non_ solo test, sitting next to a solo test, because it inherits `group2`'s solo status.
- Other cases are considered in the tests, in this PR.

### DIFF
--- a/lib/src/backend/declarer.dart
+++ b/lib/src/backend/declarer.dart
@@ -69,6 +69,9 @@ class Declarer {
   /// Whether [build] has been called for this declarer.
   bool _built = false;
 
+  /// Whether a solo test or solo group has been encountered.
+  bool _soloSeen = false;
+
   /// The current zone-scoped declarer.
   static Declarer get current => Zone.current[#test.declarer];
 
@@ -88,12 +91,25 @@ class Declarer {
   declare(body()) => runZoned(body, zoneValues: {#test.declarer: this});
 
   /// Defines a test case with the given name and body.
-  void test(String name, body(), {String testOn, Timeout timeout, skip,
-      Map<String, dynamic> onPlatform, tags}) {
+  void test(String name, body(),
+      {String testOn,
+      Timeout timeout,
+      skip,
+      solo,
+      Map<String, dynamic> onPlatform,
+      tags}) {
     _checkNotBuilt("test");
 
+    if (solo) {
+      _soloSeen = true;
+    }
+
     var metadata = _metadata.merge(new Metadata.parse(
-        testOn: testOn, timeout: timeout, skip: skip, onPlatform: onPlatform,
+        testOn: testOn,
+        timeout: timeout,
+        skip: skip,
+        solo: solo,
+        onPlatform: onPlatform,
         tags: tags));
 
     _entries.add(new LocalTest(_prefix(name), metadata, () async {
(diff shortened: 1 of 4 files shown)


nex3:
We've decided not to go in this direction.

matanlurey:
Can we have some background why? This is still my #1 issue with package:test, and we've had clients outright refuse to convert from unittest to test because they want this exact functionality.

PR #447, opened 2016-08-04
Old unittest use pkg matcher

kevmoo:
Will let us migrate folks internally – and likely help some external folks, too.

CC @jakemac53


nex3:
The matcher 0.12.0 API is not the same as the matcher API that we pulled into unittest; that's why we pulled it in in the first place. Doing this would be a breaking change.

PR #448, opened 2016-08-08
move nestingMiddleware to path_handler.dart

jakemac53:
This removes a dependency on `shelf` and `path_handler.dart` from `utils.dart`, so that we can more easily separate out test-only deps from test-runner deps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.12.15+3
+
+* Move `nestingMiddleware` to `lib/src/util/path_handler.dart` to enable a
+  cleaner separation between test-runner files and test writing files.
+
 ## 0.12.15+2
 
 * Support running without a `packages/` directory.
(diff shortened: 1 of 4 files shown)


kevmoo:
Closed with 5ee63c0

nex3:
This isn't the right place for `nestingMiddleware`. The `path_handler.dart` file has a single specific purpose: exposing the `PathHandler` class. It's not just a dumping-ground for all shelf-related stuff.

What's the problem with including shelf in the set of stuff that's reachable from `lib/test.dart`? Shelf if perfectly web-safe.

jakemac53:
It looks like the issue was `package:shelf/shelf_io.dart`. We can talk offline, but a general glob of all the shelf libs isn't safe because of that. Something like https://github.com/dart-lang/dev_compiler/issues/620 would resolve these issues.

PR #454, opened 2016-08-24
only decode data if its a String

jakemac53:
This is coming through as a `Map` for me (at least sometimes), and thus fails at runtime in ddc.


nex3:
If this is coming through as a string, something is broken elsewhere in the system. We need to find and fix the root cause. Do you have a consistent external reproduction?

jakemac53:
This is only an issue if opening the test directly instead of an iframe, as thats not a supported way of running tests I am closing this pr.

PR #467, opened 2016-09-28
Fix a bug in niceDuration().

### LINKED ISSUE #463: niceDuration(const Duration(minutes: 5)) -> "5 minutes, 5 seconds"
As of 3230bd991805b46ac4e78ea91626d4e0f0c26b0f

nex3:
It was printing the wrong number of seconds for durations longer than a
minute.

Closes #463


kevmoo-github:
Can haz tests?

kevmoo:
Tests?

nex3:
This is only surfaced in an error message, and I generally try to avoid writing tests for specific error messages when possible. They tend to be very fragile, since the tests always have to be updated even for routine error changes.

PR #468, opened 2016-09-28
Normalize the dependency on package:matcher

matanlurey:
Since there should not be any backwards-incompatible changes before 1.0.0, this should be safe.

This allows fixing the kludge in `package:matcher` where the version number cannot be incremented:

https://github.com/dart-lang/matcher/commit/831729359e015e1795254d1fb42f7c4803720f03

/cc @kevmoo

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -14,6 +14,7 @@ dependencies:
   collection: '^1.8.0'
   glob: '^1.0.0'
   http_multi_server: '>=1.0.0 <3.0.0'
+  matcher: '^0.12.0'
   package_resolver: '^1.0.0'
   path: '^1.2.0'
   pool: '^1.2.0'
@@ -30,10 +31,6 @@ dependencies:
   string_scanner: '>=0.1.1 <2.0.0'
   web_socket_channel: '^1.0.0'
   yaml: '^2.0.0'
-
-  # Use a tight version constraint to ensure that a constraint on matcher
-  # properly constrains all features it provides.
-  matcher: '>=0.12.0 <0.12.1'
 dev_dependencies:
   fake_async: '^0.1.2'
   scheduled_test: '^0.12.5'


kevmoo:
@nex3 thoughts?

nex3:
As the pubspec comment states, the narrow constraint is there for a reason. If one package exports another, the exporter must have a narrow constraint on the exportee. This is because users only put *their* constraints on the exporter, and they need that constraint to constrain the exported features as well.

Let's look at `test` and `matcher` as a concrete example. Suppose we release `matcher` 0.12.1 with [`isBefore()` and `isAfter()` matchers](https://github.com/dart-lang/matcher/issues/34). There are two ways this could work, depending on `test`'s constraint is on `matcher`:

* If `test` 0.12.15 had a loose constraint, `matcher: "^0.12.0"`, then we wouldn't need to release a new `test` version. A downstream user runs `pub upgrade` and gets `test` 0.12.15 and `matcher` 0.12.1. They look at `test`'s documentation and see `isBefore()` and `isAfter()`, and they start using them. They may not even know they were exported from a different package! They make sure their pubspec specifies the latest version of test, `test: "^0.12.15"`, just like they've been taught.

  But then something changes in their package resolution: someone else with a different SDK downloads their app, they change constraint elsewhere in their pubspec, whatever. *Something* happens and pub decides to select `matcher` 0.12.0 instead of 0.12.1. This is perfectly acceptable based on the constraints as they exist; the application just depends on `test: "^0.12.15"` and test just depends on `matcher: "^0.12.0"`. But it breaks their tests, because they were using `isBefore()` and `isAfter()`. They did everything right, and semver let them down. That's bad!

* But what if `test` had a tight constraint, `matcher: ">=0.12.0 <0.12.1"`? When `matcher` 0.12.1 is released, `test` kicks out its own new version. `test` 0.12.16 has the constraint `matcher: ">=0.12.1 <0.12.2"`. A downstream user runs `pub upgrade` and gets `test` 0.12.**16** and `matcher` 0.12.1 and uses `isBefore()` and `isAfter()`. They make sure their pubspec specifies `test: "^0.12.16"`. Now they're guaranteed to only get package versions that actually support the APIs they use.

Hope that helps!

PR #469, opened 2016-09-28
Fix some new analysis issues.

nex3:


lib/src/runner/browser/browser_manager.dart
```diff
@@ -225,7 +225,7 @@ class BrowserManager {
           sink.close();
         }));
 
-    return await _pool.withResource(() async {
+    return await _pool.withResource/*<Future<RunnerSuite>>*/(() async {
```
jmesserly: Ouch! If you want to add a bug note, this one is covered by:
https://github.com/dart-lang/sdk/issues/27284

nex3: I've sort of given up on bug notes for analysis issues like this TBH. It's too much work to track down and verify which one attaches to which bug. At some point when inference is more solid I'll just go through looking for explicit generic annotations and seeing if I can remove them.

PR #472, opened 2016-09-29
Stop using deprecated analyzer apis

kevmoo:


nex3:
I'm not comfortable constraining test to only one major version of `analyzer`. It revs so often and is depended on so widely that this will almost certainly cause version lock and prevent users from upgrading `test`. The deprecation hints aren't causing any harm, and we can easily update in the future when fewer packages are using analyzer versions that don't support the new import scheme.

matanlurey:
These APIs no longer will exist internally - which means we can never update third_party.

nex3:
I suggest re-adding them internally for the time-being, even if they just export the new locations. Removing them seems like a needlessly breaking change.

matanlurey:
You realize we can't roll-back the analyzer internally, ever? It's basically impossible.

nex3:
I'm not suggesting rolling it back, I'm suggesting adding a file in `lib/src/generated/ast.dart` that exports `lib/dart/ast/ast.dart`.

matanlurey:
Is there a guarantee that the APIs are exactly 1:1 (and will continue to be so, forever)?

nex3:
I hope so! If not, how can we safely make this change in test?

bwilkerson:
> I'm suggesting adding a file in lib/src/generated/ast.dart that exports lib/dart/ast/ast.dart.

That file already exists.

Kevin was just trying to remove the uses of the deprecated code before they go away internally because otherwise it will require a multi-step process to get a future version of analyzer rolled internally, which hurts our velocity.

nex3:
What I'm asking is, why are they going away at all? If the APIs still exist, what does it cost to keep the old deprecated filenames around? Removing them is a breaking change, one that concretely affects the biggest user of `analyzer`, and it'll cause a lot of pain.

bwilkerson:
Because we need to draw a clean distinction between the public and private APIs of analyzer, and src/generated is ambiguous.

nex3:
Certainly we shouldn't encourage new users to use it, but there's a substantial concrete benefit to allowing old code to continue using it for a while. That seems to me to outweigh concerns about cleanliness.

bwilkerson:
I'm not sure how long "a while" is :-), but the new location has existed since January 18th. I'm guessing Kevin could have widened the constraints a bit.

kevmoo:
Brian: I actually dug in – I think the lower bound I put is the required
one. Could double-check, though.

nex3:
I basically mean "as long as is feasible". The benefit we get from keeping `test`'s constraint on `analyzer` as wide as possible is very large.

kevmoo:
FWIW – 0.27.2 works – released on Feb 16, 2016

Updated my branch

PR #474, opened 2016-09-29
Don't capture stack traces by default.

### LINKED ISSUE #457: `pub run test` slow initialization.
`pub run test` initializes really slow.

When I write:

```bash
pub run test -p vm test/library_test.dart
```

runner waits around 25 seconds and then run everything in 7 seconds.

At the same time, writing

```bash
dart test/library_test.dart
```

will end in just 4 seconds.

I don't mind a bit slower execution, but having to wait half of minute is too much, and this time tends to increase as I add more tests.

You can check everything yourself on [dikmax/md_proc](https://github.com/dikmax/md_proc/tree/rewrite) branch `rewrite`.

nex3:
Stack traces are now only captured for the JSON reporter. Capturing
traces was slow enough to be a serious problem for test suites with many
tests.

Closes #457

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.12.15+8
+
+* Make test suites with thousands of tests load much faster on the VM (and
+  possibly other platforms).
+
 ## 0.12.15+7
 
 * Fix a bug where tags would be dropped when `on_platform` was defined in a
--- a/lib/src/backend/declarer.dart
+++ b/lib/src/backend/declarer.dart
@@ -38,6 +38,9 @@ class Declarer {
   /// The stack trace for this group.
   final Trace _trace;
 
+  /// Whether to collect stack traces for [GroupEntry]s.
+  final bool _collectTraces;
+
   /// The set-up functions to run for each test in this group.
   final _setUps = new List<AsyncFunction>();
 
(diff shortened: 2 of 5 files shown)


lib/src/backend/declarer.dart
```diff
@@ -77,10 +80,16 @@ class Declarer {
...
   /// If [metadata] is passed, it's used as the metadata for the implicit root
   /// group.
-  Declarer([Metadata metadata])
-      : this._(null, null, metadata == null ? new Metadata() : metadata, null);
+  ///
+  /// If [collectTraces] is `true`, this will set [GroupEntry.trace] for all
+  /// entries built by the declarer. Note that this can be noticeably slow when
+  /// thousands of tests are being declared (see #457).
+  Declarer({Metadata metadata, bool collectTraces: false})
+      : this._(null, null, metadata == null ? new Metadata() : metadata,
```
jmesserly: perhaps: `metadata ?? new Metadata` ?

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.12.15+8
```
jmesserly: the pubspec version below is `0.12.16-dev`, was that meant to be consistent?

lib/src/backend/declarer.dart
nex3: Done.

CHANGELOG.md
nex3: Unfortunate, but intended. Ideally the pubspec would be `0.12.15+8-dev`, but that's not allowed by semver, so we use the next-lowest dev version instead. If we release before adding any new features, it'll match the changelog.

jmesserly: ah, gotcha, interesting convention. So not necessarily intending to release right away. Makes sense.

nex3: Yeah, `-dev` always means "not intended for immediate release".

lib/src/backend/declarer.dart
```diff
@@ -77,10 +80,15 @@ class Declarer {
...
   /// If [metadata] is passed, it's used as the metadata for the implicit root
   /// group.
-  Declarer([Metadata metadata])
-      : this._(null, null, metadata == null ? new Metadata() : metadata, null);
+  ///
+  /// If [collectTraces] is `true`, this will set [GroupEntry.trace] for all
+  /// entries built by the declarer. Note that this can be noticeably slow when
+  /// thousands of tests are being declared (see #457).
+  Declarer({Metadata metadata, bool collectTraces: false})
+      : this._(null, null, metadata? ?? new Metadata(), collectTraces, null);
```
kevmoo: @nex3 OOPS! `metadata? ??` – breakage

PR #482, opened 2016-10-06
Remove hide for never-shipped pkg/async ForkableStream

kevmoo:


nex3:
I still intend to get `ForkableStream` into the async package at some point.

PR #486, opened 2016-10-10
Throw a better error for async groups.

### LINKED ISSUE #485: Fail with a better message or allow `async` groups
Async test cases are pretty common, its really easy to also throw in an `async` keyword on a group definition out of habit. When that happens the tests fail and the message isn't as helpful as it could be.
We should either allow async groups or we should make it more clear what the mistake was.

nex3:
Closes #485


natebosch:
LGTM.
Thanks!

PR #492, opened 2016-11-16
Factor out a SuiteConfiguration object.

nex3:
This tracks configuration that makes sense to apply to test suites, as
opposed to Configuration which applies to an entire run and Metadata
which applies to individual tests. This is intended to make it easier
for load separate configurations for separate targets in a live test
runner.

@goderbauer

### DIFF
--- a/lib/src/runner/configuration.dart
+++ b/lib/src/runner/configuration.dart
@@ -457,10 +457,8 @@ class Configuration {
     if (chosenPresets.isEmpty || presets.isEmpty) return this;
 
     var newPresets = new Map<String, Configuration>.from(presets);
-    var merged = chosenPresets.fold(empty, (merged, preset) {
-      if (!newPresets.containsKey(preset)) return merged;
-      return merged.merge(newPresets.remove(preset));
-    });
+    var merged = chosenPresets.fold(empty, (merged, preset) =>
+        merged.merge(newPresets.remove(preset) ?? Configuration.empty));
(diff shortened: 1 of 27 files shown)


lib/src/runner/configuration.dart
```diff
@@ -605,4 +450,26 @@ class Configuration {
...
           value: (config1, config2) => config1.merge(config2));
+
+  /// Returns a copy of this [Configuration] with all [chosenPresets] resolved
+  /// against [presets].
+  Configuration _resolvePresets() {
+    if (chosenPresets.isEmpty || presets.isEmpty) return this;
+
+    var newPresets = new Map<String, Configuration>.from(presets);
+    var merged = chosenPresets.fold(empty, (merged, preset) {
+      if (!newPresets.containsKey(preset)) return merged;
```
jmesserly: if `null` isn't possible in presets, then you can combine these two operations in one dictionary lookup:
```dart
var newPreset = newPresets.remove(preset);
return newPreset != null ? merged.merge(newPreset) : merged;
```
https://api.dartlang.org/stable/1.20.1/dart-core/Map/remove.html

whether it's worth a small tweak like that, not sure. :)

nex3: All small tweaks are worth it if they're good. I can even make it cleaner, since `Configuration.merge(Configuration.empty)` is a no-op. Done.

Issue #109, opened 2015-05-07
Add support for server plugins


nex3:
It should be possible for users to create plugins that provide additional handlers for the server that can communicate with browsers, as well as the ability to hook into the communication channel with the test page. This is especially useful for test infrastructure that straddles the browser/console boundary, for example google/webdriver.dart#76.

Blocked on #48.

PR #509, opened 2016-12-19
Add support for hybrid VM/browser tests.

nex3:
This adds `spawnHybridUri()` and `spawnHybridCode()` methods that spawn
VM isolates even when called from the browser. It still needs some
additional work to close the isolates when the test ends by default.

See #109


lib/src/runner/hybrid_listener.dart
```diff
@@ -0,0 +1,88 @@
...
+import "package:stream_channel/stream_channel.dart";
+
+import "../util/remote_exception.dart";
+import "../utils.dart";
+
+/// A sink transformer that wraps data and error events so that errors can be
+/// decoded after being JSON-serialized.
+final _transformer = new StreamSinkTransformer.fromHandlers(
+    handleData: (data, sink) {
+      ensureJsonEncodable(data);
```
natebosch: what happens if this throws? Should we be putting an error on the sink?

lib/src/runner/hybrid_listener.dart
```diff
@@ -0,0 +1,88 @@
...
+      } on NoSuchMethodError catch (_) {
+        _sendError(channel, "No top-level hybridMain() function defined.");
+        return;
+      } catch (error, stackTrace) {
+        _sendError(channel, error, stackTrace);
+        return;
+      }
+
+      if (main is! Function) {
+        _sendError(channel, "Top-level hybridMain getter is not a function.");
```
natebosch: [nit] this usage of 'getter' might not match what some readers expect. I might phrase this as "Top-level hybridMain is not a function."

lib/src/frontend/spawn_hybrid.dart
```diff
@@ -0,0 +1,166 @@
...
+/// isolate will be printed as though they came from the test that created the
+/// isolate.
+///
+/// Code in the hybrid isolate is not considered to be running in a test
+/// context, so it can't access test functions like `expect()` and
+/// `expectAsync()`.
+///
+/// **Note**: If you use this API, be sure to add a dependency on the
+/// **`stream_channel` package, since you're using its API as well!
+StreamChannel spawnHybridCode(String dartCode, {Object message}) {
```
kevmoo: DBC: This seems cool – but feel scary.

Is there a strong case for this over just giving folks the URI-based method?

lib/src/frontend/spawn_hybrid.dart
```diff
@@ -0,0 +1,166 @@
...
+          var error = RemoteException.deserialize(message['error']);
+          sink.addError(error.error, error.stackTrace);
+          break;
+      }
+    }), new StreamSinkTransformer.fromHandlers(handleData: (message, sink) {
+      ensureJsonEncodable(message);
+      sink.add(message);
+    }));
+
+/// Spawns a VM isolate for the given [uri], which may be a [Uri] or a [String].
```
kevmoo: Can be URI or String.

We need union types. 😄

lib/src/frontend/spawn_hybrid.dart
```diff
@@ -0,0 +1,166 @@
...
+
+import 'package:async/async.dart';
+import 'package:path/path.dart' as p;
+import 'package:stream_channel/stream_channel.dart';
+
+import '../backend/invoker.dart';
+import '../util/remote_exception.dart';
+import '../utils.dart';
+
+/// A transformer that handles messages from the spawned isolate and ensures
```
nex3: Done.

lib/src/frontend/spawn_hybrid.dart
```diff
@@ -0,0 +1,166 @@
...
+/// that messages sent to it are JSON-encodable.
+final _transformer = new StreamChannelTransformer(
+    new StreamTransformer.fromHandlers(handleData: (message, sink) {
+      switch (message['type']) {
```
nex3: Done.

lib/src/frontend/spawn_hybrid.dart
nex3: More specifically, in practice URLs are represented as Strings and Uri objects with similar frequency in Dart. I try to make my APIs handle either representation to avoid making my users write lots of annoying `.toString()` or `Uri.parse()` calls.

lib/src/frontend/spawn_hybrid.dart
```diff
@@ -0,0 +1,166 @@
...
+    }));
+
+/// Spawns a VM isolate for the given [uri], which may be a [Uri] or a [String].
+///
+/// This allows browser tests to spawn servers with which they can communicate
+/// to test client/server interactions. It can also be used by VM tests to
+/// easily spawn an isolate.
+///
+/// The Dart file at [uri] must define a top-level `hybridMain()` function that
+/// takes a `StreamChannel` argument and, optionally, an `Object` argument to
```
nex3: You can, but the Stream API doesn't make it particularly easy to get the first event without also canceling the stream subscription. This also mirrors the Isolate spawn APIs more closely.

lib/src/frontend/spawn_hybrid.dart
```diff
@@ -0,0 +1,166 @@
...
+  Uri parsedUrl;
+  if (uri is Uri) {
+    parsedUrl = uri;
+  } else if (uri is String) {
+    parsedUrl = Uri.parse(uri);
+  } else {
+    throw new ArgumentError.value(uri, "uri", "must be a Uri or a String.");
+  }
+
+  String uriString;
```
nex3: Done.

lib/src/runner/hybrid_listener.dart
nex3: This is called synchronously from the user's `Sink.add()` call, so if `ensureJsonEncodable()` throws here they'll get a helpful stack trace. I'll add a comment to that effect.

lib/src/runner/hybrid_listener.dart
nex3: Done.

lib/src/frontend/spawn_hybrid.dart
nex3: Can you elaborate on "scary"?

This is useful for the common situation where the hybrid isolate is very simple. It lets users keep the full definition of their test in one file.

PR #510, opened 2016-12-20
Add an addTearDown() function.

### LINKED ISSUE #93: Add the ability to add to the tearDown for a running test
This is an idea that comes from `scheduled_test`. It makes it possible to create testing APIs that can clean up after themselves without the user having to explicitly pipe them to the tearDown. For example, a `Sandbox` class might tear down the sandbox directory; `ScheduledProcess` uses this to kill dangling processes.

nex3:
See #109
Closes #93

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 0.12.18
 
+* Add an `addTearDown()` function, which allows tests to register additional
+  tear-down callbacks as they're running.
+
 * Add the `spawnHybridUri()` and `spawnHybridCode()` functions, which allow
   browser tests to run code on the VM.
 
--- a/lib/src/backend/declarer.dart
+++ b/lib/src/backend/declarer.dart
@@ -105,15 +105,24 @@ class Declarer {
         tags: tags));
 
     _entries.add(new LocalTest(_prefix(name), metadata, () async {
-      // TODO(nweiz): It might be useful to throw an error here if a test starts
-      // running while other tests from the same declarer are also running,
-      // since they might share closurized state.
+      var parents = [];
+      for (var declarer = this; declarer != null; declarer = declarer._parent) {
+        parents.add(declarer);
+      }
+
+      // Register all tear-down functions in all declarers. Iterate through
+      // parents outside-in so that the Invoker gets the functions in the order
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## 0.12.18
 
+* Add an `addTearDown()` function, which allows tests to register additional
+  tear-down callbacks as they're running.
+
```
natebosch: Are you intentionally wanting `<p>` tags inside the `<li>`?

I wouldn't normally expect an extra blank line between bullet points

lib/src/backend/declarer.dart
```diff
@@ -105,15 +105,24 @@ class Declarer {
         tags: tags));
 
     _entries.add(new LocalTest(_prefix(name), metadata, () async {
-      // TODO(nweiz): It might be useful to throw an error here if a test starts
-      // running while other tests from the same declarer are also running,
-      // since they might share closurized state.
+      var parents = [];
```
natebosch: is this an intentionally dynamically typed list?

CHANGELOG.md
nex3: I think it makes the changelog entries more readable, both in markdown and in HTML.

lib/src/backend/declarer.dart
nex3: No, done.

test/frontend/add_tear_down_test.dart
```diff
@@ -0,0 +1,325 @@
...
+        expect(tearDownRun, isTrue);
+      });
+    });
+  });
+
+  test("multiples run in reverse order", () {
+    return expectTestsPass(() {
+      var tearDown1Run = false;
+      var tearDown2Run = false;
+      var tearDown3Run = false;
```
natebosch: [nit] does having 3 tear downs instead of 2 add any extra safety? Would cut down on clutter if we used only 2.

nex3: The idea is that it makes it a lot less likely to accidentally pass if the order is shuffled (17% vs 50%)—which is at least plausible given how much asynchrony is in play.

Issue #50, opened 2015-03-31
Support debugging VM tests


nex3:
The test runner should make it very easy to run an interactive debugger for VM tests, whether it be via the command line, observatory, or an IDE. This is blocked on [issue 23320](https://code.google.com/p/dart/issues/detail?id=23320), which will allow the test runner to enable the observatory server at runtime.

kevmoo:
@nex3 FYI: update this issue with description RE observatory

kevmoo:
Also would like/need ability to start an isolate in checked mode - https://code.google.com/p/dart/issues/detail?id=21791

johnpryan:
[sdk issue 23320](https://code.google.com/p/dart/issues/detail?id=23320) is marked as 'not planned'.  Could we get some clarification on what that means exactly?

It's still possible to debug tests by creating an HTML file with `<script>` tags and debug in WebStorm, but that file needs to be maintained, and you cant use the `-n` flag or `solo_` to filter tests.

kevmoo:
See my comments here: https://github.com/dart-lang/pub/issues/1018#issuecomment-157204503

Short version: we'll mark this work

alorenzen:
Any update on this?

nex3:
The blocking issue is still blocking.

nex3:
The [`Service`](https://api.dartlang.org/dev/1.21.0-dev.6.0/dart-developer/Service-class.html) class should enable this, so I'm going to unmark this as blocked.

nex3:
This is currently blocked on dart-lang/sdk#31308.

PR #521, opened 2017-01-04
Support debugging VM tests.

nex3:
This doesn't fully support the JSON reporter yet.

See #50

### DIFF
--- a/lib/src/runner/vm/platform.dart
+++ b/lib/src/runner/vm/platform.dart
@@ -3,16 +3,25 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:developer';
+import 'dart:io';
 import 'dart:isolate';
 
 import 'package:path/path.dart' as p;
 import 'package:stream_channel/stream_channel.dart';
+import 'package:vm_service_client/vm_service_client.dart';
 
 import '../../backend/test_platform.dart';
 import '../../util/dart.dart' as dart;
 import '../configuration.dart';
+import '../configuration/suite.dart';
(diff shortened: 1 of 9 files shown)


lib/src/runner/vm/platform.dart
```diff
@@ -21,29 +30,64 @@ class VMPlatform extends PlatformPlugin {
...
 
-    var isolate;
-    var channel = StreamChannelCompleter.fromFuture(() async {
-      var receivePort = new ReceivePort();
+    var receivePort = new ReceivePort();
+    Isolate isolate;
+    try {
+      isolate = await _spawnIsolate(path, receivePort.sendPort);
+    } catch (error) {
+      receivePort.close();
```
jmesserly: just a question: does this need to be closed in case of success (at some point later in the function)? or does IsolateChannel/something else handle that.

pubspec.yaml
```diff
@@ -39,3 +40,9 @@ dev_dependencies:
   http: '^0.11.0'
   js: '^0.6.0'
   scheduled_test: '^0.12.5'
+
+dependency_overrides:
```
jmesserly: did you mean to check this in (e.g. so local developers always test against git?)

lib/src/runner/vm/platform.dart
nex3: `IsolateChannel` handles it. It assumes total control of the port.

pubspec.yaml
nex3: It's intended for now. I'll get rid of it once dart-lang/vm_service_client#13 lands.

PR #530, opened 2017-02-02
Deprecate throws and Throws in favor of throwsA().

nex3:


matanlurey:
I realize this hasn't been removed yet - but have we evaluated use internally and externally?

nex3:
@kevmoo It's still possible to write `throwsA(anything)`, but the worse option shouldn't be easier than the better option.

@matanlurey I think this is the right change regardless of use.

matanlurey:
@nex3 Well, unless you're planning on keeping <1.0 around for a long time it could impact SDK/package rolls, so just wondering what the plans are re: actually removing it (not just marking deprecated). Could we also include some sort of trivial script that updates all `throws` to `throwsA(anything)` for our users?

nex3:
That sounds useful, but I don't have the expertise (or time) to write such a script. I don't think we're actually likely to release a breaking change soon, though, so there'll be plenty of time to gradually migrate code.

PR #532, opened 2017-02-06
Add stream matchers.

nex3:


lib/src/frontend/stream_matchers.dart
```diff
@@ -0,0 +1,372 @@
...
+/// Returns a [streamMatcher] that matches the stream if at least one of
+/// [matchers] matches.
+///
+/// If multiple matchers match the stream, this chooses the matcher that
+/// consumes as many events as possible.
+///
+/// If any matchers match the stream, no errors from other matchers are thrown.
+/// If no matchers match and multiple matchers threw errors, the first error is
+/// re-thrown.
+StreamMatcher emitsAnyOf(Iterable matchers) {
```
munificent: Error if the iterable is empty?

lib/src/frontend/stream_matchers.dart
```diff
@@ -0,0 +1,372 @@
...
+    StreamQueue consumedMost;
+    for (var i = 0; i < matchers.length; i++) {
+      futures.add(() async {
+        var copy = transaction.newQueue();
+
+        String result;
+        try {
+          result = await streamMatchers[i].matchQueue(copy);
+        } catch (error, stackTrace) {
+          if (error != null) {
```
munificent: `error` should never be null. Is this case needed?

lib/src/frontend/stream_matchers.dart
```diff
@@ -0,0 +1,372 @@
...
+        newResult += newResult.contains("\n") ? "\n" : " ";
+        newResult += "because it $result";
+      }
+      return newResult;
+    }
+  }, description);
+}
+
+/// Returns a [StreamMatcher] that matches any number of
+/// events followed by events that match [matcher].
```
munificent: Nit: The wrapping here looks arbitrary.

lib/src/frontend/stream_matchers.dart
```diff
@@ -0,0 +1,372 @@
...
+      var result = await streamMatcher.matchQueue(copy);
+      if (result == null) return true;
+      failures.add(result);
+      return false;
+    });
+
+    while (await queue.hasNext) {
+      if (await tryHere()) return null;
+      await queue.next;
+    }
```
munificent: This function is quadratic, right? Or I guess, more specifically O(mn) where m is the stream length and n is the matcher's length.

That's OK, but it would probably be good to mention it since it can get slow with long streams and matchers.

Or you could adapt Boyer-Moore to this. 😛

lib/src/frontend/stream_matchers.dart
nex3: Done.

lib/src/frontend/stream_matchers.dart
nex3: Oops, this was supposed to check `firstError`.

lib/src/frontend/stream_matchers.dart
nex3: Done.

lib/src/frontend/stream_matchers.dart
nex3: The quadradic behavior is only relevant when the matcher can match long strings of events *and* when it in practice it matches many long prefixes without satisfying the entire matcher. I don't think that's likely enough to come up in practice to warrant documenting.

brianegan:
Late to the party, but yay! Nice work @nex3!

PR #533, opened 2017-02-07
Add a printOnFailure() function.

nex3:
This prints debugging information only if the test actually fails.


nex3:
`test.py` is a completely different system than the `test` package.

efortuna:
gotcha. thanks!

On Tue, Feb 7, 2017 at 2:34 PM, Natalie Weizenbaum <notifications@github.com

PR #546, opened 2017-02-21
Replace "Future expect()" with "expectLater()".

nex3:


lib/src/frontend/expect.dart
```diff
@@ -11,6 +11,12 @@ import '../backend/invoker.dart';
 import '../utils.dart';
 import 'async_matcher.dart';
 
+/// A future that emits `null`.
+///
+/// We cache and re-use this value to avoid adding a new microtask hit for each
```
natebosch: have we confirmed that this solves the flutter scenario?

Another pattern we could consider is to return `null` from `_expect` and always guarantee a `Future` from `expectLater` with something like `=> _expect(...) ?? _emptyFuture;`

nex3: Since we only instantiate this once, it definitely won't add more than one microtask unless it gets listeners. As long as they add something like `setUpAll(() => expect(null, isNull))` it should work fine.

That said, while I'm hoping to mitigate the microtask-counting scenario, finding a surefire and highly usable solution isn't really a goal for me. The number of microtasks in the event queue at any given point is deep into "not guaranteed" territory.

PR #553, opened 2017-02-23
Larger iframe

eukreign:
This pull request changes the default iframe size from a tiny 300px by 150px to a much larger 800px by 600px. This makes it more practical to write unit tests for code that uses `document.elementFromPoint()` or anything else where the limitations of the viewport create a problem.


nex3:
Can this be done in the CSS file rather than in code?

eukreign:
It appears to work with just CSS. Is that what you would prefer?

nex3:
Yeah, I think it would be a bit cleaner.

eukreign:
Changed to use CCS.

There was a suggestion on dart-lang/TALK-general to make this configurable. Would be nice I guess but I'd rather the hard-coded version landed now and a separate pull request could be made with a configurable version in the future.

Personally, I just want something bigger than 300x150, whether that something is 800x600 or 1024x768 makes no difference.

nex3:
I'd prefer to avoid adding configuration if we can find a value that works for everyone.

eukreign:
According to statcounter 1366x768 is the most common screen size for desktops now-a-days.

http://gs.statcounter.com/screen-resolution-stats/desktop/worldwide

donny-dont:
Would it be possible to have pub set the size of content shell and then size the iframe to be the size of the window? That way there wouldn't be any configuration on pubs end.

elmcrest:
@donny-dont what about  machines where no screen is attached?

eukreign:
@elmcrest There is always a screen, even if xvfb. `content_shell` cannot actually run headless, it needs at least a fake screen.

eukreign:
After thinking about this a bit more, I think there are two main use cases:

1. You just want some sane screen size that is representative of the common screen size. That's what this pull request is about.

2. You are building some kind of framework that needs to respond or handle different screen sizes. This would require being able to set the size inside of your unit tests. Configuring the size with pub or anywhere else like that wouldn't really be helpful.

nex3:
It's always possible—and will be more reliable—to create an `iframe` from within the test if it needs fine-grained control over the viewport size. I think having a reasonably-sized default and letting the configuration be done that way is reasonable.

PR #559, opened 2017-03-06
Print "[E]" after failing tests.

### LINKED ISSUE #554: Add an option to print a failure summary after tests run
At the moment, there is no clean way to copy-paste failures after a run.

Even printing "FAILURE" on the line when a new failure is found would be useful – then one could search/grep

nex3:
Closes #554

### DIFF
--- a/lib/src/runner/reporter/compact.dart
+++ b/lib/src/runner/reporter/compact.dart
@@ -198,7 +198,7 @@ class CompactReporter implements Reporter {
   void _onError(LiveTest liveTest, error, StackTrace stackTrace) {
     if (liveTest.state.status != Status.complete) return;
 
-    _progressLine(_description(liveTest), truncate: false);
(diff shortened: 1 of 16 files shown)


lib/src/runner/reporter/compact.dart
```diff
@@ -198,7 +198,7 @@ class CompactReporter implements Reporter {
   void _onError(LiveTest liveTest, error, StackTrace stackTrace) {
     if (liveTest.state.status != Status.complete) return;
 
-    _progressLine(_description(liveTest), truncate: false);
+    _progressLine(_description(liveTest) + " $_bold$_red[E]$_noColor");
```
kevmoo: no longer truncated – intentional?

nex3: Done.

PR #560, opened 2017-03-06
Remove a dependency override.

nex3:


pubspec.yaml
```diff
@@ -44,4 +44,3 @@ dev_dependencies:
 
 dependency_overrides:
   scheduled_test: '^0.12.5'
-  stack_trace: {path: ../stack_trace}
```
kevmoo: Does the existing constraint – `^1.2.1` – capture the right thing here?

nex3: Yes. We're not using any new features—the tests just happen to depend on the stringification specifics.

PR #565, opened 2017-03-11
Release 0.12.20+3.

nex3:


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: test
-version: 0.12.20+2
+version: 0.12.20+3
```
kevmoo: scheduled_test dependency override?

nex3: Since it's just overriding a dev dependency, pub won't complain.

PR #572, opened 2017-03-23
fix the strong mode errors under lib

jakemac53:
This allows this package to be compiled with dartdevc


lib/src/runner/browser/platform.dart
```diff
@@ -443,7 +443,7 @@ class BrowserPlatform extends PlatformPlugin {
   /// Returns a [Future] that completes once the server is closed and its
   /// resources have been fully released.
   Future close() => _closeMemo.runOnce(() async {
-    var futures = _browserManagers.values.map((future) async {
+    var futures = _browserManagers.values.map<Future<dynamic>>((future) async {
```
kevmoo: Don't need `<dynamic>` here

jakemac53: done

PR #575, opened 2017-03-24
Correctly kill Dartium on Linux

grouma:
Verified on Ubuntu 14.04:

➜  test git:(LinuxFixes) ✗ pub run test
00:20 +92: test/runner/hybrid_test.dart: spawnHybridCode() kills the isolate when the test closes the channel
  Skip: Enable when sdk#28081 is fixed.
00:20 +92 ~1: test/runner/hybrid_test.dart: spawnHybridCode() kills the isolate when the hybrid isolate closes the channel
  Skip: Enable when sdk#28081 is fixed.
01:23 +343 ~2: test/runner/browser/content_shell_test.dart: starts content shell with the given URL
  Skip: Failing with mysterious WebSocket issues.
09:16 +919 ~3: All tests passed!

Also verified no impact on Mac.

### DIFF
--- a/lib/src/runner/browser/browser.dart
+++ b/lib/src/runner/browser/browser.dart
@@ -112,11 +112,15 @@ abstract class Browser {
     _process.then((process) {
       // Dartium has a difficult time being killed on Linux. To ensure it is
       // properly closed, find all children processes and kill those first.
-      if (Platform.isLinux) {
-        var result = Process.runSync('pgrep', ['-P', '${process.pid}']);
-        for(var pid in '${result.stdout}'.split('\n')){
-          Process.runSync('kill', ['-9', pid]);
+      try{
+        if (Platform.isLinux) {
+          var result = Process.runSync('pgrep', ['-P', '${process.pid}']);
+          for(var pid in '${result.stdout}'.split('\n')){
+            Process.runSync('kill', ['-9', pid]);
+          }
         }
+      }catch(e){
+        print('Failed to kill browser children: $e');
       }
       process.kill();
     });
--- a/lib/src/runner/browser/browser.dart
+++ b/lib/src/runner/browser/browser.dart
@@ -115,11 +115,11 @@ abstract class Browser {
       try{
         if (Platform.isLinux) {
           var result = Process.runSync('pgrep', ['-P', '${process.pid}']);
-          for(var pid in '${result.stdout}'.split('\n')){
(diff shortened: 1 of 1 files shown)


grouma:
Yup, I tried that and it unfortunately did not work. Neither did 'pkill -TERM -P <process id>' reliably which is quite odd.

lib/src/runner/browser/browser.dart
```diff
@@ -108,7 +108,18 @@ abstract class Browser {
   /// exceptions.
   Future close() {
     _closed = true;
-    _process.then((process) => process.kill());
+
+    _process.then((process) {
+      // Dartium has a difficult time being killed on Linux. To ensure it is
+      // properly closed, find all children processes and kill those first.
+      if (Platform.isLinux) {
+        var result = Process.runSync('pgrep', ['-P', '${process.pid}']);
```
kevmoo: I'd wrap this whole thing in `try...catch`. Possible that an exec doesn't exist on the target platform or security...who knows...

grouma: Done.

lib/src/runner/browser/browser.dart
```diff
@@ -112,11 +112,15 @@ abstract class Browser {
...
-        for(var pid in '${result.stdout}'.split('\n')){
-          Process.runSync('kill', ['-9', pid]);
+      try{
+        if (Platform.isLinux) {
+          var result = Process.runSync('pgrep', ['-P', '${process.pid}']);
+          for(var pid in '${result.stdout}'.split('\n')){
+            Process.runSync('kill', ['-9', pid]);
+          }
         }
+      }catch(e){
```
kevmoo: space between `}` and `catch`

the print here makes me sad...but...eh...

terrylucas:
Do you use the dartfmt?

kevmoo:
> Do you use the dartfmt?

Can't yet...

lib/src/runner/browser/browser.dart
kevmoo: @grouma this will be fun to grab when you format

grouma: Done.

grouma:
@terrylucas first order of business was getting the tests to pass on Linux. Next is syncing in the changes to google3 and finally I'll format the entire package so that styling issues will no longer be a problem.

PR #580, opened 2017-03-31
Add type info to prevent DDC compile time failure with 1.23.0-dev.11.0

keertip:


keertip:
Gary, would like to get a version published so that this can be pulled into google3.  So should I add an entry to CHANGELOG?

PR #581, opened 2017-03-31
Fix null conditional

grouma:


grouma:
@kevmoo, yes. Necessary to resolve some DDC issues.

PR #584, opened 2017-04-05
Conditional chaining

grouma:
Introduce a new flag --chain-stack-traces which providers users with a way to conditionally enable stack trace chaining. This will be useful for those users that want improved test performance at the cost of debuggability.


lib/src/backend/invoker.dart
```diff
@@ -289,6 +289,11 @@ class Invoker {
     _controller.addError(error, stackTrace);
     removeAllOutstandingCallbacks();
 
+    if (!liveTest.test.metadata.chainStackTraces) {
+      _printsOnFailure.add("Consider enabling the flag chain-stack-traces to "
```
matanlurey: Could we make this a little more specific, perhaps:

"Consider ...\n For example, pub run test --chain-stack-traces"

matanlurey:
Not expecting for this PR, but any thought about auto-rerunning with chaining enabled when a test failed but keeping it off by default?

kevmoo:
> Not expecting for this PR, but any thought about auto-rerunning with chaining enabled when a test failed but keeping it off by default?

@matanlurey – file a bug (feature request) ?

matanlurey:
Filed https://github.com/dart-lang/test/issues/585.

lib/src/backend/invoker.dart
grouma: Done.

lib/src/runner/configuration/args.dart
```diff
@@ -85,6 +85,12 @@ final ArgParser _parser = (() {
           'Implies --concurrency=1 and --timeout=none.\n'
           'Currently only supported for browser tests.',
       negatable: false);
+  parser.addFlag("chain-stack-traces",
+      help: 'Chains stack traces to provide greater exception details\n'
```
kevmoo: `Chains` -> `Chained`?

grouma: Done.

PR #588, opened 2017-04-10
Extensible reporters and test platforms

grouma:
Adding hooks for new reporters and test platforms. These hooks are not intended to be a part of the external API and thus are aptly documented.

### DIFF
--- a/lib/src/backend/test_platform.dart
+++ b/lib/src/backend/test_platform.dart
@@ -46,7 +46,8 @@ class TestPlatform {
       isBrowser: true, isJS: true);
 
   /// A list of all instances of [TestPlatform].
-  static var all = new UnmodifiableListView<TestPlatform>(_all);
+  static final UnmodifiableListView<TestPlatform> all =
+      new UnmodifiableListView<TestPlatform>(_allPlatforms);
 
   /// Finds a platform by its identifier string.
   ///
@@ -86,7 +87,7 @@ class TestPlatform {
   String toString() => name;
 }
 
-List<TestPlatform> _all = [
+final List<TestPlatform> _allPlatforms = [
   TestPlatform.vm,
   TestPlatform.dartium,
   TestPlatform.contentShell,
@@ -108,12 +109,12 @@ TestPlatform registerTestPlatform(String name, String identifier,
     bool isJS: false,
     bool isBlink: false,
     bool isHeadless: false}) {
-  TestPlatform platform = new TestPlatform._(name, identifier,
+  var platform = new TestPlatform._(name, identifier,
       isDartVM: isDartVM,
       isBrowser: isBrowser,
       isJS: isJS,
       isBlink: isBlink,
       isHeadless: isHeadless);
-  _all.add(platform);
+  _allPlatforms.add(platform);
   return platform;
 }
--- a/lib/src/backend/test_platform.dart
+++ b/lib/src/backend/test_platform.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:collection';
+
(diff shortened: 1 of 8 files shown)


lib/src/backend/test_platform.dart
```diff
@@ -44,16 +46,7 @@ class TestPlatform {
...
-    vm,
-    dartium,
-    contentShell,
-    chrome,
-    phantomJS,
-    firefox,
-    safari,
-    internetExplorer
-  ];
+  static var all = new UnmodifiableListView<TestPlatform>(_all);
```
kevmoo: `final` – also add a type annotation `List<TestPlatform>` since this is public

lib/src/backend/test_platform.dart
```diff
@@ -92,3 +85,35 @@ class TestPlatform {
 
   String toString() => name;
 }
+
+List<TestPlatform> _all = [
```
kevmoo: `final`. Also, put a type argument on the array – so the reified type is correct `<TestPlatform>[...`

Also `_all` is a bit too vague.

lib/src/backend/test_platform.dart
```diff
@@ -92,3 +85,35 @@ class TestPlatform {
...
+///
+/// This constructs and globally registers a new TestPlatform with the provided
+/// details.
+TestPlatform registerTestPlatform(String name, String identifier,
+    {bool isDartVM: false,
+    bool isBrowser: false,
+    bool isJS: false,
+    bool isBlink: false,
+    bool isHeadless: false}) {
+  TestPlatform platform = new TestPlatform._(name, identifier,
```
kevmoo: `var` here is fine

lib/src/runner/configuration/args.dart
```diff
@@ -92,17 +93,18 @@ final ArgParser _parser = (() {
           'debuggability.',
       defaultsTo: true);
 
+  Map reporterDescriptions = {};
```
kevmoo: Put types on the right when possible.

`var reporterDescriptions = <String, String>{};`

lib/src/runner/configuration/reporters.dart
```diff
@@ -0,0 +1,55 @@
...
+  final String description;
+  final ReporterFactory factory;
+  ReporterDetails(this.description, this.factory);
+}
+
+/// All reporters and their corresponding details.
+final UnmodifiableMapView<String, ReporterDetails> allReporters =
+    new UnmodifiableMapView<String, ReporterDetails>(_allReporters);
+
+final Map<String, ReporterDetails> _allReporters = {
```
kevmoo: Same here. Types not needed on left side for privates. Put on the right.

lib/src/backend/test_platform.dart
grouma: I'm told that since all the values are TestPlatform the reified type will already be correct.

kevmoo: I guess, yes. Inferred.

lib/src/runner/configuration/reporters.dart
grouma: Done.

lib/src/runner/configuration/args.dart
grouma: Done.

lib/src/backend/test_platform.dart
grouma: Done.

lib/src/backend/test_platform.dart
grouma: Done.

nex3:
How does registering new `TestPlatform`s work with platform selectors? Selectors are parsed outside the main isolate, so the set of available platforms needs to be communicated to them somehow in order for those platforms to be selectable.

PR #596, opened 2017-05-01
Create psuedo-public entrypoints for bootstrapping vm/browser tests.

jakemac53:
This proposed change takes the code generated by the pub serve transformer and migrates all the business logic into two new "public" entry points. The methods exposed are marked as `@deprecated` (just to discourage direct use) and have a comment explaining that they should not be used directly. I also prefixed the names with `internal` as yet another signal not to rely on these files directly.

The future dartdevc transformer also doesn't support importing files from lib/src directly which are not included transitively by at least one public entry point (defined as a .dart file under `lib` but not `lib/src`). We could solve this in a couple other ways, but I think this is the most logical.

This also simplifies the transformer implementation, and more importantly it makes it so the transformer doesn't generate code with imports to dependencies that you don't depend on explicitly.

It also enables analyzer errors/warnings for these files, and reduces the amount of code that is just in a raw string.

### DIFF
--- a/lib/bootstrap/browser.dart
+++ b/lib/bootstrap/browser.dart
@@ -3,7 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 import "../src/runner/plugin/remote_platform_helpers.dart";
 import "../src/runner/browser/post_message_channel.dart";
-import '../src/utils.dart';
+import "../src/utils.dart";
 
 /// Bootstraps a browser test to communicate with the test runner.
 ///
--- a/lib/bootstrap/browser.dart
+++ b/lib/bootstrap/browser.dart
@@ -1,0 +1,16 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import "../src/runner/plugin/remote_platform_helpers.dart";
+import "../src/runner/browser/post_message_channel.dart";
+import '../src/utils.dart';
+
+/// Bootstraps a browser test to communicate with the test runner.
+///
+/// This should NOT be used directly, instead use the `test/pub_serve`
+/// transformer which will bootstrap your test and call this method.
+@deprecated
+void internalBootstrapBrowserTest(AsyncFunction originalMain) {
+  var channel = serializeSuite(() => originalMain);
+  postMessageChannel().pipe(channel);
+}
--- a/lib/bootstrap/vm.dart
+++ b/lib/bootstrap/vm.dart
@@ -1,0 +1,23 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import "dart:isolate";
+
+import "package:stream_channel/stream_channel.dart";
+
+import "../src/runner/plugin/remote_platform_helpers.dart";
+import "../src/runner/vm/catch_isolate_errors.dart";
+import "../src/utils.dart";
+
+/// Bootstraps a vm test to communicate with the test runner.
+///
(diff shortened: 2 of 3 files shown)


lib/bootstrap/browser.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import "../src/runner/plugin/remote_platform_helpers.dart";
+import "../src/runner/browser/post_message_channel.dart";
+import '../src/utils.dart';
```
natebosch: [nit] pick single or double quotes for the whole file

kevmoo:
Rebase on latest and you should be good...

nex3:
I'm skeptical of this. Putting private files that are never intended to be used by any other package outside of `src/` smells like a hack—`src/` means exactly "files that are only intended to be used by this package". If DDC is interpreting it to mean "files transitively imported by lib/", that goes against pub's semantics.

jakemac53:
The transformer adds imports to these files to files that ultimate live in other packages. Because of that I think this is reasonable - even if package:test is the one actually adding the imports.

The number of use-cases that don't fit with the current strategy for DDC is very small (this was the only one I could find), and I think it encourages good patterns in general. That being said I would like to support this use case - and once we add a way of configuring ddc explicitly then we should revisit this imo.

jakemac53:
Another possibility is actually moving this to a real public api. For instance the server from package:build_runner could use it to work as a pub serve replacement this way. Having this hook makes it so I wouldn't have to copy the bootstrapping code from the transformer (which is fragile at best) to do that. Same goes for any other build system that might want to generate code that is usable with the package:test test runner.

nex3:
> That being said I would like to support this use case - and once we add a way of configuring ddc explicitly then we should revisit this imo.

If this is on the roadmap, I'm satisfied.

> Another possibility is actually moving this to a real public api. For instance the server from package:build_runner could use it to work as a pub serve replacement this way. Having this hook makes it so I wouldn't have to copy the bootstrapping code from the transformer (which is fragile at best) to do that. Same goes for any other build system that might want to generate code that is usable with the package:test test runner.

At some point I do want to expose more of test's internals as public APIs. But I'm not confident that it's stable enough to do that yet without producing a lot of breaking changes as we add new features.

jakemac53:
> If this is on the roadmap, I'm satisfied.

Yes, for this use case specifically I think we can just add a way of telling ddc about additional "entrypoints" that don't fit the default definition (files under `lib` but not `lib/src`, files under any other folder with a `main`). 

The main question is how we want that to actually be configured (pubspec.yaml or some other file?). It will be hard to walk back anything we do here which is why we punted on it for this first release. The implementation is trivial though once we decide on how we want to handle it.

> At some point I do want to expose more of test's internals as public APIs. But I'm not confident that it's stable enough to do that yet without producing a lot of breaking changes as we add new features.

What about marking it as `@experimental`? We do already have use cases today where we have copied this code, so that would at least be a bit better?

nex3:
> The main question is how we want that to actually be configured (pubspec.yaml or some other file?). It will be hard to walk back anything we do here which is why we punted on it for this first release. The implementation is trivial though once we decide on how we want to handle it.

I don't have an opinion on this at the moment, although I may develop one as I learn more about how DDC/Bazel work in general.

> What about marking it as `@experimental`? We do already have use cases today where we have copied this code, so that would at least be a bit better?

In my experience, no amount of documentation about a feature being unstable or experimental will keep people from relying on it and giving me hell when it changes.

jakemac53:
See also https://github.com/dart-lang/pub/issues/1606 which is related

jakemac53:
Ok, I did end up adding support for this as I guess there are some use cases with angular as well for importing files from src which aren't reachable from any entrypoint, see https://github.com/dart-lang/pub/pull/1608.

Once that is merged into an sdk release we can move these files under `src` if you would like.

PR #600, opened 2017-05-02
add dart:isolate import to bootstrap code, prepare for release

jakemac53:


jakemac53:
The tests were passing previously because in production mode I think it just dropped these types on the floor, and the file isn't analyzed because its just a string.

Issue #599, opened 2017-05-01
pkg/test VM tests flaky timeout on Travis


kevmoo:
But not locally.

grouma:
After some initial debugging it appears that the tests hang do to some unawaited futures. I intend to add a workaround until I can track down the root cause.

grouma:
Further flakiness is due to the fact that travis can execute the tests slowly which causes extra prints in the reports. I haven't determined a good fix for this yet.

grouma:
This was closed with https://github.com/dart-lang/test/pull/608

grouma:
Accidentally closed. This should remain open until the root cause is found.

grouma:
It appears the flaky test consistently resides in shard 3: https://travis-ci.org/dart-lang/test/builds/260833022

srawlins:
@grouma, you say "extra prints in the reports", can you elaborate?

At dart-lang/markdown#191, we're seeing a similar issue, and @kevmoo found that it can be worked around by using `--reporter expanded` instead of `--reporter compact`.

kevmoo:
Tiny delta. For markdown, the tests don't complete and we consistently get a `Unhandled exception:` output...

Not sure...

grouma:
The test runner will give command line updates every second or so. If the test took longer to run it would give additional output compared to what was expected. This caused tests to fail. I resolved the issue by updating tests to use more lenient matchers. For example: https://github.com/dart-lang/test/commit/d6cd857a9a7e5b026e999d63c967c94eb9914a92

PR #601, opened 2017-05-02
Explicitly exit after tests are run

grouma:


grouma:
Related to https://github.com/dart-lang/test/issues/599

natebosch:
Feels hacky - are we going to keep digging on the root cause?

grouma:
Yes. There are tons of unawaited futures in the package though so it may take a while to track down. Lets discuss when you are in the office on potential ways to track down the offending future.

Hixie:
FYI this patch broke Flutter because we have code that calls the test package's `main()` method and tries to do extra work (most notably collecting coverage) afterwards, and instead now our script just exits suddenly. We're pinning to an earlier version for now, but do you have a recommended alternative solution we should use?

natebosch:
@grouma I think I remember you making some progress on this - do you know if we'll be able to drop the explicit exit now?

grouma:
I had thought I resolved some flakiness but unfortunately couldn't pinpoint the root cause. I added a fix after this patch to only explicitly close on Travis. Update to the latest version and you shouldn't have issues.

natebosch:
> Update to the latest version and you shouldn't have issues.

Except on Travis...

Hixie:
Yeah, on Travis is where we ran into this.

grouma:
Spoke with Jacob and he explained the situation. Sorry for the confusion. My belief right now is that the flakiness comes only from the tests inside package:test. We have a couple levels of indirection in which we spin up a test runner which spins up another test runner etc. Things may have gotten better recently with some of the refactors and Natalie's changes. I'll do some experimentation and see if we can remove the hack all together. If not I'll use a custom flag which we'll set only in our Travis configuration. Expect a new version some time tomorrow.

PR #602, opened 2017-05-03
Lots of different cleanups

kevmoo:
Note we still have 3 failures – something is broken in 1.24


kevmoo:
> Note we still have 3 failures – something is broken in 1.24

Or maybe not. Anyway...

test/runner/compact_reporter_test.dart
```diff
@@ -490,6 +490,12 @@ $tests
       return line.substring(indentation);
     }).join("\n");
 
-    expect(actual, equals(expected));
+    // In Dart 1.24, stack traces with Future constructors output as
+    // `new Future` instead of `Future.Future`.
+    // Support running tests in both old and new styles.
```
natebosch: how long do we expect we'll keep backwards compatibility?

kevmoo: At least until 1.24 goes gold – and hopefully a bit longer. A release or two?

natebosch:
Did you intend to keep using the same branch/pull request for the other changes?

PR #609, opened 2017-05-08
Exit early on Travis

grouma:
Some internal coverage tools require the process to remain open. Hack to ensure Travis is not flaky while not impacting these internal tools.

### DIFF
--- a/lib/src/executable.dart
+++ b/lib/src/executable.dart
@@ -185,7 +185,9 @@ transformers:
 
   // TODO(grouma) - figure out why the executable can hang in the travis
   // environment. https://github.com/dart-lang/test/issues/599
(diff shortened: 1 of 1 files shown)


eseidelGoogle:
My understanding is that the addition of the exit() to package:test main() also broke flutter_tools ability to collect coverage, since previously we would do work after running tests:
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/test/runner.dart#L88
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/commands/test.dart#L204
Obviously this change makes it only break on travis. :)  But that still breaks our coverage reporting. :/

Issue #611, opened 2017-05-15
Analyzer errors with new Dart SDK (Dart VM version: 1.24.0-dev.4.0)


grouma:
error • Element 'StreamGroup' from SDK library 'async.dart' is implicitly hidden by 'stream_group.dart' at lib/src/runner.dart:215:12 • conflicting_dart_import
  error • Element 'StreamGroup' from SDK library 'async.dart' is implicitly hidden by 'stream_group.dart' at lib/src/runner/engine.dart:169:35 • conflicting_dart_import
  error • Element 'StreamGroup' from SDK library 'async.dart' is implicitly hidden by 'stream_group.dart' at lib/src/runner/loader.dart:93:12 • conflicting_dart_import

keertip:
@grouma , this was fixed with Lasse's cl  https://codereview.chromium.org/2872263003/

grouma:
This should now be resolved.

PR #612, opened 2017-05-15
Allow analyzer errors for dev SDK

grouma:
See https://github.com/dart-lang/test/issues/611 for more details


grouma:
This is no longer necessary.

Issue #615, opened 2017-05-15
Add a chain-stack-traces configuration field


nex3:
Everything that's configurable via the command line should also be configurable via [configuration files](https://github.com/dart-lang/test/blob/master/doc/configuration.md). #584 added a command-line flag for controlling whether stack traces are chained, but a corresponding configuration field is still necessary.

grouma:
This was closed with #616

PR #616, opened 2017-05-16
Add file configuration for chain-stack-traces

grouma:
Adds a configuration field for chain-stack-traces. See the following issue for more details.

https://github.com/dart-lang/test/issues/615

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -100,8 +100,10 @@ verbose_trace: true
 ### `chain_stack_traces`
 
 This boolean field controls whether or not stack traces are chained. 
-Disabling stack trace chaining will improve performance at the cost of
-debuggability. 
+Disabling [`stack trace chaining`][stack trace chaining] will improve
+performance for heavily asynchronous code at the cost of debuggability. 
+
+[stack trace chaining]: https://github.com/dart-lang/stack_trace/blob/master/README.md#stack-chains
 
 ```yaml
 chain_stack_traces: false 
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -27,6 +27,7 @@ tags:
 * [Test Configuration](#test-configuration)
   * [`timeout`](#timeout)
   * [`verbose_trace`](#verbose_trace)
+  * [`chain_stack_traces`](#chain_stack_traces)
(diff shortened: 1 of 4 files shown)


test/runner/configuration/top_level_test.dart
```diff
@@ -148,6 +148,39 @@ void main() {
...
+              throw "oh no";
+            });
+          }
+    """)
+        .create();
+
+    var test = runTest(["test.dart"]);
+    test.stdout.expect(containsInOrder([
+      "00:00 +0: failure",
+      "00:00 +0 -1: failure [E]",
```
nex3: This is technically a race condition—it's unlikely but possible that running this test will take longer than a second. Best to just omit the time for good measure, especially since you're already using `containsInOrder()`.

grouma: Good call. I actually had to fix some similar race conditions when providing support for Travis. See https://github.com/dart-lang/test/pull/608

doc/configuration.md
```diff
@@ -96,6 +97,16 @@ itself. It defaults to `false`.
 verbose_trace: true
 ```
 
+### `chain_stack_traces`
+
+This boolean field controls whether or not stack traces are chained. 
+Disabling stack trace chaining will improve performance at the cost of
+debuggability.
```
nex3: Consider linking [the stack chain docs](https://github.com/dart-lang/stack_trace/blob/master/README.md#stack-chains) here so people have a better idea of what debuggability they get.

doc/configuration.md
```diff
@@ -96,6 +97,16 @@ itself. It defaults to `false`.
 verbose_trace: true
 ```
 
+### `chain_stack_traces`
+
+This boolean field controls whether or not stack traces are chained. 
+Disabling stack trace chaining will improve performance at the cost of
```
nex3: "improve performance for heavily asynchronous code". I don't want people disabling stack chains in synchronous code for an imagined perf gain :stuck_out_tongue:.

grouma: Good call!

doc/configuration.md
grouma: Done

Issue #441, opened 2016-06-23
Add an option to retry a test


nex3:
Some tests are inherently flaky in ways that aren't feasible to fix by modifying the test itself. In order to support running in environments where flakiness is undesirable, such as CI servers, the user should be able to mark tests to be retried a certain number of times in case they fail.

I think the best way for this to work is to add a named `int retry` parameter to `test()` and `group()`, an `@Retry(int times)` metadata annotation, and a `retry: !int` test-level field to the configuration file. A test experiencing an error—that is, any non-`TestFailure` exception—would be re-run that many times before actually being marked as a failure.

We could add more detailed logic to the parameter to allow users to specify exactly which exception type(s) cause retries, but I expect retrying mostly to be specified in configuration files (e.g. saying "all browser tests get three retries when running on Travis"), where there's not a good cross-platform way of specifying Dart types.

nex3:
I believe since #632 landed this is largely done. One last piece which would be nice is a `--no-retry` flag and corresponding runner configuration, which would be useful for iterating locally on a test with retry configuration that's failing non-flakily.

grouma:
This was completed with: https://github.com/dart-lang/test/pull/633

PR #618, opened 2017-05-16
Test level retry

grouma:
Adds test level retry logic.

Partly addresses https://github.com/dart-lang/test/issues/441.

Group level and annotation support will follow.


lib/src/backend/metadata.dart
```diff
@@ -193,6 +200,7 @@ class Metadata {
         _skip = skip,
         _verboseTrace = verboseTrace,
         _chainStackTraces = chainStackTraces,
+        _retry = retry,
```
nex3: Add a [`RangeError.checkNotNegative()`](https://api.dartlang.org/stable/1.23.0/dart-core/RangeError/checkNotNegative.html) call to the body as well. Same for `new Metadata.parse()`.

lib/src/runner/engine.dart
```diff
@@ -340,8 +340,11 @@ class Engine {
   ///
   /// If [countSuccess] is `true` (the default), the test is put into [passed]
   /// if it succeeds. Otherwise, it's removed from [liveTests] entirely.
+  ///
+  /// [attempt] is used to determine if this test should be retried, defaults to
+  /// 1.
```
nex3: Maybe "The [attempt] argument is only passed when [_runLiveTest] calls itself recursively. It indicates the number of times the test has been run, including the current time. It defaults to 1."

lib/src/runner/engine.dart
```diff
@@ -340,8 +340,11 @@ class Engine {
...
   Future _runLiveTest(LiveSuiteController suiteController, LiveTest liveTest,
-      {bool countSuccess: true}) async {
+      {bool countSuccess: true, int attempt: 1}) async {
```
nex3: We generally avoid setting default values for non-boolean arguments, since it makes it difficult for callers to explicitly specify "the default value". Even in cases like this, where no callers are doing that, I like to use `??` in the body for consistency.

lib/src/runner/live_suite_controller.dart
```diff
@@ -106,8 +106,12 @@ class LiveSuiteController {
   /// is `true` (the default), the test is put into [passed] if it succeeds.
   /// Otherwise, it's removed from [liveTests] entirely.
   ///
+  /// Likewise, if [countFailure] is `true` (the default), the test is put into
+  /// [failed] if it fails. Otherwise, it's removed from [liveTests] entirely.
```
nex3: Another way to handle this would be to treat all retries as part of the same test, and just have that test not complete until all retries are complete—you'd probably handle that in the Invoker rather than the Engine.

If we decide to go with the `countFailures` approach, I think that means a breaking change to the JSON reporter protocol—it does have a notion of "hidden" tests currently, but the documentation guarantees that "only successful tests will be hidden" which won't be true after this. That indicates to me that a single-test approach would be less disruptive.

grouma: I played around with trying to move the logic into the Invoker but was unfortunately unsuccessful. Ensuring the errors are printed, dealing with the timeout, and correctly setting the state resulted in a lot of additional logic. I quite like the idea of simply cloning the test and reusing all the existing logic. That being said I understand your concerns so lets chat in person and see if the difficulty of moving the logic into the Invoker was simply a result of my lack of experience with this code base.

grouma:
I took a second crack at it and moved the logic over to the invoker. It was a lot more straightforward than my initial attempt. All tests pass; admittedly not many cover the retry branch and therefore I'm nervous I'm missing some edge conditions, especially around the timeout logic.

lib/src/backend/invoker.dart
```diff
@@ -348,6 +352,12 @@ class Invoker {
         await _outstandingCallbacks.noOutstandingCallbacks;
         if (_timeoutTimer != null) _timeoutTimer.cancel();
 
+        if (liveTest.state.result != Result.success &&
+            _runCount < liveTest.test.metadata.retry + 1) {
+          _onRun();
```
natebosch: have you explored what happens to the stack trace for these tests? I'm curious how things look when we go through the same Chain.capture() multiple tmes

grouma:
Yes @nex3 called it out in her initial review. It is dependent on which approach we take. I have a pending change with that functionality.

lib/src/backend/invoker.dart
nex3: `Chain.capture()` is a no-op if it's called within a capture zone.

nex3:
Does this work for `LoadSuite`s? If so, can you add a test?

grouma:
This doesn't work with LoadSuite tests. It'll require some extra logic.

lib/src/backend/metadata.dart
```diff
@@ -193,10 +200,12 @@ class Metadata {
...
         tags = new UnmodifiableSetView(tags == null ? new Set() : tags.toSet()),
         onPlatform =
             onPlatform == null ? const {} : new UnmodifiableMapView(onPlatform),
         forTag = forTag == null ? const {} : new UnmodifiableMapView(forTag) {
+    if (retry != null) RangeError.checkNotNegative(retry);
```
grouma: Done.

grouma:
@nex3 can you take a second look at this when you get a chance.

lib/src/backend/invoker.dart
```diff
@@ -348,6 +352,14 @@ class Invoker {
         await _outstandingCallbacks.noOutstandingCallbacks;
         if (_timeoutTimer != null) _timeoutTimer.cancel();
 
+        if (liveTest.state.result != Result.success &&
+            _runCount < liveTest.test.metadata.retry + 1) {
+          _controller
+              .message(new Message.retry("Retry: ${liveTest.test.name}"));
```
nex3: Are retry messages really worth a separate message type? We only have one for skips because we really want to print them in yellow text to match skipped tests.

lib/src/backend/metadata.dart
nex3: Looks like you may have missed an upload for this?

lib/src/runner/live_suite_controller.dart
```diff
@@ -128,6 +128,7 @@ class LiveSuiteController {
         _failed.add(liveTest);
       } else if (countSuccess) {
         _passed.add(liveTest);
+        _failed.remove(liveTest);
```
nex3: Add a comment here explaining why a failed test might start succeeding.

lib/src/runner/parse_metadata.dart
```diff
@@ -60,6 +60,7 @@ class _Parser {
     var skip;
     Map<PlatformSelector, Metadata> onPlatform;
     Set<String> tags;
+    var retry;
```
nex3: Nit: `int` (`skip` is `var` because it can be a string or a bool)

test/runner/runner_test.dart
```diff
@@ -292,6 +292,82 @@ $_usage""");
...
+            }, retry: 5);
+          }
+          """)
+          .create();
+
+      var test = runTest(["test.dart"]);
+      test.stdout.expect(consumeThrough(contains("+1: All tests passed!")));
+      test.shouldExit(0);
+    });
+  });
```
nex3: These should probably be in `retry_test.dart`.

test/runner/runner_test.dart
```diff
@@ -292,6 +292,82 @@ $_usage""");
...
+
+          import 'package:test/test.dart';
+
+          int attempt = 0;
+          void main() {
+            test("failure", () {
+            attempt++;
+            if (attempt != 2){
+             throw new TestFailure("oh no");
+            }
```
nex3: A simpler way to do this might be to print a message each time the test runs, and use stream matchers to assert that the message is only printed once.

test/runner/runner_test.dart
```diff
@@ -292,6 +292,82 @@ $_usage""");
...
+              "test.dart",
+              """
+          import 'dart:async';
+
+          import 'package:test/test.dart';
+
+          int attempt = 0;
+          void main() {
+            test("failure", () {
+             attempt++;
```
nex3: Nit: +2 indentation within `test()` (also below).

test/runner/runner_test.dart
```diff
@@ -292,6 +292,82 @@ $_usage""");
...
+          .file(
+              "test.dart",
+              """
+          import 'dart:async';
+
+          import 'package:test/test.dart';
+
+          int attempt = 0;
+          void main() {
+            test("failure", () {
```
nex3: This should probably have a different name, since it succeeds :wink:.

test/runner/timeout_test.dart
```diff
@@ -59,6 +59,36 @@ void main() {
...
+    runCount++;
+    if (runCount <=2) {
+      await new Future.delayed(new Duration(milliseconds: 1000));
+    }
+  }, retry: 3);
+}
+''')
+        .create();
+
+    var test = runTest(["--timeout=0.016x", "test.dart"]);
```
nex3: It's probably cleaner to set a time-based timeout here, rather than a factor-based one.

lib/src/backend/invoker.dart
grouma: My thought process is that we should print the retries in yellow as well. Most teams would likely want to reduce flakiness as much as possible, so having them stand out would be beneficial.

lib/src/backend/metadata.dart
grouma: Done here and elsewhere now :P

lib/src/runner/live_suite_controller.dart
grouma: Done.

lib/src/runner/parse_metadata.dart
grouma: Done.

test/runner/runner_test.dart
grouma: Done. Wish there was a way to specify inline code blocks that were formatted better :(

test/runner/runner_test.dart
grouma: Done.

test/runner/runner_test.dart
grouma: Noted. Gonna leave this as is though.

test/runner/timeout_test.dart
grouma: Done.

test/runner/runner_test.dart
grouma: I wasn't too sure on the distinction between runner test and the other various test file. I can see how runner_test can become a kitchen sink. I have moved these tests over.

lib/src/backend/invoker.dart
nex3: Unlike skips, there are valid reasons to mark a test as retried for the long term—for example, that's probably what we'll do for our browser tests, since browsers can sometimes just fail or time out arbitrarily. I don't want to force a super-visible message for those cases.

test/runner/runner_test.dart
nex3: Long ago there was a proposal for tagged template strings in Dart that could maybe work that way, but it never went anywhere :disappointed:. We could conceivably add a special case to the formatter to look for `"""// Dart` or something, but I doubt @munificent would go for that.

test/runner/runner_test.dart
nex3: Yeah, I've been trying to make more feature-specific files to avoid kitchen-sink-itis, but I haven't migrated all the old tests.

lib/src/backend/invoker.dart
grouma: Removed the special message type.

grouma:
Discussed offline. I have added logic to ignore errors propagated from previous test runs and have added a  corresponding test.

lib/src/backend/invoker.dart
```diff
@@ -357,7 +371,8 @@ class Invoker {
             // Use the invoker as a key so that multiple invokers can have different
             // outstanding callback counters at once.
             _counterKey: outstandingCallbacksForBody,
-            _closableKey: true
+            _closableKey: true,
+            #runCount: _runCount + 1
```
nex3: I think it's a little confusing that this gets incremented both here and in `new Future()`. Can we increment it once before we called `runZonedWithValue()`?

test/runner/retry_test.dart
```diff
@@ -0,0 +1,148 @@
...
+            "test.dart",
+            """
+          @Retry(3)
+
+          import 'dart:async';
+
+          import 'package:test/test.dart';
+
+
+          int attempt = 0;
```
nex3: Nit: `var attempt`

We generally don't type variables that have initializers, since they'll be inferred anyway. Also below.

test/runner/retry_test.dart
```diff
@@ -0,0 +1,148 @@
...
+              int attempt = 0;
+              Completer completer = new Completer();
+              void main() {
+                test("failure", () {
+                  attempt++;
+                  if (attempt == 1) {
+                    completer.future.then((_) => throw 'some error');
+                    throw new TestFailure("oh no");
+                  }
+                  completer.complete(null);
```
nex3: I think this should only be completed during the second attempt. It would also be good to wait a little bit to ensure that the `throw` is reached.

test/runner/retry_test.dart
```diff
@@ -0,0 +1,148 @@
...
+
+          import 'dart:async';
+
+          import 'package:test/test.dart';
+
+
+          int attempt = 0;
+          void main() {
+            test("failure", () {
+             attempt++;
```
nex3: Nit: +1 indent (also below).

If it helps, you can copy these to a file, format that with `--indent 6`, and then copy them back here.

test/runner/retry_test.dart
```diff
@@ -0,0 +1,148 @@
...
+                }, retry: 5);
+          }
+          """)
+          .create();
+
+      var test = runTest(["test.dart"]);
+      test.stdout.expect(consumeThrough(contains("+1: All tests passed!")));
+      test.shouldExit(0);
+    });
+  });
```
nex3: We should also test what happens if a retry test fails after it reports success. Something like

```dart
void main() {
  var completer1 = new Completer();
  var completer2 = new Completer();
  test("first", () {
    completer1.future.then((_) {
      completer2.complete();
      throw "oh no";
    });
  });

  test("second", () async {
    completer1.complete();
    await completer2.future;
  });
}
```

lib/src/backend/invoker.dart
grouma: Done.

test/runner/retry_test.dart
grouma: Done.

test/runner/retry_test.dart
grouma: Done.

test/runner/retry_test.dart
grouma: This will only get completed in the second attempt since the throw will happen in the first attempt. I have added an await to ensure the throw is reached.

test/runner/retry_test.dart
grouma: Done.

PR #620, opened 2017-05-17
Only run a single Travis check for each PR.

nex3:


natebosch:
Interesting. Should we be doing this in all of our repositories?

nex3:
> Interesting. Should we be doing this in all of our repositories?

For sure.

Issue #24, opened 2015-02-24
Add a doesNotComplete matcher


nex3:
`unittest` should provide a `doesNotComplete` matcher to use against `Future`s which asserts that they never fire and produces a useful error if they do.

"Never" is a tricky concept to capture, but doing something like the `pumpEventQueue()` function that's floating around should be good enough for most uses.

PR #623, opened 2017-05-19
Add Matcher for doesNotComplete

grouma:
This closes https://github.com/dart-lang/test/issues/24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 ## 0.12.21
 
+* Add a `doesNotComplete` matcher that asserts that a Future never completes.
+
 * `throwsA()` and all related matchers will now match functions that return
   `Future`s that emit exceptions.
 
 * Respect `onPlatform` for groups.
-
-* Expose new matcher `doesNotComplete`
 
 ## 0.12.20+13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 * Respect `onPlatform` for groups.
 
+* Expose new matcher `doesNotComplete`
+
 ## 0.12.20+13
 
 * Upgrade to package:matcher 0.12.1
--- a/lib/src/frontend/future_matchers.dart
+++ b/lib/src/frontend/future_matchers.dart
@@ -103,20 +103,16 @@ class _DoesNotComplete extends AsyncMatcher {
   }
 
   /*FutureOr<String>*/ matchAsync(item) {
-    if (item is! Future) return '$item is not a Future';
+    if (item is! Future) fail('$item is not a Future');
 
-    var value;
-    var isCompleted = false;
     item.then((value) {
       value = value;
-      isCompleted = true;
+      fail('$item completed with a value of $value');
     });
-    return _pumpEventQueue(_timesToPump)
-        .then((_) => isCompleted ? 'completed with a value of $value' : null);
+    return _pumpEventQueue(_timesToPump);
   }
 
   Description describe(Description description) {
-    description.add('does not complete');
     return description;
   }
 }
--- a/lib/src/frontend/future_matchers.dart
+++ b/lib/src/frontend/future_matchers.dart
@@ -114,7 +114,7 @@ class _DoesNotComplete extends AsyncMatcher {
   }
 
   /*FutureOr<String>*/ matchAsync(item) {
-    if (item is! Future) return "was not a Future";
+    if (item is! Future) return null;
 
     var value;
 
--- a/lib/src/frontend/future_matchers.dart
+++ b/lib/src/frontend/future_matchers.dart
@@ -89,12 +89,12 @@ class _Completes extends AsyncMatcher {
 /// the matcher will complete.
 ///
 /// This returns an [AsyncMatcher], so [expect] won't complete until
-/// [timesToPump] passes through the event queue.
+/// [_timesToPump] passes through the event queue.
 final Matcher doesNotComplete = const _DoesNotComplete(20);
 
 class _DoesNotComplete extends AsyncMatcher {
-  final int timesToPump;
-  const _DoesNotComplete(this.timesToPump);
+  final int _timesToPump;
+  const _DoesNotComplete(this._timesToPump);
 
   Future _pumpEventQueue(times) {
     if (times == 0) return new Future.value();
@@ -110,7 +110,7 @@ class _DoesNotComplete extends AsyncMatcher {
       value = value;
       isCompleted = true;
     });
-    return _pumpEventQueue(timesToPump)
+    return _pumpEventQueue(_timesToPump)
         .then((_) => isCompleted ? 'completed with a value of $value' : null);
   }
 
--- a/lib/src/frontend/future_matchers.dart
+++ b/lib/src/frontend/future_matchers.dart
@@ -92,18 +92,6 @@ class _Completes extends AsyncMatcher {
 /// [timesToPump] passes through the event queue.
 final Matcher doesNotComplete = const _DoesNotComplete(20);
 
-/// Matches a [Future] that does not complete.
-///
-/// Note that this creates an asynchronous expectation. The call to
-/// `expect()` that includes this will return immediately and execution will
-/// continue. Later, after [timesToPump]  passes through the event queue,
-/// the matcher will complete.
-///
-/// This returns an [AsyncMatcher], so [expect] won't complete until
-/// [timesToPump] passes through the event queue.
-Matcher doesNotCompleteAfter(int timesToPump) =>
-    new _DoesNotComplete(timesToPump);
-
 class _DoesNotComplete extends AsyncMatcher {
   final int timesToPump;
   const _DoesNotComplete(this.timesToPump);
--- a/lib/src/frontend/future_matchers.dart
+++ b/lib/src/frontend/future_matchers.dart
@@ -85,14 +85,10 @@ class _Completes extends AsyncMatcher {
 ///
 /// Note that this creates an asynchronous expectation. The call to
 /// `expect()` that includes this will return immediately and execution will
-/// continue. Later, after 20  passes through the event queue,
-/// the matcher will complete.
-///
-/// This returns an [AsyncMatcher], so [expect] won't complete until
-/// the provided number of passes through the event queue.
+/// continue.
 final Matcher doesNotComplete = const _DoesNotComplete(20);
 
-class _DoesNotComplete extends AsyncMatcher {
+class _DoesNotComplete extends Matcher {
   final int _timesToPump;
   const _DoesNotComplete(this._timesToPump);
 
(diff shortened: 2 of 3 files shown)


lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,57 @@ class _Completes extends AsyncMatcher {
...
+final Matcher doesNotComplete = const _DoesNotComplete(20);
+
+/// Matches a [Future] that does not complete.
+///
+/// Note that this creates an asynchronous expectation. The call to
+/// `expect()` that includes this will return immediately and execution will
+/// continue. Later, after [timesToPump]  passes through the event queue,
+/// the matcher will complete.
+///
+/// This returns an [AsyncMatcher], so [expect] won't complete until
```
natebosch: why doesn't the function signature tell us this?

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,57 @@ class _Completes extends AsyncMatcher {
...
+  final int timesToPump;
+  const _DoesNotComplete(this.timesToPump);
+
+  Future _pumpEventQueue(times) {
+    if (times == 0) return new Future.value();
+    return new Future(() => _pumpEventQueue(times - 1));
+  }
+
+  /*FutureOr<String>*/ matchAsync(item) {
+    if (item is! Future) return null;
```
natebosch: maybe add a comment "// non-Future instances cannot 'complete'"

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,57 @@ class _Completes extends AsyncMatcher {
...
+
+    var value;
+
+    return new Future(() async {
```
natebosch: `new Future(() async{` feels weird to me. What is our goal there? Do we need it to wait both an event loop (new Future) and a microtask (async)?

If we don't need to double up we can either:

```dart
return () async {
    ...
   await _pumpEventQueue(timeToPump);
  return notCompleted ? null : 'completed with a value of $value';
}();
```

or

```dart
var value;
var notCompleted = true;
item.then((value) {
  value = value;
  notCompleted = false;
});
return _pumpEventQueue(timeToPump)
    .then((_) => notCompleted ? null : 'completed with a value of $value');
```

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,57 @@ class _Completes extends AsyncMatcher {
...
+      var notCompleted = true;
```
natebosch: [nit] prefer positive boolean names

grouma: Done.

lib/src/frontend/future_matchers.dart
grouma: Done.

lib/src/frontend/future_matchers.dart
grouma: The other AsyncMatchers do this as well. I'm not sure to be honest. Left it for consistency and was intending to discuss this with @nex3

lib/src/frontend/future_matchers.dart
grouma: No need for the additional microtask. The interface expects a Future not a closure so the second approach is the correct thing to do. Fixed.

natebosch:
Oh - almost forgot. Add a changelog entry

lib/src/frontend/future_matchers.dart
nex3: We don't export `AsyncMatcher` because I don't want to commit to the API, especially since I'm planning a substantial refactor to the upstream `matcher` API. I don't want to declare the return type as something we don't export.

The documentation comments for async matchers should probably be updated to avoid mentioning `AsyncMatcher` directly, though.

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,54 @@ class _Completes extends AsyncMatcher {
...
+/// Matches a [Future] that does not complete.
+///
+/// Note that this creates an asynchronous expectation. The call to
+/// `expect()` that includes this will return immediately and execution will
+/// continue. Later, after [timesToPump]  passes through the event queue,
+/// the matcher will complete.
+///
+/// This returns an [AsyncMatcher], so [expect] won't complete until
+/// [timesToPump] passes through the event queue.
+Matcher doesNotCompleteAfter(int timesToPump) =>
```
nex3: I'm not sure it's worth making this a separate function. I think it's pretty unlikely that users will want to have fine-grained control over how long the test runs, and if they do they can just write `expect(pumpEventQueue(n), completes)`.

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,54 @@ class _Completes extends AsyncMatcher {
...
+    new _DoesNotComplete(timesToPump);
+
+class _DoesNotComplete extends AsyncMatcher {
+  final int timesToPump;
```
nex3: Style thing: I'd personally make this private. It doesn't *really* matter, since the class itself is private, but whether a member is private is still a good signal for people reading the code as to whether it's accessed outside of the declaring class.

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,54 @@ class _Completes extends AsyncMatcher {
...
+  const _DoesNotComplete(this.timesToPump);
+
+  Future _pumpEventQueue(times) {
```
nex3: Can you add a TODO to make this a public, top-level function?

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,54 @@ class _Completes extends AsyncMatcher {
...
+    if (times == 0) return new Future.value();
+    return new Future(() => _pumpEventQueue(times - 1));
+  }
+
+  /*FutureOr<String>*/ matchAsync(item) {
+    if (item is! Future) return null; // non-Future instances cannot 'complete'
```
nex3: I think non-futures should be an error here—it's probably a mistake if you pass one in.

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,54 @@ class _Completes extends AsyncMatcher {
...
+
+    var value;
+    var isCompleted = false;
+    item.then((value) {
+      value = value;
+      isCompleted = true;
+    });
```
nex3: This should produce an error even if the future completes after `_pumpEventQueue()`.

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,43 @@ class _Completes extends AsyncMatcher {
...
+
+/// Matches a [Future] that does not complete.
+///
+/// Note that this creates an asynchronous expectation. The call to
+/// `expect()` that includes this will return immediately and execution will
+/// continue. Later, after 20  passes through the event queue,
+/// the matcher will complete.
+///
+/// This returns an [AsyncMatcher], so [expect] won't complete until
+/// [_timesToPump] passes through the event queue.
```
natebosch: Is `_timesToPump` in scope here?

lib/src/frontend/future_matchers.dart
grouma: Done

lib/src/frontend/future_matchers.dart
grouma: It is not. That's what I get for find replace. Updated the comment.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 
 * Respect `onPlatform` for groups.
 
+* Expose new matcher `doesNotComplete`
```
nex3: Maybe "Add a `doesNotComplete` matcher that asserts that a Future never completes."? It's important to describe what new features do, not just that they exist.

Also, I'd move this above the previous changelog entry; I generally list features before bug fixes, because more people care about new features.

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,54 @@ class _Completes extends AsyncMatcher {
...
+
+/// Matches a [Future] that does not complete.
+///
+/// Note that this creates an asynchronous expectation. The call to
+/// `expect()` that includes this will return immediately and execution will
+/// continue. Later, after 20  passes through the event queue,
+/// the matcher will complete.
+///
+/// This returns an [AsyncMatcher], so [expect] won't complete until
+/// [timesToPump] passes through the event queue.
```
nex3: There are other ways to keep the test open. The underlying mechanism is [`Invoker.add/removeOutstandingCallback()`](https://github.com/dart-lang/test/blob/master/lib/src/backend/invoker.dart#L150-L170), but in this case it might be simpler to just run `expect(_pumpEventQueue(), completes)` within the matcher body.

grouma: Oh neat. Changed.

CHANGELOG.md
grouma: Done.

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,37 @@ class _Completes extends AsyncMatcher {
...
+  }
+
+  Description describe(Description description) {
+    description.add("does not complete");
+    return description;
+  }
+
+  @override
+  bool matches(item, Map matchState) {
+    if (item is! Future) fail('$item is not a Future');
```
nex3: This should return false and describe the error in `describeMismatch()`.

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,37 @@ class _Completes extends AsyncMatcher {
...
+    item.then((value) {
+      value = value;
+      fail('$item completed with a value of $value');
```
nex3: It's very unlikely that the `item.toString()` will contain anything useful. Maybe just `"Future was not expected to complete, but it completed to $value."`.

grouma: Yeah I thought about that since it would mostly be <Instance of _Future> but opted to keep it just in case it was useful. I've removed it though.

lib/src/frontend/future_matchers.dart
grouma: Done.

lib/src/frontend/future_matchers.dart
```diff
@@ -80,3 +80,44 @@ class _Completes extends AsyncMatcher {
...
+  Description describe(Description description) {
+    description.add("does not complete");
+    return description;
+  }
+
+  @override
+  bool matches(item, Map matchState) {
+    if (item is! Future) return false;
+    item.then((value) {
+      value = value;
```
nex3: Wait what is this line doing?

grouma: Artifact from AsyncMatcher approach.

PR #625, opened 2017-05-23
Enable browser tests on Travis.

nex3:
As far as I can tell, all tests are passing other than the few that are skipped due to dart-lang/sdk#29693 and an as-yet-unidentified issue with signal-handling for browsers that seems to only appear on Travis. It's possible that there will be some flakes, though; if that seems to happen frequently, we may want to delay landing this until #618 is done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: dart
-sudo: false
+sudo: required
+
+# We require PhantomJS 2.x, which is only available on Trusty.
+dist: trusty
 
 dart:
  - dev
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,16 +1,42 @@
...
+    - ttf-kochi-gothic
+    - ttf-dejavu
+    - ttf-indic-fonts
+    - fonts-tlwg-garuda
+
+before_install:
+  # TODO: remove this and use "sudo: false" when travis-ci/travis-ci#4714 is
+  # fixed.
+  - sudo apt-get update -yq
+  - sudo sh -c "echo ttf-mscorefonts-installer msttcorefonts/accepted-mscorefonts-eula select true | debconf-set-selections"
```
grouma: What browser needs this? I thought for Dartium to work we simply needed mincho and gothic on top of a trusty install?

natebosch:
"The job exceeded the maximum time limit for jobs, and has been terminated."

Do we know which tests are causing the timeout?

grouma:
Looks like you are running into similar issues I did. Take a look at the performance of the jobs. For some odd reason the browser tests are taking 45+ minutes as opposed to ~15. One way to get around this is to break up the browsers into several tasks to better parallelize them. Ideally though we figure out why things are so much slower in the Travis environment though.

.travis.yml
nex3: Content shell. I'll add a comment.

nex3:
Looks like there's one browser flake, which is impressive across three different runs of the entire test suite, but still more than I'd like. I'm putting this on hold until we have retry support.

grouma:
I think that makes sense. I'll try to address the retry issues today.

kevmoo:
Woo hoo! Green baby! https://travis-ci.org/dart-lang/test/builds/240195419

grouma:
Kind of sucks that the browser tests take ~20 minutes to run. May be worth breaking each browser out into a separate task.

kevmoo:
Depends how much of that is due to the browser vs dart2js time...

.travis.yml
```diff
@@ -1,24 +1,62 @@
...
+
 dart_task:
+ # Browser tests take particularly long on Dartium, so we split them up into different tasks.
  - test: --platform dartium
    install_dartium: true
- - test: --platform vm -x content-shell -x chrome -x firefox -x dartium -x phantomjs --timeout 4x
- - dartfmt
- - dartanalyzer
+ - test: --platform vm -x browser
+ - test: --platform vm -t content-shell -x chrome -x firefox -x phantomjs -x dartium --timeout 4x
```
kevmoo: Should this be limited to just stable – since that's the only platform we're installing?

kevmoo: ...Er, I guess this is testing the vm side of the integration – so it only matters that content_shell is usable...right?

nex3: That's right.

kevmoo:
Oops - https://travis-ci.org/dart-lang/test/jobs/240492348#L798

kevmoo:
Looking good!

nex3:
It looks like the throughput for four and five shards is close enough to be noise, whereas six shards takes slightly longer. I think we should go with five shards; I suspect as we increase the number of tests, more sharding will become more valuable, and this gives us a head start.

kevmoo:
SGTM

On Wed, Jun 7, 2017 at 2:43 PM, Natalie Weizenbaum <notifications@github.com

PR #627, opened 2017-06-01
Only print browser errors once per browser.

### LINKED ISSUE #621: Validate browsers are installed [on travis, etc]
As mentioned on https://github.com/dart-lang/test/issues/619, we don't always have guarantees that all browsers being run by `pub run test` always are installed on the current machine. For example, on a non-OSX machine there will never be `safari`, and by default some browsers are not installed on environments like travis.

Suggestions:

- [ ] Collect a list of all browsers being executed in a given session

(i.e. from `dart_test.yaml` or `pub run test -p ...`)

- [ ] Validate all browsers are installed, failing early if one or more are missing and/or
- [ ] Skip all browsers that are not installed

nex3:
Closes #621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 * Respect `onPlatform` for groups.
 
+* Only print browser load errors once per browser.
+
 ## 0.12.20+13
 
 * Upgrade to package:matcher 0.12.1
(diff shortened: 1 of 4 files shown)


matanlurey:
Patched this down, huge improvement when you are missing a browser. Thanks!

(Will defer to @grouma for LGTM)

PR #628, opened 2017-06-01
Add timeout for deserializing a test suite

grouma:
In Dartium and Content-Shell, run time errors can cause the test to hang as we wait to deserialize the test suite. Add a timeout to prevent this. 

Ideally we would propagate any console errors but I have been unsuccessful in doing so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 * Respect `onPlatform` for groups.
 
+* Gracefully timeout when attempting to deserialize a test suite.
+
 ## 0.12.20+13
 
 * Upgrade to package:matcher 0.12.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 * Only print browser load errors once per browser.
 
-* Gracefully timeout when attempting to deserialize a test suite.
+* Gracefully time out when attempting to deserialize a test suite.
 
 ## 0.12.20+13
 
--- a/lib/src/runner/plugin/platform_helpers.dart
+++ b/lib/src/runner/plugin/platform_helpers.dart
@@ -24,7 +24,7 @@ import '../runner_test.dart';
 
 typedef StackTrace _MapTrace(StackTrace trace);
 
-Duration _deserializeTimeout = new Duration(seconds: 30);
+Duration _deserializeTimeout = new Duration(seconds: 45);
 
 /// A helper method for creating a [RunnerSuiteController] containing tests
 /// that communicate over [channel].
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 
 * Only print browser load errors once per browser.
 
+* Gracefully timeout when attempting to deserialize a test suite.
```
nex3: Nit: "time out". It's a noun without a space, and a verb with one.

giphy 1

lib/src/runner/plugin/platform_helpers.dart
```diff
@@ -23,6 +24,8 @@ import '../runner_test.dart';
 
 typedef StackTrace _MapTrace(StackTrace trace);
 
+Duration _deserializeTimeout = new Duration(seconds: 45);
```
nex3: Make this final and remove the type annotation.

lib/src/runner/plugin/platform_helpers.dart
```diff
@@ -108,7 +111,12 @@ Future<RunnerSuiteController> deserializeSuite(
       });
 
   return new RunnerSuiteController(
-      environment, suiteConfig, await completer.future,
+      environment,
+      suiteConfig,
+      await completer.future.timeout(_deserializeTimeout, onTimeout: () {
+        throw new ApplicationException(
+            "Timed out trying to deserialize the test suite.");
```
nex3: I don't think "deserialize" is going to make a lot of sense to our users here. Maybe

```
"Timed out while loading the test suite.\n"
"It's likely that there's a missing import or syntax error."
```

test/runner/browser/runner_test.dart
```diff
@@ -25,6 +25,16 @@ void main() {
...
 
+final _importFailure = """
+import 'package:test/test.dart';
+import 'blah' as bad_library;
+
+void main() {
+  bad_library.main();
+  test("failure", () => throw new TestFailure("oh no"));
+}
+""";
```
nex3: Since this is only used in one test, just include it inline.

Also, add a comment mentioning why bad imports cause this error.

CHANGELOG.md
grouma: TIL.

lib/src/runner/plugin/platform_helpers.dart
grouma: Done.

lib/src/runner/plugin/platform_helpers.dart
grouma: Done. I always have difficulty coming up with the correct verbiage. Thanks!

test/runner/browser/runner_test.dart
grouma: Good call. Done.

grouma:
Should this timeout be configurable or based off of the test runner timeout? It can be flaky in Travis given how slow things run there.

nex3:
To answer your question about configuration, I'd prefer to have a conservative default if we can than to expose more configuration surface area. If 45s turns out to be too long for Travis, we can adjust it upwards—after all, this is just a last-measure fallback.

PR #632, opened 2017-06-06
Add retry configuration

grouma:


doc/configuration.md
```diff
@@ -139,6 +140,16 @@ tags:
...
 [global configuration file](#global-configuration).
 
+### `retry`
+
+This int field controls how many times a test is retried upon failure. 
+
+```yaml
+tags:
+  chrome:
+    retry: 3 # retry chrome failures 3 times
```
nex3: Nit: Capitalization and period.

doc/configuration.md
```diff
@@ -31,6 +31,7 @@ tags:
   * [`js_trace`](#js_trace)
   * [`skip`](#skip)
   * [`test_on`](#test_on)
+  & [`retry`](#retry)
```
nex3: `*`

doc/configuration.md
```diff
@@ -139,6 +140,16 @@ tags:
 This field is not supported in the
 [global configuration file](#global-configuration).
 
+### `retry`
```
nex3: The order of sections should match the order in the table of contents.

doc/configuration.md
```diff
@@ -139,6 +140,16 @@ tags:
...
+
+This int field controls how many times a test is retried upon failure. 
+
+```yaml
+tags:
+  chrome:
+    retry: 3 # retry chrome failures 3 times
+```
+
```
nex3: Mention that this isn't supported in the global config file.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: test
-version: 0.12.21
+version: 0.12.21-dev
```
nex3: 0.12.22-dev. We want this version to be greater than  the last released version, so if people depend on it via git the versioning semantics are correct.

test/runner/configuration/top_level_error_test.dart
```diff
@@ -45,6 +45,14 @@ void main() {
     test.shouldExit(exit_codes.data);
   });
 
+  test("rejects an invalid retry", () {
+    d.file("dart_test.yaml", JSON.encode({"retry": "flup"})).create();
+
+    var test = runTest(["test.dart"]);
+    test.stderr.expect(containsInOrder(["retry must be an int", "^^^^^^"]));
+    test.shouldExit(exit_codes.data);
+  });
```
nex3: Also test that it rejects negative numbers.

pubspec.yaml
grouma: Ugh. Lots of silly mistakes. I promise I'll learn the Github workflow soon enough!

test/runner/configuration/top_level_error_test.dart
grouma: Done.

test/runner/configuration/top_level_error_test.dart
```diff
@@ -53,6 +53,15 @@ void main() {
     test.shouldExit(exit_codes.data);
   });
 
+  test("rejects an negative retry values", () {
+    d.file("dart_test.yaml", JSON.encode({"retry": -1})).create();
+
+    var test = runTest(["test.dart"]);
+    test.stderr
+        .expect(containsInOrder(["Not greater than or equal to 0"]));
```
nex3: Test that the error contains "^^" too, so we can be reasonably sure the error is pointing to the location in the YAML file. You may need to add some logic to the parser to eagerly verify the bounds.

doc/configuration.md
nex3: Global config is intended to allow users to configure things that may be system-wide—for example, their computer may be particularly slow, so they could set `timeout: 1.5x` globally. Or they may personally prefer verbose stack traces, so they could set `verbose_trace: true` globally.

But we don't want to let their global configuration make tests run differently on their system; that would be a recipe for painful bugs. The line between those two is subjective, but broadly we try to only allow global configuration that has to do with their system as a whole or their personal preferences.

test/runner/configuration/top_level_error_test.dart
grouma: I have added the additional check. I opted to go with a disjoint helper instead of overriding the existing _getInt function.

PR #638, opened 2017-06-09
Trim Travis tests

grouma:
Dartium tests are included in the sharded tests so explicitly running with platform Dartium should be redundant. Also remove 1.22.1 given the SDK release today.

We should have a discussion around which versions of the SDK we would like to run the browser tests. I'm under the opinion that we lose some confidence and only run browser tests for the stable SDK. I'm not a fan of the very long Travis runs. It also negatively impacts other teams as we starve Travis of resources.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ dart:
  - dev
  - stable
  - 1.23.0
- - 1.22.1
 
 # Content shell needs these fonts.
 addons:
@@ -39,11 +38,7 @@ before_install:
   - ln -s `pwd`/`echo drt-linux-*`/content_shell bin/content_shell
 
 dart_task:
- # Browser tests take particularly long on Dartium, so we split them up into different tasks.
- - test: --platform dartium
-   install_dartium: true
-
- # Split the tests into four shards to help parallelize them across Travis workers.
+ # Split the tests into five shards to help parallelize them across Travis workers.
  - test: --preset travis --total-shards 5 --shard-index 0
    install_dartium: true
  - test: --preset travis --total-shards 5 --shard-index 1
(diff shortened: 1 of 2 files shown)


nex3:
`--platform dartium` isn't the same as running tests that exercise our Dartium support. It ensures that the tests that don't depend on `dart:io` (mostly frontend and backend stuff) pass in a browser context.

I'm nervous about declaring that we support SDK versions when we don't run tests against those SDKs. I think for most smaller packages that's probably fine, but `test` changes quickly *and* is widely-used. If we decide to only test on the last-but-one version of the SDK, we should declare in the pubspec that we only support that SDK.

kevmoo:
> --platform dartium isn't the same as running tests that exercise our Dartium support. It ensures that the tests that don't depend on dart:io (mostly frontend and backend stuff) pass in a browser context.

Ah! So the sharded runs are only VM tests (effectively).

So we're not running the pure browser tests against firefox, etc. K.

> I'm nervous about declaring that we support SDK versions when we don't run tests against those SDKs.

Understood. I guess this is the price of good testing.

Hrm...we could always see if about getting another 5 instances on Travis...

nex3:
> Hrm...we could always see if about getting another 5 instances on Travis...

This is probably worth the money, now that we're switching so many more packages over.

grouma:
+1 For getting more Travis instances. Money well spent in my opinion.

As for this pull, I'll add back in the platform Dartium test but we should keep the removal of 1.22.1 correct?

nex3:
I'm okay with removing 1.22, but if we do that please also update the SDK constraint so we only declare compatibility with SDK versions we test against.

kevmoo:
Sure. 1.23 has been out for 2 months. We have a VERY quick upgrade path.

On Fri, Jun 9, 2017 at 3:39 PM, Natalie Weizenbaum <notifications@github.com

PR #639, opened 2017-06-10
Fix flakiness 

grouma:
I believe this will resolve our flakiness issue. We can get into a condition where the browser is unable to startup. We then complete the onExit with an ApplicationException. Since we swallow all errors, the onExit never completes which keeps the test runner open.

### DIFF
--- a/lib/src/runner/browser/browser.dart
+++ b/lib/src/runner/browser/browser.dart
@@ -126,6 +126,8 @@ abstract class Browser {
     });
 
     // Swallow exceptions. The user should explicitly use [onExit] for these.
-    return onExit.catchError((_) {});
+    // However, surface application exceptions which are thrown due to startup
+    // issues with the browser.
+    return onExit.catchError((_) {}, test: (e) => e is! ApplicationException);
   }
 }
--- a/lib/src/executable.dart
+++ b/lib/src/executable.dart
@@ -182,12 +182,6 @@ transformers:
   } finally {
     await close();
   }
-
-  // TODO(grouma) - figure out why the executable can hang in the travis
-  // environment. https://github.com/dart-lang/test/issues/599
-  if (Platform.environment["TRAVIS"] == "true") {
-    exit(exitCode);
-  }
 }
 
 /// Print usage information for this command.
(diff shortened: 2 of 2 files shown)


grouma:
Looks like Travis is still flaky so adding back in the corresponding hack. This fix still resolves a certain reproducible issue on my Mac. Hopefully it resolves https://github.com/dart-lang/angular_test/issues/66

lib/src/runner/browser/browser.dart
```diff
@@ -126,6 +126,8 @@ abstract class Browser {
     });
 
     // Swallow exceptions. The user should explicitly use [onExit] for these.
```
kevmoo: @nex3 who/what is "The user" in this case?

nex3: "The user" is the code that uses the `Browser` subclass.

lib/src/runner/browser/browser.dart
```diff
@@ -126,6 +126,8 @@ abstract class Browser {
...
-    return onExit.catchError((_) {});
+    // However, surface application exceptions which are thrown due to startup
+    // issues with the browser.
+    return onExit.catchError((_) {}, test: (e) => e is! ApplicationException);
```
nex3: I don't think this is the right place to handle this error. As the comment indicates, code that uses `Browser` is expected to handle errors itself through the `onExit` getter. That suggests that it's that code's responsibility to add special handling for application exceptions.

I believe the right place to address this is in `BrowserManager.start()`.

grouma: Spoke offline. The close errors happen in a different zone than what is seen in the BrowserManager. @nex3 Will investigate potential solutions.

nex3: > The close errors happen in a different zone than what is seen in the BrowserManager.

I don't think that's true. There was a zone difference in the test you were running that interacted with `Browser` on its own, but that was specific to the test environment you were using—I think it's unrelated to the bug that's causing flakiness on Travis. If you want to test that, you'll probably need to invoke `test` as an executable that loads a browser.

grouma: Indeed. What I intended to say is that the issue raised in this pull request is indeed an actual issue which you'll be looking at. I will close this pull until I can get a better handle on what is causing the flakiness.

PR #640, opened 2017-06-12
Fix for content shell on Mac

grouma:
The latest version of Content-Shell on Mac has a bug in which error messages do not properly propagate. If we see evidence of this issue, we then must assume Content-Shell started up correctly. Also, as an additional safety measure, provide a timeout while we try to read the error stream. It is assumed that if we did not find a critical error while waiting on the error stream, Content-Shell started successfully.

This will resolve: https://github.com/dart-lang/angular_test/issues/63

### DIFF
--- a/lib/src/runner/browser/content_shell.dart
+++ b/lib/src/runner/browser/content_shell.dart
@@ -60,10 +60,12 @@ class ContentShell extends Browser {
         // make sure it's not expired and that the remote debugging port worked.
         // Any errors from this will always come before the "Running without
         // renderer sandbox" message.
-        while (await stderr
-                .moveNext()
-                .timeout(_errorTimeout)
-                .catchError((_) => false) &&
+        while (await stderr.moveNext().timeout(_errorTimeout).catchError((_) =>
+                throw new ApplicationException(
+                    "Error starting up content_shell.\n"
+                    "Ensure you are using the latest version:\n"
+                    "http://gsdview.appspot.com/dart-archive/channels/stable/"
+                    "release/latest/dartium/")) &&
             !stderr.current.endsWith("Running without renderer sandbox") &&
             !stderr.current.contains("Running without the SUID sandbox") &&
             // Error messages on Mac can get gobbled, we must assume that it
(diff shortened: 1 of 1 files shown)


lib/src/runner/browser/content_shell.dart
```diff
@@ -59,9 +60,16 @@ class ContentShell extends Browser {
...
-        while (await stderr.moveNext() &&
+        while (await stderr
+                .moveNext()
+                .timeout(_errorTimeout)
+                .catchError((_) => false) &&
             !stderr.current.endsWith("Running without renderer sandbox") &&
-            !stderr.current.contains("Running without the SUID sandbox")) {
+            !stderr.current.contains("Running without the SUID sandbox") &&
+            // Error messages on Mac can get gobbled, we must assume that it
+            // it started up successfully.
```
nex3: Content shell starts up successfully without producing *any* output on standard error? Do certain versions do this reliably, or does it happen at random? In cases where this happens, how do we tell if `bind()` failed?

Assuming the answers are "yes", "at random", and "we can't", I guess we'll just work around it as best we can and hope we can remove content shell support soon. In that case, we probably want a shorter timeout—10s should be totally fine even on very slow systems.

grouma: It doesn't produce any meaningful output other than the 'kqueue' message above. As far as I can tell, content-shell 50 reliably does this on Mac only. However, other versions may also have this issue. When it does happen, unfortunately we can't tell if the bind was successful.

I've reduced the timeout. Looking forward to deprecating content-shell.

nex3: If it does report a "kqueue" message, can we just use that as a signal to stop listening rather than having a timeout?

Have you verified that it doesn't print anything if you're binding to a port that's definitely already allocated?

grouma: Discussed offline. "kqueue" is the only type of message we receive for content-shell version 50 on Mac. This creates various issues. Adding timeout logic with an application exception encouraging users to make sure they are using the most recent version of content-shell.

PR #642, opened 2017-06-13
add dartdevc tests to the pub serve tests

jakemac53:
Added dartdevc to most of the tests.

Note that I didn't add it to the stack trace tests as those don't seem to pass currently, we can talk offline about what the expectation is for those.

### DIFF
--- a/test/runner/pub_serve_test.dart
+++ b/test/runner/pub_serve_test.dart
@@ -97,8 +97,8 @@ void main() {
       await pub.kill();
     });
 
-    test("runs those tests on Chrome", () async {
-      var pub = await runPubServe();
+    testWithCompiler("runs those tests on Chrome", (compiler) async {
+      var pub = await runPubServe(args: ['--web-compiler', compiler]);
       var test = await runTest([_pubServeArg, '-p', 'chrome']);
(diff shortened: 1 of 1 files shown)


nex3:
Looks like this is breaking tests.

You'll need to add logic to disable the DDC tests for older Dart versions—I'd suggest returning early if `Platform.version` is lower than 1.24.

jakemac53:
Ya, looks like its also broken in dev currently due to `--unsafe-angular2-whitelist` being removed. I think I will just wait on the next dev release (which should contain a fix for that) to merge this.

I also added the version check to only run dartdevc on >=1.24.0

test/runner/pub_serve_test.dart
```diff
@@ -343,3 +344,23 @@ void main() {
     await test.shouldExit(1);
   });
 }
+
+/// The list of supported compilers for the current [Platform.version].
+final Iterable<String> _compilers = () {
+  var compilers = ['dart2js'];
+  var sdkVersion = new Version.parse(
+      Platform.version.substring(0, Platform.version.indexOf(' ')));
```
nex3: This would probably be cleaner with `String.split()`.

PR #643, opened 2017-06-13
Run tear-downs in the same error zone as tests.

nex3:
We used to create a new error zone for each call to
waitForOutstandingCallbacks(), which meant that asynchronous errors
couldn't be safely passed around within a single test. Now we have a
single error handler that manually checks the zone in which the error
was thrown.

### DIFF
--- a/test/backend/declarer_test.dart
+++ b/test/backend/declarer_test.dart
@@ -283,6 +283,22 @@ void main() {
 
       await _runTest(tests.single, shouldFail: true);
     });
+
+    test("runs in the same error zone as the test", () {
+      return expectTestsPass(() {
+        Future future;
+        tearDown(() {
+          // If the tear-down is in a different error zone than the test, the
+          // error will try to cross the zone boundary and get top-leveled.
+          expect(future, throwsA("oh no"));
+        });
+
+        test("test", () {
+          future = new Future.error("oh no");
+          expect(future, throwsA("oh no"));
+        });
+      });
+    });
   });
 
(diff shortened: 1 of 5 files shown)


test/backend/declarer_test.dart
```diff
@@ -283,6 +283,22 @@ void main() {
...
+      return expectTestsPass(() {
+        Future future;
+        tearDown(() {
+          // If the tear-down is in a different error zone than the test, the
+          // error will try to cross the zone boundary and get top-leveled.
+          expect(future, throwsA("oh no"));
+        });
+
+        test("test", () {
+          future = new Future.error("oh no");
```
grouma: What happens if this was defined outside the test?

nex3: It will be top-leveled and cause the load suite to fail. There's no way around that, since we have to run each test in its own error zone.

grouma: Noted. Thanks for this fix! Looks very clean to me.

lrhn: Just curiosity: Why do you need to run each test in its own error zone?
Is it to prevent errors leaking between tests? Otherwise, you could potentially run everything in the same error-zone, and then multiplex the handling depending on a zone variable. Each test could still have its own handler, it's just registered on the global test error handler, not on the individual zone.

nex3: We want to make sure that no errors escape the tests that produced them, since that provides important context for tracking them down and debugging.

PR #645, opened 2017-06-14
Enable and fix await-only-futures lint

kevmoo:


kevmoo:
The SDK constraint change will be subsumed by https://github.com/dart-lang/test/pull/638

pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: test
-version: 0.12.22
+version: 0.12.23-dev
 author: Dart Team <misc@dartlang.org>
 description: A library for writing dart unit tests.
 homepage: https://github.com/dart-lang/test
 environment:
-  sdk: '>=1.14.0 <2.0.0'
+  sdk: '>=1.22.1 <2.0.0'
```
grouma: Should we just wait for https://github.com/dart-lang/test/pull/638 to go through? It'll be less tests to run too :D

kevmoo: Sure – just hurry up! 😁

grouma: Waiting for @nex3 to provide the LGTM.

kevmoo:
> I'm wondering if we should add a lint for unawaited futures as well.

@grouma that lint exists, but there are a LOT of false positives in pkg/test

You can exclude these one-by-one with line-level comments, thought.

Worth looking into

kevmoo:
@nex3 want me should wait for Travis, or we feel okay about this?

nex3:
I think we're fine to just land it, since Travis passed before the merge.

Issue #416, opened 2016-04-04
Add the ability to customize which packages' frames are folded


nex3:
Right now, we have some hard-coded logic for folding stack chains that come from test errors. We fold core library frames and frames from the `test`, `stack_trace`, and `stream_channel` packages. We should make this set of packages customizable, at least by allowing users to add to it. The config for this could look something like:

``` yaml
folded_packages:
- pool
- async
```

I'm not super happy with that name—it only makes sense to people who know that "folding" is a thing that can be done to stack traces. But I also don't want to make it too verbose, and I'm having trouble coming up with a better alternative.

kevmoo:
A thought: this will rarely be configured. Verbose is not a huge issue for something rarely used.

For rarely used features, discoverability is more important IMHO

nex3:
@kevmoo Do you have a name suggestion?

kevmoo:
`stack_trace_folded_packages` ?

nex3:
I think the following would be a good format for this:

```yaml
fold_stack_frames:
  only:
  - pool
  - async

# or
fold_stack_frames:
  except:
  - pool
  - async
```

PR #650, opened 2017-06-16
Configurable package folding

grouma:
- Allow StackTraceMapper to be serialized and deserialized
- Send StackTraceMapper to remote isolate
- Move all test chaining to the remote isolate
- Enable new configurations for stack trace folding
- Add corresponding tests

This will close https://github.com/dart-lang/test/issues/416

### DIFF
--- a/lib/src/frontend/test_chain.dart
+++ b/lib/src/frontend/test_chain.dart
@@ -39,4 +39,4 @@ Chain terseChain(StackTrace stackTrace, {bool verbose: false}) {
 
 /// Converts [stackTrace] to a [Chain] following the test's configuration.
 Chain testChain(StackTrace stackTrace) => terseChain(stackTrace,
-    verbose: Invoker.current.liveTest.test.metadata.verboseTrace);
+    verbose: Invoker.current?.liveTest?.test?.metadata?.verboseTrace ?? true);
--- a/lib/src/frontend/test_chain.dart
+++ b/lib/src/frontend/test_chain.dart
@@ -1,3 +1,7 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
 import 'package:stack_trace/stack_trace.dart';
 
 typedef StackTrace Mapper(StackTrace trace);
--- a/lib/src/frontend/test_chain.dart
+++ b/lib/src/frontend/test_chain.dart
@@ -4,26 +4,39 @@
 
 import 'package:stack_trace/stack_trace.dart';
 
-typedef StackTrace Mapper(StackTrace trace);
+import '../backend/invoker.dart';
+import '../util/stack_trace_mapper.dart';
 
 /// Converts [trace] into a Dart stack trace
-Mapper currentMapper = (trace) => trace;
+StackTraceMapper _mapper;
 
 /// The list of packages to fold when producing [StackTrace]s.
-Set<String> exceptPackages = new Set.from(['test', 'stream_channel']);
+Set<String> _exceptPackages = new Set.from(['test', 'stream_channel']);
 
 /// If non-empty, all packages not in this list will be folded when producing
 /// [StackTrace]s.
-Set<String> onlyPackages = new Set();
+Set<String> _onlyPackages = new Set();
 
-/// Converts [stackTrace] to a [Chain] following the test's configuration.
-Chain testChain(StackTrace stackTrace, {bool verbose: false}) {
-  var testTrace = currentMapper(stackTrace);
+void configureTestChaining(
+    {StackTraceMapper mapper,
+    Set<String> exceptPackages,
+    Set<String> onlyPackages}) {
+  if (mapper != null) _mapper = mapper;
+  if (exceptPackages != null) _exceptPackages = exceptPackages;
+  if (onlyPackages != null) _onlyPackages = onlyPackages;
+}
+
+Chain terseChain(StackTrace stackTrace, {bool verbose: false}) {
+  var testTrace = _mapper?.mapStackTrace(stackTrace) ?? stackTrace;
   if (verbose) return new Chain.forTrace(testTrace);
   return new Chain.forTrace(testTrace).foldFrames((frame) {
-    if (onlyPackages.isNotEmpty) {
-      return !onlyPackages.contains(frame.package);
+    if (_onlyPackages.isNotEmpty) {
+      return !_onlyPackages.contains(frame.package);
     }
-    return exceptPackages.contains(frame.package);
+    return _exceptPackages.contains(frame.package);
   }, terse: true);
 }
+
+/// Converts [stackTrace] to a [Chain] following the test's configuration.
+Chain testChain(StackTrace stackTrace) => terseChain(stackTrace,
+    verbose: Invoker.current.liveTest.test.metadata.verboseTrace);
--- a/lib/src/frontend/test_chain.dart
+++ b/lib/src/frontend/test_chain.dart
@@ -1,0 +1,25 @@
+import 'package:stack_trace/stack_trace.dart';
+
+typedef StackTrace Mapper(StackTrace trace);
+
+/// Converts [trace] into a Dart stack trace
+Mapper currentMapper = (trace) => trace;
+
+/// The list of packages to fold when producing [StackTrace]s.
+Set<String> exceptPackages = new Set.from(['test', 'stream_channel']);
+
+/// If non-empty, all packages not in this list will be folded when producing
+/// [StackTrace]s.
+Set<String> onlyPackages = new Set();
+
+/// Converts [stackTrace] to a [Chain] following the test's configuration.
+Chain testChain(StackTrace stackTrace, {bool verbose: false}) {
+  var testTrace = currentMapper(stackTrace);
+  if (verbose) return new Chain.forTrace(testTrace);
+  return new Chain.forTrace(testTrace).foldFrames((frame) {
+    if (onlyPackages.isNotEmpty) {
+      return !onlyPackages.contains(frame.package);
+    }
+    return exceptPackages.contains(frame.package);
+  }, terse: true);
+}
--- a/lib/src/frontend/test_chain.dart
+++ b/lib/src/frontend/test_chain.dart
@@ -10,13 +10,19 @@ import '../util/stack_trace_mapper.dart';
 /// Converts [trace] into a Dart stack trace
 StackTraceMapper _mapper;
 
-/// The list of packages to fold when producing [StackTrace]s.
+/// The list of packages to fold when producing [Chain]s.
 Set<String> _exceptPackages = new Set.from(['test', 'stream_channel']);
 
 /// If non-empty, all packages not in this list will be folded when producing
-/// [StackTrace]s.
+/// [Chain]s.
 Set<String> _onlyPackages = new Set();
 
+/// Configure the resources used for test chaining.
+///
+/// [mapper] is used to convert traces into Dart stack traces.
+/// [exceptPackages] is the list of packages to fold when producing a [Chain].
+/// [onlyPackages] is the list of packages to keep in a [Chain]. If non-empty,
+/// all packages not in this will be folded.
 void configureTestChaining(
     {StackTraceMapper mapper,
     Set<String> exceptPackages,
@@ -26,6 +32,10 @@ void configureTestChaining(
   if (onlyPackages != null) _onlyPackages = onlyPackages;
 }
 
+/// Returns [stackTrace] converted to a [Chain] with all irrelevant frames
+/// folded together.
+///
+/// If [verbose] is `true`, returns the chain for [stackTrace] unmodified.
 Chain terseChain(StackTrace stackTrace, {bool verbose: false}) {
   var testTrace = _mapper?.mapStackTrace(stackTrace) ?? stackTrace;
   if (verbose) return new Chain.forTrace(testTrace);
--- a/lib/src/runner/configuration.dart
+++ b/lib/src/runner/configuration.dart
@@ -322,11 +322,8 @@ class Configuration {
             : Uri.parse("http://localhost:$pubServePort"),
         _concurrency = concurrency,
         _paths = _list(paths),
-        _foldTraceExcept = foldTraceExcept != null
-            ? new Set.from(_list(foldTraceExcept))
-            : null,
-        _foldTraceOnly =
-            foldTraceOnly != null ? new Set.from(_list(foldTraceOnly)) : null,
+        _foldTraceExcept = _set(foldTraceExcept),
+        _foldTraceOnly = _set(foldTraceOnly),
         _filename = filename,
         chosenPresets =
             new UnmodifiableSetView(chosenPresets?.toSet() ?? new Set()),
@@ -365,6 +362,14 @@ class Configuration {
     var list = new List/*<T>*/ .unmodifiable(input);
     if (list.isEmpty) return null;
     return list;
+  }
+
+  /// Returns a set from [input].
+  static Set<T> _set<T>(Iterable<T> input) {
+    if (input == null) return null;
+    var set = new Set<T>.from(input);
+    if (set.isEmpty) return null;
+    return set;
   }
 
   /// Returns an unmodifiable copy of [input] or an empty unmodifiable map.
--- a/lib/src/runner/configuration.dart
+++ b/lib/src/runner/configuration.dart
@@ -97,6 +97,15 @@ class Configuration {
   /// See [shardIndex] for details.
   final int totalShards;
 
+  /// The list of packages to fold when producing [StackTrace]s.
+  List<String> get exceptPackages => _exceptPackages ?? [];
+  final List<String> _exceptPackages;
+
+  /// If non-empty, all packages not in this list will be folded when producing
+  /// [StackTrace]s.
+  List<String> get onlyPackages => _onlyPackages ?? [];
+  final List<String> _onlyPackages;
+
   /// The paths from which to load tests.
   List<String> get paths => _paths ?? ["test"];
   final List<String> _paths;
@@ -198,6 +207,8 @@ class Configuration {
       int shardIndex,
       int totalShards,
       Iterable<String> paths,
+      Iterable<String> exceptPackages,
+      Iterable<String> onlyPackages,
       Glob filename,
       Iterable<String> chosenPresets,
       Map<String, Configuration> presets,
@@ -238,6 +249,8 @@ class Configuration {
         shardIndex: shardIndex,
         totalShards: totalShards,
         paths: paths,
+        exceptPackages: exceptPackages,
+        onlyPackages: onlyPackages,
         filename: filename,
         chosenPresets: chosenPresetSet,
         presets: _withChosenPresets(presets, chosenPresetSet),
@@ -290,6 +303,8 @@ class Configuration {
       this.shardIndex,
       this.totalShards,
       Iterable<String> paths,
+      Iterable<String> exceptPackages,
+      Iterable<String> onlyPackages,
       Glob filename,
       Iterable<String> chosenPresets,
       Map<String, Configuration> presets,
@@ -307,6 +322,8 @@ class Configuration {
             : Uri.parse("http://localhost:$pubServePort"),
         _concurrency = concurrency,
         _paths = _list(paths),
+        _exceptPackages = _list(exceptPackages),
+        _onlyPackages = _list(onlyPackages),
         _filename = filename,
         chosenPresets =
             new UnmodifiableSetView(chosenPresets?.toSet() ?? new Set()),
@@ -382,6 +399,8 @@ class Configuration {
         shardIndex: other.shardIndex ?? shardIndex,
         totalShards: other.totalShards ?? totalShards,
         paths: other._paths ?? _paths,
+        exceptPackages: other._exceptPackages ?? _exceptPackages,
+        onlyPackages: other._onlyPackages ?? _onlyPackages,
         filename: other._filename ?? _filename,
         chosenPresets: chosenPresets.union(other.chosenPresets),
         presets: _mergeConfigMaps(presets, other.presets),
@@ -412,6 +431,8 @@ class Configuration {
       int shardIndex,
       int totalShards,
       Iterable<String> paths,
+      Iterable<String> exceptPackages,
+      Iterable<String> onlyPackages,
       Glob filename,
       Iterable<String> chosenPresets,
       Map<String, Configuration> presets,
@@ -450,6 +471,8 @@ class Configuration {
         shardIndex: shardIndex ?? this.shardIndex,
         totalShards: totalShards ?? this.totalShards,
         paths: paths ?? _paths,
+        exceptPackages: exceptPackages ?? _exceptPackages,
+        onlyPackages: onlyPackages ?? _onlyPackages,
         filename: filename ?? _filename,
         chosenPresets: chosenPresets ?? this.chosenPresets,
         presets: presets ?? this.presets,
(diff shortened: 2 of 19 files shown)


lib/src/runner/configuration/load.dart
```diff
@@ -74,6 +74,30 @@ class _ConfigurationLoader {
   Configuration _loadGlobalTestConfig() {
     var verboseTrace = _getBool("verbose_trace");
     var chainStackTraces = _getBool("chain_stack_traces");
+    var foldStackFrames = _getMap("fold_stack_frames", key: (keyNode) {
+      _validate(keyNode, "fold_stack_frames key must be a string",
```
nex3: I'd make this 'Must be "only" or "except".', which makes it clearer how to fix the failure. You don't need to include the key name, since the span exception will make it clear where in the file the error is.

This will also let you use the same error for this and for the value check.

lib/src/runner/configuration/load.dart
```diff
@@ -74,6 +74,30 @@ class _ConfigurationLoader {
...
+      if (valueNode is! YamlList ||
+          (valueNode as YamlList).firstWhere((value) => value is! String,
+                  orElse: () => null) !=
+              null) {
+        throw new SourceSpanFormatException(
+            'Invalid option for fold_stack_frames value: '
+            'Must be a list of strings.',
+            valueNode.span,
+            _source);
+      }
```
nex3: Make this a `_validate()` call.

lib/src/runner/configuration/load.dart
```diff
@@ -74,6 +74,30 @@ class _ConfigurationLoader {
...
+            'Invalid fold_stack_frames key: '
+            'Must either be `only` or `except`.',
+            keyNode.span,
+            _source);
+      }
+      return keyNode.value;
+    }, value: (valueNode) {
+      if (valueNode is! YamlList ||
+          (valueNode as YamlList).firstWhere((value) => value is! String,
+                  orElse: () => null) !=
```
nex3: Couldn't this `firstWhere()` call just be `any()`?

We should also verify that each element is a valid package name (which is to say, a valid Dart identifier).

lib/src/runner/configuration/load.dart
```diff
@@ -74,6 +74,30 @@ class _ConfigurationLoader {
...
+              null) {
+        throw new SourceSpanFormatException(
+            'Invalid option for fold_stack_frames value: '
+            'Must be a list of strings.',
```
nex3: I'd make this "Folded packages must be strings."

lib/src/runner/remote_listener.dart
```diff
@@ -16,7 +16,9 @@ import '../backend/suite.dart';
 import '../backend/test.dart';
 import '../backend/test_platform.dart';
 import '../util/remote_exception.dart';
+import '../util/stack_trace_mapper.dart';
 import '../utils.dart';
+import '../frontend/test_chain.dart';
```
nex3: Nit: imports should be alphabetical.

lib/src/runner/remote_listener.dart
```diff
@@ -105,7 +124,7 @@ class RemoteListener {
   static void _sendError(StreamChannel channel, error, StackTrace stackTrace) {
     channel.sink.add({
       "type": "error",
-      "error": RemoteException.serialize(error, stackTrace)
+      "error": RemoteException.serialize(error, testChain(stackTrace))
```
nex3: This should probably choose whether to produce a verbose trace based on the metadata once it's available, and default to verbose before then.

lib/src/runner/runner_test.dart
```diff
@@ -30,16 +28,9 @@ class RunnerTest extends Test {
...
   final MultiChannel _channel;
 
-  /// The function used to reformat errors' stack traces.
-  final _MapTrace _mapTrace;
-
-  RunnerTest(
-      this.name, this.metadata, Trace trace, this._channel, _MapTrace mapTrace)
-      : trace = trace == null ? null : new Trace.from(mapTrace(trace)),
-        _mapTrace = mapTrace;
+  RunnerTest(this.name, this.metadata, this.trace, this._channel);
```
nex3: I love how much cleaner this is...

lib/src/util/stack_trace_mapper.dart
```diff
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
```
nex3: Nit: newline between different import types.

lib/src/util/stack_trace_mapper.dart
```diff
@@ -17,14 +18,69 @@ class StackTraceMapper {
   /// The URI of the SDK root from which dart2js loaded its sources.
   final Uri _sdkRoot;
 
-  StackTraceMapper(String contents,
+  /// The contents of the source map.
+  final String _contents;
```
nex3: I'd call this `_mapContents`, to match `_mapUrl`.

lib/src/util/stack_trace_mapper.dart
```diff
@@ -17,14 +18,69 @@ class StackTraceMapper {
...
+
+  /// The URI of the source map.
```
nex3: Nit: "URL" (if you can fix the `_sdkRoot` docs too that would be great)

Technically, [URI is a deprecated term](https://url.spec.whatwg.org/#goals)!

lib/src/util/stack_trace_mapper.dart
```diff
@@ -17,14 +18,69 @@ class StackTraceMapper {
...
+  final Uri _mapUrl;
+
+  StackTraceMapper(this._contents,
       {Uri mapUrl, SyncPackageResolver packageResolver, Uri sdkRoot})
-      : _mapping = parseExtended(contents, mapUrl: mapUrl),
+      : _mapping = parseExtended(_contents, mapUrl: mapUrl),
```
nex3: We don't ever actually use this in the runner isolate, right? Can we make it lazy so that we don't spend time parsing it there?

lib/src/util/stack_trace_mapper.dart
```diff
@@ -17,14 +18,69 @@ class StackTraceMapper {
...
+      'packageConfigMap':
+          _serializablePackageConfigMap(_packageResolver.packageConfigMap),
+      'packageRoot': _packageResolver.packageRoot?.toString(),
+      'mapUrl': _mapUrl?.toString(),
+    };
+  }
+
+  /// Returns a Future which will resolve to a [StackTraceMapper] contained in
+  /// the provided serialized representation.
+  static Future<StackTraceMapper> deserialize(Map serialized) async {
```
nex3: If you use the [`SyncPackageResolver`](https://www.dartdocs.org/documentation/package_resolver/1.0.2/package_resolver/SyncPackageResolver-class.html) constructors, you should be able to make this synchronous.

lib/src/util/stack_trace_mapper.dart
```diff
@@ -17,14 +18,69 @@ class StackTraceMapper {
...
+    String packageRoot = serialized['packageRoot'] ?? '';
```
nex3: I'd use `as String` here—it helps call out that a dynamic cast is happening.

lib/src/util/stack_trace_mapper.dart
```diff
@@ -17,14 +18,69 @@ class StackTraceMapper {
...
+  }
+
+  static Map<String, String> _serializablePackageConfigMap(
+      Map<String, Uri> packageConfigMap) {
+    if (packageConfigMap == null) return null;
+    var result = {};
+    for (var key in packageConfigMap.keys) {
+      result[key] = '${packageConfigMap[key]}';
+    }
+    return result;
```
nex3: You can use [`mapMap()`](https://www.dartdocs.org/documentation/collection/1.14.1/collection/mapMap.html) here and below to make these a bit cleaner.

lib/src/util/stack_trace_mapper.dart
```diff
@@ -17,14 +18,69 @@ class StackTraceMapper {
...
+                .asSync
+            : await new PackageResolver.config(_deserializePackageConfigMap(
+                    serialized['packageConfigMap']))
+                .asSync,
+        mapUrl: Uri.parse(serialized['mapUrl']));
+  }
+
+  static Map<String, String> _serializablePackageConfigMap(
+      Map<String, Uri> packageConfigMap) {
+    if (packageConfigMap == null) return null;
```
nex3: Do we ever expect this to be `null`? Can `new PackageResolver.config()` even handle a `null` value?

lib/src/util/stack_trace_mapper.dart
```diff
@@ -17,14 +18,69 @@ class StackTraceMapper {
...
+            ? await new PackageResolver.root(
+                    Uri.parse(serialized['packageRoot']))
+                .asSync
+            : await new PackageResolver.config(_deserializePackageConfigMap(
+                    serialized['packageConfigMap']))
+                .asSync,
+        mapUrl: Uri.parse(serialized['mapUrl']));
+  }
+
+  static Map<String, String> _serializablePackageConfigMap(
```
nex3: Document this.

"serializable" -> "serialize"

test/frontend/test_chain_test.dart
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
```
nex3: 2017

lib/src/frontend/test_chain.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:stack_trace/stack_trace.dart';
```
nex3: Don't forget the copyright header.

test/frontend/test_chain_test.dart
```diff
@@ -0,0 +1,103 @@
...
+import 'package:test/test.dart';
+
+void main() {
+  test("failure", () async{
+    await new Future((){});
+    await new Future((){});
+    throw "oh no";
+  });
+}
+""";
```
nex3: Just create this file in a `setUp()` call.

test/runner/configuration/top_level_error_test.dart
```diff
@@ -13,6 +13,49 @@ import 'package:test/test.dart';
...
+        .create();
+
+    var test = await runTest(["test.dart"]);
+    expect(
+        test.stderr,
+        containsInOrder(
+            ["Invalid option for fold_stack_frames value", "^^^^^^"]));
+    await test.shouldExit(exit_codes.data);
+  });
+
```
nex3: Test what happens if both `only` and `except` keys are set.

test/frontend/test_chain_test.dart
```diff
@@ -0,0 +1,103 @@
...
+                "except": ["stream_channel"]
+              }
+            }))
+        .create();
+    await d.file("test.dart", _asyncFailure).create();
+
+    var test = await runTest(["test.dart"]);
+    var testOutput = (await test.stdoutStream().toList());
+    for (var output in testOutput) {
+      expect(output.contains('package:stream_channel'), isFalse);
```
nex3: I'd use the `neverEmits()` matcher for this.

lib/src/runner/remote_listener.dart
grouma: Done, I think? Invoker might be out of scope at this point.

lib/src/runner/configuration.dart
```diff
@@ -386,6 +386,31 @@ class Configuration {
     if (this == Configuration.empty) return other;
     if (other == Configuration.empty) return this;
 
+    List<String> foldTraceOnly = other._foldTraceOnly ?? _foldTraceOnly;
+    if (other._foldTraceOnly != null && _foldTraceExcept != null) {
+      foldTraceOnly = new Set.from(other._foldTraceExcept)
```
nex3: These fields should probably be stored as sets to begin with; that'll make this logic much cleaner.

lib/src/runner/configuration.dart
```diff
@@ -307,6 +322,8 @@ class Configuration {
             : Uri.parse("http://localhost:$pubServePort"),
         _concurrency = concurrency,
         _paths = _list(paths),
+        _exceptPackages = _list(exceptPackages),
+        _onlyPackages = _list(onlyPackages),
```
nex3: We should also ensure this at the configuration level, so users of the API can make simplifying assumptions. Also, make sure to document that.

lib/src/runner/configuration.dart
```diff
@@ -386,6 +386,31 @@ class Configuration {
...
+          .intersection(new Set.from(_foldTraceOnly))
+          .toList();
+    }
+
+    List<String> foldTraceExcept = other._foldTraceExcept ?? _foldTraceExcept;
+    if (other._foldTraceExcept != null && _foldTraceExcept != null) {
+      foldTraceExcept = new Set.from(other._foldTraceExcept)
+          .union(new Set.from(_foldTraceExcept))
+          .toList();
+    }
```
nex3: If you can assume that only one of `only` and `except` is null, you can clean up all these conditionals:

```dart
if (_foldTraceOnly != null) {
  if (other._foldTraceExcept != null) {
    // difference
  } else {
    // intersection
  }
} else {
  if (other._foldTraceExcept != null) {
    // union
  } else {
    // difference
  }
}

File: lib/src/runner/configuration/load.dart [hunk 25]
```diff
@@ -289,6 +266,32 @@ class _ConfigurationLoader {
         excludeTags: excludeTags);
   }
 
+  Map _loadFoldedStackFrames() {
```
nex3: Document this. In particular, it's not clear what the semantics of the return value are.

lib/src/runner/configuration/load.dart
```diff
@@ -289,6 +266,32 @@ class _ConfigurationLoader {
...
+    bool foldOptionSet = false;
```
nex3: Don't type-annotate local variables.

lib/src/runner/configuration/load.dart
```diff
@@ -289,6 +266,32 @@ class _ConfigurationLoader {
...
+    return _getMap("fold_stack_frames", key: (keyNode) {
+      _validate(keyNode, "Must be a string", (value) => value is String);
+      if (keyNode.value != "only" && keyNode.value != "except") {
+        throw new SourceSpanFormatException(
+            'Must be "only" or "except".', keyNode.span, _source);
+      }
```
nex3: Any statement of the form `if (something about value) { throw }` should probably be a call to `_validate()` instead.

lib/src/runner/remote_listener.dart
```diff
@@ -113,6 +102,13 @@ class RemoteListener {
     return controller.foreign;
   }
 
+  /// Returns a [Set] from a JSON serialized list.
+  static Set<String> _setFromSerializedList(List<String> list) {
```
nex3: Nit: I'd call this `_deserializeSet()` to match other serialization functions.

lib/src/util/stack_trace_mapper.dart
```diff
@@ -3,84 +3,81 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'package:collection/collection.dart';
+
```
nex3: This isn't where the newline should go. See [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/style#do-place-dart-imports-before-other-imports).

lib/src/util/stack_trace_mapper.dart
```diff
@@ -3,84 +3,81 @@
...
 import 'package:package_resolver/package_resolver.dart';
 import 'package:source_map_stack_trace/source_map_stack_trace.dart' as mapper;
 import 'package:source_maps/source_maps.dart';
 
 /// A class for mapping JS stack traces to Dart stack traces using source maps.
 class StackTraceMapper {
   /// The parsed source map.
-  final Mapping _mapping;
+  Mapping _mapping;
```
nex3: Mention in the documentation that this is initialized lazily in `mapStackTrace()`.

lib/src/util/stack_trace_mapper.dart
```diff
@@ -3,84 +3,81 @@
...
+            ? new SyncPackageResolver.root(Uri.parse(serialized['packageRoot']))
+            : new SyncPackageResolver.config(
+                _deserializePackageConfigMap(serialized['packageConfigMap'])),
         mapUrl: Uri.parse(serialized['mapUrl']));
   }
 
-  static Map<String, String> _serializablePackageConfigMap(
+  /// Converts a [packageConfigMap] into a format suitable for JSON
+  /// serialization.
+  static Map<String, String> _serializePackageConfigMap(
```
nex3: These shouldn't be static. Generally, don't make helpers static unless they need to be used in other static methods.

lib/src/runner/configuration.dart
grouma: Cleaned it up slightly. I think it is easier to understand broken out. It's much easier to read now that they are sets.

lib/src/runner/configuration.dart
grouma: Done.

lib/src/runner/configuration.dart
```diff
@@ -386,6 +386,31 @@ class Configuration {
     if (this == Configuration.empty) return other;
     if (other == Configuration.empty) return this;
 
+    List<String> foldTraceOnly = other._foldTraceOnly ?? _foldTraceOnly;
```
grouma: Done.

lib/src/runner/configuration/load.dart
grouma: Done.

lib/src/runner/configuration/load.dart
grouma: Done.

lib/src/runner/configuration/load.dart
grouma: Done.

lib/src/runner/remote_listener.dart
grouma: Done.

lib/src/util/stack_trace_mapper.dart
grouma: Done.

lib/src/runner/remote_listener.dart
grouma: Done.

lib/src/frontend/test_chain.dart
```diff
@@ -26,6 +32,10 @@ void configureTestChaining(
   if (onlyPackages != null) _onlyPackages = onlyPackages;
```
nex3: These `!= null` checks are unnecessary.

lib/src/runner/configuration.dart
```diff
@@ -322,8 +322,11 @@ class Configuration {
             : Uri.parse("http://localhost:$pubServePort"),
         _concurrency = concurrency,
         _paths = _list(paths),
-        _foldTraceExcept = _list(foldTraceExcept),
-        _foldTraceOnly = _list(foldTraceOnly),
+        _foldTraceExcept = foldTraceExcept != null
+            ? new Set.from(_list(foldTraceExcept))
```
nex3: Just add a `_set()` helper method.

lib/src/runner/configuration.dart
nex3: I still think nested ifs would be clearer. Broken out, the differences between each condition get lost especially when some conditions duplicate logic in others.

lib/src/runner/configuration/load.dart
```diff
@@ -266,14 +266,18 @@ class _ConfigurationLoader {
         excludeTags: excludeTags);
   }
 
+  /// Returns a map representation of the `fold_stack_frames` configuration.
+  ///
+  /// The key `except` will correspond to the list of packages to fold.
+  /// The key `only` will correspond to the list of packages to keep in a
+  /// test [Chain].
```
nex3: If the map only has string keys, the type annotation should indicate that.

lib/src/runner/configuration/load.dart
nex3: This still needs to be done.

lib/src/frontend/test_chain.dart
grouma: The null check is necessary since we don't check for nullness in terseChain.

lib/src/runner/configuration.dart
grouma: Done

lib/src/runner/configuration/load.dart
grouma: Done.

lib/src/runner/configuration/load.dart
grouma: Done.

lib/src/runner/configuration.dart
grouma: Done.

grouma:
@nex3 I think I have addressed all open comments. This UI is kind of terrible though.

lib/src/runner/configuration.dart
```diff
@@ -395,20 +395,19 @@ class Configuration {
...
-    }
-
-    if (other._foldTraceExcept != null && _foldTraceOnly != null) {
-      foldTraceOnly = _foldTraceOnly.difference(other._foldTraceExcept);
-    }
-    if (other._foldTraceOnly != null && _foldTraceOnly != null) {
-      foldTraceOnly = other._foldTraceOnly.intersection(_foldTraceOnly);
-    }
-
     var foldTraceExcept = other._foldTraceExcept ?? _foldTraceExcept;
```
nex3: At this point, I don't think it makes sense to set these by default since they'll *always* either be overridden or end up as `null`.

lib/src/runner/configuration/load.dart
```diff
@@ -287,11 +300,19 @@ class _ConfigurationLoader {
...
+      _validate(
+          valueNode,
+          "Folded packages must be strings",
+          (valueList) =>
+              valueList is YamlList &&
+              !(valueList).any((value) => value is! String));
+
+      _validate(
+          valueNode,
+          "Invalid package identifier",
```
nex3: "identifier" -> "name"

lib/src/runner/configuration/load.dart
```diff
@@ -287,11 +300,19 @@ class _ConfigurationLoader {
...
       return keyNode.value;
     }, value: (valueNode) {
-      if (valueNode is! YamlList ||
-          (valueNode as YamlList).any((value) => value is! String)) {
-        throw new SourceSpanFormatException(
-            'Folded packages must be strings.', valueNode.span, _source);
-      }
+      _validate(
+          valueNode,
+          "Folded packages must be strings",
```
nex3: These messages need trailing periods.

lib/src/runner/configuration/load.dart
```diff
@@ -287,11 +300,19 @@ class _ConfigurationLoader {
...
+          (valueList) =>
+              valueList is YamlList &&
+              !(valueList).any((value) => value is! String));
```
nex3: `valueList.every((value) => value is String)`

Same below.

lib/src/runner/configuration.dart
grouma: Discussed offline. There is a case where we want to default.

lib/src/runner/configuration/load.dart
grouma: Done.

lib/src/runner/configuration/load.dart
grouma: Done.

lib/src/runner/configuration/load.dart
grouma: Done.

lib/src/runner/configuration/load.dart
```diff
@@ -302,16 +302,16 @@ class _ConfigurationLoader {
     }, value: (valueNode) {
       _validate(
           valueNode,
-          "Folded packages must be strings",
+          "Folded packages must be strings.",
           (valueList) =>
               valueList is YamlList &&
-              !(valueList).any((value) => value is! String));
+              (valueList).every((value) => value is String));
```
nex3: The parentheses around `valueList` are unnecessary.

grouma: Originally was casting. Fixed.

PR #653, opened 2017-06-20
Add support for IPv6

grouma:


lib/src/util/io.dart
```diff
@@ -142,13 +146,31 @@ Future<T> getUnusedPort<T>(FutureOr<T> tryPort(int port)) async {
 /// This has a built-in race condition: another process may bind this port at
 /// any time after this call has returned. If at all possible, callers should
 /// use [getUnusedPort] instead.
-Future<int> getUnsafeUnusedPort() async {
-  var socket = await RawServerSocket.bind(InternetAddress.LOOPBACK_IP_V4, 0);
+Future<int> getUnsafeUnusedPort({bool IPv6: false}) async {
```
kevmoo: `ipV6` ? Maybe ask @munificent about guidelines here...

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   makes it possible to safely share `Future`s and `Stream`s between tests and
   their tear-downs.
 
+* Add support for IPv6
```
kevmoo: Add a period.

lib/src/util/io.dart
```diff
@@ -142,13 +146,31 @@ Future<T> getUnusedPort<T>(FutureOr<T> tryPort(int port)) async {
...
 }
 
+/// Returns whether this computer supports binding to IPv6 addresses.
+Future<bool> _supportsIPv6() async {
+  try {
+    var socket = await ServerSocket.bind(InternetAddress.LOOPBACK_IP_V6, 0,
+        v6Only: true);
+    socket.close();
+    return true;
+  } on SocketException catch (_) {
```
kevmoo: I *think* you can remove `catch (_)` if you have `on` – I think.

lib/src/util/io.dart
munificent: Tricky. It's "IPv6" in prose. According to the guidelines, the acronym would get lowercased, but it's not clear what that means for the "v". I think either `ipv6` or `ipV6` would work.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   makes it possible to safely share `Future`s and `Stream`s between tests and
   their tear-downs.
 
+* Add support for IPv6.
```
nex3: We should probably be clearer about the circumstances under which this actually caused problems. Maybe "Properly allocate ports when debugging Chrome and Dartium in an IPv6-only environment."

lib/src/util/io.dart
```diff
@@ -130,8 +130,12 @@ void warn(String message, {bool color}) {
 /// applications that don't print out the bound port.
 Future<T> getUnusedPort<T>(FutureOr<T> tryPort(int port)) async {
   T value;
+  var supportsIPv6 = false;
+  if (await _supportsIPv6()) {
+    supportsIPv6 = true;
+  }
```
nex3: This could just be `var supportsIPv6 = await _supportsIPv6`, but it seems perverse to bind a socket here when we're already doing so in `getUnsafeUnusedPort()`. Maybe `getUnsafeUnusedPort()` should try to bind to IPv6 first and record if that doesn't work.

CHANGELOG.md
grouma: Done.

lib/src/util/io.dart
grouma: Removed the unnecessary conditional. I think it makes sense to be in here as I don't want getUnsafedUnusedPort to make multiple binds since it is called in a while.

nex3: I don't understand that reasoning. If you recorded whether IPv6 binding worked, you'd only call `bind()` twice the first time through the loop.

grouma: Discussed offline. Had concerns with storing state in this utility file. Given that it is guaranteed to be system wide I have modified it.

grouma:
Friendly ping to @nex3. I had to make similar changes inside Google3 to unblock other teams. I don't like leaving the two packages out of sync.

lib/src/util/io.dart
```diff
@@ -143,9 +142,9 @@ Future<T> getUnusedPort<T>(FutureOr<T> tryPort(int port)) async {
 /// This has a built-in race condition: another process may bind this port at
 /// any time after this call has returned. If at all possible, callers should
 /// use [getUnusedPort] instead.
-Future<int> getUnsafeUnusedPort({bool ipV6: false}) async {
+Future<int> getUnsafeUnusedPort() async {
   var socket;
-  if (ipV6) {
+  if (await _supportsIPv6) {
```
grouma: Done.

lib/src/util/io.dart
```diff
@@ -137,36 +137,33 @@ Future<T> getUnusedPort<T>(FutureOr<T> tryPort(int port)) async {
   return value;
 }
 
+/// Whether this computer supports binding to IPv6 addresses.
+bool _supportsIPv6 = null;
```
nex3: Nit: `var`

lib/src/util/io.dart
```diff
@@ -137,36 +137,33 @@ Future<T> getUnusedPort<T>(FutureOr<T> tryPort(int port)) async {
...
 /// This has a built-in race condition: another process may bind this port at
 /// any time after this call has returned. If at all possible, callers should
 /// use [getUnusedPort] instead.
 Future<int> getUnsafeUnusedPort() async {
   var socket;
-  if (await _supportsIPv6) {
-    socket = await RawServerSocket.bind(InternetAddress.LOOPBACK_IP_V6, 0,
-        v6Only: true);
-  } else {
+  if (_supportsIPv6 ?? true) {
```
nex3: `?? true` skeeves me out a little bit. Maybe make the variable `_maySupportIPv6` and have it default to `true`?

lib/src/util/io.dart
grouma: Done.

lib/src/util/io.dart
grouma: Done.

PR #654, opened 2017-06-22
Await some unawaited futures

grouma:
A handful of these obviously should be awaited, namely the file creation ones. However, many of these do not obviously need to be awaited.


lib/src/executable.dart
```diff
@@ -161,7 +161,7 @@ transformers:
     if (signalSubscription == null) return;
     signalSubscription.cancel();
     signalSubscription = null;
-    stdinLines.cancel(immediate: true);
+    await stdinLines.cancel(immediate: true);
```
nex3: I don't think this should matter, and it would be nice to get the runner started closing as soon as possible. Also, since this is in the signal-handling infrastructure, it's definitely not causing bot issues.

lib/src/runner/engine.dart
```diff
@@ -529,8 +529,8 @@ class Engine {
   Future close() async {
     _closed = true;
     if (_closedBeforeDone != null) _closedBeforeDone = true;
-    _onSuiteAddedController.close();
-    _suiteController.close();
+    await _onSuiteAddedController.close();
+    await _suiteController.close();
```
nex3: I think these are the most likely candidates for the source of the flakiness. It's possible that there was a race condition where their events hadn't fully propagated, which would mean that the `close()` calls below didn't hit all the relevant tests.

It would be an interesting experiment to try `await`ing *only* these and seeing if that clears up the flakiness.

lib/src/executable.dart
grouma: Removed.

lib/src/runner/engine.dart
grouma: I removed this await and did a couple Travis runs. Unfortunately I could not get into the flaky state.

grouma:
Hmmm. Removing that await caused one of the tests to time out. I'm going to try and verify that is the root cause of the flakiness.

PR #657, opened 2017-06-23
Remove unused imports

grouma:
These warnings will cause issues with Google3. Upgrading the warnings to errors and resolving the issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.12.23+1
+
+* Remove unused imports. 
+
 ## 0.12.23
 
 * Add a `fold_stack_frames` field for `dart_test.yaml`. This will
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.12.23+1
```
nex3: It's not worth publishing this version, since it doesn't have any user-visible changes.

analysis_options.yaml
```diff
@@ -1,5 +1,13 @@
 analyzer:
   strong-mode: true
+errors:
+  unused_element: error
+  unused_import: error
+  unused_local_variable: error
+  dead_code: error
```
nex3: Why are you making these errors?

CHANGELOG.md
grouma: Discussed offline. Necessary for a Google3 sync.

analysis_options.yaml
grouma: Added a comment. Necessary to not break Google3.

PR #663, opened 2017-07-06
Add a Node platform.

nex3:


natebosch:
Could we add more details to the commit message?

nex3:
Sure! What are you looking for in particular?

natebosch:
Mainly hoping for a breakdown of what are changes to existing extension points to support this vs the pieces that would go in for any new platform.

nex3:
Done.

lib/src/runner/node/platform.dart
```diff
@@ -0,0 +1,194 @@
...
+      var response = await (await _http.getUrl(url)).close();
+
+      if (response.statusCode != 200) {
+        // We don't care about the response body, but we have to drain it or
+        // else the process can't exit.
+        response.listen(null);
+
+        throw new LoadException(
+            suitePath,
+            "Error getting $url: ${response.statusCode} "
```
efortuna: consider using single quotes here and below

lib/src/runner/node/platform.dart
```diff
@@ -0,0 +1,194 @@
...
+            "${response.reasonPhrase}\n"
+            'Make sure "pub serve" is serving the test/ directory.');
+      }
+
+      return await UTF8.decodeStream(response);
+    } on IOException catch (error) {
+      var message = getErrorMessage(error);
+      if (error is SocketException) {
+        message = "${error.osError.message} "
```
efortuna: okay I'm going to stop posting about this quote thing, but check out other places in these files that it's happening.

nex3:
I think the styleguide intentionally *doesn't* provide guidance on string quotes, because keeping them consistent is a lot of overhead for little benefit. This package certainly doesn't strive for consistency in general. I'd prefer to leave them as-is.

PR #672, opened 2017-08-04
Fix travis hack

grouma:
Forcing an exit on Travis is negatively impacting those that collect code coverage after a test is run. The belief is that this hack is necessary only for a subset of tests within package:test itself. Updating the logic so that it makes use of a more specific environment variable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ dart:
  - stable
  - 1.23.0
 
+env:
+ - FORCE_TEST_EXIT=true
+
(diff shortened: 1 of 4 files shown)


grouma:
I thought I found the issue in https://github.com/dart-lang/test/pull/654 but fixing it didn't seem to impact the flakiness (I'll still merge this change later today). Now that I know which shard is consistently causing issues hopefully I can use that information to root cause the problem. That being said, this issues hasn't been a huge priority.

PR #678, opened 2017-08-30
WIP: latest packages

kevmoo:
And other cleanup

Note the last commit: will remove this before committing. There are a lot of other dependencies to fix, but need to start somewhere.


natebosch:
Do the tests just not like dependency overrrides or did something actually break?

PR #684, opened 2017-09-18
Zone.strong

floitschG:
The `dart:async` zone API is in the process of becoming strong-mode clean.
This patch fixes a breakage introduced by this change.

The change should have no semantic difference, even when running with the non-strong zone API.


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: test
-version: 0.12.24+5
+version: 0.12.24+6
```
grouma: Do we need to update the minimum sdk version as well?

floitschG: No. This change should maintain the same semantics.

floitschG:
I added a Changelog entry.

nex3:
@floitschG It looks like as of 2.0.0-dev.1.0, `test` is still experiencing a bunch of Zone-related errors ([for example](https://travis-ci.org/dart-lang/test/jobs/277993772#L1056-L1095)). Are there further changes needed?

PR #689, opened 2017-09-27
Fix spawnHybridUri() in the browser

### LINKED ISSUE #688: Relative url in spawnHybridUri function
Documentation of spawnHybridUri function says:
```dart
/// If [uri] is relative, it will be interpreted relative to the `file:` URL for
/// the test suite being executed. If it's a `package:` URL, it will be resolved
/// using the current package's dependency constellation.
```
I placed my hybrid test and spawned server in the package test directory. Then in hybrid test i call (as is shown in the example):
```dart
final channel = spawnHybridUri('server.dart');
```
After test started I got 'server.dart is not found'. And in the error message I can see that server.dart is expected to be in <server root>/test/test/server.dart
After following changes everything was o'kay:
```dart
final channel = spawnHybridUri('../server.dart');
```
What is the reason of this behaivour:
* incorrect documentation
* bug in spawnHybridUri
* there is something I do not understand

nex3:
Closes #688

### DIFF
--- a/lib/src/frontend/spawn_hybrid.dart
+++ b/lib/src/frontend/spawn_hybrid.dart
@@ -98,10 +98,17 @@ StreamChannel spawnHybridUri(uri, {Object message, bool stayAlive: false}) {
 
   String absoluteUri;
   if (parsedUrl.scheme.isEmpty) {
-    var suitePath = Invoker.current.liveTest.suite.path;
-    absoluteUri = p.url.join(
-        p.url.dirname(p.toUri(p.absolute(suitePath)).toString()),
-        parsedUrl.toString());
+    // If we're running in a browser context, the working directory is already
+    // relative to the test file, whereas on the VM the working directory is the
+    // root of the package.
+    if (p.style == p.Style.url) {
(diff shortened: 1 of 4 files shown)


lib/src/frontend/spawn_hybrid.dart
```diff
@@ -98,10 +98,17 @@ StreamChannel spawnHybridUri(uri, {Object message, bool stayAlive: false}) {
...
   if (parsedUrl.scheme.isEmpty) {
-    var suitePath = Invoker.current.liveTest.suite.path;
+    // If we're running in a browser context, the working directory is already
+    // relative to the test file, whereas on the VM the working directory is the
+    // root of the package.
+    if (p.style == p.Style.url) {
+      absoluteUri = p.absolute(parsedUrl.toString());
+    } else {
+      var suitePath = Invoker.current.liveTest.suite.path;
     absoluteUri = p.url.join(
```
grouma: Is the indentation off here?

nex3: Oops, forgot to reformat. Fixed.

Issue #391, opened 2016-02-01
Add configurable platform support


stevenroose:
Chromium and Chrome should have similar run properties. However Chromium is open-source, while Chrome is branded. This makes Chromium more widely available on Linux systems and the preferred choice of many Linux users.

Since they would probably only differ in the name of the command, it's easy to add Chromium support as a separate platform, but alternatively, the `chrome` platform could just use `google-chrome` or `chromium`, whichever is available. (I assume something similar should already happen in order to support `google-chrome-beta`.)

nex3:
I'm going to retarget this issue to refer to general support for configuring platforms. This is blocked on #46, since that's where the configuration will live.

I envision the ability to define new platforms that are small configurations of existing ones, like so:

```yaml
platforms:
  chromium:
    type: chrome
    name: Chromium
    executable: chromium
```

The "type" is an existing platform whose spawning logic can be re-used. The name is an optional human-readable name for that platform, and the executable is the executable name.

In practice, finding the executable will likely be a little more complicated than that, because different operating systems have different locations and executable path conventions. We'd probably want to support something like this:

```yaml
platforms:
  chromium:
    type: chrome
    name: Chromium
    executable:
      linux: chromium
      mac_os: /Applications/Chromium.app/Contents/MacOS/Chromium
      windows: Chromium\Application\chrome.exe
```

Absolute paths would be resolved absolutely, executable names would use the executable path, and for Windows in particular relative paths would check against `LOCALAPPDATA`, `PROGRAMFILES`, and `PROGRAMFILES(X64)`.

You'd also be able to use this to override the executable locations of existing platforms:

```yaml
platforms:
  dartium:
    executable:
      linux: dartium
      mac_os: /Applications/Dartium.app/Contents/MacOS/Chromium
      windows: Dartium\Application\chrome.exe
```

stevenroose:
That sounds awesome!

nex3:
Another thing I should mention: derived platforms would inherit the platform selectors of their parents, so `@TestOn("chrome")` would also allow a test to run on Chromium. They would also introduce their own platform selector variable, so `@TestOn("chromium")` would work as well.

nex3:
Initial configuration file support has landed, so this is no longer blocked.

StepFischer:
Is the topic of configurable platforms still in the work or is it abandoned?

nex3:
This is definitely still something we plan to do (which is why this issue is still open :wink:).

nex3:
[...]

PR #693, opened 2017-10-03
Don't use TestPlatform.all in PlatformSelector

nex3:
We want users to be able to dynamically define new platforms, which
means we need infrastructure in place for piping those platforms to
places that previously assumed TestPlatform.all was a full list of
available platforms. PlatformSelector is the trickiest example, since
it's parsed in a number of different places and needs to provide useful
feedback to users when they use an undefined platform.

This splits parsing and platform validation into two separate steps.
Validation will be done immediately after parsing when the selectors
come from top-level annotations or parameters passed to test() or
group(), but selectors defined in configuration files are now parsed
only after all configuration is parsed. This will allow new platforms to
be defined *and* referenced in configuration files.

See #99
See #391

### DIFF
--- a/lib/src/runner/configuration/args.dart
+++ b/lib/src/runner/configuration/args.dart
@@ -10,6 +10,7 @@ import 'package:boolean_selector/boolean_selector.dart';
 import '../../backend/test_platform.dart';
 import '../../frontend/timeout.dart';
 import '../configuration.dart';
+import 'platform_selection.dart';
 import 'reporters.dart';
 import 'values.dart';
 
@@ -17,7 +18,7 @@ import 'values.dart';
 final ArgParser _parser = (() {
   var parser = new ArgParser(allowTrailingOptions: true);
 
-  var allPlatforms = TestPlatform.all.toList();
+  var allPlatforms = TestPlatform.all.toList()..remove(TestPlatform.vm);
   if (!Platform.isMacOS) allPlatforms.remove(TestPlatform.safari);
   if (!Platform.isWindows) allPlatforms.remove(TestPlatform.internetExplorer);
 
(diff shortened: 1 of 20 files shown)


lib/src/runner/configuration/args.dart
```diff
@@ -10,14 +10,15 @@ import 'package:boolean_selector/boolean_selector.dart';
...
+import 'platform_selection.dart';
 import 'reporters.dart';
 import 'values.dart';
 
 /// The parser used to parse the command-line arguments.
 final ArgParser _parser = (() {
   var parser = new ArgParser(allowTrailingOptions: true);
 
-  var allPlatforms = TestPlatform.all.toList();
+  var allPlatforms = TestPlatform.all.toList()..remove(TestPlatform.vm);
```
grouma: Why do we remove the vm platform here?

nex3: Because we manually include it in the list so we can add "(default)" after it.

nex3:
> Do you have an example of how one would define a new platform?

Actually defining a platform isn't supported yet; this is just laying the groundwork. But my plan is laid out in #391.

lib/src/runner/configuration/args.dart
grouma: Ahh I see.

nex3:
> Noted. We'll have to be careful when we pull this into Google3. As long as this: https://github.com/dart-lang/test/blob/master/lib/src/backend/test_platform.dart#L111 still works we should be good.

As this CL stands, that should continue to work, but I'm hoping to refactor it a bit to make it more similar to other plugin API stuff we expose.

PR #694, opened 2017-10-03
Send test platforms to the remote listener

nex3:
This ensures that the remote listener has access to any platforms that
are dynamically loaded in the test runner, so they can be used in
platform selectors.

See #99
See #391


lib/src/runner/browser/platform.dart
```diff
@@ -192,8 +192,8 @@ class BrowserPlatform extends PlatformPlugin {
   ///
   /// This will start a browser to load the suite if one isn't already running.
   /// Throws an [ArgumentError] if [browser] isn't a browser platform.
-  Future<RunnerSuite> load(
-      String path, TestPlatform browser, SuiteConfiguration suiteConfig) async {
+  Future<RunnerSuite> load(String path, TestPlatform browser,
```
grouma: I believe this will be breaking for Bolt internally. So again we'll have to be cautious with the next roll.

nex3: Yeah, and there are more changes coming down that will slightly change the plugin API in other ways. It should all be straightforward to migrate, though.

PR #695, opened 2017-10-04
Remove registerTestPlatform()

nex3:
Now that the Loader is the canonical place for determining which
TestPlatforms are valid, we no longer need a separate mechanism for
registering platforms. Adding a plugin to the loader is sufficient.

See #391


grouma:
I fail to see how we will register a new platform in the loader. Can you give a small code snippet?

nex3:
```dart
loader.registerPlatformPlugin([
  new TestPlatform("name", "identifier")
], () => new MyPlatformPlugin());
```

Or, using the top-level hack API:

```dart
registerPlatformPlugin([
  new TestPlatform("name", "identifier")
], () => new MyPlatformPlugin());
```

PR #696, opened 2017-10-10
Validate platform selectors against string sets

nex3:
This makes it easier to add platforms that extend existing platforms
without running into serialization headaches. It also cleans up the
Declarer API and means we're sending less redundant data between
isolates.

See #391

### DIFF
--- a/lib/src/backend/platform_selector.dart
+++ b/lib/src/backend/platform_selector.dart
@@ -8,8 +8,8 @@ import 'package:source_span/source_span.dart';
 import 'operating_system.dart';
 import 'test_platform.dart';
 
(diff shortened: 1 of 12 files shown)


lib/src/backend/platform_selector.dart
```diff
@@ -11,6 +11,7 @@ import 'test_platform.dart';
 /// The set of statically-known valid variable names.
 final _validVariables =
```
grouma: I'm thinking we should change this variable name since we now pass a `validVariables` argument in validate. Maybe `_staticValidVariables`?

nex3: Done.

PR #697, opened 2017-10-12
Add a define_platforms configuration field

nex3:


lib/src/backend/test_platform.dart
```diff
@@ -109,6 +140,14 @@ class TestPlatform {
   Object serialize() {
     if (builtIn.contains(this)) return identifier;
 
+    if (parent != null) {
```
grouma: [optional] create the map with the fields name and identifier then conditionally append the parent field before returning.

doc/configuration.md
```diff
@@ -567,6 +570,91 @@ when running on a particular operating system, use [`on_os`](#on_os) instead.
...
+### Browser Settings
+
+All built-in browser platforms provide the same settings that can be set using
+[`define_platforms`](#define_platforms), which control how the browser
+executable is invoked.
+
+#### `executable`
+
+The `executable` field tells the test runner where to look for the executable to
+use to start the browser. It has three sub-keys, one for each supported operating
```
grouma: Should we mention that the sub-keys don't all have to be defined?

nex3: Done.

lib/src/backend/test_platform.dart
nex3: I think that would make it less clear that the child map isn't the same structure as the base-level map.

PR #701, opened 2017-10-16
Forbid define_platforms in the global config file

nex3:


nex3:
That's tricky... the fact that it's not even updating the test run time suggests that something synchronous is deadlocking the main isolate, but it's hard to imagine what that would be. It's possible that it's some kind of pernicious VM issue.

PR #703, opened 2017-10-18
Support configuring the Node platform

nex3:
See #391

### DIFF
--- a/lib/src/runner/node/platform.dart
+++ b/lib/src/runner/node/platform.dart
@@ -171,9 +171,10 @@ class NodePlatform extends PlatformPlugin
   /// Starts the Node.js process for [platform] with [jsPath].
   Future<Process> _startProcess(TestPlatform platform, String jsPath) async {
     try {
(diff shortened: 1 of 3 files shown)


lib/src/runner/node/platform.dart
```diff
@@ -141,7 +165,19 @@ class NodePlatform extends PlatformPlugin {
...
     }
 
-    return new Pair(await Process.start(_executable, [jsPath]), mapper);
+    return new Pair(await _startProcess(platform, jsPath), mapper);
+  }
+
+  /// Starts the Node.js process for [platform] with [jsPath].
+  Future<Process> _startProcess(TestPlatform platform, String jsPath) async {
+    try {
+    return await Process.start(_settings[platform].executable, [jsPath]);
```
grouma: Is this formatted correctly? looks like the return should be indented.

nex3: Oops, forgot to reformat!

Issue #377, opened 2015-12-18
Add a surround() function.


nex3:
Given Dart's Zone-based asynchrony, there are a number of use-cases for being able to wrap entire tests—including the setup and teardown—in a single zone (see for example [this thread](https://github.com/dart-lang/observe/pull/84#discussion_r47805618)). A straightforward way to accomplish this would be to provide a `void surround(Future callback(Future runTest()))` function that defines a callback to run around the test and any inner setups or teardowns. For example:

```dart
void main() {
  surround((runTest) {
    return runZoned(runTest, zoneSpecification: myZoneSpecificiation);
  });

  // ..
}
```

There are a lot of edge-cases that would need to be worked out, and it's possible that one of them will make this not work. But it's something to keep in mind.

natebosch:
I think this is something we should push on.

Some open questions I have:

- Naming - my current inclination is `wrapCallbacks`. Are there other options?
- What level of granularity should this apply to? I think for the sake of conceptual integrity it makes sense to work like `setUp` does today. You can define multiple levels of wrapping at each `group` and they would nest in the same order as groups. This has the potential to be abused and lead to confusing test patterns, but it's also I think the simplest building block that would solve all use cases uniformly.

cc @jakemac53 @grouma for thoughts.

cc @mehmetf for anything we should be thinking about in terms of solving the `flutter_test` goals. Does this need to even be per-test? I could see this as an additional named argument to `test()` if we need that level of granularity.

mehmetf:
Yes, we would need this at the granularity of test for `flutter_test` use case. It creates a new `FakeAsync` object per test to ensure isolation.

jakemac53:
> Yes, we would need this at the granularity of test for `flutter_test` use case. It creates a new `FakeAsync` object per test to ensure isolation.

I believe the proposed API would already support that without a callback to `test` - it looks like it would be invoked once for each test and its corresponding setUp/tearDown invocations.

jakemac53:
@natebosch how does this feature interact with `setUpAll` and `tearDownAll`?

natebosch:
> I believe the proposed API would already support that without a callback to `test` - it looks like it would be invoked once for each test and its corresponding setUp/tearDown invocations.

[...]

PR #713, opened 2017-10-25
Make addTearDown() play nice with setUpAll()

### LINKED ISSUE #712: addTearDown() should add to tearDownAll() in setUpAll()
Currently, if you call `addTearDown()` in a `setUpAll()`, it will run after all the `setUpAll()`s complete but before any tests run. This isn't very useful behavior, and almost certainly not what the user intends. We should instead have it add to the `tearDownAll()` list in the same way it would normally add to `tearDown()`.

Inspired by https://github.com/dart-lang/test_descriptor/issues/11.

nex3:
addTearDown() now adds a tearDownAll() rather than adding a plain
tearDown() when called from within setUpAll() (or tearDownAll()). This
is technically a behavioral change, but the old behavior was so
useless I doubt anyone was relying on it.

Closes #712

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 0.12.27
 
+* When `addTearDown()` is called within a call to `setUpAll()`, it runs its
+  callback after *all* tests instead of running it after the `setUpAll()`
+  callback.
+
 * When running in an interactive terminal, the test runner now prints status
   lines as wide as the terminal and no wider.
 
--- a/lib/src/backend/declarer.dart
+++ b/lib/src/backend/declarer.dart
@@ -155,10 +155,14 @@ class Declarer {
         }
       }
 
-      await Invoker.current.waitForOutstandingCallbacks(() async {
-        await _runSetUps();
-        await body();
-      });
+      await runZoned(
+          () => Invoker.current.waitForOutstandingCallbacks(() async {
+                await _runSetUps();
+                await body();
+              }),
+          // Make the declarer visible to running tests so that they'll throw
+          // useful errors when calling `test()` and `group()` within a test.
+          zoneValues: {#test.declarer: this});
     }, trace: _collectTraces ? new Trace.current(2) : null, guarded: false));
   }
 
@@ -224,6 +228,10 @@ class Declarer {
     _tearDownAlls.add(callback);
   }
 
+  /// Like [tearDownAll], but called from within a running [setUpAll] test to
+  /// dynamically add a [tearDownAll].
+  void addTearDownAll(callback()) => _tearDownAlls.add(callback);
+
   /// Finalizes and returns the group being declared.
   ///
   /// **Note**: The tests in this group must be run in a [Invoker.guard]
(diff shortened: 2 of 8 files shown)


nex3:
cc @jakemac53

kevmoo:
@nex3 – need to make travis happy, though

nex3:
It looks like the zones here are trickier than I thought. We want to run the tear-down logic in the same error zone as `setUpAll()` so that errors can be transmitted between them, but we model them as separate tests, which means they have separate error zones. I tried to work around this by running `addTearDown()` callbacks in `setUpAll()`'s error zone, but then this looks like `setUpAll()` is succeeding at first and then failing later, which we want to avoid.

I can see a few possible solutions:

1. Bite the bullet and run tear-downs in different error zones. I don't like this because it violates user expectations, and because it means some uses of `addTearDown()` will work in `setUp()` but break in `setUpAll()`.

2. Do some clever zone variable munging to run the tear-downs in the `setUpAll()` error zone but have them report errors to the `tearDownAll()` invoker. This would probably be the least disruptive solution, but 
it would probably involve a lot of ugly duplicated code and a very messy zone hierarchy that may cause difficulty down the line.

3. Merge the virtual tests for `setUpAll()` and `tearDownAll()` tests into a single virtual test. This would be a substantial ontological change in two ways:

    * The virtual test would need a way to signal to the `Engine` that it's finished running set-ups, and to receive a signal back that it's time to run tear-downs. There's no way to do this with the existing `LiveTest` API, so we'd probably need to create new `Test` and `LiveTest` subclasses that would need their own serialization logic.

    * The virtual test would run concurrently with all other tests in the suite. This is currently possible, since we can't guarantee that one test has stopped running asynchronously when we move on to the next one, but right now we always at least report the completion of one test before we start the next one.

      It should be easy to update our own reporters to handle this, but there's a risk that it will violate the assumptions of consumers of the JSON reporter. It doesn't directly violate any existing guarantees, so it's not strictly a breaking change, but we should make sure consumers are aware of it.

I think I favor option 3. It's the most work, but in exchange it keeps the zone semantics of each test (relatively) simple and lays the groundwork for a `surroundAll()` feature *a la* #377.

@grouma Thoughts?

@devoncarew @jwren @alexander-doroshko Do you have insights on how disruptive it would be to IntelliJ if the JSON reporter indicated that multiple tests from a given suite were running at the same time?

grouma:
Discussed offline. Option one is a nonstarter as it can deadlock tests and overall be a poor experience. As for option two, I'm not a fan of complicating our zone code as few folks have a strong grasp of zones and their implementation details. We have recently have been running into issues with zone behavior between the VM and DDC and it hasn't been easy to follow. It looks like we are pigeon holed to option 3. Hopefully the change isn't too much work and will result in more maintainable code.

nex3:
After talking with @grouma, we've come up with a couple more options:

4. Run `setUpAll()` and `tearDownAll()` in the same context as `main()`, rather than creating virtual tests. This means they'd automatically be run in the same error zone. However, errors in `main()` are treated as errors loading the test, which means they immediately kill the entire test suite, which isn't the behavior we want. This would also require running `main()` in a test context with an `Invoker`, which we don't do currently.

5. Have a single error zone for the entire suite, rather than one per test. This means that *all* tests are in the same error zone. The downside is that direct uses of the backend API get more complicated, since users are obligated to manually set up this error handling. On the other hand, it means its users don't have to worry about error zones at all, which is a win.

I think 5 may be the best option, if we can get it to work. Currently we don't expose the backend API at all, and I'm okay with making it somewhat more difficult to use in exchange for lower code complexity.

alexander-doroshko:
>  @alexander-doroshko Do you have insights on how disruptive it would be to IntelliJ if the JSON reporter indicated that multiple tests from a given suite were running at the same time?

I'm far from understanding all the details you discuss here, but the answer to this question is: it shouldn't be a problem. Happy to test this use case if you can give me a project to play with.

nex3:
Fixed Travis; @grouma PTAL

PR #715, opened 2017-10-27
Fix test hang and prep for release - 0.12.26+1

grouma:
We declared an incorrect constraint for package stack_trace. This was fixed in https://github.com/dart-lang/test/pull/714

Also fixing potential test hangs by manually closing IO streams.

Preparing for a new release.

### DIFF
--- a/lib/src/runner/browser/browser.dart
+++ b/lib/src/runner/browser/browser.dart
@@ -51,12 +51,12 @@ abstract class Browser {
   Future get onExit => _onExitCompleter.future;
   final _onExitCompleter = new Completer();
 
-  /// IO streams for the underlying browser process.
-  final _ioStreams = <StreamSubscription>[];
+  /// Standard IO streams for the underlying browser process.
+  final _ioSubscriptions = <StreamSubscription>[];
 
   Future _drainAndIgnore(Stream s) async {
     try {
-      _ioStreams.add(s.listen(null, cancelOnError: true));
+      _ioSubscriptions.add(s.listen(null, cancelOnError: true));
     } on StateError catch (_) {}
   }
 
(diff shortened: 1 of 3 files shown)


lib/src/runner/browser/browser.dart
```diff
@@ -51,9 +51,12 @@ abstract class Browser {
   Future get onExit => _onExitCompleter.future;
   final _onExitCompleter = new Completer();
 
+  /// IO streams for the underlying browser process.
+  final _ioStreams = <StreamSubscription>[];
```
nex3: Nit: I'd call this `_ioSubscriptions`.

lib/src/runner/browser/browser.dart
```diff
@@ -51,9 +51,12 @@ abstract class Browser {
   Future get onExit => _onExitCompleter.future;
   final _onExitCompleter = new Completer();
 
+  /// IO streams for the underlying browser process.
```
nex3: Nit: "IO" -> "Standard IO"

lib/src/runner/browser/browser.dart
```diff
@@ -120,6 +123,12 @@ abstract class Browser {
   Future close() {
     _closed = true;
 
+    // If we don't manually close the stream the test runner can hang.
+    // For example this happens with Chrome Headless.
```
nex3: Did you file a `dart:io` bug for this? If so, refer to it here.

lib/src/runner/browser/browser.dart
grouma: Done.

lib/src/runner/browser/browser.dart
grouma: Done.

lib/src/runner/browser/browser.dart
grouma: Done.

PR #719, opened 2017-11-03
Fix fuzzy arrow warnings since they will become errors before too long.

munificent:
Context: https://github.com/dart-lang/sdk/issues/29630

### DIFF
--- a/lib/src/runner/browser/browser_manager.dart
+++ b/lib/src/runner/browser/browser_manager.dart
@@ -272,7 +272,7 @@ class BrowserManager {
   }
 
   /// The callback for handling messages received from the host page.
-  void _onMessage(Map message) {
+  void _onMessage(message) {
     switch (message["command"]) {
       case "ping":
         break;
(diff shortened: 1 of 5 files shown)


lib/src/runner/browser/browser_manager.dart
```diff
@@ -272,7 +272,7 @@ class BrowserManager {
   }
 
   /// The callback for handling messages received from the host page.
-  void _onMessage(Map message) {
```
natebosch: I don't understand this. Is it possible to fix the place where this is used rather than loosen the argument?

nex3: I'd prefer this as well. I'd like to avoid dynamic calls, even though in practice `--no-implicit-dynamic` isn't really in shape to be enabled yet.

munificent: > Is it possible to fix the place where this is used rather than loosen the argument?

Done, though it feels a little silly in this case. `_onMessage` is only called by that one place and `Map` isn't super precise either.

PR #721, opened 2017-11-08
Add flushEventQueue() and neverCalled

nex3:


natebosch:
FWIW I think `pumpEventQueue` is less misleading than `flushEventQueue` - after a `pump` I'd expect that there could still be events if it was very full, whereas after a `flush` I'd expect it to be empty. The doc comment is clear but I think this still has more potential for confusion than if the name is a closer match to behavior.

grouma:
LGTM but I agree with natebosch@ that `pumpEventQueue` is more descriptive than `flushEventQueue` especially when considering that `flushEventQueue` has an optional argument.

nex3:
Renamed to `pumpEventQueue()`.

PR #727, opened 2017-11-30
Include browser standard output in crash messages

### LINKED ISSUE #726: Better error messaging when a browser fails.
When a browser fails the test runner will print out the cryptic exit code, e.g. 

`Failed to run Chrome: Chrome failed with exit code -6`

Often times there is useful information on the `stderr` and `stdout`. It would be very useful to pipe this information to the console as well.

Likely need to modify the logic here: https://github.com/dart-lang/test/blob/master/lib/src/runner/browser/browser.dart#L78

nex3:
Closes #726

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.12.28+1
+
+* When a browser crashes, include its standard output in the error message.
+
 ## 0.12.28
 
 * Add a `pumpEventQueue()` function to make it easy to wait until all
--- a/lib/src/runner/browser/browser.dart
+++ b/lib/src/runner/browser/browser.dart
@@ -3,9 +3,11 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:stack_trace/stack_trace.dart';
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.12.28+1
```
natebosch: This doesn't match the change in the pubspec version

nex3: That's intentional—`0.12.28+1-dev` isn't a valid semantic version, so we use the next closest option, `0.12.29-dev`.

natebosch: Do we intend to publish in this state?
If not, why do we need a valid semantic version?

nex3: No, the point of the `-dev` flag is to indicate that we *don't* intend to publish. However, pub will still parse the version number, and it may even be relevant for version solving if users have a git dependency on this package.

grouma: I missed this the first time around. What is preventing us from publishing?

nex3: I'm just waiting for a few more changes to land so that we can publish them all together.

grouma: sg

PR #728, opened 2017-12-01
Add support for importing from node_modules on Node.js

nex3:


daniel-v:
Agree, cool feature! 
@nex3 would you mind sharing your use-case for this?

nex3:
The immediate use-case is for testing Dart Sass, but in general it's important for Node.js interop to be able to actually use Node.js packages.

PR #733, opened 2017-12-13
Fix a couple of strong mode runtime cast errors.

munificent:
In strong mode, a generic type instantiated with dynamic is not a
subtype of all types. You can't pass a List<dynamic> to something
expecting, say, List<int>.

These errors are usually detected statically, and most of those have
been fixed. However, sometimes this becomes a runtime cast, as in:

```dart
main() {
  // Store a List<dynamic> in a variable of type dynamic.
  dynamic d = [];

  // Implicit runtime downcast from dynamic to List<String>.
  List<String> s = d;
}
```

In order to ease the migration to strong mode, DDC has been ignoring
these cast failures when they involve certain commonly used types. We
are now in the process of actively fixing those errors.

More context: https://github.com/dart-lang/sdk/issues/27223

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.12.30
+
+* Fix strong mode runtime cast failures.
+
 ## 0.12.29
 
 * Node.js tests can now import modules from a top-level `node_modules`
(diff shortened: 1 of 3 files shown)


munificent:
@vsm, this takes the change you were testing locally and upstreams it into the package.

I validated this by:

1. Change DDC to not ignore these errors. Build the SDK.
2. Run `pub run test test/runner/pub_serve_test.dart -n dartdevc`.
3. Behold the tests fail because of cast errors.
4. Fix the errors in `lib/src/runner/remote_listener.dart`.
5. Run `pub run test test/runner/pub_serve_test.dart -n dartdevc` again.
6. All pass. 🎉

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.12.30
```
nex3: I think this should be 0.12.29+1

munificent: Done. (And in the pubspec.)

PR #736, opened 2017-12-18
Bump to 0.12.30 before publishing.

munificent:
Publishing the current stuff so I can roll it internally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.12.30
+## 0.12.29+1
 
 * Fix strong mode runtime cast failures.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.12.29+1
+## 0.12.30
```
nex3: Why not 0.12.29+1?

munificent: The pubspec was already at 0.12.30-dev.

nex3: That's just because `0.12.29+1-dev` isn't valid. There aren't any changes in the changelog that would require a minor version bump.

munificent: Ah, I see. Done.

PR #739, opened 2017-12-19
fix analysis issues

grouma:


natebosch:
@kevmoo  - maybe we shouldn't be changing for bleeding edge SDK before it's imported to google?

grouma:
Current SDK inside Google3 is "v2_0_0_dev_8_0". What version of the SDK did @kevmoo's change require?

lib/src/frontend/expect.dart
```diff
@@ -157,6 +157,9 @@ Future _expect(actual, matcher,
     reason ??= '$e at $trace';
   }
   fail(formatter(actual, matcher, reason, matchState, verbose));
+  // Fix missing return. The above fail() will always throw, but static
+  // analysis doesn't know this.
+  throw "Unreachable";
```
kevmoo: Add an `ignore: ...xyz` here for the dev SDK where this is reported...maybe w/ a todo?

kevmoo:
We figured it out -dev.12 I believe

nex3:
I don't want to regress analysis on bleeding edge. I'd rather temporarily have changes in Google3 only while we're waiting for the SDK to be rolled in than temporarily have changes externally as well.

kevmoo:
Google3 is now at -dev.18 – should be good, right?

kevmoo:
...or just delete the branch entirely, @grouma ?

Issue #48, opened 2015-03-31
Provide a programmatic API


nex3:
Currently the only APIs that `unittest` exposes publicly are the frontend APIs used to define tests: `test()`, `group()`, `expect()`, and so on. However, it has a lot of backend code that would be useful for external packages to extend the platform. Plugins are the most obvious example of this, since they need to at least implement some backend classes and likely call out to other APIs, but one could also imagine something like a custom runner that spins up a long-lived server and re-runs tests when their sources change.

grouma:
This is quite generic and is covered by other issues.

PR #741, opened 2017-12-28
Custom reporter

DmitryKrutskikh:
Added ability to define user custom reporter

### DIFF
--- a/lib/test.dart
+++ b/lib/test.dart
@@ -13,6 +13,7 @@ import 'src/frontend/timeout.dart';
 import 'src/runner/configuration/suite.dart';
 import 'src/runner/engine.dart';
 import 'src/runner/plugin/environment.dart';
+import 'src/runner/reporter.dart';
 import 'src/runner/reporter/expanded.dart';
 import 'src/runner/runner_suite.dart';
 import 'src/utils.dart';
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -13,3 +13,5 @@ packages
 # Include when developing application packages.
 pubspec.lock
 .packages
+
+.idea
```
kevmoo: Please revert this file change. `.idea` files should be ignored in your global .gitignore file

DmitryKrutskikh: @kevmoo done

lib/test.dart
```diff
@@ -41,6 +42,12 @@ export 'src/frontend/utils.dart';
 /// This is used if a test file is run directly, rather than through the runner.
 Declarer _globalDeclarer;
 
+typedef Reporter UserDefinedReporterCreateCallback(Engine engine);
+UserDefinedReporterCreateCallback _userDefinedReporterCallback;
+void setUserDefinedReporter(UserDefinedReporterCreateCallback callback) {
```
natebosch: Avoid methods named `setSomething`. You can have a `set` method at the top level, but since this isn't doing any extra work we can also make the variable public instead of private.

DmitryKrutskikh: @natebosch done

nex3:
This approach won't work, for a couple reasons:

* It only works when running tests via `dart path/to/test.dart`. This is a semi-unofficial way of invoking the test runner, so it doesn't make sense to add features that only work there.

* It exposes the private internal `Engine` and `Reporter` APIs. Even when #48 is done, I don't want to make these packages part of the API for the `test` package itself because additional API surface makes it harder to make large changes without breaking the world.

Issue #749, opened 2018-01-08
Should be able to override the default line length when wrapping the test runner


jakemac53:
If you want to wrap the test runner in a different process the compact reporter doesn't work properly because the line length falls back to the default of 200 which ends up doing weird things if your terminal width is <200 lines.

We should be able to override this with an explicit value, or come up with some other solution for when the test runner isn't directly connected to a terminal.

One way to deal with this would be a global environment variable, DEFAULT_TERMINAL_LINES or similar.

nex3:
Closing since the VM now supports spawning with a PTY.

PR #750, opened 2018-01-08
Add --line-length flag which overrides the maximum line length

jakemac53:
Potential fix for https://github.com/dart-lang/test/issues/749

Unblocks https://github.com/dart-lang/build/issues/821

### DIFF
--- a/lib/src/executable.dart
+++ b/lib/src/executable.dart
@@ -75,13 +75,13 @@ main(List<String> args) async {
   try {
     configuration = new Configuration.parse(args);
   } on FormatException catch (error) {
-    _printUsage(error: error.message);
+    _printUsage(Configuration.empty, error.message);
     exitCode = exit_codes.usage;
     return;
   }
 
   if (configuration.help) {
-    _printUsage(configuration: configuration);
+    _printUsage(configuration);
     return;
   }
 
@@ -128,9 +128,9 @@ main(List<String> args) async {
       .toList();
   if (undefinedPresets.isNotEmpty) {
     _printUsage(
-        configuration: configuration,
-        error: "Undefined ${pluralize('preset', undefinedPresets.length)} "
-            "${toSentence(undefinedPresets.map((preset) => '"$preset"'))}.");
+        configuration,
+        "Undefined ${pluralize('preset', undefinedPresets.length)} "
+        "${toSentence(undefinedPresets.map((preset) => '"$preset"'))}.");
     exitCode = exit_codes.usage;
     return;
   }
@@ -151,9 +151,9 @@ transformers:
   if (!configuration.explicitPaths &&
       !new Directory(configuration.paths.single).existsSync()) {
     _printUsage(
-        configuration: configuration,
-        error: 'No test files were passed and the default "test/" '
-            "directory doesn't exist.");
+        configuration,
+        'No test files were passed and the default "test/" '
+        "directory doesn't exist.");
     exitCode = exit_codes.data;
     return;
   }
@@ -204,7 +204,7 @@ transformers:
 ///
 /// If [error] is passed, it's used in place of the usage message and the whole
 /// thing is printed to stderr instead of stdout.
-void _printUsage({Configuration configuration, String error}) {
+void _printUsage(Configuration configuration, [String error]) {
   var wordWrap = configuration?.wordWrap ?? (message) => message;
   var output = stdout;
 
--- a/lib/src/executable.dart
+++ b/lib/src/executable.dart
@@ -205,7 +205,6 @@ transformers:
 /// If [error] is passed, it's used in place of the usage message and the whole
 /// thing is printed to stderr instead of stdout.
 void _printUsage(Configuration configuration, [String error]) {
-  var wordWrap = configuration?.wordWrap ?? (message) => message;
   var output = stdout;
 
   var message = "Runs tests in this package.";
@@ -214,7 +213,7 @@ void _printUsage(Configuration configuration, [String error]) {
     output = stderr;
   }
 
-  output.write("""${wordWrap(message)}
+  output.write("""${configuration.wordWrap(message)}
 
 Usage: pub run test [files or directories...]
 
--- a/lib/src/executable.dart
+++ b/lib/src/executable.dart
@@ -75,13 +75,13 @@ main(List<String> args) async {
   try {
     configuration = new Configuration.parse(args);
   } on FormatException catch (error) {
-    _printUsage(error.message);
+    _printUsage((line) => line, error.message);
     exitCode = exit_codes.usage;
     return;
   }
 
   if (configuration.help) {
-    _printUsage();
+    _printUsage(configuration.wordWrap);
     return;
   }
 
@@ -127,7 +127,9 @@ main(List<String> args) async {
       .where((preset) => !configuration.knownPresets.contains(preset))
       .toList();
   if (undefinedPresets.isNotEmpty) {
-    _printUsage("Undefined ${pluralize('preset', undefinedPresets.length)} "
+    _printUsage(
+        configuration.wordWrap,
+        "Undefined ${pluralize('preset', undefinedPresets.length)} "
         "${toSentence(undefinedPresets.map((preset) => '"$preset"'))}.");
     exitCode = exit_codes.usage;
     return;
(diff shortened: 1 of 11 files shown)


nex3:
This seems like a pretty hacky way to work around a lack of functionality in `dart:io`. The way other languages handle this is to just attach the TTY to the child process so that it can accurately query information from it, and Dart should do the same. dart-lang/sdk#24716 tracks exposing this functionality.

I'm not opposed to exposing a workaround for this in `test`, but I'd like to see some commitment to addressing the underlying issue as well so we can remove the workaround.

jakemac53:
> This seems like a pretty hacky way to work around a lack of functionality in dart:io. The way other languages handle this is to just attach the TTY to the child process so that it can accurately query information from it, and Dart should do the same. dart-lang/sdk#24716 tracks exposing this functionality.
> 
> I'm not opposed to exposing a workaround for this in test, but I'd like to see some commitment to addressing the underlying issue as well so we can remove the workaround.

Agreed on all fronts, but I don't see a better workaround in the short/medium term.

nex3:
As far as workarounds go, I'd rather add a flag than an environment variable.

@mit-mit @anders-sandholm Can we get a commitment to work on this in `dart:io` once the VM team's code yellow is over?

jakemac53:
> As far as workarounds go, I'd rather add a flag than an environment variable.

The reason I went with the environment variable is this is a top level field. It would be relatively difficult to support changing its value with a flag outside of adding a global public setter for it which sounds worse.

Another option could be to make it a setting on the `CompactReporter` itself but then you could get in an inconsistent state if something else started using the `lineLength` and didn't know about this other flag.

jakemac53:
I also considered using the `-D` "environment" variables but we can't pass those through `pub run ...`. That might be a bit cleaner than an actual environment variable for this case if we wanted to look into adding support for passing those through pub run generally.

nex3:
> The reason I went with the environment variable is this is a top level field.

Pretty much everywhere in the runner has access to the `Configuration` object, so it should be possible to move it into that.

jakemac53:
> Pretty much everywhere in the runner has access to the Configuration object, so it should be possible to move it into that.

Ok, I can update it to be a field on there. What do you think about a `--line-length` flag and corresponding `lineLength` field on `Configuration`? It would default to the existing `lineLength` top level field if not supplied.

jakemac53:
Actually it looks like the top level `wordWrap` function also uses the top level `lineLength` getter. I would probably need to change that to take a required lineLength argument as well, at which point it seems like deleting the top level `lineLength` field altogether might make the most sense?

I can just move that logic to the `Configuration` class instead.

nex3:
I'd say get rid of the top-level `lineLength` and make `wordWrap()` a method on `Configuration`.

jakemac53:
Ok, PTAL. If you are generally ok with this I can update tests etc.

lib/src/executable.dart
```diff
@@ -200,7 +204,7 @@ transformers:
 ///
 /// If [error] is passed, it's used in place of the usage message and the whole
 /// thing is printed to stderr instead of stdout.
-void _printUsage([String error]) {
+void _printUsage(String wordWrap(String message), [String error]) {
```
nex3: I think it would be cleaner to just pass in the `Configuration` here.

lib/src/runner/configuration.dart
```diff
@@ -27,6 +28,19 @@ import 'configuration/values.dart';
 /// The key used to look up [Configuration.current] in a zone.
 final _currentKey = new Object();
 
+/// The default line length for output.
+final int _defaultLineLength = () {
```
nex3: This should go in `lib/src/configuration/values.dart`.

lib/src/runner/configuration.dart
```diff
@@ -124,6 +138,10 @@ class Configuration {
   Glob get filename => _filename ?? defaultFilename;
   final Glob _filename;
 
+  /// The maximum length to print per line.
+  int get lineLength => _lineLength ?? _defaultLineLength;
```
nex3: Add a TODO to make this global again when dart-lang/sdk#24716 is fixed.

lib/src/runner/configuration/args.dart
```diff
@@ -82,6 +82,8 @@ final ArgParser _parser = (() {
   parser.addOption("timeout",
       help: 'The default test timeout. For example: 15s, 2x, none',
       defaultsTo: '30s');
+  parser.addOption('line-length',
+      help: 'Overrides the max width of lines to print.');
```
nex3: Mark this as hidden and add a comment that it will be removed in the future and that anyone using it should notify the `test` authors so they can be warned when removal is impending.

lib/src/executable.dart
jakemac53: Done, I changed it to an optional named arg as well since there is one location where we don't have a configuration (if it fails to parse). Also changed error to a named instead of positional arg as a result of that.

lib/src/runner/configuration.dart
jakemac53: done

lib/src/runner/configuration.dart
jakemac53: done

lib/src/runner/configuration/args.dart
jakemac53: Done, I also added a comment with a list of current users of this feature (currently just package:build_runner) so it can be tracked.

lib/src/runner/configuration/args.dart
```diff
@@ -82,8 +82,14 @@ final ArgParser _parser = (() {
...
       help: 'The default test timeout. For example: 15s, 2x, none',
       defaultsTo: '30s');
+  // Existing users to be notified prior to removal:
+  //   - package:build_runner
   parser.addOption('line-length',
-      help: 'Overrides the max width of lines to print.');
+      help: 'Overrides the max width of lines to print.\n'
+          'This is a temporary option which will be removed in the future.'
+          'Please notify the package:test authors if you are using this option '
+          'so you can be notified before it is removed.',
```
nex3: Adding a help string doesn't really do anything for a hidden option. I'd just move this into a comment.

lib/src/executable.dart
nex3: You can just use `Configuration.empty` when the args fail to get the default configuration.

lib/src/runner/configuration/args.dart
jakemac53: done

lib/src/executable.dart
jakemac53: done

lib/src/runner/configuration/args.dart
```diff
@@ -82,14 +82,14 @@ final ArgParser _parser = (() {
...
+  //
+  // Existing users:
   //   - package:build_runner
   parser.addOption('line-length',
-      help: 'Overrides the max width of lines to print.\n'
-          'This is a temporary option which will be removed in the future.'
-          'Please notify the package:test authors if you are using this option '
-          'so you can be notified before it is removed.',
-      hide: true);
+      help: 'Overrides the max width of lines to print.', hide: true);
```
nex3: I'd move this help string to the comment as well.

jakemac53: done

jakemac53:
Added a unittest for the Configuration class, did you want an integration test as well?

jakemac53:
nvm went ahead and added an integration test as well, PTAL

lib/src/executable.dart
```diff
@@ -200,7 +204,8 @@ transformers:
 ///
 /// If [error] is passed, it's used in place of the usage message and the whole
 /// thing is printed to stderr instead of stdout.
-void _printUsage([String error]) {
+void _printUsage(Configuration configuration, [String error]) {
+  var wordWrap = configuration?.wordWrap ?? (message) => message;
```
nex3: I don't think you need the null-checks here anymore.

test/runner/compact_reporter_test.dart
```diff
@@ -163,6 +163,15 @@ void main() {
         +0 -1: Some tests failed.""");
   });
 
+  test("Respects the --line-length argument", () {
+    return _expectReport("""
+        test('A test with a pretty long name', () {});""", """
```
nex3: Nit: Since this `test()` invocation is just one line, there's no need to make it a multiline string.

lib/src/executable.dart
jakemac53: done

test/runner/compact_reporter_test.dart
jakemac53: done

kevmoo:
ProcessStartMode.INHERIT_STDIO landed in 2.0.0-dev.20 - dart-lang/sdk@6523896

jakemac53:
Closing this pr as I have confirmed everything works as expected with ProcessStartMode.INHERIT_STDIO!

PR #754, opened 2018-01-20
prep for release

grouma:

PR #755, opened 2018-01-22
Prepare for fix of https://github.com/dart-lang/sdk/issues/31963

stereotype441:
When https://github.com/dart-lang/sdk/issues/31963 is fixed, the analyzer will become more strict about reporting the `TOP_LEVEL_FUNCTION_LITERAL_BLOCK` hint.  This change will prevent the test package from triggering the hint once it is reported more strictly.


grouma:
Will this need to be published immediately? If so can you update the pubspec and changelog as well?

grouma:
Just realized https://github.com/dart-lang/test/pull/754 didn't go in yet. LGTM.

PR #756, opened 2018-01-26
Don't use Stream.firstWhere(defaultValue:) as it is being renamed

srawlins:
As Stream.firstWhere's `defaultValue` parameter is being renamed `orElse`, and this is the only use case in this package, and it's for Dartium, I reckon it will make a few things easier if we just refactor this line to use try/catch instead of `defaultValue`. This file will presumably be deleted after Dart 2.0 final is released.

### DIFF
--- a/lib/src/runner/browser/dartium.dart
+++ b/lib/src/runner/browser/dartium.dart
@@ -150,6 +150,9 @@ class Dartium extends Browser {
     /// check whether it's actually connected to an isolate, indicating that
     /// it's the observatory for the main page. Once we find the one that is, we
     /// cancel the other requests and return it.
+    ///
+    /// Use try/catch rather than `defaultValue` in `firstWhere` since the
+    /// parameter name will change in Dart 2.0.
     try {
       return (await inCompletionOrder(operations)
           .firstWhere((url) => url != null)) as Uri;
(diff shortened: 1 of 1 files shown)


lib/src/runner/browser/dartium.dart
```diff
@@ -150,8 +150,12 @@ class Dartium extends Browser {
     /// check whether it's actually connected to an isolate, indicating that
     /// it's the observatory for the main page. Once we find the one that is, we
     /// cancel the other requests and return it.
-    return (await inCompletionOrder(operations)
-        .firstWhere((url) => url != null, defaultValue: () => null)) as Uri;
+    try {
+      return (await inCompletionOrder(operations)
+          .firstWhere((url) => url != null)) as Uri;
+    } on StateError catch (_) {
+      return null;
```
nex3: Add a comment about why you're doing `try`/`catch` rather than `orElse`.

srawlins: Done! Thanks

zoechi: wouldn't `var result = ...where((url) => url != null)); return result.isNotEmpty ? result.first : null;` be better than `try`/`catch`? I'd expect `isNotEmpty` only to look for the first result like `firstWhere` instead of materializing the whole result (but I haven't checked myself yet).

nex3: You can't call multiple read methods on the same stream. `result.isNotEmpty` will create a subscription, which will make `result.first` fail.

zoechi: I see. Didn't realize it's a stream. Sorry for the noise.

PR #758, opened 2018-02-07
Don't load test suites with excluded tags

### LINKED ISSUE #753: Don't load test suites whose tags are excluded
Currently, if a suite has a tag that's excluded by the configuration, that test suite is still loaded even though we theoretically know in advance that we won't run any of its tests. This can take a meaningful amount of time on platforms that require compilation.

We can see this if we run `pub run test -p chrome -x tag` with the following test file:

```dart
@Tags(const ["tag"])

void main() {
  test("success", () {});
}
```

nex3:
Closes #753

### DIFF
--- a/lib/src/runner/loader.dart
+++ b/lib/src/runner/loader.dart
@@ -211,6 +211,10 @@ class Loader {
       return;
     }
 
+    if (_config.suiteDefaults.excludeTags.evaluate(suiteConfig.metadata.tags)) {
+      return;
+    }
+
     if (_config.pubServeUrl != null && !p.isWithin('test', path)) {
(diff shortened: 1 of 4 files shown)


lib/src/runner/loader.dart
```diff
@@ -211,6 +211,10 @@ class Loader {
       return;
     }
 
+    if (_config.suiteDefaults.excludeTags.evaluate(suiteConfig.metadata.tags)) {
+      return;
+    }
```
grouma: I'm guessing the return can't fit on a single line under 80 characters? Hard to tell in the Github review UI.

nex3: Yep.

PR #762, opened 2018-02-12
Remove unnecessary cast

keertip:
In 2.0.0-dev.23.0, having the explicit cast is an error. 

@nex3, waiting for travis to see if this passes with stable.


matanlurey:
I'm guessing this PR fixing Dartium is now stale.

PR #771, opened 2018-02-22
Fix a hang when a Node test failed to compile

nex3:
This wasn't caught by Travis due to our workaround for #599.

### DIFF
--- a/lib/src/runner/node/platform.dart
+++ b/lib/src/runner/node/platform.dart
@@ -96,29 +96,35 @@ class NodePlatform extends PlatformPlugin
       TestPlatform platform, SuiteConfiguration suiteConfig) async {
     var server = await MultiServerSocket.loopback(0);
 
-    var pair = await _spawnProcess(path, platform, suiteConfig, server.port);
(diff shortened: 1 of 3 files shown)


lib/src/runner/node/platform.dart
```diff
@@ -96,29 +96,35 @@ class NodePlatform extends PlatformPlugin
...
+          .transform(jsonDocument)
+          .transformStream(
+              new StreamTransformer.fromHandlers(handleDone: (sink) {
+        if (process != null) process.kill();
+        sink.close();
+      }));
+
+      return new Pair(channel, pair.last);
+    } catch (_) {
+      server.close().catchError((_) {});
```
grouma: Should we add additional messaging here notifying that the error is coming from the node compilation or will that be obvious from the stderr and current error?

nex3: I don't think it's actually possible for this to throw, and if it does we don't actually care about the error. I just have `catchError()` here to be absolutely sure that it doesn't confuse the user.

Issue #772, opened 2018-02-22
Tests hang on sdk versions >2.0.0-dev.28.0


jakemac53:
Interestingly, if you use `--pause-after-load` the tests don't hang, which indicates to me a possible race condition somewhere.

We are also using `--precompiled`, I haven't confirmed if it is related to that specifically or not. Example travis run [here](https://travis-ci.org/dart-lang/build/jobs/344839092).

You can repro this by checking out the build repo, changing to the `e2e_example` directory, and running `pub run build_runner test -- -p chrome`.

jakemac53:
Fwiw, this also does not reproduce on windows, only on linux/mac.

grouma:
I did a little debugging and it appears that this issue is only occurring in Chrome Headless. Normally we would forward any errors to the console but Headless gets in an odd state where it dies and restarts immediately. I was able to capture the initial error though:

[...]

PR #773, opened 2018-02-23
Remove headless flag

grouma:
We are seeing issues with the increased SDK size and headless: https://github.com/dart-lang/test/issues/772#issuecomment-367910465

Disabling headless for now to unblock our users. We can add an option in the future.


grouma:
Travis is failing due to https://github.com/dart-lang/test/issues/774. Squashing and merging regardless to unblock our users.

PR #775, opened 2018-02-23
Allow Chrome's headless mode to be configured

nex3:
Also re-enable headless mode by default.

See #772 for an example of a failure that only occurs in headless
mode.


nex3:
@matanlurey If you add

```yaml
override_platforms:
  chrome:
    headless: false
```

to Angular's `dart_test.yaml`, that'll keep headless mode disabled for you once this change gets released.

Issue #723, opened 2017-11-21
Share dart_test.yaml configuration across packages


matanlurey:
AngularDart (https://github.com/dart-lang/angular) uses a mono repository, which means we have several packages in this single repository with a shared set of analysis settings, and we manually try to keep dependencies in sync as well.

I'm working on using `dart_test.yaml` instead of passing duplicate flags across `pub run test` invocations. For example, I'd like to define the following presets/tags and share them across multiple packages:

```yaml
presets:
  # Only run tests that need to run within the browser.
  # * ... make the stack traces nicer.
  # * ... all tests in the browser likely need code generation.
  # * ... we currently only run these tests on chrome (today).
  browser:
    js_trace: true
    include_tags: codegen
    platform: chrome

  # Only run tests that need to run within the standalone VM.
  compiler:
    platform: vm

  # When running on travis:
  # * ... skip tests that don't work in this environment.
  # * ... increase timeouts (tests run slower here).
  # * ... do not use concurrency (we run out of resources).
  travis:
    exclude_tags: skip_on_travis
    timeout: 1.5x
    concurrency: 1

tags:
  # Tests that require code generation to be executed.
  #
  # * ... it's expected that these are manually tagged by the author.
  # * ... pass `--precompiled build` to `pub run test`.
  codegen:
    precompiled: build

  # Tests that are expected to fail on travis. These should be kept tightly
  # curated with tracking bugs, and are used to temporarily get our build
  # status to "green".
  skip_on_travis:
```

The only option I see today is writing a script to keep the following in sync:

* pkg_a/dart_test.yaml
* pkg_b/dart_test.yaml
* pkg_c/dart_test.yaml

The simplest solution I can think of is add a `--config` attribute to specify the path for `dart_test.yaml`. It could even default to current path logic if not specified. If added, I would use it like this:

```sh
$ pub run test --config ../dart_test_common.yaml -P compiler
```

Another alternative would be able to include and merge configuration files:

```yaml
include: ../dart_test_common.yaml
```

In my opinion, `--config` looks easier and would be more bullet-proof (don't have to deal with how to merge files/keys, for example), but I'm open to discussing both and any other options.

nex3:
I'm leaning towards the `include` option because it fits `test`'s goal of not requiring users to pass a bunch of boilerplate flags to get tests working right. We already have logic for merging config files (we use this for merging global with local config and for merging command-line flags with file configuration), so the implementation would be straightforward.

matanlurey:
[...]

PR #777, opened 2018-02-24
Start adding an include: ... option.

matanlurey:
(Potentially) closes https://github.com/dart-lang/test/issues/723.

I realize there are a lot of moving parts in this package I don't really understand (local config, global config, runner config), and many different ways I could instrument and test this.

... so this is my "best start", and I'd love feedback on how we can get this in.


lib/src/runner/configuration.dart
```diff
@@ -50,6 +50,10 @@ class Configuration {
   bool get pauseAfterLoad => _pauseAfterLoad ?? false;
   final bool _pauseAfterLoad;
 
+  /// The path to a file to load as a base configuration before this one.
+  String get includePath => _includePath;
+  final String _includePath;
```
natebosch: No need for a private field and getter when it's final.

matanlurey: Good catch, done!

grouma:
Looks like you broke the build as it is looking for an invalid field. The approach looks good to me though. You'll want to document this change in the readme and the changelog.

matanlurey:
Yup, my mistake. Running again now @grouma

matanlurey:
Looks like a flake but otherwise green.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+* Added an `include` field for configuration, which may specify another YAML
```
nex3: Please give this a `## 0.12.32` header and update the pubspec to `0.12.32-dev`.

lib/src/runner/configuration.dart
```diff
@@ -50,6 +50,9 @@ class Configuration {
   bool get pauseAfterLoad => _pauseAfterLoad ?? false;
   final bool _pauseAfterLoad;
 
+  /// The path to a file to load as a base configuration before this one.
+  final String includePath;
```
nex3: Rather than adding a field that only gets set and used in `load.dart`, it's probably cleaner to just handle the `include` field specially in the `load()` function.

lib/src/runner/configuration/load.dart
```diff
@@ -49,13 +49,18 @@ Configuration load(String path, {bool global: false}) {
...
   if (document.value == null) return Configuration.empty;
 
-  if (document is! Map) {
-    throw new SourceSpanFormatException(
-        "The configuration must be a YAML map.", document.span, source);
+  if (document is YamlMap) {
+    var loader = new _ConfigurationLoader(document, source, global: global);
+    var config = loader.load();
+    if (config.includePath != null) {
+      var basePath = p.normalize(p.join(path, '../', config.includePath));
```
nex3: I'd write this as `p.join(p.dirname(path), config.includePath)`.

test/runner/configuration/include_test.dart
```diff
@@ -0,0 +1,30 @@
...
+          concurrency: 3
+        '''),
+      ]),
+    ]).create();
+    var path = p.join(d.sandbox, 'repo', 'pkg', 'dart_test.yaml');
+    var config = new Configuration.load(path);
+    expect(config.filename.pattern, 'test_*.dart');
+    expect(config.concurrency, 3);
+  });
+}
```
nex3: Also test:

* That fields specified in both configs are merged. 
* That an included file can include another file.
* That a non-string value is handled gracefully.
* That a non-existent path is handled gracefully.
* That an `include` field isn't allowed in a test configuration context.

matanlurey: Done!

lib/src/runner/configuration/load.dart
matanlurey: Wow, much better. Done.

lib/src/runner/configuration.dart
matanlurey: I wasn't sure if its useful to have this just to be able to make configuration issues more discoverable (i.e. "this file originated from that file"). Happy to make that change though, so done.

CHANGELOG.md
matanlurey: Done

test/runner/configuration/include_test.dart
```diff
@@ -27,4 +26,64 @@ void main() {
...
+          include: 3 # Oops!
+        '''),
+      ]),
+    ]).create();
+    var path = p.join(d.sandbox, 'repo', 'pkg', 'dart_test.yaml');
+    expect(() => new Configuration.load(path), throwsFormatException);
+  });
+
+  test('should not allow an include field in a test config context', () async {
+    expect(() => new Configuration.parse(['--include=dart_test.yaml']),
```
nex3: I meant like:

```yaml
tags:
  foo:
    include: ../dart_test.yaml
```

Come to think of it, we should also test that:

```yaml
presets:
  foo:
    include: preset.yaml
```

works as expected.

matanlurey: Done, thanks!

test/runner/configuration/include_test.dart
```diff
@@ -83,7 +83,30 @@ void main() {
...
+      d.dir('pkg', [
+        d.file('dart_test.yaml', r'''
+          presets:
+            foo:
+              include: ../dart_test.yaml
+        '''),
+      ]),
+    ]).create();
+    var path = p.join(d.sandbox, 'repo', 'pkg', 'dart_test.yaml');
+    expect(() => new Configuration.load(path), returnsNormally);
```
nex3: As written, this should fail due to `../dart_test.yaml` not existing.

This test should assert that the preset contains configuration from the corresponding file.

lib/src/runner/configuration/load.dart
```diff
@@ -49,13 +49,25 @@ Configuration load(String path, {bool global: false}) {
...
+    var config = loader.load();
+    var includePathNode = document.nodes["include"];
+    if (includePathNode != null && includePathNode.value is! String) {
+      throw new SourceSpanFormatException(
+          "The 'include' field must be omitted or a String",
+          includePathNode.span,
+          source);
+    }
+    if (includePathNode != null) {
+      var basePath = p.join(p.dirname(path), includePathNode.value);
```
nex3: I just realized: this should parse `path` as a URL, to ensure that it's cross-platform. We should also document that it's a relative `file:` URL.

nex3:
Also, it looks like there's still no test for the included file not existing.

test/runner/configuration/include_test.dart
matanlurey: Took a bit of refactoring, but done now.

lib/src/runner/configuration/load.dart
matanlurey: Done.

lib/src/runner/configuration/load.dart
```diff
@@ -537,7 +564,12 @@ class _ConfigurationLoader {
     _validate(node, "$name must be a map.", (value) => value is Map);
     var loader = new _ConfigurationLoader(node, _source,
         global: _global, runnerConfig: runnerConfig ?? _runnerConfig);
-    return loader.load();
+    var config = loader.load();
+    var include = (node.value as YamlMap)['include'];
+    if (include != null) {
+      config = _processIncludeIfAny(config, node);
+    }
```
nex3: Could this logic be moved into `_ConfigurationLoader.load()`, so it doesn't need to be duplicated between here and the top-level `load()` method?

Also, if `_processIncludeIfAny()` is in `_ConfigurationLoader`, it can use the `_parseNode()` helper method which encapsulates some of the logic it's doing.

lib/src/runner/configuration/load.dart
```diff
@@ -51,23 +51,50 @@ Configuration load(String path, {bool global: false}) {
...
-    if (includePathNode != null && includePathNode.value is! String) {
+    return _processIncludeIfAny(loader.load(), document);
+  }
+  throw new SourceSpanFormatException(
+      "The configuration must be a YAML map.", document.span, source);
+}
+
+/// If an `include` node is contained in [node], merges and returns [config].
+Configuration _processIncludeIfAny(Configuration config, YamlMap node) {
+  var source = node.span.sourceUrl.toFilePath(windows: Platform.isWindows);
```
nex3: I generally use `p.fromUri()` instead of `Uri.toFilePath()`. It's terser, it's more portable, and it's guaranteed to match other path operations' understandings of the current platform. It'll also automatically throw a nice error for non-file URLs. Same below.

lib/src/runner/configuration/load.dart
```diff
@@ -51,23 +51,50 @@ Configuration load(String path, {bool global: false}) {
...
+    } on FormatException catch (_) {
       throw new SourceSpanFormatException(
-          "The 'include' field must be omitted or a String",
+          "Not a valid file URL: \"${includePathNode.value}\"",
           includePathNode.span,
           source);
     }
-    if (includePathNode != null) {
-      var basePath = p.join(p.dirname(path), includePathNode.value);
+    if (includeValue.isAbsolute) {
```
nex3: I don't think we necessarily want to forbid absolute file URLs. I could imagine someone wanting to write `include: file:///usr/share/dart_test.yaml` in their global config, for example.

lib/src/runner/configuration/load.dart
matanlurey: Wow thanks, did not even know that exists.

lib/src/runner/configuration/load.dart
matanlurey: I'm not 100% sure of the best way to get it to use _parseNode. Do you want to take a look at the current state and suggest changes? I did move and refactor into `.load`. PTAL.

lib/src/runner/configuration/load.dart
matanlurey: Done.

lib/src/runner/configuration/load.dart
```diff
@@ -83,7 +81,41 @@ class _ConfigurationLoader {
...
+      Uri includeValue;
+      try {
+        includeValue = Uri.parse(includePathNode.value);
+      } on FormatException catch (_) {
+        throw new SourceSpanFormatException(
+            "Not a valid file URL: \"${includePathNode.value}\"",
+            includePathNode.span,
+            source);
+      }
+      var includePath = p.fromUri(includeValue);
```
nex3: You could write everything up to here as:

```dart
var includeNode = _document.nodes["include"];
if (includeNode == null) return Configuration.empty;

var includePath = _parseNode(includeNode, "include path", p.fromUri);
```

lib/src/runner/configuration/load.dart
```diff
@@ -83,7 +81,41 @@ class _ConfigurationLoader {
   Configuration load() => _loadGlobalTestConfig()
       .merge(_loadLocalTestConfig())
       .merge(_loadGlobalRunnerConfig())
-      .merge(_loadLocalRunnerConfig());
+      .merge(_loadLocalRunnerConfig())
+      .merge(_loadIncludeConfig());
```
nex3: Merging this last means the included config will override the local config, which isn't what we want.

This also makes me realize we should test that the local config takes precedence when they can't be merged.

matanlurey: Can you help me understand this?

I can move the `_loadIncludeConfig` after `_loadGlobalTestConfig`.

Then what? What should I test?

lib/src/runner/configuration/load.dart
matanlurey: Looks great, done!

lib/src/runner/configuration/load.dart
nex3: I think `_loadIncludeConfig()` should come first in the list, so that all other configuration overrides it.

You could test that

```yaml
# dart_test.yaml
include: other.yaml
concurrency: 5
```

```yaml
# other.yaml
concurrency: 10
```

ends up with concurrency 5.

matanlurey: Thanks! Done!

lib/src/runner/configuration/load.dart
```diff
@@ -81,10 +79,30 @@ class _ConfigurationLoader {
 
   /// Loads the configuration in [_document].
   Configuration load() => _loadGlobalTestConfig()
+      .merge(_loadIncludeConfig())
```
nex3: This should be the first in line:

```dart
_loadIncludeConfig()
      .merge(_loadGlobalTestConfig())
      .merge(_loadLocalTestConfig())
      .merge(_loadGlobalRunnerConfig())
      .merge(_loadLocalRunnerConfig());
```

`config1.merge(config2)` gives the fields in `config2` precedence over those in `config1`. We want the local config to be able to override any of the fields in the included one, which means we need to load the included config first and then merge local fields into it.

matanlurey: Makes sense. Done!

nex3:
[Looks like something may have broken in the merge from master](https://travis-ci.org/dart-lang/test/jobs/347048809).

matanlurey:
Oops. PTAL again @nex3

matanlurey:
Looks like failures are more Dartium flakes (?). Ping :)

test/runner/configuration/include_test.dart
```diff
@@ -0,0 +1,148 @@
...
+      d.dir('pkg', [
+        d.file('dart_test.yaml', r'''
+          tags:
+            foo:
+              include: ../dart_test.yaml
+        '''),
+      ]),
+    ]).create();
+    var path = p.join(d.sandbox, 'repo', 'pkg', 'dart_test.yaml');
+    expect(() => new Configuration.load(path), throwsFormatException);
```
nex3: I ran this locally as a sanity check, and it looks like this is getting a "cannot open file" error rather than a "this field isn't allowed here" error. Maybe these throw checks should be a little more specific about what error message is being produced...

matanlurey: Done. PTAL.

nex3:
Waiting to land this until #780 gets the tests green.

nex3:
LGTM, waiting on a clean test run then I'll land it.

nex3:
Thanks Matan!

matanlurey:
@nex3 Thanks for the review + helpful comments. Excited to start using this!

PR #778, opened 2018-02-27
Add a SuitePlatform class

nex3:
This encapsulates all the information about the platform on which a
suite runs, and which platform selectors can select for. Currently it
just includes the runtime and the operating system, but eventually it
could include whether we're running internally to Google, or which
compiler is in use.

This also renames `TestPlatform` to `Runtime`, which clarifies the
distinction between `TestPlatform` and `SuitePlatform`, and uses a
more specific word rather than a more general one. Unfortunately, it
does mean that we have some user-facing interfaces (the `--platform`
flag, the `override_platforms` and `define_platforms` configuration
fields) that now use different terminology than the test runner.

It would be possible to migrate those to refer to "runtime" as well,
but seeing as `--platform` in particular is very widely-used *and*
matches up nicely with `--preset` (`-p` vs `-P`) I'm inclined to think
it's not worth the effort.


lib/src/backend/suite_platform.dart
```diff
@@ -0,0 +1,42 @@
...
+import 'runtime.dart';
+
+/// The platform on which a test suite is loaded.
+class SuitePlatform {
+  /// The runtime that hosts the suite.
+  final Runtime runtime;
+
+  /// The operating system on which the suite is running.
+  ///
+  /// This will always be [OperatingSystem.none] if `runtime.isBrowser` is
```
grouma: Should we add another value to operating system that is 'Browser'?

lib/src/runner/configuration/args.dart
```diff
@@ -65,7 +65,7 @@ final ArgParser _parser = (() {
       abbr: 'p',
       help: 'The platform(s) on which to run the tests.\n'
           '[vm (default), '
-          '${allPlatforms.map((platform) => platform.identifier).join(", ")}]',
+          '${allRuntimes.map((runtime) => runtime.identifier).join(", ")}]',
```
grouma: Should we add some internal comments here stating platforms map to runtimes due to historical reasons?

lib/src/backend/suite_platform.dart
grouma: Fair enough.

PR #780, opened 2018-03-01
Stop testing content shell

nex3:
These tests have been flaking all over the place, and with the total
lack of support for content shell coming so soon it doesn't make sense
to spend the resources tracking down and fixing the root cause.


nex3:
I think we should leave it as long as we support 1.24 and it's not causing mysterious test failures.

PR #787, opened 2018-03-19
Workaround for async SDK issue

grouma:
- Provide work around for dart-lang/sdk/issues/32556
- Prep for release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.12.32+2
+
+* Provide a work around for dart-lang/sdk/issues/32556.
+
 ## 0.12.32+1
 
 * Fix a bug that broke content shell on Dart 1.24.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.12.32+2
+
+* Provide a work around for dart-lang/sdk/issues/32556.
```
nex3: This doesn't really communicate to downstream users what changed. Those users don't care what the SDK bug was, they care what user-visible thing was going wrong that is now fixed. Consider writing something like "Work around an SDK bug that caused timeouts in asynchronous code."

Also, for future reference, it's probably a good idea to use a link that users can follow rather than something like `dart-lang/sdk/issues/32556`.

grouma: Done.

grouma:
Added back the "unnecessary cast". This will have to be removed when dev becomes stable. I will also need to merge with a failing Travis due to https://github.com/dart-lang/test/issues/788.

PR #789, opened 2018-03-19
Refuse to run VM or Node tests with --precompiled

### LINKED ISSUE #784: Fail if `--precompiled` is used with `-p vm`
Today the behavior seems to be silently ignore the `--precompiled` and the resulting test failures (from non-importable files) are confusing. If we can support precompiled with the VM platform that would be awesome, but assuming it's not easy we should fall back on failing explicitly.

nex3:
Closes #784

### DIFF
--- a/lib/src/runner/loader.dart
+++ b/lib/src/runner/loader.dart
@@ -248,6 +248,11 @@ class Loader {
 
         try {
           var plugin = await memo.runOnce(_platformCallbacks[platform.runtime]);
+          if (platformConfig.precompiledPath != null &&
+              !platform.runtime.isBrowser) {
+            throw "--precompiled is only supported for browser platforms.";
+          }
+
           _customizePlatform(plugin, platform.runtime);
           var suite = await plugin.load(path, platform, platformConfig,
               {"platformVariables": _runtimeVariables.toList()});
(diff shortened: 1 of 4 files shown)


matanlurey:
Can we verify this with `angular`? We have packages we run:

`pub run build_runner test`

... that have a combination of VM and Browser tests today.

natebosch:
@matanlurey - those tests likely aren't running the way we'd want today anyway. I think it's better to fail fast than to do something unexpected, and we can separately run `pub run build_runner test -- -p chrome` and `pub run test -p vm` to get the current behavior.

matanlurey:
@natebosch Right, I just mean our build would go red. It would be nice to fix our build first.

natebosch:
Checking whether https://github.com/dart-lang/angular/pull/1092 is green, then will try with dependency overrides to pick up this change.

nex3:
> Did some more digging. Turns out that `pub run test -p vm` will attempt to run tests that don't work even if `dart_test.yaml` specifies that the only platform is `chrome`.

The `platform` field in `dart_test.yaml` specifies the default value for the `--platform` command-line flag, and is overridden if `--platform` is passed explicitly. The `test_on` field indicates which platforms the package supports at all, and it will cause `test` to print a warning and refuse to run tests if the user specifies an unsupported platform.

> How about for now we make this a warning instead and we can figure out a better long term solution later.

@matanlurey Will this work for Angular?

matanlurey:
Whatever makes/keeps https://github.com/dart-lang/angular/pull/1092 green is OK with me.

nex3:
@natebosch PTAL

nex3:
Will merge once the next dev release comes out and makes the tests green.

PR #790, opened 2018-03-19
Don't compile tests by hand if their precompiled file doesn't exist

### LINKED ISSUE #785: Fail if running with `--precompiled` and there is a missing .js file
Today the behavior seems to be running a compile with Dart2JS and running the test with that version which can silence issues that we expect to be failures.

nex3:
Closes #785


nex3:
Holding off on merging this until the next SDK dev release gets the tests green.

PR #799, opened 2018-04-05
simplify --precompiled to serve the precompiled directory directly 

jakemac53:
Essentially, this just passes the `--precompiled` directory through to the `Loader` as the `root`, which causes the server to serve that directory directly.

This means no special handlers are set up for the precompiled js, and the precompiled directory is treated as the source of truth, not the root directory (the root dir is not served at all any more). All tests are assumed to be precompiled if the --precompiled flag is passed, and no compilation will ever happen by package:test in that scenario.

I still need to test this with bazel to make sure it still works there as well, and I'm sure I probably broke some package:test tests as well, but sending this out for now to get a first look.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.12.34-dev
+
+* The `--precompiled` flag now serves all sources directly from the precompiled
+  directory, and will never attempt to do its own compilation.
+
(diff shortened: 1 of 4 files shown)


jakemac53:
Fwiw just tested internally with a basic test and everything works! :tada:

PR #801, opened 2018-04-06
Support --precompiled for vm tests

jakemac53:
Note that this does require a `.packages` file in the root of the precompiled directory, which build_runner does not currently provide.


lib/src/runner/vm/platform.dart
```diff
@@ -52,7 +53,38 @@ class VMPlatform extends PlatformPlugin {
   /// This isolate connects an [IsolateChannel] to [message] and sends the
   /// serialized tests over that channel.
   Future<Isolate> _spawnIsolate(String path, SendPort message) async {
-    if (_config.pubServeUrl == null) {
+    if (_config.suiteDefaults.precompiledPath != null) {
```
natebosch: [nit] this method is hard for me to follow, it looks like there are 3 top level cases. Could you see how it works to split this out?

```dart
 Future<Isolate> _spawnIsolate(String path, SendPort message) {
  if (_config.suiteDefaults.precompiledPath != null) {
    return _someNiceName(path, message);
  } else if (_config.pubServeUrl != null) {
    return _spawnIsolateWithPubUri(path, message);
  } else {
    return _someOtherNiceName(path, message);
  }
}

File: test/runner/precompiled_test.dart [hunk 2]
```diff
@@ -0,0 +1,103 @@
...
+
+      var test = await runTest(
+          ["-p", "chrome", "--precompiled=precompiled/", "test.dart"]);
+      expect(test.stdout,
+          containsInOrder(["+0: success", "+1: All tests passed!"]));
+      await test.shouldExit(0);
+    });
+  });
+
+  group("vm tests", () {
```
natebosch: Does this make sense with the tag "chrome" on line 6?

jakemac53:
Merging on red with same pre-existing failures.

PR #807, opened 2018-04-10
fix node tests in subdirs

jakemac53:
This was using the basename of the file before, oops!

### DIFF
--- a/lib/src/runner/node/platform.dart
+++ b/lib/src/runner/node/platform.dart
@@ -194,9 +194,7 @@ class NodePlatform extends PlatformPlugin
       int socketPort,
       String precompiledPath) async {
     StackTraceMapper mapper;
-    var jsPath =
-        p.join(precompiledPath, p.basename(testPath) + ".node_test.dart.js");
-
+    var jsPath = p.join(precompiledPath, testPath + ".node_test.dart.js");
     if (!suiteConfig.jsTrace) {
       var mapPath = jsPath + '.map';
       var resolver = await SyncPackageResolver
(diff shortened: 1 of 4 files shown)


lib/src/runner/node/platform.dart
```diff
@@ -194,9 +194,7 @@ class NodePlatform extends PlatformPlugin
       int socketPort,
       String precompiledPath) async {
     StackTraceMapper mapper;
-    var jsPath =
-        p.join(precompiledPath, p.basename(testPath) + ".node_test.dart.js");
-
+    var jsPath = p.join(precompiledPath, testPath + ".node_test.dart.js");
```
natebosch: single quotes for consistency with the line 2 below.

Would also prefer `'$testPath.node_test.dart.js'`

lib/src/runner/node/platform.dart
```diff
@@ -248,7 +246,7 @@ class NodePlatform extends PlatformPlugin
     try {
       return await Process.start(settings.executable,
           settings.arguments.toList()..add(jsPath)..add(socketPort.toString()),
-          environment: {'NODE_PATH': nodePath});
+          environment: {'NODE_PATH': nodePath}w);
```
natebosch: ?

pulyaevskiy: Curious what `w` means in this case?

jakemac53: lol

jakemac53: it means my finger slipped ;)

lib/src/runner/node/platform.dart
jakemac53: done, in general theres a crazy mix of single/double quotes here so eventually we might want to fix that repo-wide, but at least its consistent within this method now...

PR #810, opened 2018-04-12
Fix Dart 2 parse bugs in the VM

natebosch:
I'm not sure why the analyzer doesn't catch these, they appear to be
parse time bugs not runtime bugs when executing in the vm with
--preview-dart-2


kevmoo:
@natebosch – file an analyzer issue for this?

Issue #766, opened 2018-02-14
Support for node platform with --precompiled flag


pulyaevskiy:
I was trying to get `build_runner` to build and run tests in [node_interop](https://github.com/pulyaevskiy/node-interop/tree/master/node_interop) package but `pub run test --precompiled` doesn't seem to support node platform yet. It always falls back to compile the source itself.

I have created [build_node_compilers](https://pub.dartlang.org/packages/build_node_compilers) package which produces Node-compatible output. If I just build my tests and run them manually (e.g. `node build/test/some_test.dart.js`) it works fine. But it'd be really great to have it running through the standard tooling.

nex3:
`--precompiled` is *not* an official part of the test runner's API. It was added for Google-internal tooling, and any use of it outside of that context is unsupported. As such, I don't want to expand the set of things it happens to support and thereby encourage more people to use it.

That said, I'm definitely interested in getting the full matrix of build/DDC/platform configurations, and I have some experimental work in progress to that end that will hopefully come out on the sooner end.

pulyaevskiy:
That makes sense, thanks for quick response. If there is anywhere I can click "Subscribe" to follow progress on this front please let me know!

nex3:
I'll try to remember to post something here once code reviews start going up.

Out of curiosity, are you interested in build mostly for the speed improvements of DDC or because you're running tests against generated code?

pulyaevskiy:
Mostly to validate that my JS bindings work in both DDC and dart2js (as well as Dart code using these bindings). There are differences in how JS interop is interpreted by both compilers, the same code can run perfectly fine in dart2js but fail badly in DDC and vice versa.

So I needed a way to run tests with both of them. For now I'm just gonna use a little shell script as a workaround:

```sh
#!/bin/sh

pub run build_runner build --output=build/

for filename in test/*_test.dart; do
    node "build/$filename.js"
done
```

jakemac53:
I am going to see if I can tackle this - we are changing the semantics and support level for --precompiled so it can be used reliably with build_runner for all scenarios.

jakemac53:
@pulyaevskiy this is published now if you want to give it a shot! Please file issues if it doesn't work for you :).

jakemac53:
hmm I checkout out the node_interop issue and I am seeing a couple problems, sent out https://github.com/dart-lang/test/pull/807 which resolves the first issue.

pulyaevskiy:
[...]

PR #811, opened 2018-04-16
Fix node bootstrapping for dart2

jakemac53:
Fixes package:test related issues for https://github.com/dart-lang/test/issues/766


jakemac53:
@kevmoo added you for approval to drop Dart 1.x support.

Technically, I could remove types instead of adding them and get something that works in both but that just feels wrong and it seems like a good time to make the switch.

PR #812, opened 2018-04-17
Drop dartium support

natebosch:


.travis.yml
```diff
@@ -38,45 +38,16 @@ before_install:
...
-      dart_task: {test: --preset travis --total-shards 5 --shard-index 3 -x dartium}
-    - dart: dev
-      dart_task: {test: --preset travis --total-shards 5 --shard-index 4 -x dartium}
-
-    - dart: dev
-      dart_task: dartfmt
-    - dart: dev
-      dart_task: dartanalyzer
+ # Split the tests into five shards to help parallelize them across Travis workers.
+ - test: --preset travis --total-shards 5 --shard-index 0 -x dartium
```
jakemac53: We shouldn't need the `-x dartium` any more right?

jakemac53:
Should we also delete the `--pub-serve` support and related tests?

natebosch:
I think this is worth landing
@jakemac53 - can you take another look?

Issue #757, opened 2018-02-06
Consider setting up travis build stages


jakemac53:
By setting up things to run in stages on travis, you can prevent some jobs from running until another stage completes successfully. This helps limit travis usage so that everybody can get their jobs running more quickly, at a cost of slightly longer travis runs in some cases. In the long run the more packages we get to do this the better it should be overall for everybody.

One potential configuration I would propose is splitting the jobs into 3 stages:

1. analyze and dartformat checks
2. run tests on the latest dart sdk
3. run tests on other sdks

The tests look like they take around 10 minutes to run each shard, and the analyze/format steps only take around 1-2 minutes, so that would lead to a best case presubmit time of somewhere around 20 minutes, but its hard to say exactly how it would pan out in practice.

Current presubmit times appear to usually be just under 30 minutes in practice, although occasionally it looks like you do get full parallelism and the tests complete in around 10 minutes, so there would be a noticable worsening of the best case scenario with separating the tests into separate stages.

nex3:
What's the benefit of running the formatter check before the other stages? We'd usually expect the behavior of tests to be otherwise unaffected by whether they're well-formatted, right?

jakemac53:
> What's the benefit of running the formatter check before the other stages? We'd usually expect the behavior of tests to be otherwise unaffected by whether they're well-formatted, right?

Its true the tests won't be affected (hopefully!), but you are going to have to push another commit anyways so it prevents the tests from running an extra time. I can see how you might want the tests to run anyways though just to get an earlier signal, and that would be fine. You could even run the formatting check last if that is what you would prefer.

Personally, I like having the build fail as quickly as possible (and do as little work as possible) for trivial errors like that. I can quickly format,  push a new commit, and then follow the tests from there.

nex3:
I often push a commit and come back in half an hour to see how the tests did, and I think I'd be pretty frustrated to find that they didn't even run because I forgot to run something trivial. Plus, if everything is green except the formatter, I can safely reformat and then land the PR without waiting for the other tests anyway.

jakemac53:
> I often push a commit and come back in half an hour to see how the tests did, and I think I'd be pretty frustrated to find that they didn't even run because I forgot to run something trivial. Plus, if everything is green except the formatter, I can safely reformat and then land the PR without waiting for the other tests anyway.

Ya, I agree that would be annoying, but it should be a rare case. Those tasks will also run first so if you come back in just a couple minutes you can make sure it actually made it past that first stage.

[...]

PR #815, opened 2018-04-17
add build stages for travis

jakemac53:
This adds a single dartfmt/dartanalyzer task in its own stage, followed by the test stage.

Fixes https://github.com/dart-lang/test/issues/757


.travis.yml
```diff
@@ -24,20 +24,29 @@ before_install:
...
- # Split the tests into five shards to help parallelize them across Travis workers.
- - test: --preset travis --total-shards 5 --shard-index 0
- - test: --preset travis --total-shards 5 --shard-index 1
- - test: --preset travis --total-shards 5 --shard-index 2
- - test: --preset travis --total-shards 5 --shard-index 3
- - test: --preset travis --total-shards 5 --shard-index 4
+jobs:
+  include:
+    - stage: analyze_and_format
+      script: dartfmt_and_analyze.sh
```
natebosch: should we skip the shell script?

`dartfmt -n --set-exit-if-changed . && dartanalyzer .`

jakemac53:
oh hmm I think maybe there was some special magic being set up with the dart `test` task before - will check these failures tomorrow

PR #816, opened 2018-04-25
Remove a number of unused members

kevmoo:
Made members only used within a library private


kevmoo:
CC @natebosch – it works!

PR #819, opened 2018-04-27
Fix for Dart 2 runtime failures

keertip:
Fixes for Dart 2 runtime failures that showed up during internal testing.

- type '_InternalLinkedHashMap<Object, SuiteConfiguration>' is not a subtype of type 'Map<BooleanSelector, SuiteConfiguration>'
- type 'Future<dynamic>' is not a subtype of type 'Future<Group>'

### DIFF
--- a/lib/src/runner/configuration/suite.dart
+++ b/lib/src/runner/configuration/suite.dart
@@ -237,8 +237,10 @@ class SuiteConfiguration {
         runtimes: other._runtimes ?? _runtimes,
         includeTags: includeTags.intersection(other.includeTags),
         excludeTags: excludeTags.union(other.excludeTags),
-        tags: _mergeConfigMaps(tags, other.tags),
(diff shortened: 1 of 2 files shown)


lib/src/runner/configuration/suite.dart
```diff
@@ -237,8 +237,10 @@ class SuiteConfiguration {
         runtimes: other._runtimes ?? _runtimes,
         includeTags: includeTags.intersection(other.includeTags),
         excludeTags: excludeTags.union(other.excludeTags),
-        tags: _mergeConfigMaps(tags, other.tags),
-        onPlatform: _mergeConfigMaps(onPlatform, other.onPlatform),
```
kevmoo: Could `_mergeConfigMaps` have generic params and handle this conversion?

keertip: yes, changed to do so.

PR #822, opened 2018-05-02
Add @isTest and @isTestGroup annotations.

scheglov:
See https://github.com/flutter/flutter-intellij/issues/2055 for the context.


natebosch:
cc @jakemac53 

I think we should put these in `package:meta`. I don't understand your reasoning for putting them here.

It so happens that `flutter_test` already depends on `package:test`, but I don't know if that is something we want to require for anyone who wants to write a test framework and have a similar outline group/test concept.

What is the reason _not_ to include these in `package:meta`?

scheglov:
It seemed too test specific to put it into language specific `package:meta`.
OTOH, there is already `@visibleForTesting` there.
So, I guess we have already crossed this bridge.
I will move it to `package:meta`.
Thanks.

PR #823, opened 2018-05-02
Remove transformer and barback dependency

jakemac53:


README.md
```diff
@@ -802,57 +802,25 @@ Windows users.
 
 [term_glyph]: https://pub.dartlang.org/packages/term_glyph
 
-### `barback`
+## Integration with build systems
```
grouma: Make this more specific to build_runner and we can generalize later as we solidify our plans.

PR #826, opened 2018-05-03
Require meta 1.1.5 and use @isTest and @isTestGroup to mark test() and group().

scheglov:
See https://github.com/flutter/flutter-intellij/issues/2055 for the context.

### DIFF
--- a/lib/test.dart
+++ b/lib/test.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 
+import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 
 import 'src/backend/declarer.dart';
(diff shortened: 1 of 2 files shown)


jakemac53:
@scheglov any desired timeline for a release?

scheglov:
I think there is no burning need to publish it, we would have to keep support for not annotated `test()` and `group()` is Analysis Server for some time anyway (for users with older `package:test`). Just publish it at some point :-)

jakemac53:
Ok sounds good, we are publishing fairly often anyways so it should be available within a couple days/weeks.

scheglov:
SG, thanks.

Issue #830, opened 2018-05-09
expectLater returns futures that are created in another zone than the calling zone


Hixie:
If you call `expectLater` in a `FakeAsync` zone and trigger one of the codepaths that returns the precached _emptyFuture, the test will hang then time out, because of the way zones work.

grouma:
Hmm this sounds awfully related to https://github.com/dart-lang/sdk/issues/32556

PR #831, opened 2018-05-09
Use a new future instead of a cached future from another zone

Hixie:
As far as I can tell the comment doesn't make sense -- you don't pay the microtask hit unless you await the future.

Fixes https://github.com/dart-lang/test/issues/830

### DIFF
--- a/lib/src/frontend/expect.dart
+++ b/lib/src/frontend/expect.dart
@@ -12,12 +12,6 @@ import '../backend/invoker.dart';
 import '../utils.dart';
 import 'async_matcher.dart';
 
-/// A future that emits `null`.
-///
-/// We cache and re-use this value to avoid adding a new microtask hit for each
-/// call to `expect()`.
-final _emptyFuture = new Future.value();
-
 /// An exception thrown when a test assertion fails.
 class TestFailure {
   final String message;
@@ -120,7 +114,7 @@ Future _expect(actual, matcher,
     }
 
     Invoker.current.skip(message);
-    return _emptyFuture;
+    return new Future.value();
   }
 
   if (matcher is AsyncMatcher) {
@@ -151,12 +145,12 @@ Future _expect(actual, matcher,
       });
     }
 
-    return _emptyFuture;
+    return new Future.value();
   }
 
   var matchState = {};
   try {
-    if (matcher.matches(actual, matchState)) return _emptyFuture;
+    if (matcher.matches(actual, matchState)) return new Future.value();
   } catch (e, trace) {
     reason ??= '$e at $trace';
   }
--- a/lib/src/frontend/expect.dart
+++ b/lib/src/frontend/expect.dart
@@ -114,7 +114,7 @@ Future _expect(actual, matcher,
     }
 
     Invoker.current.skip(message);
-    return new Future.value();
+    return new Future.sync(() {});
   }
 
   if (matcher is AsyncMatcher) {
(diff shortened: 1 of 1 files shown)


Hixie:
Interestingly, it was added in https://github.com/dart-lang/test/pull/546 which suggests that this might be needed to not break Flutter in some other way. So maybe this PR is not a good fix for the zone bug.

natebosch:
@Hixie - here's the original thread with a link to a Travis failure that I believe is the issue `_emptyFuture` was intended to solve.

Can you confirm whether this change causes a similar problem?

grouma:
This sounds related to https://github.com/dart-lang/sdk/issues/32556

We had to do a hacky work around here: https://github.com/dart-lang/test/pull/787 Can the same approach apply?

natebosch:
Another pattern I'd be curious to try:

```dart
_emptyCallback() {}
final _emptyFuture = new Future.sync(_emptyCallback):
```

tvolkert:
afaik, `Future.sync()` just delegates to `Future.value()` under the hood...

natebosch:
> Can you post the link?

😳 

https://github.com/dart-lang/test/pull/529#issuecomment-280796048

tvolkert:
Ah, yes, I remember that!

So here's the test/line that was affected: 

https://github.com/flutter/flutter/blob/50a1d91ff39df06c7aed57f7242968a3cf95fa7a/packages/flutter/test/services/system_chrome_test.dart#L22

tvolkert:
Ok, in https://github.com/flutter/flutter/pull/17415, I've created a test that:

1. fails without this PR
2. passes with this PR

Further, I've verified that this PR does not regress system_chrome_test.dart.

tldr: this fix LGTM

natebosch:
@grouma can you confirm that the quiver tests and package:fake_async tests also still pass with this change? I would only expect this to make things better for those packages, not worse, but we should double check.

Hixie:
(I don't have commit rights in this repo, so when this is ready to land, please feel free to land it. :-) Thanks!)

grouma:
package:fake_async fails here: https://github.com/dart-lang/fake_async/blob/master/test/fake_async_test.dart#L461

package:quiver fails here: https://github.com/google/quiver-dart/blob/master/test/testing/async/fake_async_test.dart#L567

natebosch:
Two other possibilities to try:

- `final _emptyFuture = new Future.value()..listen((_) {});`
- Return `null` instead of a future and default it from `expectLater`.

grouma:
Note that nbosch@ meant:
`final _emptyFuture = new Future.value()..then((_) {});
`
I have confirmed that all tests pass in each package with either option. I'm having a difficult time confirming that either approach resolves the test in flutter/flutter#17415. I'm not sure how to do a dependency override for the flutter_test tests. @tvolkert or @Hixie can you confirm that either approach resolves your issue?

tvolkert:
@grouma adding a dependency override at the bottom of https://github.com/flutter/flutter/blob/master/packages/flutter_test/pubspec.yaml#L77 was working for me yesterday - does it work for you?

(you'll need to run `flutter packages upgrade` after you add it)

grouma:
Ran a bunch of tests with the various options. The only thing that passes for all tests in `package:async`, `package:quiver` and `package:flutter_test` is `return new Future.sync(() {});`. I guess `Future.value` isn't equivalent.

PR #834, opened 2018-05-18
Fix several Dart 2 runtime type errors

kevmoo:
Prepare to release 0.12.38+1

### DIFF
--- a/lib/src/runner/configuration/load.dart
+++ b/lib/src/runner/configuration/load.dart
@@ -371,7 +371,7 @@ class _ConfigurationLoader {
           (valueList) =>
               valueList.every((value) => _packageName.hasMatch(value)));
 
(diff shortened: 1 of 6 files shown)


lib/src/runner/configuration/load.dart
```diff
@@ -371,7 +371,7 @@ class _ConfigurationLoader {
           (valueList) =>
               valueList.every((value) => _packageName.hasMatch(value)));
 
-      return valueNode.value;
+      return new List<String>.from(valueNode.value);
```
natebosch: [optional] could also be `valueNode.value.cast<String>();`

kevmoo: Then you'd have to cast `value` to a List, since it's dynamic...trying to keep it simple

PR #839, opened 2018-05-30
Use public analyzer APIs.

scheglov:


lib/src/util/dart.dart
```diff
@@ -78,3 +86,84 @@ SourceSpan contextualizeSpan(
...
+    }
+    return parseResult.unit;
+  }
+
+  /// Return import and export directives in the file with the given [path].
+  List<UriBasedDirective> parseImportsAndExports(String path) {
+    var unit = parse(path);
+    var uriDirectives = <UriBasedDirective>[];
+    for (var directive in unit.directives) {
+      if (directive is UriBasedDirective) {
```
bwilkerson: This will include part directives. Is that intentional?

scheglov: No, actually we don't need it at all.
Removing.

lib/src/util/dart.dart
```diff
@@ -78,3 +86,72 @@ SourceSpan contextualizeSpan(
 
   return file.span(start, contextRunes.offset);
 }
+
+class AnalysisSessionManager {
```
kevmoo: This is a LOT of new code. What changed here?

kevmoo:
@scheglov – FYI - tests likely failing due to https://github.com/dart-lang/test/issues/840

lib/src/util/dart.dart
scheglov: The library `package:analyzer/analyzer.dart` is not API that we want to support. We migrating everyone to the new API that we think will be compatible with the way CFE will be integrated into Analyzer.

Specifically, `parseDirectives(contents)` will not work. You can only parse in the context of some analysis session, because parsing result depend on analysis options. Analysis options can be only created from `analysis_options.yaml`, to avoid duplication and exposing additional APIs. Also, there is no way to parse a file partially, to avoid creating parseXYZ for every kind of node. You get the whole unit parsed or resolved.

grouma: You explanation makes sense but this is still a lot of boilerplate. Is there not a better alternative instead of forcing all users to duplicate the boiler plate below?

bwilkerson: If this _does_ prove to be needed by multiple clients, perhaps we should make this a utility class in 'analyzer' so that the code doesn't need to be repeated.

natebosch: This does look very similar to https://github.com/dart-lang/pub/pull/1912, should we consider centralizing it?

bwilkerson:
`AnalysisSessionManager` has been incorporated into analyzer (now named `AnalysisContextCollection`, so at the very least this PR needs to be updated.

kevmoo:
@grouma @scheglov – this seems quite outdated. Can close?

grouma:
Yeah. Gonna close.

PR #846, opened 2018-06-01
Discussion: use package:stream_transform

natebosch:
Removes some of the noise of explicitly dealing with the `sink` so that
the business logic is more prevalent where possible. There are still
usages of `new StreamTransformer.fromHandlers` and similar where the
intent can't be expressed by the lower level building blocks.


jakemac53:
Do we pick up any additional dependencies with this?

grouma:
I'm not a fan of adding another dependency but if it can provide further benefit than what is done in this PR I'm not opposed to it.

natebosch:
It is a new dependency on `stream_transform` but no further new transitive deps. I had though that `build_runner_core` already depended on it which would mitigate the problem with adding a dep, but it doesn't...

Issue #847, opened 2018-06-01
Reporter default varies across platforms


Hixie:
I got confused recently because the default reporter is different on Windows (-rexpanded) than on other platforms (-rcompact).

We should either document this more clearly (including e.g. in the --help text), including the reasoning, or, even better if possible, make the default consistent throughout.

jakemac53:
Fwiw, I checked with the regular command prompt on my windows box and the compact reporter seems to work fine, so I don't think there is a technical limitation.

I also noticed that the colors appear to be turned off on windows.

jakemac53:
@nex3 are there any outstanding issues you know of with the compact reporter on windows?

nex3:
I know we changed some details about how the VM prints output on Windows since these defaults were decided, so it's plausible that the compact reporter works now.

jakemac53:
Both colors and the compact reporter seem to work on my windows box with the built in command prompt and in the vscode console, and the compact reporter appears to work on appveyor as well [before](https://ci.appveyor.com/project/kevmoo/build/build/1.0.1795) - [after](https://ci.appveyor.com/project/kevmoo/build/build/1.0.1796). I think the appveyor log viewer just doesn't respect colors, because its going through the same logic to determine if it should use colors.

I changed all the windows detection logic to instead use `stdout.supportsAnsiEscapes`. We could also simply check if its a terminal but I think this is better? There are some scenarios I believe where that doesn't get recognized properly though.

jakemac53:
This is changed and published - we will keep an eye out for any users reporting that they lost the compact reporting and/or colors and file corresponding issues on the SDK to get better support for `stdout.supportsAnsiEscapes` (it has relatively basic rules right now).

PR #849, opened 2018-06-05
Change the default reporter logic to use `stdout.supportsAnsiEscapes`

jakemac53:
Enable the compact reporter and colors based on `stdout.supportsAnsiEscapes` instead of platform.

Tested in a regular windows command prompt, vscode console on windows, appveyor, and my local windows box. All appear to work as expected.

Fixes https://github.com/dart-lang/test/issues/847.

### DIFF
--- a/lib/src/runner/configuration/reporters.dart
+++ b/lib/src/runner/configuration/reporters.dart
@@ -43,7 +43,8 @@ final _allReporters = <String, ReporterDetails>{
       (_, engine) => JsonReporter.watch(engine)),
 };
 
-final defaultReporter = inTestTests ? 'expanded' : 'compact';
+final defaultReporter =
+    inTestTests ? 'expanded' : canUseSpecialChars ? 'compact' : 'expanded';
 
 /// **Do not call this function without express permission from the test package
 /// authors**.
--- a/lib/src/runner/configuration/reporters.dart
+++ b/lib/src/runner/configuration/reporters.dart
@@ -43,8 +43,7 @@ final _allReporters = <String, ReporterDetails>{
       (_, engine) => JsonReporter.watch(engine)),
 };
 
-final defaultReporter =
-    inTestTests ? 'expanded' : (Platform.isWindows ? 'expanded' : 'compact');
+final defaultReporter = inTestTests ? 'expanded' : 'compact';
 
 /// **Do not call this function without express permission from the test package
 /// authors**.
--- a/lib/src/util/io.dart
+++ b/lib/src/util/io.dart
@@ -78,14 +78,9 @@ final _tempDir = Platform.environment.containsKey("_UNITTEST_TEMP_DIR")
     ? Platform.environment["_UNITTEST_TEMP_DIR"]
     : Directory.systemTemp.path;
 
-// TODO(nweiz): Make this check [stdioType] once that works within "pub run".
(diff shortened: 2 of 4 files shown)


lib/src/util/io.dart
```diff
@@ -78,14 +78,9 @@ final _tempDir = Platform.environment.containsKey("_UNITTEST_TEMP_DIR")
...
 
-// TODO(nweiz): Make this check [stdioType] once that works within "pub run".
-/// Whether "special" strings such as Unicode characters or color escapes are
-/// safe to use.
-///
-/// On Windows or when not printing to a terminal, only printable ASCII
-/// characters should be used.
-bool get canUseSpecialChars =>
-    Platform.operatingSystem != 'windows' && !inTestTests;
+// Whether or not the current terminal supports ansi escape codes. Otherwise
```
natebosch: missing a slash, also add a blank line after the first sentence.

lib/src/runner/configuration/reporters.dart
```diff
@@ -43,8 +43,7 @@ final _allReporters = <String, ReporterDetails>{
       (_, engine) => JsonReporter.watch(engine)),
 };
 
-final defaultReporter =
-    inTestTests ? 'expanded' : (Platform.isWindows ? 'expanded' : 'compact');
+final defaultReporter = inTestTests ? 'expanded' : 'compact';
```
jakemac53: oh I just realized I should use the canUseSpecialChars bool here too

lib/src/util/io.dart
jakemac53: done

jakemac53:
@natebosch can you tests this with zsh?

natebosch:
Gave this a try on my environment and didn't see anything broken.

nex3:
This causes tests to run in expanded mode, without colors, on Travis. Also within Emacs' terminal emulator, although that's admittedly pretty narrow.

natebosch:
Yeah we had noticed that in Travis too and couldn't decide whether it was worth switching back to get colors in Travis. Emacs and any place where term isn't `xterm*` also won't see color or the compact reporter.

FWIW It I _think_ it  should be safe to set your `$TERM` to `xterm-256color` within emacs. I do that for tmux since the Dart VM doesn't have proper capability checking.

@jakemac53 - should we flip back to logic that will work in travis and elsewhere?

nex3:
I strongly vote for flipping back the logic. This makes looking at errors in Travis *much* more painful, and as far as I know there were no false positives with the old logic.

jakemac53:
The old logic was flawed on windows though - it had lots of false negatives there.

We could do a combination - always turn it on if not windows, and on windows use the new logic.

jakemac53:
It also seems like the real problem here lies in the implementation of `stdout.supportsAnsiEscapes`, so it could potentially be resolved from that end.

natebosch:
>It also seems like the real problem here lies in the implementation of `stdout.supportsAnsiEscapes`, so it could potentially be resolved from that end.

https://github.com/dart-lang/sdk/issues/31606

I think we're _very_ unlikely to get that fixed. We might be able to get more special cases hardcoded in but using proper capability tracking would be a lot of work.

I'm in favor of using special logic for windows again.

Issue #842, opened 2018-05-31
Tons of failing tests with `--preview-dart-2`


grouma:
I ran the following:
`DART_VM_OPTIONS=--preview-dart-2 pub run test`

and found tons of failing tests:
`08:46 +796 ~10 -159: Some tests failed.`

This needs to be addressed in the near future.

zoechi:
I get
```
 (dart2) $ pub run test
Unable to spawn isolate: file:///Users/zoechi/.pub-cache/hosted/pub.dartlang.org/analyzer-0.30.0+4/lib/src/dart/element/element.dart:4968:5: Error: Declared type variables of 'GenericFunctionTypeElementImpl::accept' doesn't match those on overridden method 'Element::accept'.
  T accept<T>(ElementVisitor<T> visitor) {
    ^
file:///Users/zoechi/.pub-cache/hosted/pub.dartlang.org/analyzer-0.30.0+4/lib/src/dart/element/element.dart:4968:33: Error: The parameter 'visitor' of the method 'GenericFunctionTypeElementImpl::accept' has type analyzer.dart.element.element::ElementVisitor<analyzer.src.dart.element.element::GenericFunctionTypeElementImpl::accept::T>, which does not match the corresponding type in the overridden method (analyzer.dart.element.element::ElementVisitor<dynamic>).
Change to a supertype of analyzer.dart.element.element::ElementVisitor<dynamic> (or, for a covariant parameter, a subtype).
  T accept<T>(ElementVisitor<T> visitor) {
                                ^
file:///Users/zoechi/.pub-cache/hosted/pub.dartlang.org/analyzer-0.30.0+4/lib/src/cancelable_future.dart:255:17: Error: Declared type variables of '_CancelableCompleterFuture::then' doesn't match those on overridden method 'Future::then'.
  Future/*<S>*/ then/*<S>*/(FutureOr/*<S>*/ onValue(T value),
                ^
file:///Users/zoechi/.pub-cache/hosted/pub.dartlang.org/analyzer-0.30.0+4/lib/src/cancelable_future.dart:255:17: Error: The return type of the method '_CancelableCompleterFuture::then' is dart.async::Future<dynamic>, which does not match the return type of the overridden method (dart.async::Future<dart.async::Future::then::R>).
Change to a subtype of dart.async::Future<dart.async::Future::then::R>.
  Future/*<S>*/ then/*<S>*/(FutureOr/*<S>*/ onValue(T value),
                ^
[...]

PR #851, opened 2018-06-06
Dart 2 fixes

grouma:
More towards https://github.com/dart-lang/test/issues/842


lib/src/frontend/spawn_hybrid.dart
```diff
@@ -20,7 +20,7 @@ import '../utils.dart';
 /// The spawned isolate sends three kinds of messages. Data messages are emitted
 /// as data events, error messages are emitted as error events, and print
 /// messages are printed using `print()`.
-final _transformer = new StreamChannelTransformer<Object, Map>(
+final _transformer = new StreamChannelTransformer<dynamic, dynamic>(
```
natebosch: maybe add a TODO or file an issue to see if we can restore the type here and flow the types through the rest of the code.

grouma: Done.

PR #855, opened 2018-06-07
Avoid `Future.sync` wrapping an async method

natebosch:


natebosch:
One other instance exists that will be changed by @grouma in another commit.

kevmoo:
@natebosch - fix unused import!

PR #858, opened 2018-06-08
Update comment explaining untyped StreamChannelTransformer

natebosch:


pubspec.yaml
```diff
@@ -45,3 +45,8 @@ dev_dependencies:
   shelf_test_handler: '^1.0.0'
   test_descriptor: '^1.0.0'
   test_process: '^1.0.0'
+
+dependency_overrides:
```
jakemac53: accident?

natebosch: haven't published the new version of stream_channel yet. Will publish it and bump the dep here once we confirm it works.

natebosch:
Ah, it looks like there is a test expecting that the user is allowed to send any json encodable type over this channel. The types on the `StreamChannelTransformer` were flat our wrong anyway, it's not always `Map` expected to flow through that channel, that's just the only type the test package used it for, but users could pick other types.

PR #861, opened 2018-06-09
Fix implicit casts in lib directory

natebosch:
There are still errors in tests which may be harder to fix so we can't
check in the change to analysis options.

This should make it easier to diagnose Dart 2 semantic issues as more of
the places where there can be failures are explicit.

Most of the changes are just adding explicit `as` statements and keeping
the semantics that already existed. In a few places an uninitialized
`var` is given a type, and in others a reused variable name with
different types is split into multiple variables.

The changes in the frontend code are the most risky, since some users
might still be using that code in a non Dart 2 runtime and relying on
looseness which is lost with an explicit cast.

### DIFF
--- a/lib/src/frontend/async_matcher.dart
+++ b/lib/src/frontend/async_matcher.dart
@@ -44,7 +44,8 @@ abstract class AsyncMatcher extends Matcher {
       Invoker.current.addOutstandingCallback();
       result.then((realResult) {
         // ignore: deprecated_member_use
-        if (realResult != null) fail(formatFailure(this, item, realResult));
+        if (realResult != null)
+          fail(formatFailure(this, item, realResult as String));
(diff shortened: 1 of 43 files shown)


lib/src/frontend/async_matcher.dart
```diff
@@ -43,8 +43,10 @@ abstract class AsyncMatcher extends Matcher {
     if (result is Future) {
       Invoker.current.addOutstandingCallback();
       result.then((realResult) {
-        // ignore: deprecated_member_use
-        if (realResult != null) fail(formatFailure(this, item, realResult));
+        if (realResult != null) {
+          // ignore: deprecated_member_use
```
kevmoo: @natebosch @grouma – here's the issue.

PR #866, opened 2018-06-11
Remove Travis hack

grouma:
Towards https://github.com/dart-lang/test/issues/599

-  Remove odd browser kill logic. The hypothesis is that it puts Chrome in an odd state that results in the exitCode future never completing.
- Add a timeout to the browser close function and throw if it does not complete.

### DIFF
--- a/lib/src/executable.dart
+++ b/lib/src/executable.dart
@@ -188,12 +188,6 @@ transformers:
   } finally {
     await close();
   }
-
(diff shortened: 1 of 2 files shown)


grouma:
Looks like the theory was not correct as we have a passing run that hung on Travis: https://travis-ci.org/dart-lang/test/jobs/391006919

PR #867, opened 2018-06-11
More Dart 2 fixes

grouma:
More towards https://github.com/dart-lang/test/issues/842

- Export `@Retry` annotation (this is technically broken now but the corresponding tests only fail under preview-dart-2)
- Fix type issue with mapping over dynamic values


lib/src/frontend/retry.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// An annotation for marking a test to be retried.
+///
+/// A test with reties enabled will be re-run if it fails for a reason
```
natebosch: typo, reties -> retries

PR #869, opened 2018-06-12
Add types to a bunch of uninitalized variables

natebosch:
Searched for lines matching `var \w+;`. This is a much weaker version of
no-implicit-dynamic. This catches a bunch of places where we otherwise
had dynamic calls and tightens up types a bit. Where we are relying on
being able to parse to multiple types and then throw a specific
excpetion or return null on bad types I added an explicit `dynamic`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,8 +106,8 @@ if a test fails, to ensure that it has a chance to clean up after itself.
 import "package:test/test.dart";
 
 void main() {
-  var server;
-  var url;
+  HttepServer server;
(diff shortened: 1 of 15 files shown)


kevmoo:
@natebosch – this is the lint you want - http://dart-lang.github.io/linter/lints/prefer_typing_uninitialized_variables.html

kevmoo:
Easy enough - https://github.com/dart-lang/test/pull/871

PR #870, opened 2018-06-12
Re-enable some skipped tests for a stale issue

natebosch:
We no longer use `mapMap` so the hints aren't printed. Fix the tests to
allow for the "Compiled" message from dart2js. Update the reason for
skipping one test which has an incorrect line number - we will need to
investigate why this line is wrong.

Allow Matcher usage in the `_print` call since the "Compiled" message is
unpredictable.

### DIFF
--- a/test/runner/browser/loader_test.dart
+++ b/test/runner/browser/loader_test.dart
@@ -158,7 +158,7 @@ void main() {
     var loadSuite = suites.first;
 
     var liveTest = (loadSuite.group.entries.single as Test).load(loadSuite);
-    // Skip the "Compiled" message from dart2js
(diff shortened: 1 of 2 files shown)


test/runner/browser/loader_test.dart
```diff
@@ -158,9 +158,10 @@ void main() {
     var loadSuite = suites.first;
 
     var liveTest = (loadSuite.group.entries.single as Test).load(loadSuite);
-    expect(liveTest.onMessage.first.then((message) => message.text),
+    // Skip the "Compiled" message from dart2js
```
grouma: [nit] Period at the end.

PR #876, opened 2018-06-12
Support VM debugging

grouma:
Closes https://github.com/dart-lang/test/issues/50

The blocking issue still has not been fixed: https://github.com/dart-lang/sdk/issues/31308

Provide a warning to use the suggested work around.


kevmoo:
Woo hoo!!!

PR #880, opened 2018-06-14
Enable the default reporter and colors always if !windows

jakemac53:
See comments in https://github.com/dart-lang/test/pull/849


jakemac53:
merging on red due to the `packageRoot` issue - locally all tests pass

PR #881, opened 2018-06-14
Add solo test and group

grouma:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.12.42
+
+* Add support for `solo` test and group.
```
kevmoo: Be more clear that this is a new param on test and group. And explain a bit of what it does?

lib/src/backend/declarer.dart
```diff
@@ -127,7 +133,8 @@ class Declarer {
       skip,
       Map<String, dynamic> onPlatform,
       tags,
-      int retry}) {
+      int retry,
+      bool solo = false}) {
```
kevmoo: Consider marking the param `@Deprecated` – so any usage **screams** to the user ... also caught if folks check for lints in their CI environments...

grouma: Good call.

CHANGELOG.md
grouma: Done.

lib/src/backend/declarer.dart
```diff
@@ -80,6 +80,12 @@ class Declarer {
   /// Whether [build] has been called for this declarer.
   bool _built = false;
 
+  /// Whether any tests and/or groups have been flagged as solo.
+  bool _solo = false;
```
natebosch: could this be

```dart
bool get _solo => _soloEntries.isNotEmpty;
```

lib/test.dart
```diff
@@ -132,21 +132,29 @@ Declarer get _declarer {
 ///
 /// If multiple platforms match, the annotations apply in order as through
 /// they were in nested groups.
+///
+/// If the `solo` flag is `true`, only tests and groups marked as
+/// "solo" will be be run. This only restricts tests *within this test
+/// suite*—tests in other suites will run as normal. We recommend that users
+/// avoid this flag if possible and instead use the test runner flag `-n` to
```
natebosch: maybe call out tags as another option?

Issue #885, opened 2018-06-15
Unable to run all test files in folder, from a folder named "packages"


weenzeel:
I'm just getting started with Dart.

I thought I would write some packages, and maybe publishing them to pub.dartlang.org.

I created a folder structure like ... / Documents / _projects / dart / packages / myPackage. I wrote some unit-tests and configured a run configuration in IntelliJ to run all tests in my test folder.

The problem is, that the runner doesn't find any testcases if the project is in a folder named packages. If I move the project folder out off the "packages" folder, everything works just fine.

There may be som good reason for this not to work, but it tripped me up for a couple of hours.

Thanks for providing this framework.

I'm using test 0.12.41, Dart 2 0.63 and

IntelliJ IDEA 2018.2 EAP (Community Edition)
Build #IC-182.2949.4, built on June 5, 2018
JRE: 1.8.0_152-release-1226-b7 x86_64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
macOS 10.13.2

grouma:
What happens when you run the tests from the command line using `pub run test`? From the documentation: "The test runner considers any file that ends with `_test.dart` to be a test file. If you don't pass any paths, it will run all the test files in your `test/` directory, making it easy to test your entire application at once."

That being said Intellij may do something different. Unfortunately I am not familiar with the Intellji Dart plugin.

weenzeel:
The result seems to be reproducible via the command line.

1. Create new project in a path .../packages/MyPoject/. The project should contain a test folder with at least one file ending in _test.dart.
2. From the commandline run pub run test .../packages/MyProject/test/
3. Result "No tests ran".

To be able to run the tests do ...

1. Create new project in a path .../OtherName/MyPoject/. The project should contain a test folder with at least one file ending in _test.dart.
2. From the commandline run pub run test .../OtherName/MyProject/test/
3. Result "tests are found and run from the commandline".

This indicates to me that the problem is not in Intellij, but in some other part of the stack.

Please let me know if I can provide any other information.

grouma:
For historical reasons the packages folder is special. A number of tools still make use of the folder today, e.g. `package:build`. It is not recommended to have your path contain this special folder as other tools might behave incorrectly.

PR #887, opened 2018-06-18
Don't ignore packages folder

grouma:
For historical reasons we used to ignore files under a `packages` directory. This is no longer necessary.

Closes: https://github.com/dart-lang/test/issues/885


natebosch:
Hmm, hit approve but then I remember we still have `packages` directories when we run `build_runner --output`. And then we run with `--precompiled` in that directory...

grouma:
Discussed offline. This is the intended behavior and will cause issues if we remove the check.

PR #890, opened 2018-06-20
Add pid to start event; bump protocol version to 0.1.1; package version to 1.1.0

### LINKED ISSUE #889: Add the `pid` to the `start` event
Currently when running without the debugger I can't find a way to get the `pid`. Is it possible to add this to the `start` JSON event? (other tools like Flutter and the analysis server do this).

The reason is that in Dart Code we use `shell=true` when spawning processes to work around all sorts of issues we've had with escaping of paths and other things. Unfortunately this means the process we have a reference to is for the shell and not the actual process we spawned. Killing that [shell process is somewhat unreliable](https://github.com/Dart-Code/Dart-Code/issues/1025) and often doesn't terminate the process spawned from it.

When debugging, we get the `pid` from observatory; but when running without the debugger we don't use observatory so we don't have this.

I'm happy to send a PR for this if you'd accept it. I did this in flutter and it was fairly straight forward (https://github.com/flutter/flutter/pull/17873).

DanTup:
Fixes #889.

I bumped the protocol version since it's a change, even though it shouldn't be breaking in any way. I moved the expected start event down in the test so I can put the real `pid` in it (so it's tested to be correct, rather than just stripped out) though not sure if there's a better way to do this.

I have some test failures locally that I don't think are related to this (maybe it's Dart2?), but let me know if you disagree and I'll do some more digging:

```
dantup-macbookpro:test dantup$ pub run test
01:12 +365 -1: test/runner/configuration/custom_platform_test.dart: override_platforms can override a browser with a basename-only executable [E]              
  Expected: <0>
    Actual: <1>
  Process `dart bin/test.dart` had an unexpected exit code.
  
  package:test/src/frontend/expect.dart 165:30               fail
  package:test/src/frontend/expect.dart 159:3                _expect
  package:test/src/frontend/expect.dart 60:3                 expect
  package:test_process                                       TestProcess.shouldExit
  test/runner/configuration/custom_platform_test.dart 75:20  main.<fn>.<fn>.<fn>
  dart:async                                                 _completeOnAsyncReturn
  test/io.dart 82:3                                          runTest
  dart:async                                                 _completeOnAsyncReturn
  test/io.dart 99:3                                          runDart
  dart:async                                                 _asyncThenWrapperHelper
  test/runner/configuration/custom_platform_test.dart 62:56  main.<fn>.<fn>.<fn>
  package:test/src/backend/declarer.dart 161:27              Declarer.test.<fn>.<fn>.<fn>
  dart:async                                                 _asyncThenWrapperHelper
  package:test/src/backend/declarer.dart 159:70              Declarer.test.<fn>.<fn>.<fn>
  package:test/src/backend/invoker.dart 249:15               Invoker.waitForOutstandingCallbacks.<fn>
  dart:async                                                 new Future
  package:test/src/backend/invoker.dart 402:15               Invoker._onRun.<fn>.<fn>.<fn>
  
  Expected: should eventually emit an event that contains 'All tests passed!'
    Actual: <Instance of '_StreamQueue'>
     Which: emitted • 00:00 +0: compiling test.dart
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.0.1
+
+* Added a new `pid` field to the StartEvent in the json runner containing the
+  pid of the VM process running the tests.
+
 ## 1.0.0
 
 * No change from `0.12.42`. We are simply signalling to users that this is a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.0.1
+## 1.1.0
 
 * Added a new `pid` field to the StartEvent in the json runner containing the
   pid of the VM process running the tests.
--- a/doc/json_reporter.md
+++ b/doc/json_reporter.md
@@ -95,6 +95,9 @@ class StartEvent extends Event {
 
   // The version of the test runner being used.
   String runnerVersion;
+
+  // The pid of the VM process running the tests.
+  int pid;
 }
 ```
 
--- a/json_reporter.schema.json
+++ b/json_reporter.schema.json
@@ -98,6 +98,7 @@
       "required": ["protocolVersion", "runnerVersion"],
       "properties": {
         "type": {"enum": ["start"]},
+        "pid": {"type": "integer", "minimum": 0},
         "protocolVersion": {"type": "string", "pattern": "^0\\.1\\."},
         "runnerVersion": {
           "oneOf": [{"type": "string"}, {"type": "null"}]
--- a/lib/src/runner/reporter/json.dart
+++ b/lib/src/runner/reporter/json.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 import 'dart:convert';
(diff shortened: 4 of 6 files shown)


DanTup:
Ugh, I remember editing that but maybe I didn't save. Added again.

I've also rebased on master (somehow I didn't update before I did this!) and updated the version to 1.0.1 and added to changelog.

DanTup:
(I don't have write access to this repo so someone will have to merge if/when happy with it :-))

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.0.1
```
natebosch: I think this should be `1.1.0` since this is a new feature, not a bug fix.

DanTup: Done!

DanTup: (not sure if we should change protocolVersion to 0.2.0 for same reason?)

PR #893, opened 2018-06-25
Drop --checked arg to dart2js

natebosch:
We'll stop getting messages like "Option '--enable-checked-mode' is not
needed in Dart 2.0."


kevmoo:
Blocked landing this, @natebosch ? Might fix some test failures related to https://github.com/dart-lang/sdk/issues/33639

PR #896, opened 2018-06-27
Add support for loading precompiled kernel files for vm tests

jakemac53:
When using the `--precompiled` flag we will now first check for an `<original-test-path>.vm_test.vm.app.dill` file, and if present load that directly in the isolate. Otherwise the `<original-test-path>.vm_test.dart` file will be used.


jakemac53:
Submitting with pre-existing failures

PR #900, opened 2018-07-06
Fix constants on legacy `unittest` branch.

lrhn:
We still have a lot of tests in the SDK repository using `unittest`.

We need either a version of `unittest` that will run when we remove the old-style constants, or to disable all those tests (rewriting is an option, just an expensive one).

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,10 +1,10 @@
 name: unittest
-version: 0.11.7
+version: 0.11.8
 author: Dart Team <misc@dartlang.org>
 description: A library for writing dart unit tests.
 homepage: https://github.com/dart-lang/old_unittest
 environment:
-  sdk: '>=1.0.0 <2.0.0'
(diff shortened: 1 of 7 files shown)


pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: unittest
-version: 0.11.7
+version: 0.11.8
 author: Dart Team <misc@dartlang.org>
 description: A library for writing dart unit tests.
 homepage: https://github.com/dart-lang/old_unittest
 environment:
-  sdk: '>=1.0.0 <2.0.0'
+  sdk: '>=2.0.0-dev.61.0 <3.0.0'
```
jakemac53: nit: this should be `<2.0.0` still, we aren't bumping packages to `<3.0.0` until Dart 2 stable is released

lrhn: Not absolutely sure, but it seems to be running for the use-cases where we are using it.

We should stop using `unittest` in the SDK completely, as soon as we find the time to fix it.

PR #902, opened 2018-07-11
Serve files outside the path during precompiled tests

jakemac53:
See https://github.com/dart-lang/build/pull/1647 for context

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.2.1
+
+* When using `--precompiled`, the test runner now allows symlinks to reach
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.2.1
```
natebosch: I think this is a new feature and it is reasonable to update your pubspec to require it. This should be 1.3.0

jakemac53: Hmm, I was thinking of it as a bug fix, but I could see it either way (fixing the bug does enable the use of symlinks which could be considered a feature?)

PR #906, opened 2018-07-16
Updated VM platform to accept custom `PackageResolver Function(path)` that can be used in spawning isolated.

liuming0:
Add resolverForDataIsolate(path) API to vm platform.dart to allow children to override which package resolver to use when spawning DataIsolate.

### DIFF
--- a/lib/src/runner/vm/platform.dart
+++ b/lib/src/runner/vm/platform.dart
@@ -30,7 +30,7 @@ class VMPlatform extends PlatformPlugin {
   /// The test runner configuration.
   final _config = Configuration.current;
 
-  /// A function that returns a [PackageResovler] that is used in spawning
+  /// A function that returns a [PackageResolver] that is used in spawning
   /// isolates based on the current test path.
   final PackageResolver Function(String path) _packageResolver;
 
@@ -114,7 +114,7 @@ class VMPlatform extends PlatformPlugin {
 }
 
 Future<Isolate> _spawnDataIsolate(String path, SendPort message,
-    {PackageResolver resolver}) async {
+    PackageResolver resolver) async {
   return await dart.runInIsolate('''
     import "dart:isolate";
 
--- a/lib/src/runner/vm/platform.dart
+++ b/lib/src/runner/vm/platform.dart
@@ -7,6 +7,7 @@ import 'dart:developer';
 import 'dart:io';
 import 'dart:isolate';
 
+import 'package:package_resolver/package_resolver.dart';
 import 'package:path/path.dart' as p;
 import 'package:stream_channel/stream_channel.dart';
 
@@ -102,12 +103,18 @@ class VMPlatform extends PlatformPlugin {
     } else if (_config.pubServeUrl != null) {
       return _spawnPubServeIsolate(path, message, _config.pubServeUrl);
     } else {
-      return _spawnDataIsolate(path, message);
+      return _spawnDataIsolate(path, message,
+          resolver: resolverForDataIsolate(path));
     }
   }
+
+  /// What [PackageResolver] should be used for data isolate by given a [path].
+  PackageResolver resolverForDataIsolate(String path) =>
+      PackageResolver.current;
 }
 
-Future<Isolate> _spawnDataIsolate(String path, SendPort message) async {
+Future<Isolate> _spawnDataIsolate(String path, SendPort message,
+    {PackageResolver resolver}) async {
   return await dart.runInIsolate('''
     import "dart:isolate";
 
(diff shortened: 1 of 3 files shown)


lib/src/runner/vm/platform.dart
```diff
@@ -102,12 +103,18 @@ class VMPlatform extends PlatformPlugin {
...
       return _spawnPubServeIsolate(path, message, _config.pubServeUrl);
     } else {
-      return _spawnDataIsolate(path, message);
+      return _spawnDataIsolate(path, message,
+          resolver: resolverForDataIsolate(path));
     }
   }
+
+  /// What [PackageResolver] should be used for data isolate by given a [path].
+  PackageResolver resolverForDataIsolate(String path) =>
```
natebosch: I would strongly prefer

```dart
class VMPlatform extends PlatformPlugin {
  final PackageResolver packageResolver;

  VMPlatform({PackageResolver packageResolver}) :
      packageResolver = packageResolver ?? PackageResolver.current;
}
```

liuming0: Done.

lib/src/runner/vm/platform.dart
```diff
@@ -29,7 +30,12 @@ class VMPlatform extends PlatformPlugin {
   /// The test runner configuration.
   final _config = Configuration.current;
 
-  VMPlatform();
+  /// A function that returns a [PackageResovler] that is used in spawning
```
natebosch: typo

lib/src/runner/vm/platform.dart
```diff
@@ -29,7 +30,12 @@ class VMPlatform extends PlatformPlugin {
...
+  /// isolates based on the current test path.
```
natebosch: is this test _case_ or test _suite_ ?

lib/src/runner/vm/platform.dart
```diff
@@ -102,12 +108,13 @@ class VMPlatform extends PlatformPlugin {
...
     } else {
-      return _spawnDataIsolate(path, message);
+      return _spawnDataIsolate(path, message, _packageResolver(path));
     }
   }
 }
 
-Future<Isolate> _spawnDataIsolate(String path, SendPort message) async {
+Future<Isolate> _spawnDataIsolate(String path, SendPort message,
+    {PackageResolver resolver}) async {
```
natebosch: this wouldn't work - it's declaring a named argument but being used as a positional argument

lib/src/runner/vm/platform.dart
liuming0: Done

lib/src/runner/vm/platform.dart
liuming0: Changed to test suite.

lib/src/runner/vm/platform.dart
liuming0: Done

liuming0:
Any idea why Travis CI build failed?

natebosch:
> Any idea why Travis CI build failed?

Need to run `dartfmt`

lib/src/runner/vm/platform.dart
```diff
@@ -29,7 +30,12 @@ class VMPlatform extends PlatformPlugin {
   /// The test runner configuration.
   final _config = Configuration.current;
 
-  VMPlatform();
+  /// A function that returns a [PackageResolver] that is used in spawning
+  /// isolates based on the current test suite path.
+  final PackageResolver Function(String path) _packageResolver;
+
+  VMPlatform({PackageResolver Function(String path) packageResolver})
```
jakemac53: The name here implies that it affects anything using a package resolver, but in fact it only affects the data isolate code path (one of 3 possible code paths).

It should probably be named accordingly?

liuming0:
Or I can apply the same to the other 2. What is the default value of packageConfig in spawnUri, current?

natebosch:
> for the pub serve isolate I believe it is all http scheme uris, so there are no `package:` uris to resolve.

I'm not exactly sure how that worked for the VM platform... was something somehow rewriting the imports?

> for the precompiled isolate it would defeat the purpose, the only thing that really does differently is provide the path to a custom .packages file. This could potentially be migrated to use the new api though and be deleted?

+1 for migrating to use the new API.

liuming0:
So what does it mean for this PR? What is my action point next?

jakemac53:
> I'm not exactly sure how that worked for the VM platform... was something somehow rewriting the imports?

Ya good point I am not really sure how this worked exactly for the VM platform.... it may have never really had any actual usage?

natebosch:
@liuming0 - not sure what the next actions are. We need to consider this change pretty carefully.

liuming0:
Ok. I will withdraw this code change. But please let me know when the new API is available.

What I need is that customizing PackageResolver resp with the test suite path.

PR #911, opened 2018-07-20
Strengthen UTF-8 decoding.

lrhn:
Decoding input bytes by doing `.map(utf8.decode)` assumes that input byte chunks are split at encoded code-unit boundaries.
It's safer to use `.transform(utf8.decoder)` which allows encodings to be split accross input chunks.

Also changed `.listen(f).asFuture()` to the equivalent `.forEach(f)`.

### DIFF
--- a/lib/src/runner/compiler_pool.dart
+++ b/lib/src/runner/compiler_pool.dart
@@ -90,8 +90,8 @@ class CompilerPool {
         var buffer = new StringBuffer();
 
         await Future.wait([
-          process.stdout.map(utf8.decode).listen(buffer.write).asFuture(),
-          process.stderr.map(utf8.decode).listen(buffer.write).asFuture(),
+          process.stdout.transform(utf8.decoder).forEach(buffer.write),
+          process.stderr.transform(utf8.decoder).forEach(buffer.write),
         ]);
 
         var exitCode = await process.exitCode;
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test
-version: 1.3.0
+version: 1.3.1-dev
 author: Dart Team <misc@dartlang.org>
 description: A library for writing dart unit tests.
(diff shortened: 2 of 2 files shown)


a14n:
@lrhn would it make sense to provide 2 lints ( prefering `stream.map(codec.decode)` over `stream.transform(codec.decoder)` and prefering `stream.forEach(f)` over `stream.listen(f).asFuture()`)  ?

lrhn:
I think preferring `Stream.forEach(f)` over `Stream.listen(f)` is always a good idea whenever you don't store the subscription anywhere, or if all you do is to call `asFuture()` on it. The `listen` method is really intended for low-level operations, and `forEach` is what corresponds to `await for` and is what you should do if you really want to act on each event.

For recommending `transform` over `map` when decoding, it's less clear that it's general enough to warrant a lint. It's quite possible to have a stream of byte-lists which are each individually encoded messages, just as it's possible to have one long stream representing a single long chunked message. There is not enough information in the types to distinguish, but you should use `map` in the former case and `transform` in the latter. There aren't that many cases at all, and a good number of them are using JSON encode/decode where it actually does seem like the events are full separate. So, low impact, high chance of false positive, so I would be wary of making a lint about this.

jakemac53:
Is there any urgency to publishing this? Any actual bugs you have ran into that motivated the change or simply general cleanup?

lrhn:
Simply general cleanup. I noticed the issue somewhere else and searched for other instances of the same problem. 
The code *probably* works because the underlying system is using line-based buffering for stdout/stderr. There is just nothing guaranteeing that behavior.

jakemac53:
ok, thanks! I will go ahead and wait to release this until the next normal release then.

lrhn:
(One possible, but not really likely, error case is if a single output line is more than 65536 bytes on Linux, then it will be broken at 65536, which may be in the middle of a UTF-8 encoding, leaving both part as invalid encodings).

Issue #915, opened 2018-07-25
Allow annotations to omit `const`


natebosch:
The CFE will already ensure the code is valid, the test package doesn't need to repeat the checks.

natebosch:
Hmm, parts of this are harder than I had hoped.

We use an unresolved `Expression` in order to figure out what constructor is being called. With `const` [this would be an `InstanceCreationExpression`](https://github.com/dart-lang/test/blob/769f0dd9ccb70a3def5834e4a37c00110fbcb068/lib/src/runner/parse_metadata.dart#L388). Without the `const` we end up with a `MethodInvocation`. Getting back to a class and constructor name isn't immediately obvious to me. @bwilkerson, @scheglov - is there a canonical way to handle this? I thought method calls were not allowed in const contexts (we're parsing inside an annotation so it should always be const) - should the analyzer be able to parse it to an `InstanceCreationExpression` in those cases?

chalin:
Any progress on this? /cc @kevmoo @kwalrath

natebosch:
I was able to fix the case for collection literals but haven't published yet.

I have not been able to fix the case for constructor calls like `@Timeout(Duration(seconds: 5))`. Parsing this is much trickier.

PR #916, opened 2018-07-25
Remove explicit checks for syntax errors

natebosch:
Towards #915

The CFE will already reject this code if it is incorrect, we don't need
to also check for syntax errors.

- Remove the explict checks for `const`
- Remove the explicit checks for using valid constructor names and
  argument types.
- Update test to remove most of the `const` in examples. Retain const
  constructor calls since we can't yet parse the resulting expression if
  it is omitted.
- Remove tests for the removed explicit checks.

### DIFF
--- a/test/runner/parse_metadata_test.dart
+++ b/test/runner/parse_metadata_test.dart
@@ -185,8 +185,8 @@ library foo;
 
     group("throws an error for", () {
       test("multiple @Tags", () {
-        new File(_path).writeAsStringSync(
-            "@Tags(['a'])\n@Tags(['b'])\nlibrary foo;");
+        new File(_path)
+            .writeAsStringSync("@Tags(['a'])\n@Tags(['b'])\nlibrary foo;");
         expect(() => parseMetadata(_path, new Set()), throwsFormatException);
       });
     });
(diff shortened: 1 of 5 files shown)


test/runner/parse_metadata_test.dart
```diff
@@ -354,9 +195,9 @@ library foo;
   group("@OnPlatform:", () {
     test("parses a valid annotation", () {
       new File(_path).writeAsStringSync("""
-@OnPlatform(const {
+@OnPlatform({
   'chrome': const Timeout.factor(2),
```
jakemac53: nit: consider omitting const for one of the constructors also

natebosch: Can't yet. See the note in the commit message: "Retain const
constructor calls since we can't yet parse the resulting expression if
it is omitted."

And details of why it's hard to support omitting this const at https://github.com/dart-lang/test/issues/915#issuecomment-407923698

jakemac53: Ah ok, maybe make a separate test and mark it as skipped then and link to the bug?

natebosch: Done

PR #921, opened 2018-08-13
Add missing return to fix analysis error with 2.1.0-dev.0.0.

keertip:


natebosch:
For future context - this is an analysis _hint_, not an error.

PR #933, opened 2018-09-13
Remove imports of `dart:io` from `test.dart` unless its available

jakemac53:
Adds conditional imports and moves some code around to eliminate non-conditional imports of `dart:io` from `test.dart`.

### DIFF
--- a/lib/src/util/io_stub.dart
+++ b/lib/src/util/io_stub.dart
@@ -1,3 +1,7 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
 import '../backend/runtime.dart';
 import '../backend/suite_platform.dart';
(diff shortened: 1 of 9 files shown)


lib/src/util/io_stub.dart
```diff
@@ -0,0 +1,5 @@
+import '../backend/runtime.dart';
```
natebosch: [nit] copyright?

lib/test.dart
```diff
@@ -39,6 +38,8 @@ export 'src/frontend/throws_matchers.dart';
 export 'src/frontend/timeout.dart';
 export 'src/frontend/utils.dart';
 
+import 'package:path/path.dart' as p;
```
natebosch: why did this move?

jakemac53: just accidental I think, was playing around with a few things and we don't have the import order lint I guess

jakemac53:
Running internal presubmit, assuming that is happy I will submit tomorrow

PR #935, opened 2018-09-14
Allow omitting const in parsed metadata

### LINKED ISSUE #915: Allow annotations to omit `const`

natebosch:
Fixes #915

When there is no `const` the expression is parsed as a
`MethodInvocation` instead of an `InstanceCreationExpression`. Update
all the places we were casting to the latter so that it can handle
either case.

Add some additional tests for import prefix cases.

### DIFF
--- a/lib/src/runner/parse_metadata.dart
+++ b/lib/src/runner/parse_metadata.dart
@@ -145,11 +145,13 @@ class _Parser {
   }
 
   /// Parses a `Timeout` constructor.
-  Timeout _parseTimeoutConstructor(InstanceCreationExpression constructor) {
-    var name = _parseConstructor(constructor, 'Timeout');
-    var args = constructor.argumentList.arguments;
-    if (name == null) return new Timeout(_parseDuration(args.first));
-    return new Timeout.factor(_parseNum(args.first));
+  Timeout _parseTimeoutConstructor(Expression constructor) {
+    var name = _findConstructorName(constructor, 'Timeout');
+    var arguments = _parseArguments(constructor);
+    if (name == null) return new Timeout(_parseDuration(arguments.first));
+    if (name == 'factor') return new Timeout.factor(_parseNum(arguments.first));
+    throw new SourceSpanFormatException(
+        'Invalid timeout', _spanFor(constructor));
   }
 
   /// Parses a `@Skip` annotation.
@@ -166,10 +168,10 @@ class _Parser {
   /// Parses a `Skip` constructor.
   ///
   /// Returns either `true` or a reason string.
-  _parseSkipConstructor(InstanceCreationExpression constructor) {
-    _parseConstructor(constructor, 'Skip');
-    var args = constructor.argumentList.arguments;
-    return args.isEmpty ? true : _parseString(args.first).stringValue;
+  _parseSkipConstructor(Expression constructor) {
+    _findConstructorName(constructor, 'Skip');
+    var arguments = _parseArguments(constructor);
+    return arguments.isEmpty ? true : _parseString(arguments.first).stringValue;
   }
 
   /// Parses a `@Tags` annotation.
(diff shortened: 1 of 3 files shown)


lib/src/runner/parse_metadata.dart
```diff
@@ -335,6 +365,62 @@ class _Parser {
...
+    var target = constructor.target;
+    if (target != null) {
+      // target could be an import prefix or a different class. Assume that
+      // named constructor on a different class won't match the class name we
+      // are looking for.
+      if (constructor.methodName.name == className) return null;
+      // target is an optionally prefixed class, method is named constructor
+      String parsedName;
+      if (target is SimpleIdentifier) parsedName = target.name;
+      if (target is PrefixedIdentifier) parsedName = target.identifier.name;
```
munificent: It would be helpful to show concrete examples of each case and which chunk of code matches them:

```dart
Timeout.factor()
prefix.Timeout()
prefix.Timeout.factor()
Timeout()
```

lib/src/runner/parse_metadata.dart
```diff
@@ -335,6 +365,62 @@ class _Parser {
...
+  /// Returns a type from [candidates] that _may_ be a type instantiated by
+  /// [constructor].
+  ///
+  /// This can be fooled - for instance the invocation `foo.Bar()` may look like
+  /// a prefixed instantiation of a `Bar` even though it is a named constructor
+  /// instantiation of a `foo`, or a method infocation on a variable `foo`, or
+  /// ...
+  ///
+  /// Similarly `Baz.another` may look like the named constructor invocation of
+  /// a `Baz`even though it is a prefixeda instantiation of an `another`, or a
```
munificent: "prefixeda" -> "prefixed".

lib/src/runner/parse_metadata.dart
```diff
@@ -335,6 +365,62 @@ class _Parser {
...
+  /// method invocation on a variable `Baz`, or ...
+  String _typeNameFromMethodInvocation(
+      MethodInvocation constructor, List<String> candidates) {
+    var methodName = constructor.methodName.name;
+    if (candidates.contains(methodName)) return methodName;
```
munificent: Show an example that this matches.

lib/src/runner/parse_metadata.dart
```diff
@@ -335,6 +365,62 @@ class _Parser {
...
+    var target = constructor.target;
+    if (target == null) return null;
+    if (target is SimpleIdentifier) {
+      if (candidates.contains(target.name)) return target.name;
```
munificent: Ditto.

lib/src/runner/parse_metadata.dart
```diff
@@ -335,6 +365,62 @@ class _Parser {
...
+    }
+    if (target is PrefixedIdentifier) {
+      // Looks some_prefix.SomeTarget.someMethod - "SomeTarget" is the only
+      // potential type name
```
munificent: "Looks" -> "Looks like".

Add a period.

lib/src/runner/parse_metadata.dart
natebosch: Done

lib/src/runner/parse_metadata.dart
natebosch: Done

lib/src/runner/parse_metadata.dart
natebosch: Done

lib/src/runner/parse_metadata.dart
natebosch: Done

lib/src/runner/parse_metadata.dart
natebosch: Done

PR #936, opened 2018-09-17
Drop optional new/const

natebosch:
- Run `dartfmt --fix`
- Limit to 2.0.0 stable and newer SDK.


a14n:
You can also add lints `unnecessary_new` and `unnecessary_const`.

natebosch:
> You can also add lints `unnecessary_new` and `unnecessary_const`.

Good call. https://github.com/dart-lang/test/pull/937

PR #945, opened 2018-10-17
Refactor test into two packages

jonahwilliams:
### Split `package:test` into `test` and `test_api`
The set of dependencies pulled in by package:test is large, and as a result so is the set of dependencies needed for flutter_test. Reducing these dependencies is critical for the code health of Flutter, and measurable helps our users by reducing the occurrences of version solving failures

This does not propose making a new package that most users would depend on - the intention is to be fully backwards compatible. The API of package:test will re-export the smaller core library. Users should to continue to use package:test as they are today, while tooling like flutter_test can build around the core library instead.

test_api:
```
lib/src/frontend/*
lib/src/backend/*

lib/src/runner/environment.dart 
lib/src/runner/plugin/hack_register_platform.dart
lib/src/runner/plugin/platform.dart
lib/src/runner/runner_suite.dart
lib/src/runner/configuration/runtime_selection.dart
lib/src/runner/configuration/suite.dart

lib/src/utils/iterable_set.dart
lib/src/utils/placeholder.dart
lib/src/utils/remote_exception.dart
lib/src/utils/stack_trace_mapper.dart
lib/src/utils/test.dart
lib/src/utils.dart
```

All imports of `test_api` into `test` have been left as implementation imports with an `ignore: implementation_imports` comment. I think we should refactor this so that we can export a `backend`, `frontend`, and `util` library from test_api, but I've left it mostly unchanged for now.

Updates:

- `StackTraceMapper` (test_api) is now an abstract class with no implementations. `JSStackTraceMapper` is the only implementation (test).
- test.dart (test_api) does not allow direct invocation and thus has no `_globalDeclarer`. test.dart (test) does still support this. The methods test, group, setUp, setUpAll, tearDown, tearDownAll have been duplicated in each library to remove the runner dependency from core.
- suite.dart (core) is modified to have a getter `isLoadSuite` which defaults to false. This is overridden by the LoadSuite class, removing the dependency on runner code from Invoker for an `is` check.
----
EDIT:
- Added full list of files
- Renamed test_core to test_api


jonahwilliams:
cc @grouma

jonahwilliams:
I'm happy to restructure the packages, it should mostly show up as "file moved" in the diff so it won't be too noisy.

Regarding the implementation imports, I mostly left them as is to reduce the size of the diff. Ideally, `test_core` would have a stable public API that `flutter_test` could depend on. Currently, Flutter is importing several implementation files, part of the reason for the tight version pinning. I don't quite know what the ideal API looks like, but as long as we can refine it in the future, I could start with a `backend.dart` and a `frontend.dart`.

jonahwilliams:
For an example of how Flutter will use this, I put together a placeholder PR here: https://github.com/flutter/flutter/pull/23136/files

in flutter_test I copied the RemoteListener (temporary): https://github.com/flutter/flutter/pull/23136/files#diff-0ad5bb27ea82c6f53e13f4fee9590e8a (not sure if there should be an implementation in test_core).

And then the flutter_tool defines a little bit of runner logic: https://github.com/flutter/flutter/pull/23136/files#diff-39538134dff328da5360691916bf0f06

package:test is still used for the executable, but I think eventually we would have own since we support a different set of options.

natebosch:
> package:test is still used for the executable, but I think eventually we would have own since we support a different set of options.

Are the reduced dependencies from `test_core` helpful today or will it wait to pay off until flutter has it's own executable?

natebosch:
I should have followed your links and I could have answered my own question :)

https://github.com/flutter/flutter/pull/23136/files#diff-6db519d06b539bb165a44ad589217e9e

I guess users add a dependency on package:test and run it that way?

jonahwilliams:
Unless I've done my math wrong it should pay off today. The test dependency is defined in `flutter_tool`, which users won't have a dependency on. Instead they just import `flutter_test` which contains only `test_core`.

`flutter_tool` invokes the test executable, passes the test script to the frontend compiler, and and runs it in an isolate.

grouma:
I took a look at this PR and for the most part it looks good to me. 

My main concern is that this change is compatible with Google3. It looks like:
` test_core = API_SRCS + ["hack_register_platform.dart", "platform.dart"]`
in third_party/dart/test. If that's the case this change should be fairly easy to roll internally.

It was hard for me to determine the exact files you were referring to when you stated: "// + several utils, based entirely on usage." Can you list them out entirely? I want to make sure we are not inadvertently adding restrictive SDK imports such as "dart:io".

As for implementation imports, I am not too concerned. All of our Google3 extensions, and as you know Flutter, have implementation imports. I prefer to stabilize on the API after things settle.

Finally, I also want to echo Nate's desire to not have nested packages.

grouma:
Also one more thing, I'm wondering if we should instead call `test_core`, `test_api`. Internally we have `third_party/dart/test` (which maps to `API_SRCS`). Users depend on that to actually write their tests. We then have `third_party/dart/test:core` which has the core functionality of running the tests. We extend core with a select set of extensions to make our various test runners, e.g. `test_runner_scuba` etc.

jonahwilliams:
I've updated the top comment with the full list of sources moved.  I did not see any imports of `dart:io`, but there is a single usage of `dart:isolate` to catch a particular exception type in `remote_exception.dart`. Not sure if that is okay.

As for naming, I don't have strong opinions here so I'm happy to change `test_core` to `test_api`.

grouma:
Thanks for the update. `dart:isolate` is a problem. This is actually a problem we have internally as well (that I was not aware of). I will have to resolve the issue likely by using conditional imports. No need to block this change on that though. Just expect me to rev the package version in the near future. One final fix required before LGTM. There are several bugs on various tools that don't properly handle `packages` directories. On the cautious side, instead of having a `packages` folder lets rename it to `pkgs`.

jonahwilliams:
QQ, should I point the version of package:test at the list version or leave it as is until we upload it?

jonahwilliams:
Oh, and what needs to be done about the dart.js file in static? It got caught by the gitignore when I moved the files but it was previously uploaded.

jonahwilliams:
Sounds good, I added the dart.js file back to static. I did update the test pubspec.yaml but I guess we'll have to roll everything anyway

grouma:
See if you can update the branch history of test.dart and utils.dart so we don't lose the history. Otherwise LGTM. Also would be good to get a passing Travis run but I'm fine with breaking head for a small period of time to make landing things easier.

jonahwilliams:
I realized I duplicated utils.dart, so I deleted the copy in `test`. Unfortunately I don't know a good way to save the history for the copy of test.dart, so I added a comment on the top import to point to test_api for now.

Hopefully we can figure out a better way to do that anyway.

Re: travis. I've had a _mostly_ successful run locally (though I was missing a lot of the required tools)

jonahwilliams:
@jakemac53 , @natebosch PTAL?

kevmoo:
Any reason to not just have `test` and `test_api` at the root? `pkgs` seems a bit superfluous. 

Also, are we sure we like `test_api` as a package name? Not `test_core`?

If you're looking for help configuring travis, may I suggestion `package:mono_repo` 😄

jonahwilliams:
I'm happy to commit to whatever names and package structure the Dart team wants

jakemac53:
I will defer to Nate and Gary on this one, as I don't have strong opinions here and they already reviewed it :). I am generally in support of the separation into two packages though.

natebosch:
@jonahwilliams - sorry for the slowness and churn here - this is an important turning point for the package and we're bikeshedding a bit in person. We'll make some decisions ASAP.

grouma:
Discussed with @natebosch. This LGTU as is. Thanks for the change! Expect a little churn over the next couple weeks as we nail down the split.

jonahwilliams:
No problem - and I'm happy to help where I can in the future, so don't hesitate to ask.

PR #949, opened 2018-10-30
Refactor package:test for use with package:test_api

grouma:
- Add changelogs to each pkg
- Update test_api to fold `test_api` frames by default
- Move test/test/frontend/* -> test_api/test/frontend/
- Move test/test/backend/* -> test_api/test/backend/
- Fork test/test/utils.dart and test/tool/dartfmt_and_analyze.sh for test_api
  - I will likely follow up this PR by creating a shared test utility directory
- Skip test broken by recent SDK change (opened corresponding issue)
- Add support for mono_repo

TODO:
- publish test_api 1.1.0
- publish test 1.5.0


pkgs/test_api/mono_pkg.yaml
```diff
@@ -0,0 +1,10 @@
+dart:
+ - dev
+
+stages:
+    - analyze_and_format:
+      - group:
+        - command: tool/dartfmt_and_analyze.sh
```
jakemac53: this doesn't need to be a group if its just one command - but note that if you do make it a group with the two explicit commands it might actually be a bit nicer:

- it will show in travis the exact dartfmt/dartanalyzer commands being ran
- it will tell you which one failed, and always run both instead of exiting when the first one fails

grouma: Done.

PR #955, opened 2018-11-01
fix test_api version

grouma:


grouma:
I haven't published yet and need to pull the VM back into core so dropping this one.

PR #959, opened 2018-11-04
Move remote_listener to test_api

grouma:
https://github.com/flutter/flutter/pull/23755 is unnecessarily forking `remote_listener.dart` which has a lot of internal logic on how to communicate with the test runner. We will expose this logic from `test_api` to better support their needs.


pkgs/test_core/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.2.0
+
+* Remote `remote_listener.dart` and `suite_channel_manager.dart` from runner
```
jonahwilliams:
```suggestion
* Remove `remote_listener.dart` and `suite_channel_manager.dart` from runner
```

PR #966, opened 2018-11-30
Allow use of analyzer version 0.34.x.

stereotype441:
The breaking changes between analyzer 0.33.x and 0.34.x do not affect
test or test_core.

### DIFF
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -1,12 +1,12 @@
 name: test
-version: 1.5.1
+version: 1.5.1+1
 author: Dart Team <misc@dartlang.org>
 description: A full featured library for writing and running Dart tests.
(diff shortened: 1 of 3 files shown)


grouma:
hint • 'analyzer' is deprecated and shouldn't be used at test/util/string_literal_iterator_test.dart:5:1 • deprecated_member_use

What's the replacement?

stereotype441:
> hint • 'analyzer' is deprecated and shouldn't be used at test/util/string_literal_iterator_test.dart:5:1 • deprecated_member_use
> 
> What's the replacement?

There isn't a suitable replacement available yet.  I've added an ignore comment to suppress the hint.

Sorry about this--we've instituted a process to make sure we stop deprecating things without providing suitable replacements, but the deprecation of `lib/analyzer.dart` happened before we started using that process.

stereotype441:
@grouma Ok, pubspec fixed.

PR #967, opened 2018-12-11
Ignore analyses warnings

grouma:


natebosch:
Change the commit message from "Fix" to "Ignore"

PR #968, opened 2018-12-19
Fix DartDocs links

srawlins:
Many of the dartdocs.org links were broken, redirecting incorrectly to bad pub.dartlang.org/documentation URIs.

Additionally, shorten many links to use [shortcut reference links](https://spec.commonmark.org/0.28/#shortcut-reference-link), where the reference ID was very similar to the link text. This makes the text more legible in its source format.

### DIFF
--- a/pkgs/test/README.md
+++ b/pkgs/test/README.md
@@ -23,11 +23,11 @@
 
 ## Writing Tests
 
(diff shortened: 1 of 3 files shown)


srawlins:
Travis failures are unrelated.

YOLO

PR #971, opened 2018-12-28
Update pkg:pedantic for all packages, fix associated unawaited lints

kevmoo:


pkgs/test/lib/src/runner/browser/browser.dart
```diff
@@ -134,7 +135,7 @@ abstract class Browser {
     // For example this happens with Chrome Headless.
     // See SDK issue: https://github.com/dart-lang/sdk/issues/31264
     for (var stream in _ioSubscriptions) {
-      stream.cancel();
+      unawaited( stream.cancel());
```
natebosch: @grouma - intentionally unawaited?

think this has anything to do with our force kill on travis?

kevmoo:
> Can you file an issue to have Gary or I look at those canceled stream subscriptions that aren't awaited?

Will do once this lands – so I can point to the code...

That's the nice thing about this lint, I think. Makes it easy to find/review all of the cases

PR #973, opened 2019-01-08
Allow for multiple calls to executable.main()

gamebox:
In Flutter, we would like to have our test runner be able to watch
a directory and re-run tests on file changes.  This requires making
multiple calls to the test_core executable, which with current
behavior causes it to throw on multiple calls to listen() on the
process signals stream.

Since the stream is properly cleaned up between runs, and we ensure
that only one subscription is active, move the _signals stream from
a final variable to a getter lambda expression.

Verified that the behavior is stable across multiple test runs within
a single process

### DIFF
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -661,4 +661,36 @@ void main() {
     expect(test.stdout, emitsThrough(contains("\u001b[31m")));
     await test.shouldExit();
   });
+
+  group("runs tests successfully more than once with --no-shutdown-on-completion flag", () {
+    test("defined in a single file", () async {
+      // final executablePath = await absolutePackagePath('test_core',
+      //     'lib/src/executable.dart');
+      await d.file("test.dart", _success).create();
+      await d.file("runner.dart", """
+import 'package:test_core/src/executable.dart' as test;
+
+void main(List<String> args) async {
+  await test.main(args);
+  await test.main(args);
+}
+      """).create();
+      var test = await runDart(
+        ["runner.dart", "--no-color", "--no-shutdown-on-completion", "--", "test.dart"],
+        description: 'dart runner.dart --no-shutdown-on-completion -- test.dart');
+      expect(test.stdout, emitsThrough(containsInOrder([
+        "+0: loading test.dart",
+        "+0: success",
+        "+1: success",
+        "All tests passed!"
+      ])));
+      expect(test.stdout, emitsThrough(containsInOrder([
+        "+0: loading test.dart",
+        "+0: success",
+        "+1: success",
+        "+1: All tests passed!",
+      ])));
+      await test.shouldExit(0);
+    });
+  });
 }
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -671,11 +671,10 @@ void main() {
 import 'package:test_core/src/executable.dart' as test;
 
 void main(List<String> args) async {
-  await test.runTests(args);
-  await test.runTests(args);
-  test.completeShutdown();
-}
-      """).create();
+  final sub = await test.runTests(args);
+  await test.runTests(args, signalSubscription: sub);
+  test.completeShutdown(sub);
+}""").create();
       var test = await runDart(["runner.dart", "--no-color", "--", "test.dart"],
           description: 'dart runner.dart -- test.dart');
       expect(
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -662,7 +662,7 @@ void main() {
     await test.shouldExit();
   });
 
-  group("runs tests successfully more than once with --no-shutdown-on-completion flag", () {
+  group("runs tests successfully more than once when calling runTests", () {
     test("defined in a single file", () async {
       // final executablePath = await absolutePackagePath('test_core',
       //     'lib/src/executable.dart');
@@ -671,13 +671,14 @@ void main() {
 import 'package:test_core/src/executable.dart' as test;
 
 void main(List<String> args) async {
-  await test.main(args);
-  await test.main(args);
+  await test.runTests(args);
+  await test.runTests(args);
+  test.completeShutdown();
 }
       """).create();
       var test = await runDart(
-        ["runner.dart", "--no-color", "--no-shutdown-on-completion", "--", "test.dart"],
-        description: 'dart runner.dart --no-shutdown-on-completion -- test.dart');
+        ["runner.dart", "--no-color", "--", "test.dart"],
+        description: 'dart runner.dart -- test.dart');
       expect(test.stdout, emitsThrough(containsInOrder([
         "+0: loading test.dart",
         "+0: success",
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -676,21 +676,24 @@ void main(List<String> args) async {
   test.completeShutdown();
 }
       """).create();
-      var test = await runDart(
-        ["runner.dart", "--no-color", "--", "test.dart"],
-        description: 'dart runner.dart -- test.dart');
-      expect(test.stdout, emitsThrough(containsInOrder([
-        "+0: loading test.dart",
-        "+0: success",
-        "+1: success",
-        "All tests passed!"
-      ])));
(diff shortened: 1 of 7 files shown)


jakemac53:
I wouldn't expect separate invocations of the executable to have any issue here, is this executable being kept alive somehow and then is `main` being manually invoked?

It would be good to get a test for the expected behavior in any case to make sure we don't regress it on accident with a similar pattern.

gamebox:
Since the declaration is at the top level the reference stays alive between
calls to main().  This is to get around that.  The reference is only used
to listen to it.  Since it is cleaned up, it should be good to go.

As for a test, what would you like to see?

grouma:
It depends on how you are calling `executable.main()`. Probably makes sense to add a runner test that calls main twice. Take a look here: https://github.com/dart-lang/test/blob/master/pkgs/test/test/runner/runner_test.dart

gamebox:
@grouma The problem with that is that runner_test.dart is creating a new process everytime.  The usecase here is calling main() as a function from a continuously running process.  Maybe I could write a similar test that creates a file that calls main twice and run that file in a process?  So like this

In test:
1. Create "test" file(s)
2. Create a fake "runner" file that does the following:
  1. Imports test_core/src/executable.dart as test
  2. Calls test.main() twice with appropriate args.
  3. Exits 0
3. Start a dart process with that "runner" file
4. Checks for expected output, and 0 exit code

This seems reasonable, but I honestly can't wrap my mind around how to import executable.dart from this "runner" file I created on the fly.

grouma:
What you have outlined sounds right to me. Take a look at the IO utilities here: https://github.com/dart-lang/test/blob/master/pkgs/test/test/io.dart#L96
It is using the current `PackageResolver` which should allow you to import executable.dart.

gamebox:
I was able to utilize the PackageResolver and import the file, but this leads to very unexpected behavior, I receive the following error which does seem to follow, since this code normally runs just fine as is:

```
[e] file:///usr/dart-test-core/pkgs/test_core/lib/src/runner/loader.dart:70:49: Error: A value of type 'VMPlatform' can't be assigned to a variable of type 'FutureOr<PlatformPlugin>'.
[e]  - 'VMPlatform' is from 'file:///usr/dart-test-core/pkgs/test_core/lib/src/runner/vm/platform.dart'.
[e]  - 'FutureOr' is from 'dart:async'.
[e]  - 'PlatformPlugin' is from 'file:///usr/dart-test-core/pkgs/test_core/lib/src/runner/platform.dart'.
[e] Try changing the type of the left hand side, or casting the right hand side to 'FutureOr<PlatformPlugin>'.
[e]     _registerPlatformPlugin([Runtime.vm], () => VMPlatform());
[e]   
```

Paths have been anonymized.  VMPlatform extends PlatformPlugin, and PlatformPlugin _is_ assignable to FutureOr<PlatformPlugin>.  Furthermore, we know this code works in normal usage.

The "runner file is as straightforward as you can get:

```dart
import 'dart:async';
import 'dart:io';
import '${await absolutePackagePath('test_core', 'lib/src/executable.dart')}' as test;

void main(List<String> args) async {
  await test.main(args);
  await test.main(args);
}
```

gamebox:
Verified by trying to do a cast at the line in loader where the exception is thrown, In this subprocess, it does not think that VMPlatform is a subtype of PlatformPlugin.  Ever seen this before @jakemac53 @grouma ?

grouma:
I have run into this issue once before. The issue is that you are mixing relative imports with absolute imports which confuses the type system. You should change the import path to be a `package:` URI. That *should* solve the issue.

Cheers to @natebosch for reminding me of this issue.

gamebox:
I already tried package URI, and thought it didn't work, turns out I was just passing the arguments in incorrectly.  After figuring out stream matcher semantics, there you go!

natebosch:
I don't like the idea of adding a special flag for this that wouldn't be used from the normal executable. I'm also very nervous anytime we are _not_ canceling subscriptions.

I'm curious what it would look like to do something like this:

- Add a new entrypoint method in `executable.dart` named something other than `main`. Perhaps `runTests`.
- Make the `stdInLines` an argument that gets passed in to `runTests`.
- Make the `main` method something that constructs a `stdInLines`, passes it to `runTests`, then cleans up afterwards.

Then to implement something that can call `runTests` multiple times your script would own the lifecycle of `stdInLInes` entirely.

Does that seem feasible?

gamebox:
Definitely should be, I'll whip it up

gamebox:
Are these test failures known flakes?  I don't see how my change could cause them...but I'm often proven wrong when I think that ;-)

natebosch:
> Are these test failures known flakes?

Does not look like it to me. I won't be able to look deeply until Monday, but at a glance my first suspicion is that turning `_signals` into a getter could be having some negative repercussions.

grouma:
The pause after load feature expects input to continue execution. Definitely seems related to your changes.

gamebox:
The only remaining failure I can't seem to debug, and it doesn't seem related to my change.  It looks like the format of an error changed.

gamebox:
Confirmed that this does not seem to be due to my change, built SDK from latest dart-sdk and ran the test that failed, and all 55 test cases in that suite passed.

```
$ ~/dart-sdk/sdk/out/ReleaseX64/dart-sdk/bin/pub run test test/runner/configuration/custom_platform_test.dart 
04:51 +55: All tests passed!  
```

grouma:
Head is failing as well: https://travis-ci.org/dart-lang/test/jobs/478792144

@natebosch was looking at the failures yesterday. I believe he said it was due to a change in the source span package. @natebosch were you able to root cause the issue / resolve it?

gamebox:
So it looks like we are waiting on someone to cut a 1.5.2 release with @natebosch 's change?

natebosch:
> So it looks like we are waiting on someone to cut a 1.5.2 release with @natebosch 's change?

Not exactly - we're waiting on someone to push fixes to the tests that were broken by the latest `source_span` release. I'll try to take a look at that soon.

gamebox:
I can do that if needed.  Let me know.

natebosch:
I think https://github.com/dart-lang/test/pull/975 should do it

gamebox:
Sounds good

gamebox:
Failure seems to be a flake. Passes on my workstation when rebased on master:HEAD and using sdk built from master:HEAD

grouma:
I triggered a rerun. fyi: click travis details and on the right is a button to restart a test.

gamebox:
Thanks @grouma

gamebox:
:tada: :tada: :tada:

pkgs/test_api/test/backend/declarer_test.dart
```diff
@@ -657,3 +657,12 @@ Future _runTest(Test test, {bool shouldFail = false}) {
 
   return liveTest.run();
 }
+
+void shouldNotFailListener(AsyncError error) {
+  registerException(error.error, error.stackTrace);
+}
+
+void shouldFailListener(AsyncError error) {
```
grouma: Are these methods used?

gamebox: Let me check

gamebox: Just checked, no references.  I can remove.

pkgs/test/test/runner/runner_test.dart
```diff
@@ -661,4 +661,39 @@ void main() {
     expect(test.stdout, emitsThrough(contains("\u001b[31m")));
     await test.shouldExit();
   });
+
+  group("runs tests successfully more than once when calling runTests", () {
+    test("defined in a single file", () async {
+      // final executablePath = await absolutePackagePath('test_core',
```
grouma: Dead code?

gamebox: Done

pkgs/test_core/lib/src/executable.dart
```diff
@@ -147,14 +164,16 @@ main(List<String> args) async {
...
+    await runner?.close();
   }
 
   // TODO(grouma) - figure out why the executable can hang in the travis
   // environment. https://github.com/dart-lang/test/issues/599
   if (Platform.environment["FORCE_TEST_EXIT"] == "true") {
     exit(exitCode);
   }
+
+  return;
```
natebosch: did something complain before you added this? It looks unnecessary to me.

gamebox: I can't really remember, I think I got a lint before.  I'll put in a PR to clean it up

PR #978, opened 2019-01-29
allow analyzer version 0.35.x

jakemac53:


pkgs/test/CHANGELOG.md
```diff
@@ -9,11 +13,11 @@
 
 ## 1.5.0
 
-* Depend on `package:test_core` for core functionality. 
+* Depend on `package:test_core` for core functionality.
```
grouma: Is this an intended change?

kevmoo: Guessing this is @jakemac53 's markdown auto-format on save thingy

kevmoo:
@jakemac53 – FYI I pushed a few things here to make it pass and ensure tests are run w/ dart_core + latest things

pkgs/test/CHANGELOG.md
jakemac53: Yep - figured I might as well leave it in since it cleans it up a bit

pkgs/test_core/pubspec.yaml
```diff
@@ -43,3 +43,8 @@ dev_dependencies:
 dependency_overrides:
   test_api:
     path: ./../test_api
+
+  # to remove before publish
+  analyzer: ^0.35.0
+  test:
```
jakemac53: There are no tests in this package, and no package:test dependency... :P

kevmoo: I know...and yet if you don't add this, you won't get the latest version of pkg:test and few other things...it's weird

jakemac53: Ah it picks up a transitive dep on package:test.

The package:test version solve does get the latest analyzer though, and that is where the tests run, so I don't think this is necessary.

kevmoo: we'll nix all of the this before publish anyway 🤷‍♂️

jakemac53: Lol so this package has a bunch of dev deps, but no tests. As far as I can tell they can all be deleted which removes the circular dep.

kevmoo:
SGTM!

PR #986, opened 2019-02-22
Drop special format for IsolateSpawnException

natebosch:
The SDK no longer outputs anything that looks like "Load Error". The
last commit which removed this pattern was
250773cb651beec1a61e126a6c0756aace9c19ca

I'm also unable to reproduce any case where the `toString()` includes
"Stack Trace:\n", and the file names don't look like they need to be
adjusted at all.


natebosch:
https://github.com/dart-lang/sdk/commit/250773cb651beec1a61e126a6c0756aace9c19ca removed the last usage of "Load Error"

Issue #972, opened 2018-12-29
Investigate unawaited calls in library code


kevmoo:
See calls to `unawaited` in https://github.com/dart-lang/test/commit/8a90f38080f4d313f9d87e633e2a5a9b36b5a1bd

Need to investigate.

CC @natebosch @grouma

jakemac53:
I glanced through these and I don't see anything concerning, its mostly just cleanup code (deleting temp dirs after browsers shut down, closing streams on shutdown, etc).

I will leave it open for now but if somebody else takes a look and agrees then I think we can close this.

jakemac53:
That would only be for debugging though right? Not awaiting them shouldn't have any negative impact.

natebosch:
Yeah, the await would be intended to give us confidence in what we're looking at when debugging, I don't expect it should cause the hang.

grouma:
I actually went through and addressed all unawaited futures a while back when I was investigating the Travis issue. Unfortunately it didn't help and that's why I abandoned the PR. I still think we should clean these up though so let's leave this open.

PR #988, opened 2019-02-23
Re-centralize the analysis_options.yaml

natebosch:
After splitting the package the file was copied to some of the package
but missing from the `test_core` package. Move it back up the directory
structure so it is picked up for all sub-packages.


natebosch:
I noticed this because I was _not_ getting the `implicit-casts: false` behavior when editing in `test_core` which leaves us at risk for regressions.

grouma:
Seems related to: https://github.com/dart-lang/test/issues/972


Nice. I wonder if this will help https://github.com/dart-lang/test/issues/599

PR #995, opened 2019-03-06
Prep for analyzer changes to support the "UI as code" feature.

stereotype441:
As part of implementing the "UI as code" feature
(https://github.com/dart-lang/language/blob/master/accepted/future-releases/unified-collections/feature-specification.md)
the analyzer is deprecating the following classes and methods:

- ForStatement and ForEachStatement (ForStatement2 is now used for
  both kinds of loops)

- ListLiteral.elements (use elements2, which has a more general return
  type)

- AstVisitor.visitForEachStatement and AstVisitor.visitForStatement
  (visitForStatement2 is now used for both kinds of loops)

- AstVisitor.visitMapLiteral and AstVisitor.visitSetLiteral
  (visitSetOrMapLiteral is now used for both kinds of literals).

See deprecation CL: https://dart-review.googlesource.com/c/sdk/+/95665/

### DIFF
--- a/pkgs/test_core/lib/src/runner/parse_metadata.dart
+++ b/pkgs/test_core/lib/src/runner/parse_metadata.dart
@@ -437,14 +437,19 @@ class _Parser {
     key ??= (expression) => expression as K;
     value ??= (expression) => expression as V;
 
-    if (expression is! MapLiteral) {
+    if (expression is! SetOrMapLiteral) {
       throw SourceSpanFormatException('Expected a Map.', _spanFor(expression));
     }
 
-    var map = expression as MapLiteral;
-
-    return Map.fromIterables(map.entries.map((e) => key(e.key)),
-        map.entries.map((e) => value(e.value)));
+    var map = <K, V>{};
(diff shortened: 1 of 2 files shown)


natebosch:
I worry about the behavior of failure case in `elements2.cast<Expression>()` in the case where the users does try to use a conditional element.

What do I need to do to make that cast fail?

stereotype441:
> I worry about the behavior of failure case in `elements2.cast<Expression>()` in the case where the users does try to use a conditional element.
> 
> What do I need to do to make that cast fail?

Good point.  To make the cast fail you need to enable the experimental flags `control-flow-collections` and `spread-collections`, and then use either the spread or control flow syntax in the list being parsed.  I'm not too familiar with this code so I'm not sure how to actually achieve that situation with the test package.  I'm happy to discuss over IM or VC if that would help.

natebosch:
Added an explicit check that the element is an `Expression` and throwing a more helpful exception if it isn't. This should point the user to exactly what they need to change.

stereotype441:
Thanks, Nate!  Your change lgtm.

PR #999, opened 2019-03-22
test & test_core: allow analyzer 0.36.0

kevmoo:


kevmoo:
@grouma – we should get these published soonish

PR #1000, opened 2019-03-22
Update matcher version and prepare to publish

natebosch:


pkgs/test/pubspec.yaml
```diff
@@ -33,8 +33,8 @@ dependencies:
   web_socket_channel: '^1.0.0'
   yaml: '^2.0.0'
   # Use an exact version until the test_api and test_core package are stable.
-  test_api: '0.2.2'
-  test_core: '0.2.1+1'
+  test_api: '0.2.3'
```
kevmoo: Missed version bump for `pkg:test` – going to do that in a separate PR?

natebosch: Was an oversight. Fixed now and merged the changelog entry from the unpublished version.

kevmoo:
We expect travis to blow up here, right? – because version constraints?

natebosch:
> We expect travis to blow up here, right? – because version constraints?

I don't think so? We have overrides.

PR #1001, opened 2019-03-26
Update imports to allow stream_channel 2.0.0

natebosch:
Version `1.7.0` adds the import so it's the minimum version required.
Allow versions through `2.x` since the breaking change will be to change
the import which will now be satisfied.


natebosch:
Note - I'm holding on this for a day while we check that some other changes planned in `package:stream_channel` don't break the usage here.

Issue #562, opened 2017-03-07
Changing window.location causes test to hang 


srawlins:
This might be expected, but I thought I'd open a bug in case. Take this Dart:

```dart
@TestOn('browser')
import 'package:test/test.dart';
import 'dart:html';

void main() {
  test('x', () {
    window.location.hash = 'seven';
  });
}
```

Run through test: `pub run test -p dartium b.dart` (can replace dartium with content-shell, chrome, ...). Test hangs.

kevmoo:
One off brainstorm: could pkg/test at least listen for navigation events. Then could mark things as fail or cancel or something...

nex3:
This is because we currently use the hrefs of incoming messages to determine which message corresponds to which worker. I'd really like to switch to using `MessageChannel`s instead, but they're hopelessly broken in Dart. If we can get https://github.com/dart-lang/sdk/issues/26349 and https://github.com/dart-lang/sdk/issues/26351 fixed, this will be really easy to fix as well.

kevmoo:
Still blocked on https://github.com/dart-lang/sdk/issues/26351

kevmoo:
Should be golden in `2.0.0-dev.44.0`+

Issue #423, opened 2016-04-15
Use MessageChannels for communicating with iframes


nex3:
Now that Firefox supports [the MessageChannel API](https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel), we should use it to communicate between the host frame and its iframes. This is not only more elegant, it will avoid issues with the Google-internal continuous test runner where multiple instances of the same suite could step on one another's events.

The best way to do this is probably to create a `MessagePortChannel` which provides a StreamChannel API for a MessagePort. It might even be worth putting this in `stream_channel` itself.

/cc @goderbauer

nex3:
This is blocked on dart-lang/sdk#26349.

nex3:
Removing the milestone because Q2 is past and there doesn't seem to be any work being done on the blocking issues.

kevmoo:
Seems a slam-dunk browser-support wise - http://caniuse.com/#search=MessageChannel

kevmoo:
Should be in `2.0.0-dev.43.0`

kevmoo:
Issue with calling `start` has been fixed. Should be golden in `2.0.0-dev.44.0`+

PR #1002, opened 2019-03-28
Use message channels for communicating with iframes

leonsenft:
The host code that runs in the browser and loads tests now uses the
Channel Messaging API to communicate with iframes. It now connects to
each test suite's iframe with its own dedicated `MessageChannel`.

This method is more robust than sending global messages through
`window.postMessage()`.

It also allows tests to modify `window.location`, which would previously
cause them to hang. This was because each iframe used `window.location`
(which is unique to each iframe) to tag messages sent using
`window.postMessage()` so that the host could identify their source.

Closes https://github.com/dart-lang/test/issues/423.
Fixes https://github.com/dart-lang/test/issues/562.


grouma:
I'm in favor of this change. Our internal testing infra should also be updated to follow this pattern. It may be easier to push that one through first.

jakemac53:
+1 to testing it out internally first just to be safe but :+1:

leonsenft:
This has landed internally.

PR #1012, opened 2019-04-04
Disable dartfmt travis tests

natebosch:


natebosch:
It's not clear this will continue to be the style - https://github.com/dart-lang/dart_style/issues/799

We want to get travis green for now.

kevmoo:
Or just ignore it until the fix is rolled – @munificent is working on it now

kevmoo:
PR in flight. I'd just skip it for now?

kevmoo:
https://github.com/dart-lang/dart_style/pull/800

natebosch:
> I'd just skip it for now?

I suppose. The nice thing about reformatting and then formatting back is that we can't forget about it...

If we skip the test we might forget to re-enable.

jakemac53:
fwiw I much prefer keeping the test and just accepting the new formatting - we will immediately know when we can fix it again. This is a lot more likely to remain "skipped" well beyond when the issue is fixed because we get zero signal about it when the fix lands.

kevmoo:
> fwiw I much prefer keeping the test and just accepting the new formatting - we will immediately know when we can fix it again. This is a lot more likely to remain "skipped" well beyond when the issue is fixed because we get zero signal about it when the fix lands.

There is a new dev SDK out w/ the "right fix". Just need to go through and enable formatting on dev-only and we should be good!

PR #1013, opened 2019-04-05
expose the parseMetadata function and all required classes to use it

jakemac53:
Towards https://github.com/dart-lang/build/issues/1554 (makes that a trivial change in build_test)

### DIFF
--- a/pkgs/test_api/CHANGELOG.md
+++ b/pkgs/test_api/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.2.5
+
+* Expose the  `Metadata`, `PlatformSelector`, `Runtime`, and `SuitePlatform`
+  classes publicly.
+
 ## 0.2.4
 
 * Allow `stream_channel` version `2.0.0`.
@@ -12,16 +17,16 @@
 
 ## 0.2.1
 
-* Add `remote_listener.dart` and `suite_channel_manager.dart`. 
+* Add `remote_listener.dart` and `suite_channel_manager.dart`.
 
 ## 0.2.0
 
-* Remove "runner" extensions. 
+* Remove "runner" extensions.
 
 
 ## 0.1.1
 
-* Update `stack_trace_formatter` to fold `test_api` frames by default. 
+* Update `stack_trace_formatter` to fold `test_api` frames by default.
 
 
 ## 0.1.0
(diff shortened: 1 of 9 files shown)


pkgs/test_core/lib/src/runner/parse_metadata.dart
```diff
@@ -24,8 +24,11 @@ import '../util/dart.dart';
 ///
 /// Throws an [AnalysisError] if parsing fails or a [FormatException] if the
 /// test annotations are incorrect.
-Metadata parseMetadata(String path, Set<String> platformVariables) =>
-    _Parser(path, platformVariables).parse();
+///
+/// If [contents] is `null` then it will be read as a [File] using [path].
+Metadata parseMetadata(String path, Set<String> platformVariables,
```
natebosch: I would make this always take `content` and update the calling locations to read the file themselves.

pkgs/test_core/lib/backend.dart
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+export 'package:test_api/src/backend/metadata.dart' show Metadata;
```
natebosch: should this be `export 'package:test_api/backend.dart' show ...;`?

pkgs/test_api/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.2.5
+
+* Expose the  `Metadata`, `PlatformSelector`, `Runtime`, and `SuitePlatform`
```
natebosch: Do you want to take a pass through other code using this and see if we can get rid of any existing `src/` imports from the other packages? Or should we lump that in with the other cleanup later?

jakemac53: I just did that for `test_core` for a few imports, but ya I think we can handle that for other packages later. We could probably search all the packages we pull internally the easiest?

pkgs/test_core/lib/backend.dart
jakemac53: done

pkgs/test_core/lib/src/runner/parse_metadata.dart
natebosch: Did this slip?

Is it still worth doing? We haven't published yet...

jakemac53: Oh weird - somehow I didn't see this comment when I was merging. I will go ahead and send out another PR for it.

PR #1014, opened 2019-04-08
travis: Run dartfmt only on dev

kevmoo:
Run analyzer (at a minimum) across all packages at their oldest supported SDK

### DIFF
--- a/pkgs/test/mono_pkg.yaml
+++ b/pkgs/test/mono_pkg.yaml
@@ -1,11 +1,15 @@
 dart:
- - dev
+  - dev
 
 stages:
     - analyze_and_format:
       - group:
-        #- dartfmt: sdk
+        - dartfmt: sdk
         - dartanalyzer: --fatal-infos --fatal-warnings .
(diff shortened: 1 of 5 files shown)


jakemac53:
It looks like we need to add `deprecated_member_use_from_same_package: ignore` to the errors section of the analysis_options.yaml.

kevmoo:
> It looks like we need to add `deprecated_member_use_from_same_package: ignore` to the errors section of the analysis_options.yaml.

Nope. Just run the right test...

pkgs/test/mono_pkg.yaml
```diff
@@ -1,11 +1,15 @@
 dart:
- - dev
+  - dev
```
natebosch: Why was this section removed in `test_api` but not here?

kevmoo: Weirdness. `mono_repo` doesn't have support for specifying the SDK in a stage w/out a group – that's the short version.

PR #1016, opened 2019-04-08
Make the FORCE_TEST_EXIT env global

natebosch:
We have had flakes on travis because this environment does not appear to be taking effect. This is probably because it's being overridden by the individual environments in each job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ language: dart
 sudo: required
 addons:
   chrome: stable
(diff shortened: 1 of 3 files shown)


kevmoo:
Weird random failure https://travis-ci.org/dart-lang/test/jobs/517486296#L528

natebosch:
> Weird random failure https://travis-ci.org/dart-lang/test/jobs/517486296#L528

Turns out we have a test which explicitly _can't_ force exit.

Issue #1003, opened 2019-03-28
Out of memory failures look like successful runs


aaronstgeorge-wf:
We've found that if the browser runs out of memory while running unit tests (due to memory leaks in the dart code under test) the test runner reports success. It seems like some sort of error should be reported.

Example:
```dart
@TestOn("browser")

import 'dart:html';
import 'dart:math';

import "package:test/test.dart";

void createMemoryLeak() {
  var a = new List.filled(100000000, pow(2, 53));
  window.onClick.listen((_) {
    print(a);
  });
}

void main() {
  test("leak a bunch of memory 1", () {
    createMemoryLeak();
  });

  test("leak a bunch of memory 2", () {
    createMemoryLeak();
  });

  test("leak a bunch of memory 3", () {
    createMemoryLeak();
  });

  test("leak a bunch of memory 4", () {
    createMemoryLeak();
  });

  test("this test should fail", () async {
    expect(true, isFalse);
  });
}
```
Running this test suite on my machine in chrome only gets through three of the tests but reports success with a `0` exit code.
```bash
~/Dev/dart/oom pub run test -p chrome test/oom_test.dart --reporter=expanded
00:00 +0: compiling test/oom_test.dart
Compiled 8,598,577 characters Dart to 1,021,698 characters JavaScript in 4.33 seconds
Dart file (/private/var/folders/mm/5_ns8ybs7pngk54qtdmcc8j40000gp/T/dart_test_WuvG1r/runInBrowser.dart) compiled to JavaScript: /private/var/folders/mm/5_ns8ybs7pngk54qtdmcc8j40000gp/T/dart_test_wdfpNc/test_W2KUPN/oom_test.dart.browser_test.dart.js

00:00 +0: leak a bunch of memory 1
01:03 +1: leak a bunch of memory 2
02:05 +2: leak a bunch of memory 3
02:11 +2: All tests passed!
~/Dev/dart/oom echo $?
0
```

grouma:
Thanks for the issue. This seems like a fun one :P

I did a little initial digging. 

For some reason the engine is reporting success here:
https://github.com/dart-lang/test/blob/master/pkgs/test_core/lib/src/runner/engine.dart#L94

This comment seems very problematic:
https://github.com/dart-lang/test/blob/master/pkgs/test_api/lib/src/backend/state.dart#L16

I don't have the context for why that decision was made so we'll have to investigate further to see how changing that assumption impacts the engine.

robbecker-wf:
[...]

PR #1024, opened 2019-04-29
Non-completed tests should cause failures

grouma:
- Update engine to consider non-completed tests as a failure run
- Update `compact` and `expanded` reporters to list the non-completed tests
- Add corresponding test
- Rev version numbers

Closes https://github.com/dart-lang/test/issues/1003

### DIFF
--- a/pkgs/test/test/runner/engine_test.dart
+++ b/pkgs/test/test/runner/engine_test.dart
@@ -5,10 +5,10 @@
 import 'dart:async';
 
 import 'package:pedantic/pedantic.dart';
+import 'package:test/test.dart';
 import 'package:test_api/src/backend/group.dart';
 import 'package:test_api/src/backend/state.dart';
 import 'package:test_core/src/runner/engine.dart';
-import 'package:test/test.dart';
 
 import '../utils.dart';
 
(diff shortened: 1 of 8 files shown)


pkgs/test/test/runner/engine_test.dart
```diff
@@ -58,6 +59,19 @@ void main() {
     engine.suiteSink.close();
   });
 
+  test("returns failure if all tests do not complete", () async {
```
natebosch: phrasing is weird. "if any test does not complete" would make more sense to me.

pkgs/test/test/runner/engine_test.dart
```diff
@@ -58,6 +59,19 @@ void main() {
...
+    var engine = declareEngine(() {
+      for (var i = 0; i < 3; i++) {
+        test("test ${i + 1}", () async {
+          await Future.delayed(Duration(seconds: 5));
```
natebosch: can we use a `Completer`  that we never `.complete()` to make this more explicit than a delay?

pkgs/test/test/runner/engine_test.dart
```diff
@@ -58,6 +59,19 @@ void main() {
...
+        });
+      }
+    });
+
+    unawaited(expectLater(engine.run(), completion(isFalse)));
```
natebosch: change `unawaited(expectLater(...))` to `expect(...)`

pkgs/test/test/runner/engine_test.dart
grouma: Discussed offline. These tests are odd so I can't open non-completed tasks. We came up with an alternative that I'll implement though.

PR #1025, opened 2019-04-30
Fix broken head

grouma:
- Skip tests that require  https://github.com/dart-lang/vm_service_client/pull/46

### DIFF
--- a/pkgs/test/test/runner/pause_after_load_test.dart
+++ b/pkgs/test/test/runner/pause_after_load_test.dart
@@ -30,8 +30,8 @@ void main() {
 import 'package:test/test.dart';
 
 void main() {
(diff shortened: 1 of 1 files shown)


pkgs/test/test/runner/pause_after_load_test.dart
```diff
@@ -30,8 +30,8 @@ void main() {
 import 'package:test/test.dart';
 
 void main() {
-  print('loaded test 2!');
```
kevmoo: seems a weird delta...

grouma: Yeah I noticed that. Just ran dartfmt though.

PR #1027, opened 2019-04-30
Don't reroute exceptions thrown in expectAsync

natebosch:
Allows using `expectAsync*` to wrap callbacks in context where they are
called with a `try/catch` without changing visible behavior.

### DIFF
--- a/pkgs/test_api/test/frontend/expect_async_test.dart
+++ b/pkgs/test_api/test/frontend/expect_async_test.dart
@@ -300,30 +300,12 @@ void main() {
     });
   });
 
-  group("with errors", () {
-    test("reports them to the current test", () async {
-      var liveTest = await runTestBody(() {
-        expectAsync0(() => throw TestFailure('oh no'))();
-      });
-
-      expectTestFailed(liveTest, 'oh no');
+  test("allows errors", () async {
+    var liveTest = await runTestBody(() {
+      expect(expectAsync0(() => throw 'oh no'), throwsA('oh no'));
     });
 
-    test("swallows them and returns null", () async {
-      Function returnValue;
-      var caughtError = false;
-      var liveTest = await runTestBody(() {
-        try {
-          returnValue = expectAsync0(() => throw TestFailure('oh no'))();
-        } on TestFailure catch (_) {
-          caughtError = true;
-        }
-      });
-
-      expectTestFailed(liveTest, 'oh no');
-      expect(returnValue, isNull);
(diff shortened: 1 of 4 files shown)


natebosch:
I'm not 100% sure of the full impact of this or whether we should consider it breaking...

natebosch:
This is covered by a test which looks like it was in place from the beginning:

https://github.com/dart-lang/test/commit/0a098e7f895842db421b7fad29eb1de0d1047ee8

There doesn't appear to be any specific discussion about this in [the review](https://codereview.chromium.org//934413002) and it's a large CL...

The `catch` and `return null` go back all the way to at least https://github.com/dart-lang/test/commit/c55a076b1d642185e5456a9792685d805133c058

Reviewed in https://codereview.chromium.org//869043002

This was a refactoring from https://codereview.chromium.org/869043002/diff/20001/lib/src/spread_args_helper.dart which did not have a `return null`.

It's hard to know for sure, but I _think_ the behavior might be a holdover from when this code served some other purpose and isn't strictly needed...

kevmoo:
Reasonable to throw it through Tap internally?

grouma:
Already ran a TAP presubmit.

kevmoo:
Sweet!

*From: *Gary Roumanis <notifications@github.com>
*Date: *Wed, May 8, 2019 at 1:46 PM
*To: *dart-lang/test
*Cc: *Kevin Moore, Comment

Already ran a TAP presubmit.

pkgs/test_api/test/frontend/expect_async_test.dart
```diff
@@ -300,30 +300,12 @@ void main() {
     });
   });
 
-  group("with errors", () {
-    test("reports them to the current test", () async {
```
jakemac53: It seems like we should restore this test (checking that async errors are reported)?

natebosch: I don't think we want them to be reported. It should be allowed to have some callback somewhere throw and not fail the test.

natebosch: Ah, I may have misunderstood.

I don't think that this test was useful for testing the behavior of async errors. That should be covered elsewhere.

jakemac53: Ok, I don't have enough context here to know it just seems wrong on the surface to delete that test :man_shrugging:

PR #1028, opened 2019-05-02
Ensure that the test state transitions to complete before the test is considered complete.

poletti-marco:
In some environments the state change events and the "test complete" event are sent over a stream that does not guarantee ordering (e.g. as independent HTTP requests when using package:sse).
Without this change, such tests will fail intermittently, when the "test complete" event happens to be delivered before the state change event.

### DIFF
--- a/pkgs/test_api/lib/src/backend/live_test_controller.dart
+++ b/pkgs/test_api/lib/src/backend/live_test_controller.dart
@@ -35,11 +35,11 @@ class _LiveTest extends LiveTest {
 
   Stream<Message> get onMessage => _controller._onMessageController.stream;
 
-  Future get onComplete => _controller.completer.future;
+  Future<void> get onComplete => _controller.onComplete;
 
-  Future run() => _controller._run();
+  Future<void> run() => _controller._run();
 
-  Future close() => _controller._close();
+  Future<void> close() => _controller._close();
 
   _LiveTest(this._controller);
 }
@@ -101,13 +101,15 @@ class LiveTestController {
   final _onMessageController = StreamController<Message>.broadcast(sync: true);
 
   /// The completer for [LiveTest.onComplete];
-  final completer = Completer();
+  final completer = Completer<void>();
 
   /// Whether [run] has been called.
   var _runCalled = false;
 
   /// Whether [close] has been called.
   bool get _isClosed => _onErrorController.isClosed;
+
+  final _stateChangedToComplete = Completer<void>();
 
   /// Creates a new controller for a [LiveTest].
   ///
@@ -158,6 +160,11 @@ class LiveTestController {
 
     _state = newState;
     _onStateChangeController.add(newState);
+
+    if (newState.status == Status.complete &&
+        !_stateChangedToComplete.isCompleted) {
+      _stateChangedToComplete.complete();
+    }
   }
 
   /// Emits message over [LiveTest.onMessage].
@@ -173,7 +180,7 @@ class LiveTestController {
 
   /// A wrapper for [_onRun] that ensures that it follows the guarantees for
   /// [LiveTest.run].
-  Future _run() {
+  Future<void> _run() {
     if (_runCalled) {
       throw StateError("LiveTest.run() may not be called more than once.");
     } else if (_isClosed) {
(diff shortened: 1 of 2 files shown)


pkgs/test_api/lib/src/backend/live_test_controller.dart
```diff
@@ -186,9 +193,15 @@ class LiveTestController {
     return liveTest.onComplete;
   }
 
+  /// Returns a future that completes when the test is complete.
+  ///
+  /// We also wait for the state to transition to Status.complete.
+  Future get onComplete =>
```
jakemac53: nit: `Future<void>` (I realize the rest of the file doesn't use this style, but `Future<void>` didn't exist when it was written)

jakemac53:
It looks like the travis failure is just due to needing a `dartfmt`

pkgs/test_api/lib/src/backend/live_test_controller.dart
poletti-marco: Now added `<void>` in a bunch of places.

poletti-marco:
The tests on Travis CI seem to be failing: https://travis-ci.org/dart-lang/test/builds/527418451
The issues don't seem related to my change (or are they?).
How can I reproduce locally?

I tried setting Travis-like env vars and running "./tool/travis.sh command_0" but I get compiler crashes.
Maybe it's because I've done the internal Dart setup instead of the open source one?

grouma:
The failures are almost certainly related to your change.

It's difficult to emulate the Travis environment entirely. That being said you can easily run the tests locally through `pub run test`.

You can also dig into the travis runs to see the specific changes and just run a single one, e.g. `pub run test test/runner/signal_test.dart`.

I ran a couple of the problematic tests locally and I'm getting the same failures. It looks like the tests are not completing so we must be overlooking some edge case.

poletti-marco:
I think I've found the issue now.
However I get linter errors in Travis CI unrelated with my change.
I also tried to fix those in this PR, but they are still reported.
Could you please advise what to do?

poletti-marco:
Actually it turns out that fixing them in this commit works.

pkgs/test/lib/src/runner/node/socket_channel.dart
```diff
@@ -3,7 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 @JS()
-library _;
+library library_name;
```
grouma: Why is this change necessary?

pkgs/test_api/lib/src/backend/live_test_controller.dart
```diff
@@ -186,9 +193,15 @@ class LiveTestController {
     return liveTest.onComplete;
   }
 
+  /// Returns a future that completes when the test is complete.
+  ///
+  /// We also wait for the state to transition to Status.complete.
+  Future<void> get onComplete =>
```
grouma: These changes should be outlined in the `CHANGELOG` and we should bump the `pubspec` version.

pkgs/test_api/lib/src/frontend/expect.dart
```diff
@@ -138,8 +138,9 @@ Future _expect(actual, matcher,
 
   var matchState = {};
   try {
-    if ((matcher as Matcher).matches(actual, matchState))
+    if ((matcher as Matcher).matches(actual, matchState)) {
```
grouma: Did we add a lint for this recently? Not opposed to the clean up but it makes the PR hard to follow. Feel free to open an issue and I'll just fix all packages in a single go. I'd drop these type of changes from this PR.

poletti-marco: It's true that it's unrelated.
Reverted these changes from here, and filed http://go/gh/dart-lang/test/issues/1032.
Once that is fixed I'll merge and make sure that Travis succeeds on this PR.

grouma: I see. I didn't realize we now make use of `package:pedantic` for lints. Changes to that package will break us. Fix is here: https://github.com/dart-lang/test/pull/1033

Sorry about that.

pkgs/test/lib/src/runner/node/socket_channel.dart
poletti-marco: To make a lint check happy. Now reverted together with the rest.

pkgs/test_api/lib/src/frontend/expect.dart
poletti-marco: The fix is in (thanks!), merged now.

pkgs/test_api/lib/src/backend/live_test_controller.dart
poletti-marco: Done.

poletti-marco:
Travis passed, this is ready to merge.

PR #1034, opened 2019-05-15
expectAsync0 fails within callbacks with non-zero count

joshprzybyszewski-wf:
Suddenly, we started seeing unexplained test failures. I'm not sure what exactly is the problem, but I got around it with using a completer to check that my callback function is only called once. Any ideas?


grouma:
Closing this in favor of https://github.com/dart-lang/test/issues/1035

PR #1036, opened 2019-05-22
Stop using deprecated analyzer API left over from "UI as code" migration.

stereotype441:
Since analyzer version 0.36.0, we no longer need to use elements2.
elements is equivalent.

### DIFF
--- a/pkgs/test_core/lib/src/runner/parse_metadata.dart
+++ b/pkgs/test_core/lib/src/runner/parse_metadata.dart
@@ -443,7 +443,6 @@ class _Parser {
     }
 
     var map = <K, V>{};
-    // ignore: deprecated_member_use
     for (var element in (expression as SetOrMapLiteral).elements) {
       if (element is MapLiteralEntry) {
         map[key(element.key)] = value(element.value);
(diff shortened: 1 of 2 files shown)


pkgs/test_core/lib/src/runner/parse_metadata.dart
```diff
@@ -464,7 +464,7 @@ class _Parser {
     var list = expression as ListLiteral;
 
     // ignore: deprecated_member_use
```
natebosch: Can we remove the ignores now?

jakemac53: That would still break build_web_compilers 1.x users

pkgs/test_core/pubspec.yaml
```diff
@@ -8,7 +8,7 @@ environment:
   sdk: ">=2.1.0 <3.0.0"
 
 dependencies:
-  analyzer: ">=0.35.3 <0.37.0"
+  analyzer: ">=0.36.0 <0.37.0"
```
kevmoo: Use `^` syntax?

pkgs/test_core/lib/src/runner/parse_metadata.dart
stereotype441: > Can we remove the ignores now?

Done!

pkgs/test_core/pubspec.yaml
stereotype441: > Use `^` syntax?

Done!

PR #1046, opened 2019-07-06
Fix deprecations and new lints

kevmoo:


natebosch:
For the future, the commit message would be more useful if it said "Ignore deprecations" instead of "Fix deprecations"...

natebosch:
Do we have an issue filed to stop using those analyzer APIs?

kevmoo:
> For the future, the commit message would be more useful if it said "Ignore deprecations" instead of "Fix deprecations"...

ack

> Do we have an issue filed to stop using those analyzer APIs?

Nope. We'd have to be careful to go through and update the min versions if moving to new APIs

natebosch:
https://github.com/dart-lang/test/issues/1050

Issue #1047, opened 2019-07-08
Update pkg:analyzer dependency in test & test_core


kevmoo:
Support <0.38.0

PR #1048, opened 2019-07-08
Depend on latest analyzer

grouma:
Closes https://github.com/dart-lang/test/issues/1047

### DIFF
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -8,7 +8,7 @@ environment:
   sdk: '>=2.2.0 <3.0.0'
 
 dependencies:
-  analyzer: ^0.37.0
+  analyzer: ">=0.36.0 <0.38.0"
   async: ^2.0.0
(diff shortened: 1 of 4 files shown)


grouma:
Travis webhooks appear to be broken. Corresponding Travis run: https://travis-ci.org/dart-lang/test/builds/555893648

pkgs/test/pubspec.yaml
```diff
@@ -8,7 +8,7 @@ environment:
   sdk: '>=2.2.0 <3.0.0'
 
 dependencies:
-  analyzer: ">=0.26.4 <0.37.0"
+  analyzer: ^0.37.0
```
jakemac53: We should just expand the upper bound if possible, otherwise people will be blocked on upgrading until all transitive deps are updated.

grouma: Done.

grouma:
https://travis-ci.org/dart-lang/test/builds/555910033

PR #1054, opened 2019-07-26
Deprecate internal packages

grouma:
Mark internal packages as deprecated to hopefully prevent accidental use.


pkgs/test_core/pubspec.yaml
```diff
@@ -32,4 +32,4 @@ dependencies:
   # properly constrains all features it provides.
   matcher: ">=0.12.5 <0.12.6"
   # Use an exact version until the test_api package is stable.
-  test_api: 0.2.6
+  test_api: 0.2.7
```
jakemac53: does this also need a dep override? or just going to submit on red?

grouma: This was an oversight.

pkgs/test_api/lib/backend.dart
```diff
@@ -2,6 +2,10 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+@Deprecated('package:test_api is not intended for general use. '
```
natebosch: Do we see this one coming up in practice?

pkgs/test_core/lib/backend.dart
```diff
@@ -2,6 +2,11 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+@Deprecated('package:test_core is not intended for general use. '
+    'Please use package:test.')
+library test_core;
```
natebosch: `test_core.backend`

pkgs/test_api/lib/backend.dart
```diff
@@ -2,6 +2,10 @@
...
+    'Please use package:test.')
+library test_api;
```
natebosch: `test_api.backend`.

I don't know if there is a requirement for unique library names - but we should aim for it.

pkgs/test_api/lib/backend.dart
grouma: Yes. I can't find it but there was a flutter bug recently where folks were importing from test_api and it caused unexpected issues.

pkgs/test_api/lib/backend.dart
grouma: Yeah I wasn't sure if having the same library name would be problematic but it didn't appear to cause issues. Updated it nonetheless.

PR #1057, opened 2019-07-29
Drop dependency on pub_semver

natebosch:
This had been used to detect feature support so that the package could
have a wider SDK constraint and have enhanced features for newer SDKs.
Drop the dependency and code checking the SDK version because:

- We now typically bump the SDK constraint rather than support multiple
  code paths.
- The remaining usage was for checking for DDC support in `pub serve`
  which doesn't exist anymore - we haven't yet deleted the tests or
  migrated them to use `build_runner serve`. This test and references to
  the old approach with both compilers is retained to make it clear what
  should be tested when we get around to migrating it.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Pass `--server-mode` to dart2js instead of `--categories=Server` to fix a
   warning about the flag deprecation.
+* Drop dependency on `pub_semver`.
 
 ## 1.6.5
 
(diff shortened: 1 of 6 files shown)


pkgs/test/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * Pass `--server-mode` to dart2js instead of `--categories=Server` to fix a
   warning about the flag deprecation.
+* Drop dependency on `pub_semver`.
```
kevmoo: Worth a changelog entry? 🤷‍♂

natebosch: Some consumers might care. The direct consumers of `test_core` probably care more, but some might care about the drop from `test` as well.

PR #1058, opened 2019-07-29
Delete extra copy of util/test.dart

natebosch:
This file is unused since there is another copy in `test_api` which is
used instead.


natebosch:
The one in use: https://github.com/dart-lang/test/blob/6b30045fe90aa518683f5ca2756e575376fce037/pkgs/test_api/lib/src/util/test.dart

Issue #1059, opened 2019-08-14
Tests are failing at head due to Utf8Decoder


grouma:
On the latest dev SDK:
`Dart VM version: 2.5.0-dev.2.0 (Sun Aug 11 20:46:20 2019 +0200) on "linux_x64"`

Seeing errors like this:
`Failed to load "test.dart": type 'Utf8Decoder' is not a subtype of type 'StreamTransformer<Uint8List, String>' of 'streamTransformer'`

See this travis run for essentially a no-op change:
https://travis-ci.org/dart-lang/test/builds/571913043?utm_source=github_status&utm_medium=notification

grouma:
cc @tvolkert

tvolkert:
Try either:

```
--- a/pkgs/test/lib/src/runner/node/platform.dart
+++ b/pkgs/test/lib/src/runner/node/platform.dart
@@ -111,8 +111,8 @@ class NodePlatform extends PlatformPlugin
       var socket = await server.first;
       // TODO(nweiz): Remove the DelegatingStreamSink wrapper when sdk#31504 is
       // fixed.
-      var channel = StreamChannel(socket, DelegatingStreamSink(socket))
-          .transform(StreamChannelTransformer.fromCodec(utf8))
+      var channel = StreamChannelTransformer.fromCodec(utf8)
+          .bind(StreamChannel(socket, DelegatingStreamSink(socket)))
           .transform(chunksToLines)
           .transform(jsonDocument)
           .transformStream(StreamTransformer.fromHandlers(handleDone: (sink) {
```

... or:

```
--- a/pkgs/test/lib/src/runner/node/platform.dart
+++ b/pkgs/test/lib/src/runner/node/platform.dart
@@ -111,7 +111,8 @@ class NodePlatform extends PlatformPlugin
       var socket = await server.first;
       // TODO(nweiz): Remove the DelegatingStreamSink wrapper when sdk#31504 is
       // fixed.
-      var channel = StreamChannel(socket, DelegatingStreamSink(socket))
+      var channel = StreamChannel(
+              socket.cast<List<int>>(), DelegatingStreamSink(socket))
           .transform(StreamChannelTransformer.fromCodec(utf8))
           .transform(chunksToLines)
           .transform(jsonDocument)
```

PR #1060, opened 2019-08-14
Fix issue with Utf8Decoder and Node platform

grouma:
This should close https://github.com/dart-lang/test/issues/1059


tvolkert:
I backed out the most invasive stream changes, but we kept a subset of them that hardly affected any packages, and the `Socket implements Stream<Uint8List>` was one of the latter changes, so it stayed.  This call site is one of only ~two places in all of Pub that were affected by that change.

nex3:
When will this be released? It's likely breaking Dart Sass's CI build.

grouma:
I'll try to get to this tomorrow.

PR #1062, opened 2019-08-16
Update test to use analyzer 0.38.0

keertip:
and prep to publish.


pkgs/test/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: test
```
grouma: you'll need to update the test_core dep

keertip: done

grouma:
You'll need to add a temporary dep override, otherwise LGTM.

PR #1076, opened 2019-09-17
test_core: allow the latest pkg:vm_service

kevmoo:
Also fix readme


pkgs/test_core/README.md
```diff
@@ -1,5 +1,5 @@
-## test_core
+A minimal package for writing and running tests also with extensions for
```
jakemac53: nit: this reads weird now, maybe `s/with/as well as`?

kevmoo: ack

PR #1079, opened 2019-09-18
Fixed #1071, catching FilesystemException when input filename is illegal

### LINKED ISSUE #1071: Exception not handled when input test file has illegal filename
Running
```
pub run test 'This is an ***illegal*** filename'
```
will cause OS Error, which is unhandled.

YizhePKU:


pkgs/test_core/lib/src/executable.dart
```diff
@@ -156,6 +156,9 @@ Future<void> _execute(List<String> args) async {
   } on FormatException catch (error) {
     stderr.writeln(error.message);
     exitCode = exit_codes.data;
+  } on FileSystemException {
```
jakemac53: This will log more nicely (gives you the path and underlying OS error information)

```suggestion
  } on FileSystemException catch (error) {
    var loadException = LoadException(error.path, error.osError);
```

pkgs/test_core/lib/src/executable.dart
```diff
@@ -156,6 +156,9 @@ Future<void> _execute(List<String> args) async {
...
+    stderr.writeln("Illegal filename");
```
jakemac53: And then log it like so:

```suggestion
    stderr.writeln(loadException.toString(color: configuration.color));
```

jakemac53:
Ah, it looks like if you move this try/catch inside of the `_loadSuites` method of `test_core/lib/src/runner.dart` (line 236) then things will work as expected.

You would want to construct a `LoadSuite.fromLoadException` and return an empty stream with that in the catch.

Then it will still run the other tests, log a nice error, and exit normally.

YizhePKU:
> Ah, it looks like if you move this try/catch inside of the `_loadSuites` method of `test_core/lib/src/runner.dart` (line 236) then things will work as expected.

Thanks! I was looking for that for quite a long time. I'll see if it works.

YizhePKU:
No, it doesn't work. I don't think _loadSuite() is even called when the exception occurred.

YizhePKU:
The change of logging was reverted because
1. LoadException is from runner.dart, which isn't where the exception was raised.
2. There is no special logging for other exceptions in executeable.dart either, e.g. ApplicationError(see line 150).

YizhePKU:
It seems that the test runner hangs because a StreamController failed to close(the future never completes), at engine.dart, line 540. Don't know why though.

jakemac53:
> * LoadException is from runner.dart, which isn't where the exception was raised.

Ah ya I forgot to also add a suggestion for the import - it should be fine to add that here.

> 2\. There is no special logging for other exceptions in executeable.dart either, e.g. ApplicationError(see line 150).

A lot of the other ones do have some sort of formatting - but I do think this is the wrong place to handle the error anyways (it is to high level).

> No, it doesn't work. I don't think _loadSuite() is even called when the exception occurred.

I was able to get this to work when I tried testing these changes out locally - the first exception I would see is here https://github.com/dart-lang/test/blob/master/pkgs/test_core/lib/src/runner.dart#L236. The whole body of this anonymous function needs to be wrapped in the try/catch so it is done per suite (which allows loading/running of other tests, only the bad one will fail).

> It seems that the test runner hangs because a StreamController failed to close(the future never completes), at engine.dart, line 540. Don't know why though.

I think this is likely because `_loadSuites` never adds a suite for this test because of the exception. It expects at least a `LoadSuite.forLoadException` it looks like (see [here](https://github.com/dart-lang/test/blob/master/pkgs/test_core/lib/src/runner.dart#L242))

jakemac53:
@YizhePKU did you want to follow up here or should we close the PR?

PR #1088, opened 2019-09-26
Implement VM coverage gathering

willdrach-wk:
Open to feedback on this. Here's how it works at the moment:

There's a new `--coverage` option that you can use to specify the output directory for your coverage. It will create a new output directory if one doesn't already exist.

The tests are run through the engine, and then get spit out to a coverage helper function. That coverage helper function:
**A)** Checks if it's a VM suite
**B)** Gathers coverage if it is
**C)** Outputs coverage to `${suite path}.vm.json`

So, for example, when I run:
```bash
pub run test --coverage hello_world test/vm/simple_repo_test.dart
```

A new file gets added at `hello_world/test/vm/simple_repo_test.dart.vm.json` with the coverage results!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,14 +18,9 @@ jobs:
       env: PKGS="pkgs/test pkgs/test_api pkgs/test_core"
       script: ./tool/travis.sh dartfmt dartanalyzer_0
     - stage: analyze_and_format
-      name: "SDK: 2.2.0; PKG: pkgs/test; TASKS: `dartanalyzer --fatal-warnings .`"
+      name: "SDK: 2.2.0; PKG: pkgs/test, pkgs/test_api, pkgs/test_core; TASKS: `dartanalyzer --fatal-warnings .`"
       dart: "2.2.0"
-      env: PKGS="pkgs/test"
-      script: ./tool/travis.sh dartanalyzer_1
-    - stage: analyze_and_format
-      name: "SDK: 2.1.0; PKGS: pkgs/test_api, pkgs/test_core; TASKS: `dartanalyzer --fatal-warnings .`"
-      dart: "2.1.0"
-      env: PKGS="pkgs/test_api pkgs/test_core"
+      env: PKGS="pkgs/test pkgs/test_api pkgs/test_core"
       script: ./tool/travis.sh dartanalyzer_1
     - stage: unit_test
       name: "SDK: dev; PKG: pkgs/test; TASKS: `xvfb-run -s \"-screen 0 1024x768x24\" pub run test --preset travis --total-shards 5 --shard-index 0`"
--- a/pkgs/test/test/runner/coverage_test.dart
+++ b/pkgs/test/test/runner/coverage_test.dart
@@ -2,10 +2,11 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+@TestOn("vm")
+
 import 'dart:convert';
 import 'dart:io';
 
-@TestOn("vm")
 import 'package:test_descriptor/test_descriptor.dart' as d;
 
 import 'package:path/path.dart' as p;
--- a/pkgs/test/test/runner/coverage_test.dart
+++ b/pkgs/test/test/runner/coverage_test.dart
@@ -1,0 +1,49 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+@TestOn("vm")
+import 'package:test_descriptor/test_descriptor.dart' as d;
+
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+
+import '../io.dart';
+
+void main() {
+  group("with the --coverage flag,", () {
+    test("gathers coverage for VM tests", () async {
+      await d.file("test.dart", """
+        import 'package:test/test.dart';
+
+        void main() {
+          test("test 1", () {
+            expect(true, isTrue);
+          });
+        }
+      """).create();
+
+      final coverageDirectory =
+          Directory(p.join(Directory.current.path, "test_coverage"));
+      expect(await coverageDirectory.exists(), isFalse,
+          reason:
+              'Coverage directory exists, cannot safely run coverage tests. Delete the ${coverageDirectory.path} directory to fix.');
+
+      var test =
+          await runTest(["--coverage", coverageDirectory.path, "test.dart"]);
+      expect(test.stdout, emitsThrough(contains("+1: All tests passed!")));
+      await test.shouldExit(0);
+
+      final coverageFile =
+          File(p.join(coverageDirectory.path, "test.dart.vm.json"));
+      final coverage = await coverageFile.readAsString();
+      final jsonCoverage = json.decode(coverage);
+      expect(jsonCoverage['coverage'], isNotEmpty);
+
+      await coverageDirectory.delete(recursive: true);
+    });
+  });
+}
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -95,8 +95,7 @@ Usage: pub run test [files or directories...]
                                   Currently only supported for browser tests.
 
     --debug                       Runs the VM and Chrome tests in debug mode.
-    --coverage=<directory>        Enables coverage gathering for the Dart VM and outputs an
-                                  lcov formatted file to the specified directory.
+    --coverage=<directory>        Gathers coverage and outputs it to the specified directory.
                                   Implies --debug.
 
     --[no-]chain-stack-traces     Chained stack traces to provide greater exception details
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -95,6 +95,10 @@ Usage: pub run test [files or directories...]
                                   Currently only supported for browser tests.
 
     --debug                       Runs the VM and Chrome tests in debug mode.
+    --coverage=<directory>        Enables coverage gathering for the Dart VM and outputs an
+                                  lcov formatted file to the specified directory.
+                                  Implies --debug.
+
     --[no-]chain-stack-traces     Chained stack traces to provide greater exception details
                                   especially for asynchronous code. It may be useful to disable
                                   to provide improved test performance but at the cost of
--- a/pkgs/test_core/lib/src/runner.dart
+++ b/pkgs/test_core/lib/src/runner.dart
@@ -71,7 +71,8 @@ class Runner {
 
   /// Creates a new runner based on [configuration].
   factory Runner(Configuration config) => config.asCurrent(() {
-        var engine = Engine(concurrency: config.concurrency, coverage: config.coverage);
+        var engine =
+            Engine(concurrency: config.concurrency, coverage: config.coverage);
 
         var reporterDetails = allReporters[config.reporter];
         return Runner._(engine, reporterDetails.factory(config, engine));
@@ -107,7 +108,8 @@ class Runner {
         }
 
         if (_config.coverage != null) {
-          var directory = await Directory(_config.coverage).create(recursive: true);
+          var directory =
+              await Directory(_config.coverage).create(recursive: true);
           print('Gathering coverage and outputting to: ${directory.path}');
         }
 
--- a/pkgs/test_core/lib/src/runner.dart
+++ b/pkgs/test_core/lib/src/runner.dart
@@ -108,9 +108,7 @@ class Runner {
         }
 
         if (_config.coverage != null) {
-          var directory =
-              await Directory(_config.coverage).create(recursive: true);
-          print('Gathering coverage and outputting to: ${directory.path}');
+          await Directory(_config.coverage).create(recursive: true);
         }
 
         bool success;
--- a/pkgs/test_core/lib/src/runner.dart
+++ b/pkgs/test_core/lib/src/runner.dart
@@ -71,7 +71,7 @@ class Runner {
 
   /// Creates a new runner based on [configuration].
   factory Runner(Configuration config) => config.asCurrent(() {
-        var engine = Engine(concurrency: config.concurrency);
+        var engine = Engine(concurrency: config.concurrency, coverage: config.coverage);
 
         var reporterDetails = allReporters[config.reporter];
         return Runner._(engine, reporterDetails.factory(config, engine));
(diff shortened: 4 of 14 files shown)


willdrach-wk:
FYI @grouma @jakemac53

I can add/fix tests depending on whether or not you think this approach works, but want to verify that first.

grouma:
This general approach seems good to me. A couple quick questions.

> There's a new --coverage option that you can use to specify the output directory for your coverage. It will create a new output directory if one doesn't already exist.

What happens if the directory already exists and is non-empty? I assume we bash over existing coverage results.

> A new file gets added at hello_world/test/vm/simple_repo_test.dart.lcov.info with the lcov results!

I'm confused by your description, is the output in LCOV format or not? If so, why the `.info` extension?

To that point, I know that this supports the VM only for now but what about the future when we support other platforms. Should we indicate where the coverage information came from in the file name, e.g. `hello_world/test/vm/simple_repo_test.dart.vm.lcov`

willdrach-wk:
> What happens if the directory already exists and is non-empty? I assume we bash over existing coverage results.

This assumption is correct. We could add a confirmation prompt of some sort depending on how valuable you think that would be.

> I'm confused by your description, is the output in LCOV format or not? If so, why the .info extension?

Sorry, I've edited this thing like 5 times already. It's outputted in JSON at this point. [Also, I believe LCOV formatted files usually have a `.info` extension.](http://ltp.sourceforge.net/coverage/lcov.php). However, that's irrelevant because they aren't formatted that way any more.

willdrach-wk:
After looking into this more beyond my simple use case, we would need to solve dart-lang/coverage#269. Gathering coverage for one suite run gathers coverage for all suite runs, because although they're run on different isolates `coverage` gets source reports for all of them.

I can make a PR for that pretty quick though, and then we can take another look at this.

cc: @evanweible-wf

grouma:
> I can make a PR for that pretty quick though, and then we can take another look at this.

Sounds good. Assign me / cc to me the PR and I'll review / publish.

.travis.yml
```diff
@@ -18,14 +18,9 @@ jobs:
...
     - stage: analyze_and_format
-      name: "SDK: 2.2.0; PKG: pkgs/test; TASKS: `dartanalyzer --fatal-warnings .`"
+      name: "SDK: 2.2.0; PKG: pkgs/test, pkgs/test_api, pkgs/test_core; TASKS: `dartanalyzer --fatal-warnings .`"
       dart: "2.2.0"
-      env: PKGS="pkgs/test"
-      script: ./tool/travis.sh dartanalyzer_1
-    - stage: analyze_and_format
-      name: "SDK: 2.1.0; PKGS: pkgs/test_api, pkgs/test_core; TASKS: `dartanalyzer --fatal-warnings .`"
-      dart: "2.1.0"
-      env: PKGS="pkgs/test_api pkgs/test_core"
```
willdrach-wk: The range for the `coverage` package is `2.2.0`, and the `test` package is already bumped to `2.2.0` so I bumped the other two as well.

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -66,6 +69,9 @@ class Engine {
   /// `false` if the tests finished running before close was called.
   bool _closedBeforeDone;
 
+  /// The coverage output directory
```
natebosch: [nit] Format as a sentence

```suggestion
  /// The coverage output directory.
```

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -293,6 +302,29 @@ class Engine {
...
+  Future<Null> _gatherCoverage(LiveSuiteController controller) async {
+    if (_coverage == null) return;
+
+    final RunnerSuite suite = controller.liveSuite.suite;
+
+    if (!suite.platform.runtime.isDartVM) return;
+
+    // for some reason the observatoryUrl URI doesn't have the query parameters populated
+    final String isolateId = Uri.splitQueryString(
+        suite.environment.observatoryUrl.toString().split('?')[1])['isolateId'];
```
natebosch: I'd prefer to let the `Uri` class handle the split on `?`. Does it work to use the `query` field on the existing URI?

```suggestion
        suite.environment.observatoryUrl.query)['isolateId'];
```

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -293,6 +302,29 @@ class Engine {
...
+
+    final Map<String, dynamic> cov = await collect(
+        suite.environment.observatoryUrl, false, false, false, Set(),
```
natebosch: [nit] Dart `2.2.0` should support Set literals.

```suggestion
        suite.environment.observatoryUrl, false, false, false, {},
```

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -293,6 +302,29 @@ class Engine {
...
+        isolateIds: Set.from([isolateId]));
```
natebosch: How about 
```suggestion
        isolateIds: {isolateId});
```

pkgs/test_core/lib/src/runner/engine.dart
willdrach-wk: I can double check. I'm not 100% sure why but I do remember the query and queryParameters members of the class being missing in the URI.

willdrach-wk:
I'll fix those few things this afternoon, thanks for the first pass review @natebosch! I'd definitely be ok with waiting for Gary on this one as well.

pkgs/test_core/lib/src/runner/engine.dart
willdrach-wk: Yeah, the query is unpopulated in this case. I'll take some time this afternoon and try and figure out why, I 100% recognize that this is a little hacky as-is.

pkgs/test_core/lib/src/runner/engine.dart
willdrach-wk: Done!

pkgs/test_core/lib/src/runner/engine.dart
willdrach-wk: Done!

pkgs/test_core/lib/src/runner/engine.dart
willdrach-wk: Done!

pkgs/test_core/lib/src/runner/engine.dart
willdrach-wk: Ah, it's because the URL is fragmented:
`http://127.0.0.1:54637/wJV71-d7700=/#/inspect?isolateId=isolates%2F980739093&objectId=libraries%2F%4022019165`

The URI parser expects the query to come before that fragment identifier (`/#/`).

@natebosch do you have a preferred way for dealing with this? I could just do a more null-safe version of this split, or we could add something to the `Environment` class, but that might be overkill.

pkgs/test_core/lib/src/runner/engine.dart
willdrach-wk: This one failed analysis, I guess the empty Set literally wasn't quite perfect at `2.2.0`. Put it back.

pkgs/test_core/lib/src/runner/engine.dart
natebosch: Oh weird. It's unfortunate that the observatory does stuff so strangely.

How about something like:

```dart
    final String isolateId =
        Uri.parse(suite.environment.observatoryUrl.fragment)
            .queryParameters['isolateId'];
```

This way we're making it explicit that we're dealing with the fragment part and treating it as if it's a URI itself. We won't end up with any trouble if there is a separate `?` in the URI.

I'm not sure if this makes us more brittle if the observatory changes it's URI patterns though.

willdrach-wk: Any big changes like that would probably go over a major version update anyways, right? Like you said, it still has the potential to be fragile but I definitely agree that this is less fragile than the alternative. I also think this suggestion is much more clear about the intentions, like you said. I'll update it!

willdrach-wk: Updated!

pkgs/test_core/lib/src/runner.dart
```diff
@@ -106,6 +107,12 @@ class Runner {
               '`DART_VM_OPTIONS=-DSILENT_OBSERVATORY=true`.');
         }
 
+        if (_config.coverage != null) {
+          var directory =
+              await Directory(_config.coverage).create(recursive: true);
+          print('Gathering coverage and outputting to: ${directory.path}');
```
evanweible-wf: [nit] I'm guessing this was just here for debugging and should be removed before merge?

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -293,6 +302,30 @@ class Engine {
...
   }
 
+  Future<Null> _gatherCoverage(LiveSuiteController controller) async {
+    if (_coverage == null) return;
+
+    final RunnerSuite suite = controller.liveSuite.suite;
+
+    if (!suite.platform.runtime.isDartVM) return;
+
+    // for some reason the observatoryUrl URI doesn't have the query parameters populated
```
evanweible-wf: I think this comment is outdated now that we've figured out it comes from the fragment, right?

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -293,6 +302,30 @@ class Engine {
...
+    final String isolateId =
+        Uri.parse(suite.environment.observatoryUrl.fragment)
+            .queryParameters['isolateId'];
+
+    final Map<String, dynamic> cov = await collect(
```
evanweible-wf: `collect()` is already typed as `Future<Map<String, dynamic>>` - can we omit the type on the LHS here?

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -293,6 +302,30 @@ class Engine {
...
+        suite.environment.observatoryUrl, false, false, false, Set(),
+        isolateIds: {isolateId});
+
+    final outfile = File('$_coverage/${suite.path}.vm.json')
+      ..createSync(recursive: true);
```
evanweible-wf: [nit] The coverage directory should already be created at this point (https://github.com/dart-lang/test/pull/1088/files#diff-01e4e27436330f7aaa3e4b86002db314R112) and writing to a file will create it if it doesn't already exist, so you should be able to remove this `createSync` call.

willdrach-wk: The lowest level directory exists, but if I'm outputting a deep file such as `coverage/test/vm/test.dart.vm.json` those other directories don't exist yet (potentially)

pkgs/test_core/lib/src/runner.dart
willdrach-wk: I think I'd want some logging to confirm that coverage is being gathered but that might just be me

pkgs/test_core/lib/src/runner/engine.dart
evanweible-wf: Ah good point, sorry I missed that.

pkgs/test_core/lib/src/runner.dart
evanweible-wf: In that case, I'm guessing there is a more robust way to log that doesn't break reporter contracts like the JSON reporter.

pkgs/test/test/runner/runner_test.dart
```diff
@@ -95,6 +95,10 @@ Usage: pub run test [files or directories...]
                                   Currently only supported for browser tests.
 
     --debug                       Runs the VM and Chrome tests in debug mode.
+    --coverage=<directory>        Enables coverage gathering for the Dart VM and outputs an
+                                  lcov formatted file to the specified directory.
```
grouma: What are the restrictions here? Can the coverage directory be an absolute path?

willdrach-wk: Yes, the coverage directory can be an absolute or relative path. I think I'm going to make this help description as simple as possible and explain how this works more in-depth in the README.

pkgs/test_core/lib/src/runner.dart
willdrach-wk: Fixed!

pkgs/test_core/lib/src/runner/engine.dart
willdrach-wk: Fixed!

pkgs/test_core/lib/src/runner/engine.dart
willdrach-wk: Fixed

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -293,6 +302,30 @@ class Engine {
...
+    // for some reason the observatoryUrl URI doesn't have the query parameters populated
+    final String isolateId =
+        Uri.parse(suite.environment.observatoryUrl.fragment)
+            .queryParameters['isolateId'];
+
+    final Map<String, dynamic> cov = await collect(
+        suite.environment.observatoryUrl, false, false, false, Set(),
+        isolateIds: {isolateId});
+
+    final outfile = File('$_coverage/${suite.path}.vm.json')
```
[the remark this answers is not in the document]
willdrach-wk: Fixed!

pkgs/test/test/runner/runner_test.dart
willdrach-wk: @grouma I updated this argument to be super simple, but added a new section to the `README` to describe a bit more how coverage works and what the limitations are, can you give this another look and let me know what you think?

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -4,9 +4,13 @@
 
 import 'dart:async';
 import 'dart:collection';
+import 'dart:convert';
+import 'dart:io';
 
 import 'package:async/async.dart' hide Result;
 import 'package:collection/collection.dart';
+import 'package:coverage/coverage.dart';
+import 'package:path/path.dart';
```
grouma: [nit] Almost all of our packages will import this with an alias: `p`

willdrach-wk: I'll fix that when I fix tests!

willdrach-wk: Fixed!

willdrach-wk:
Heyo @grouma, added a test. Sorry it took a bit, I wanted to add a legitimate test instead of just a "it ran without failing" kinda test. Let me know if you want any more changes!

grouma:
LGTM. Let's add a changelog entry to the various packages and a rev the pubspec version to prep for publish.

willdrach-wk:
@grouma done!

pkgs/test/test/runner/coverage_test.dart
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+@TestOn("vm")
```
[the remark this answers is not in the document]
grouma: Fix this minor thing and I'll merge.

Issue #1090, opened 2019-10-09
Tests are failing to compile to dart2js/DDC in 1.9.0 due to coverage additions


greglittlefield-wf:
Tests are failing to compile since it seems that there's a transitive dependency on `dart:io` when importing `package:test/test.dart`.
```
$ pbr test -- -p chrome test/foo_test.dart                              
[INFO] Generating build script completed, took 298ms
[WARNING] Throwing away cached asset graph due to Dart SDK update.
[INFO] Cleaning up outputs from previous builds. completed, took 108ms
[INFO] Generating build script completed, took 65ms
[INFO] Creating build script snapshot... completed, took 11.2s
[INFO] Building new asset graph completed, took 961ms
[INFO] Checking for unexpected pre-existing outputs. completed, took 2ms
[WARNING] build_web_compilers:entrypoint on test/foo_test.dart.browser_test.dart:
Skipping compiling test_package|test/foo_test.dart.browser_test.dart with ddc because some of its
transitive libraries have sdk dependencies that not supported on this platform:

test_core|lib/src/runner/engine.dart
coverage|lib/src/run_and_collect.dart
coverage|lib/src/resolver.dart
coverage|lib/src/hitmap.dart
coverage|lib/src/collect.dart
coverage|lib/src/util.dart

https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-skipped-compiling-warnings

[INFO] Running build completed, took 3.8s
[INFO] Caching finalized dependency graph completed, took 112ms
[INFO] Creating merged output dir `/var/folders/lr/jt65cx4d21gdjslx_q0jh9940000gn/T/build_runner_testCknCyp/` completed, took 258ms
[INFO] Writing asset manifest completed, took 3ms
[INFO] Succeeded after 4.2s with 1215 outputs (4037 actions)
Running tests...

00:00 +0 -1: compiling test/foo_test.dart [E]                                                                                                                                                                                                                                          
  Failed to load "test/foo_test.dart": Failed to load script at http://localhost:50726/jirbScLqo%2B2vpXnZk426tdcPCXPBPzho/test/foo_test.dart.browser_test.dart.js.
00:00 +0 -1: Some tests failed.  
```

[...]

PR #1091, opened 2019-10-09
Conditionally import coverage logic

grouma:
Closes https://github.com/dart-lang/test/issues/1090


pkgs/test_core/lib/src/runner/coverage.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'dart:io';
+
+import 'package:coverage/coverage.dart';
+import 'package:path/path.dart' as p;
+
+import 'live_suite_controller.dart';
+import 'runner_suite.dart';
+
+/// Collects coverage and outputs to the [coverage] path.
+Future<Null> gatherCoverage(
```
natebosch: `Future<void>`

PR #1095, opened 2019-10-14
Mark non-solo tests as "skipped"

grouma:
While using solo, mark non-solo tests as "skipped". This improves the reporting of these skipped tests.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.9.2
+
+* Depend on the latest `package:test_api` and `package:test_core`.
+
 ## 1.9.1
 
 * Depend on latest `test_core`.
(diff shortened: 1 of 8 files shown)


pkgs/test/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.9.2
+
+* Depend on the latest `package:test_api` and `package:test_core`.
```
natebosch: We might consider expanding this, since it's the one most users will see.

pkgs/test/test/runner/solo_test.dart
```diff
@@ -2,11 +2,10 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:test/test.dart';
```
natebosch: Why move this? I'm guessing it broke the `@TestOn` annotation behavior for this file.

grouma: Unintentional. VS Code's auto sort feature. I'll need to clean this up.

PR #1099, opened 2019-10-21
Use latest matcher

natebosch:


natebosch:
Ugh, I didn't realize that the error matchers are in `package:matcher`... the throws matchers are here...

I'm going to need to publish `0.12.7` and upgrade straight to that here since I want to implement a `throwsAssertionError`

natebosch:
Skipping the assertion error matchers for now.

Issue #1100, opened 2019-10-22
Add support for writing a reporter to a file


natebosch:
See comment on https://github.com/dart-lang/build/pull/2508#pullrequestreview-304885406

natebosch:
Closed by #1129

PR #1101, opened 2019-10-22
Allow injecting an IOSink into the reporters

natebosch:
Towards #1100 

Prepare for a world where we a reporter can be configured to write directly to a file.


evanweible-wf:
Nice!

Maybe `--reporter` could be turned into a multi-option, and each reporter value could support an optional filename, e.g.:

```
pub run test --reporter compact --reporter json:results.json
```

That would be similar to how you explicitly provide port mappings for `webdev serve`.. but maybe it's not explicit enough to make it obvious what's happening. It also makes me wonder what the behavior would be if you only specify one reporter and it's a file reporter.. do you still enable the default compact reporter? Or do you replace that with the file reporter and require that someone run `--reporter compact --reporter json:results.json` to get both?

You also could add a separate option like the `--file-reporter` one you mentioned, even though under the hood it would just be adding another reporter to `MultiplexReporter`.

natebosch:
> but maybe it's not explicit enough to make it obvious what's happening. It also makes me wonder what the behavior would be if you only specify one reporter and it's a file reporter

Yeah overall I do worry that the arguments will be confusing. We could consider making the argument very narrow and explicit. It would only solve the specific use case we know exists, but that might be enough... `--json-output some_file.json`. We wouldn't really need the capability for _any_ reporter to write to a file in that case (it's not super sensible for the compact reporter to write to a file anyway).

@grouma - what do you think?

If we can't think of a use case for a reporter other than the json one it might be the simplest way to handle it.

evanweible-wf:
Fwiw, the reason we're interested in writing the json reporter output to a file is so that we can convert it to xunit, but assuming there's still a long term plan to support pluggable reporters (https://github.com/dart-lang/test/issues/12), then it would eventually be useful to be able to specify a reporter by name that should write to file.

grouma:
> Yeah overall I do worry that the arguments will be confusing. We could consider making the argument very narrow and explicit. It would only solve the specific use case we know exists, but that might be enough... `--json-output some_file.json`. We wouldn't really need the capability for _any_ reporter to write to a file in that case (it's not super sensible for the compact reporter to write to a file anyway).
> 
> @grouma - what do you think?

I agree that we should keep the flag simple like you suggest. It we support multiple reporters with multiple output files then we have to do some validation to ensure that the reporters do not conflict. Also like you said, it doesn't make sense for some reporters to write to a file.

As for the approach, LGTM.

natebosch:
Updated to `IOSink`. I think we can land this as prep since it doesn't yet impact the arg parsing.

pkgs/test_core/lib/src/runner/reporter/expanded.dart
```diff
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io';
```
natebosch: @jakemac53 - this import is a problem. I think we use the expanded reporter by default on the web when you invoke a `_test.dart` file directly rather than go through the runner.

I think I may need to go back to `Sink<List<int>>`. Any concerns?

natebosch: On second though, that will cause a problem here:

https://github.com/dart-lang/test/blob/5a3d81a1235a8f515064e3514f98f5aa7c3e5721/pkgs/test_core/lib/test_core.dart#L62-L63

I think it might not work to add an import to `dart:io` there?

PR #1106, opened 2019-10-31
Add test randomization

kathrinschuler:
Add a --test-randomize-ordering-seed flag to run tests in random order

### DIFF
--- a/pkgs/test/test/runner/configuration/randomize_order_test.dart
+++ b/pkgs/test/test/runner/configuration/randomize_order_test.dart
@@ -64,7 +64,16 @@ void main() {
     await test.shouldExit(0);
 
     // Shuffle when passed random
-    // TODO(kathrins): How can to mock the behaviour of random?
+    test =
+        await runTest(["test.dart", "--test-randomize-ordering-seed=random"]);
+    expect(
+        test.stdout,
+        emitsInAnyOrder([
+          contains("Shuffling test order with --test-randomize-ordering-seed"),
+          isNot(contains(
+              "Shuffling test order with --test-randomize-ordering-seed=0"))
+        ]));
+    await test.shouldExit(0);
   });
 
   test("shuffles each suite with the same seed", () async {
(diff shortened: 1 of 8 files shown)


pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -105,6 +106,10 @@ final ArgParser _parser = (() {
       help: "Don't re-run tests that have retry set.",
       defaultsTo: false,
       negatable: false);
+  parser.addOption("test-randomize-ordering-seed",
```
grouma: Maybe make mention that we only accept a range of values.

pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -204,6 +209,17 @@ class _Parser {
       }
     }
 
+    var testRandomizeOrderingSeed =
+        _parseOption('test-randomize-ordering-seed', (value) {
+      var seed = value == 'random'
+          ? Random().nextInt(4294967295)
+          : int.parse(value).toUnsigned(32);
+      if (seed != null && seed > 0) {
+        print("Shuffling test order with --test-randomize-ordering-seed=$seed");
```
kathrinschuler: @grouma I am printing the seed here. I can use some other logging mechanism than `print` if you want.

grouma: I missed this print. 

[nit] Use single quotes for consistency.

I think printing here is fine but @natebosch may have thoughts on if this should use the reporter API in some fashion.

pkgs/test_core/lib/src/runner/configuration/args.dart
kathrinschuler: Adding "(must be a 32bit unsigned integer)"

pkgs/test/test/runner/configuration/randomize_order_test.dart
```diff
@@ -0,0 +1,148 @@
...
+          "+0: test 1",
+          "+1: test 2",
+          "+2: test 3",
+          "+3: test 4",
+          "+4: All tests passed!"
+        ]));
+    await test.shouldExit(0);
+
+    // Shuffle when passed random
+    // TODO(kathrins): How can to mock the behaviour of random?
```
grouma: It's going to be a little difficult to mock the random because the internal logic isn't exposed cleanly. I'd just confirm that using `random` results in logging a non-zero seed 32 uint seed.

pkgs/test_core/lib/src/runner/configuration/args.dart
natebosch: I don't think we can use the `print` when a client is using the json reporter.

Why do we need to print this? I would think the logging outside of the test package would remember the seed.

grouma: Discussed offline. We can add it to the report when someone requests it but for now what we have is fine. As for why it is required, someone can provide `random` and therefore we need to log the seed that was used.

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -320,7 +322,13 @@ class Engine {
       }
 
       if (!_closed && setUpAllSucceeded) {
-        for (var entry in group.entries) {
+        // shuffle the group entries
+        var entries = List<GroupEntry>.from(group.entries);
```
natebosch: [nit] I don't think we want to use the `.from` constructor. This can be any of `group.entries.toList()`, `List.of(group.entries)`, or `[...group.entries]`. The `group.entries.toList()` is probably the most idiomatic choice.

kathrinschuler:
Thank you both for the quick review!

PR #1107, opened 2019-11-07
Extend synthetic test timeout

grouma:
Closes https://github.com/flutter/flutter/issues/28840#issuecomment-469458482

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.9.4
+
+* Depend on the latest `package:test_core`.
+* Depend on the latest `package:test_api`.
+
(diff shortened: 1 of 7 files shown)


pkgs/test/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.9.4
+
+* Depend on the latest `package:test_core`.
```
natebosch: I think we should copy in the relevant change. This is the changelog users will see first, we shouldn't make them jump through extra hoops to figure out what "latest" means, then find the changelog for that version number in 2 other packages to see what changed.

We've done the same with `package:matcher` rolls in the past - we copy/paste the interesting changelog bits here.

Issue #39, opened 2015-03-23
Support a default HTML wrapper


travissanderson-wf:
For our use case we want to be able to provide a default HTML file instead of one for every test file.  See have this PR into test_runner.dart google/test_runner.dart#35 for how we are accomplishing that there.

natebosch:
We have default HTML files which are automatically provided when the test does not have it's own.

travissanderson-wf:
The purpose of this request was for us to be able to provide a customized default HTML file besides the auto-generated one that the runner creates.  Is that capability available now?

natebosch:
Ah I see, no that isn't an option today. Will leave this open for consideration.

travissanderson-wf:
👍

natebosch:
I think what this would take is adding some support for configuration in `dart_test.yaml` to point to a template to use instead of the default one provided by the test runner. We won't have bandwidth to implement this any time soon but we'd be happy to review a pull request adding this.

wytrych:
@natebosch are you still open for a PR on this feature? I'd be happy to look into this.

jakemac53:
Another option here would be to do this with a builder - it would be pretty trivial but would require the use of build_runner. However that is already a dependency for almost any web project.

That way the `test` package doesn't have to own this logic (or add the deps required to support it), and it would be easier for people to iterate? It would also allow for multiple templating languages etc.

wytrych:
Using `build_runner` is doing something very different though, it tries to build the whole package first, not only pulling in the necessary dependencies. In a big project that is a big difference.

jakemac53:
> Using `build_runner` is doing something very different though, it tries to build the whole package first, not only pulling in the necessary dependencies. In a big project that is a big difference.

You can't run the tests (in debug mode at least) until the whole package is built anyways - so I think for most users this wouldn't matter.

Note that it does also now have the ability to only build a specific app or glob of files via the `--build-filter` option.

wytrych:
Hmm when I run a single test it only builds the files that are actually imported from that test (and their dependency graph ofc)

jakemac53:
If you run `pub run test -p chrome` then yes it does that - but it compiles with dart2js which isn't easy to debug. It isn't really something we recommend people to do.

Also, as I mentioned you can now do the same with `build_runner` (only compile a single test). We need better documentation around this to be sure but its possible to do.

jakemac53:
[...]

PR #1127, opened 2019-12-27
Add custom-html-template-path config option

wytrych:
This allows for reusing one template file across all tests in use cases
where external scripts or html elements are required for all tests. The
possibility to still use local html files per test file is retained.

Fixes https://github.com/dart-lang/test/issues/39

### DIFF
--- a/pkgs/test/README.md
+++ b/pkgs/test/README.md
@@ -528,7 +528,8 @@ If you want to share the same HTML file across all tests, you can provide a
 This file should follow the rules above, except that instead of the link tag
 add exactly one `{testScript}` in the place where you want the template processor to insert it.
 
-The template can't end with `.html` since the test loader will not able to process it and an error will be thrown.
+The template can't be named like any test file, as that would clash with using the
+custom HTML mechanics. In such a case, an error will be thrown.
 
 For example:
 
--- a/pkgs/test/README.md
+++ b/pkgs/test/README.md
@@ -526,7 +526,9 @@ the following HTML file:
 If you want to share the same HTML file across all tests, you can provide a
 `custom-html-template-path` configuration option to your configuration file.
 This file should follow the rules above, except that instead of the link tag
-add exactly one `{testScript}` in the place where you want the template processor to insert it.
+add exactly one `{{testScript}}` in the place where you want the template processor to insert it.
+
+You can also optionally use any number of `{{testName}}` placeholders which will be replaced by the test filename.
 
 The template can't be named like any test file, as that would clash with using the
 custom HTML mechanics. In such a case, an error will be thrown.
@@ -542,8 +544,8 @@ custom-html-template-path: html_template.html.tpl
 <!-- html_template.html.tpl -->
 <html>
   <head>
-    <title>Custom HTML Template</title>
-    {testScript}
+    <title>{{testName}} Test</title>
+    {{testScript}}
     <script src="packages/test/dart.js"></script>
   </head>
   <body>
--- a/pkgs/test/README.md
+++ b/pkgs/test/README.md
@@ -10,6 +10,7 @@
 * [Asynchronous Tests](#asynchronous-tests)
   * [Stream Matchers](#stream-matchers)
 * [Running Tests With Custom HTML](#running-tests-with-custom-html)
+  * [Providing a custom HTML template](#providing-a-custom-html-template)
 * [Configuring Tests](#configuring-tests)
   * [Skipping Tests](#skipping-tests)
   * [Timeouts](#timeouts)
@@ -493,7 +494,9 @@ By default, the test runner will generate its own empty HTML file for browser
 tests. However, tests that need custom HTML can create their own files. These
 files have three requirements:
 
-* They must have the same name as the test, with `.dart` replaced by `.html`.
+* They must have the same name as the test, with `.dart` replaced by `.html`. You can also
+  provide a configuration path to an html file if you want it to be reused across all tests.
+  See [Providing a custom HTML template](#providing-a-custom-html-template) below.
 
 * They must contain a `link` tag with `rel="x-dart-test"` and an `href`
   attribute pointing to the test script.
@@ -517,6 +520,37 @@ the following HTML file:
   </body>
 </html>
 ```
+
+### Providing a custom HTML template
+
+If you want to share the same HTML file across all tests, you can provide a
+`custom-html-template-path` configuration option to your configuration file.
+This file should follow the rules above, except that instead of the link tag
+add exactly one `{testScript}` in the place where you want the template processor to insert it.
+
+The template can't end with `.html` since the test loader will not able to process it and an error will be thrown.
+
+For example:
+
+```yaml
+custom-html-template-path: html_template.html.tpl
+```
+
+```html
+<!doctype html>
+<!-- html_template.html.tpl -->
+<html>
+  <head>
+    <title>Custom HTML Template</title>
+    {testScript}
+    <script src="packages/test/dart.js"></script>
+  </head>
+  <body>
+    // ...
+  </body>
+</html>
+```
+
 
 ## Configuring Tests
 
--- a/pkgs/test/lib/src/runner/browser/platform.dart
+++ b/pkgs/test/lib/src/runner/browser/platform.dart
@@ -191,7 +191,7 @@ class BrowserPlatform extends PlatformPlugin
       var template = _config.customHtmlTemplatePath ?? _defaultTemplatePath;
       var contents = File(template).readAsStringSync();
       var processedContents = contents
-          .replaceFirst('{testScript}', link)
+          .replaceFirst('{{testScript}}', link)
           .replaceAll('{{testName}}', testName);
       return shelf.Response.ok(processedContents,
           headers: {'Content-Type': 'text/html'});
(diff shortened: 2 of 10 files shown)


wytrych:
@googlebot I signed it!

wytrych:
@jakemac53 Please take a look

wytrych:
I am not sure why the travis build is failing. The `dartanalyzer --fatal-infos --fatal-warnings` command does indeed error for me locally with 8755 errors, but none of this is related to this PR

wytrych:
It looks like something is wrong with my `dartanalyzer` showing all those errors, but nevertheless, I've pulled to dev version of dart and I don't see any hints/errors/warnings in my IntelliJ dart analyzer, so I'm a bit lost on what's wrong.

kevmoo:
@grouma ?

kevmoo:
@wytrych great idea!

pkgs/test/lib/src/runner/browser/platform.dart
```diff
@@ -185,16 +185,23 @@ class BrowserPlatform extends PlatformPlugin
...
-        <head>
-          <title>${htmlEscape.convert(test)} Test</title>
-          $link
-          <script src="packages/test/dart.js"></script>
-        </head>
-        </html>
-      ''', headers: {'Content-Type': 'text/html'});
+      if (_config.customHtmlTemplatePath != null) {
+        var contents = File(_config.customHtmlTemplatePath).readAsStringSync();
+        var processedContents = contents.replaceFirst('{testScript}', link);
```
jakemac53: Consider changing this to look for `{{testScript}}` instead of single curlies as I think that is a bit more common for template placeholders.

jakemac53: We may also want to allow for a test name placeholder `{{testName}}` (for title tags like below). This would not be required though like the `{{testScript}}` is.

This would also allow the default html file to be itself a template if you wish to do that but it isn't necessary.

pkgs/test/lib/src/runner/browser/platform.dart
```diff
@@ -276,6 +292,15 @@ class BrowserPlatform extends PlatformPlugin
     return suite;
   }
 
+  void checkHtmlCorrectness(String htmlPath, String path) {
```
jakemac53: nit: make this private

pkgs/test/lib/src/runner/browser/platform.dart
jakemac53: Also consider adding a comment here that we already check during the loading phase that this placeholder exists exactly once.

pkgs/test/test/runner/browser/runner_test.dart
```diff
@@ -2,8 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-@TestOn('vm')
+import 'dart:convert';
 
+@TestOn('vm')
```
jakemac53: This annotation needs to move to the first import (or add a library tag and annotate that)

pkgs/test/test/runner/browser/runner_test.dart
```diff
@@ -207,8 +208,145 @@ void main() {
...
+          test.stdout,
+          containsInOrder([
+            '-1: compiling test.dart [E]',
+            'Failed to load "test.dart": "test.html" must contain '
+                '<script src="packages/test/dart.js"></script>.'
+          ]));
+      await test.shouldExit(1);
+    }, tags: 'chrome');
+
     // TODO(nweiz): test what happens when a test file is unreadable once issue
```
jakemac53: nit: move this back to the bottom of the group or delete it

pkgs/test_core/lib/src/runner/configuration.dart
```diff
@@ -43,6 +43,10 @@ class Configuration {
   bool get help => _help ?? false;
   final bool _help;
 
+  /// Custom HTML template file
+  String get customHtmlTemplatePath => _customHtmlTemplatePath;
```
jakemac53: Just make the final variable public here - it already has no public setter.

The other ones have getters because they do some additional logic :)

pkgs/test/README.md
```diff
@@ -518,6 +521,37 @@ the following HTML file:
...
 ```
 
+### Providing a custom HTML template
+
+If you want to share the same HTML file across all tests, you can provide a
+`custom-html-template-path` configuration option to your configuration file.
+This file should follow the rules above, except that instead of the link tag
+add exactly one `{testScript}` in the place where you want the template processor to insert it.
+
+The template can't end with `.html` since the test loader will not able to process it and an error will be thrown.
```
jakemac53: I am a bit confused by why it errors - does it glob for all html files at some point? I wouldn't expect the test loader to ever try to load this file.

pkgs/test/lib/src/runner/browser/platform.dart
wytrych: I'm not sure I understand the `This would also allow the default html file to be itself a template ` bit. You mean the HTML that is in the else branch, so it would be processed in the same way as the user provided, thus simplifying the logic?

jakemac53: Yes exactly, it would simplify/unify the logic a bit. It would also be (negligibly) slower since it has to reach out to disk. You would likely use `Isolate.resolvePackageUri` to get the absolute file uri for the template file.

But again this is just a suggestion and not a requirement, I am fine with it as is :).

pkgs/test/README.md
wytrych: Sorry, this might be my bad writing. It will throw because I made it so in https://github.com/dart-lang/test/pull/1127/files/343af3e063c748e1323177e82854dbb0aaf1a33f#diff-b9a58c9d38842cb742b063d60ff6fb18R257 But the point of this is that if it is an `.html` file it will be served directly by shelf and will not be handled by the wrapper in `platform.dart`

jakemac53: This file should only ever need to be read from disk as a part of the `platform.dart` code though (it doesn't go through the shelf handler) - and it should be fine for shelf to serve it directly even if it was requested for some reason?

jakemac53:
Oh also one quick note please don't force-push your branch just push commits - we will do a squash and merge.

This makes it easier to review changes :).

pkgs/test/README.md
wytrych: Hmm, so the issue I was having is that the `suiteUrl` is generated in the `load` method by taking the dart file name and replacing `.dart` with `.html`. If shelf then tries to load that html file but it doesn't exist on disk, it will go into the wrapper handler and generate the file.
However if it does, it will be served directly, so it is not possible to process it, so it can't really be used as a template anymore. I came to the conclusion that it's less confusing if the template can't be an html file, because then it would be hard to understand without looking at the internal mechanics why it isn't processed. I decided then to disallow that during the config parsing phase.

Sure, it is kind of an edge case since the template file would have to be named like the test file, but I think it's nice to separate the old mechanics of having a custom html file for the test and having a template file. But if you think it should be allowed, then I'll remove that check.

jakemac53: Ah ok I get where the error came from now - IMO I don't think that is a common enough case to block all `.html` file extensions though, cc @natebosch wdyt?

natebosch: +1. If anything we should check that it doesn't match exactly the `.html` file for an existing test case.

jakemac53: Ah ya we could change the check to that instead - I like that.

wytrych:
Thanks @jakemac53 and @natebosch  I pushed fixes

pkgs/test/lib/src/runner/browser/static/default.html.tpl
```diff
@@ -0,0 +1,8 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>{{testName}}</title>
```
jakemac53:
```suggestion
    <title>{{testName}} Test</title>
```

To match exactly what it was before

jakemac53:
Going to let @grouma take a pass as well before merging (I think he is back Monday)

pkgs/test/lib/src/runner/browser/platform.dart
```diff
@@ -126,7 +128,11 @@ class BrowserPlatform extends PlatformPlugin
   /// Mappers for Dartifying stack traces, indexed by test path.
   final _mappers = <String, StackTraceMapper>{};
 
+  /// The default template for html tests
```
grouma: [nit] End in a period. (here and elsewhere)


https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

wytrych:
Thanks @grouma, I pushed the fixes.

jakemac53:
@grouma are there other changes we want to roll into this release?

wytrych:
Do you have an idea on when you will be releasing?

jakemac53:
It should be soon but we have a couple other prs we would like to merge, https://github.com/dart-lang/test/pull/1129 and ~https://github.com/dart-lang/test/pull/1131~

wytrych:
Cool, thanks!

grouma:
Just published.

PR #1129, opened 2020-01-06
Add file-based reporter support (via CLI option and yaml config)

evanweible-wf:
An attempt at #1100. Uses the `MultiplexReporter` from @natebosch's first approach to updating test reporters to write to a sink rather than printing directly to stdout/stderr, found here: https://github.com/dart-lang/test/pull/1101/commits/d61faf5ce4189648eb1e3e497752d942167cedfc#diff-d851d2a0e552a896ac5c4cce9f974d19

This allows a file reporter to be configured in addition to the stdout reporter.

This file-based reporter can be configured via a CLI option as well as the `dart_test.yaml` static config.

CLI:
```bash
--file-reporter <reporter>:<filepath>

# Example:
pub run test --file-reporter json:reports/tests.json
```

Static config:
```yaml
file_reporters:
  json: reports/tests.json
```

---

I'm not sure this is the best way to expose this functionality, I just took a stab at it based on the conversations from the previous PR. My main reason for taking the more general file-based reporter approach rather than targeting the JSON reporter in particular is the hope that the test runner will eventually support pluggable reporters (#12). That being said, maybe it would make more sense to take an intermediate step and use a more specific config option for just the JSON reporter.

Let me know what you all think! If there's agreement on an approach, I can work on adding tests and changelog/docs.

@grouma @natebosch

### DIFF
--- a/pkgs/test/test/runner/json_file_reporter_test.dart
+++ b/pkgs/test/test/runner/json_file_reporter_test.dart
@@ -12,6 +12,7 @@ import 'package:path/path.dart' as p;
 import 'package:test_descriptor/test_descriptor.dart' as d;
 
 import 'package:test/test.dart';
+import 'package:test_core/src/util/exit_codes.dart' as exit_codes;
 
 import '../io.dart';
 import 'json_reporter_utils.dart';
@@ -84,6 +85,22 @@ void main() {
         testDoneJson(5, result: 'failure'),
       ]
     ], doneJson(success: false));
+  });
+
+  group('reports an error if --file-reporter', () {
+    test('is not in the form <reporter>:<filepath>', () async {
+      var test = await runTest(['--file-reporter=json']);
+      expect(test.stderr,
+          emits(contains('option must be in the form <reporter>:<filepath>')));
+      await test.shouldExit(exit_codes.usage);
+    });
+
+    test('targets a non-existent reporter', () async {
+      var test = await runTest(['--file-reporter=nope:output.txt']);
+      expect(
+          test.stderr, emits(contains('"nope" is not a supported reporter')));
+      await test.shouldExit(exit_codes.usage);
+    });
   });
 }
 
--- a/pkgs/test/test/runner/json_file_reporter_test.dart
+++ b/pkgs/test/test/runner/json_file_reporter_test.dart
@@ -1,0 +1,140 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@TestOn('vm')
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+
+import 'package:test_descriptor/test_descriptor.dart' as d;
+
+import 'package:test/test.dart';
+
+import '../io.dart';
+import 'json_reporter_utils.dart';
+
+void main() {
+  test('runs successful tests with a stdout reporter and file reporter', () {
+    return _expectReports('''
+      test('success 1', () {});
+      test('success 2', () {});
+      test('success 3', () {});
+    ''', '''
+      +0: success 1
+      +1: success 2
+      +2: success 3
+      +3: All tests passed!''', [
+      [
+        suiteJson(0),
+        testStartJson(1, 'loading test.dart', groupIDs: []),
+        testDoneJson(1, hidden: true),
+      ],
+      [
+        groupJson(2, testCount: 3),
+        testStartJson(3, 'success 1', line: 6, column: 7),
+        testDoneJson(3),
+        testStartJson(4, 'success 2', line: 7, column: 7),
+        testDoneJson(4),
+        testStartJson(5, 'success 3', line: 8, column: 7),
+        testDoneJson(5),
+      ]
+    ], doneJson());
+  });
+
+  test('runs failing tests with a stdout reporter and file reporter', () {
+    return _expectReports('''
+      test('failure 1', () => throw new TestFailure('oh no'));
+      test('failure 2', () => throw new TestFailure('oh no'));
+      test('failure 3', () => throw new TestFailure('oh no'));
+    ''', '''
+      +0: failure 1
+      +0 -1: failure 1 [E]
+        oh no
+        test.dart 6:31  main.<fn>
+
+      +0 -1: failure 2
+      +0 -2: failure 2 [E]
+        oh no
+        test.dart 7:31  main.<fn>
+
+      +0 -2: failure 3
+      +0 -3: failure 3 [E]
+        oh no
+        test.dart 8:31  main.<fn>
+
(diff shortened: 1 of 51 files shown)


pkgs/test_core/lib/src/runner/configuration.dart
```diff
@@ -73,6 +73,15 @@ class Configuration {
   String get reporter => _reporter ?? defaultReporter;
   final String _reporter;
 
+  /// The map of file reporters where the key is the name of the reporter and
+  /// the value is the filepath to which its output should be written.
+  final Map<String, String> fileReporters;
```
evanweible-wf: Realistically right now there would only ever be one json file reporter entry, but if there are additional reporters added/supported in the future, this would support multiple output formats (e.g. json and xunit).

pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -124,6 +124,10 @@ final ArgParser _parser = (() {
       defaultsTo: defaultReporter,
       allowed: reporterDescriptions.keys.toList(),
       allowedHelp: reporterDescriptions);
+  parser.addOption('file-reporter',
```
evanweible-wf: Made this a single option to start with, but in the future this could be updated to a multi option to support multiple output formats.

grouma:
I'll let @natebosch drive this one but in general I like the approach as it seems very flexible without too much complexity.

pkgs/test_core/lib/src/runner/debugger.dart
```diff
@@ -85,7 +85,7 @@ class _Debugger {
   /// Whether [close] has been called.
   bool _closed = false;
 
-  bool get _json => _config.reporter == 'json';
```
natebosch: I think it only care about whether it is allowed to write to `stdout`, so we probably don't need to worry about whether json is used as a file reporter.

natebosch:
This is looking good to me. I think we want at least one test so we don't regress. I think the `reporter:filepath` arguments look like a good approach.

pkgs/test_core/lib/src/runner/configuration/load.dart
```diff
@@ -234,6 +235,18 @@ class _ConfigurationLoader {
       _error('Unknown reporter "$reporter".', 'reporter');
     }
 
+    var fileReporters = _getMap('file_reporter', key: (keyNode) {
```
evanweible-wf: As currently written, this technically supports multiple file reporter mappings. Should I rename this to `file_reporters`?

natebosch: Good call. Yeah I think `file_reporters` makes sense in the yaml.

pkgs/test_core/lib/src/runner/debugger.dart
evanweible-wf: Fixed, although I left the other check related to the json reporter just below this as I think collecting traces would need to happen when the json file reporter is used, too.

pkgs/test/test/runner/json_reporter_test.dart
```diff
@@ -598,207 +598,5 @@ import 'package:test/test.dart';
...
-  for (var value in expected) {
-    expect(stdoutLines.map(decodeLine), containsAllInOrder(value));
-  }
-}
-
-/// Returns the event emitted by the JSON reporter providing information about
-/// all suites.
-///
-/// The [count] defaults to 1.
-Map _allSuites({int count}) {
```
evanweible-wf: I moved all of these out to a separate file so that they could be shared between this test and the file reporter test.

evanweible-wf: If that's too much moved code for not a good enough reason, I could also move the file reporter tests into this test file.

evanweible-wf:
@natebosch alright I've addressed comments and added tests

pkgs/test/test/runner/json_file_reporter_test.dart
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
```
natebosch: [nit] 2020

pkgs/test/test/runner/json_file_reporter_test.dart
```diff
@@ -0,0 +1,140 @@
...
+  }).join('\n');
+
+  // Un-indent the expected string.
+  var indentation = stdoutExpected.indexOf(RegExp('[^ ]'));
+  stdoutExpected = stdoutExpected.split('\n').map((line) {
+    if (line.isEmpty) return line;
+    return line.substring(indentation);
+  }).join('\n');
+
+  // In Dart 1.24, stack traces with Future constructors output as
```
natebosch: I think we can probably drop this bit.

```dart
expect(actual, equals(stdoutExpected));
```

pkgs/test/test/runner/json_reporter_test.dart
natebosch: The move looks good to me.

pkgs/test/test/runner/json_reporter_utils.dart
```diff
@@ -0,0 +1,220 @@
+import 'dart:convert';
```
natebosch: [nit] Copyright statement

pkgs/test/test/runner/json_file_reporter_test.dart
natebosch: Ah, looks like we need to update the content too.

See https://github.com/dart-lang/test/pull/1133

evanweible-wf: 👍 I'll merge master once that's in.

evanweible-wf: Hmm nevermind, the two tests I added don't cover stack traces anyway

pkgs/test_core/lib/src/runner.dart
```diff
@@ -74,9 +75,24 @@ class Runner {
         var engine =
             Engine(concurrency: config.concurrency, coverage: config.coverage);
 
-        var reporterDetails = allReporters[config.reporter];
+        Reporter createFileReporter(String reporterName, String filepath) {
+          final sink = (File(config.fileReporters[reporterName])
+                ..createSync(recursive: true))
+              .openWrite();
+          return allReporters[reporterName]
+              .factory(config, engine, sink, onDone: sink.flush);
```
natebosch: How about adding a `List<IOSink>` field on this class, and flushing/closing them around here https://github.com/dart-lang/test/blob/eb8aa6b518860efc105e446b66334f531c8118ba/pkgs/test_core/lib/src/runner.dart#L233

evanweible-wf: Yeah that is much cleaner 👍

evanweible-wf:
@natebosch I think this is ready for a final review. 7403c96 fixed the test failure and the latest commit updates changelogs/docs.

PR #1131, opened 2020-01-06
Drop pedantic dependency

dnfield:
Flutter pins package dependencies, including all transitive dependencies.

Because test_api depends on pedantic, Flutter pins that dependency.

However, pedantic is only depended on for the `unawaited` function, which is a one-line piece of code.

This will enable resolution of https://github.com/flutter/flutter/issues/48246 without requiring Flutter to unpin its dependencies, and with a very small overhead on test_api to not re-import pedantic :)  This helps downstream consumers who want to depend on the package for the actual analysis options files.

Also related: https://github.com/dart-lang/pedantic/issues/46

### DIFF
--- a/pkgs/test_api/pubspec.yaml
+++ b/pkgs/test_api/pubspec.yaml
@@ -13,7 +13,6 @@ dependencies:
   collection: ^1.8.0
   meta: ^1.1.5
   path: ^1.2.0
-  pedantic: ^1.0.0
   source_span: ^1.4.0
   stack_trace: ^1.9.0
   stream_channel: ">=1.7.0 <3.0.0"
--- a/pkgs/test_api/pubspec.yaml
+++ b/pkgs/test_api/pubspec.yaml
@@ -25,6 +25,7 @@ dependencies:
 
 dev_dependencies:
   fake_async: ^1.0.0
+  pedantic: ^1.0.0
   test_descriptor: ^1.0.0
   test_process: ^1.0.0
   test: any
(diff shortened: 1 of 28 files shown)


pkgs/test_api/pubspec.yaml
```diff
@@ -13,7 +13,6 @@ dependencies:
   collection: ^1.8.0
   meta: ^1.1.5
   path: ^1.2.0
-  pedantic: ^1.0.0
```
natebosch: This will need to be moved to `dev_dependencies` instead of dropped entirely. We still use it for analysis options.

dnfield: Done.  

I wasn't sure if I should bump any versions with this, so I'm leaving that for now but would be happy to add it and/or update a changelog if applicable.

natebosch: This looks fine, we'll drop the `-dev` and get everything ready to publish separately.

natebosch:
Hmm, we might need to find other places importing both this library and `package:pedantic` and add some `show` to the imports maybe.

```
error • The name 'unawaited' is defined in the libraries 'package:pedantic/pedantic.dart' and 'package:test_api/src/utils.dart' at lib/src/runner/browser/browser.dart:138:7 • ambiguous_import
  error • The name 'unawaited' is defined in the libraries 'package:pedantic/pedantic.dart' and 'package:test_api/src/utils.dart' at lib/src/runner/node/platform.dart:128:7 • ambiguous_import
```

dnfield:
Looks like those users are doing implementation imports...

natebosch:
Yeah, when we split this package we had to take some shortcuts and ended up with a lot of `lib/src/` imports. We use narrow constraints between these packages to limit the risk. @grouma or I can take a look at making sure it gets rolled out safely.

dnfield:
Updated this to cover test_core and test as well.

PR #1137, opened 2020-01-09
Use await for setUp callbacks

sigurdm:
This improves the stacktraces for exceptions thrown in a setUp() callback.


natebosch:
Can you give an example of the difference?

sigurdm:
Sorry - I cannot reproduce the behavior I thought I saw yesterday. Not sure how I got convinced otherwise....

Issue #1141, opened 2020-01-11
test_api: crashes w/ dart2js with latest deployment


kevmoo:
See

https://travis-ci.org/dart-lang/yaml/jobs/635401215#L240
https://travis-ci.org/dart-lang/crypto/jobs/635401249#L240
https://travis-ci.org/dart-lang/typed_data/jobs/635046175#L240

Might be a dart2js issue that was caused by recent release... Only happens with Dart 2.3

CC @natebosch @grouma @sigmundch 

```
../../../.pub-cache/hosted/pub.dartlang.org/test_api-0.2.12/lib/src/backend/platform_selector.dart:13:7:
Internal Error: The compiler crashed when compiling this element.
final _universalValidVariables = {
      ^
The compiler is broken.
When compiling the above element, the compiler crashed. It is not
possible to tell if this is caused by a problem in your program or
not. Regardless, the compiler should not crash.
The Dart team would greatly appreciate if you would take a moment to
report this problem at http://dartbug.com/new.
Please include the following information:
* the name and version of your operating system,
* the Dart SDK build number (2.3.0), and
* the entire message you see here (including the full stack trace
  below as well as the source location above).
The compiler crashed: NoSuchMethodError: The method 'add' was called on null.
Receiver: null
Tried calling: add(Instance of 'VariableDeclaration')
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:50:5)
#1      StaticTypeVisitor.visitVariableDeclaration (file:///b/s/w/ir/cache/builder/sdk/pkg/compiler/lib/src/ir/static_type.dart:1361:23)
#2      VariableDeclaration.accept (package:kernel/ast.dart:4647:35)
#3      StaticTypeBase.visitNode (file:///b/s/w/ir/cache/builder/sdk/pkg/compiler/lib/src/ir/static_type_base.dart:76:18)
#4      StaticTypeVisitor.visitBlock (file:///b/s/w/ir/cache/builder/sdk/pkg/compiler/lib/src/ir/static_type.dart:956:22)
#5      Block.accept (package:kernel/ast.dart:3941:35)
#6      StaticTypeBase.visitNode (file:///b/s/w/ir/cache/builder/sdk/pkg/compiler/lib/src/ir/static_type_base.dart:76:18)
#7      StaticTypeVisitor.visitBlockExpression (file:///b/s/w/ir/cache/builder/sdk/pkg/compiler/lib/src/ir/static_type.dart:928:5)
[...]

PR #1143, opened 2020-01-14
Avoid for-loop elements at the top level

natebosch:
Towards #1141

Work around the bug in the SDK by switching back to `..addAll`. After we
publish this version that works we can bump the min SDK.

Prepare to publish.

### DIFF
--- a/pkgs/test_core/pubspec.yaml
+++ b/pkgs/test_core/pubspec.yaml
@@ -1,6 +1,6 @@
 name: test_core
-version: 0.2.18-dev
-description: A basic library for writing tests and running them on the VM.
+version: 0.2.18
+description: A basicanlibrary for writing tests and running them on the VM.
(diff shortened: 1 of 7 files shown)


pkgs/test_core/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: test_core
-version: 0.2.18-dev
-description: A basic library for writing tests and running them on the VM.
+version: 0.2.18
+description: A basicanlibrary for writing tests and running them on the VM.
```
jakemac53: Sp

natebosch: 😨

kevmoo:
Let's land this? We're hitting more cases

https://travis-ci.org/dart-lang/html/jobs/637664720#L240

natebosch:
Going to merge on red. phantom_js tests are broken on master too. https://github.com/dart-lang/test/issues/1146

Issue #899, opened 2018-07-05
Add a more clear error message when there is no `.dart.js` file with `--precompiled`


MikeMitterer:
In one of my packages I have this small test:

```dart
@TestOn("chrome")
library test.unit.core.utils;

import 'package:test/test.dart';
import 'dart:html' as dom;

main() {
    final dom.DivElement div = new dom.DivElement();

       test('> isDomElement', () async {
            expect(div,isNotNull);
        });
}
```

I run the test with `pub run build_runner clean && pub run build_runner test -- test/unit/core/dom_test.dart`

It fails with
```dart
...
[INFO] Running build completed, took 18.1s
[INFO] Caching finalized dependency graph completed, took 58ms
[INFO] Creating merged output dir `/var/folders/m4/9rg16xhd2rlbln753n9zk1540000gn/T/build_runner_test3FvmBH/` completed, took 10.2s
[INFO] Writing asset manifest completed, took 11ms
[INFO] Succeeded after 28.4s with 169 outputs (209 actions)
Running tests...

00:01 +0 -1: compiling test/unit/core/dom_test.dart [E]
  Failed to load "test/unit/core/dom_test.dart": Failed to load script at http://localhost:50353/Y8%2FO%2Fdy%2FRql%2FlrHTGE%2BHNzC1SDEC0qb%2B/test/unit/core/dom_test.dart.browser_test.dart.js.
00:01 +0 -1: Some tests failed.
```
It drives me crazy - almost every day one of the language/core_lib updates breaks something else...

MikeMitterer:
F... - it was a dependency problem!!!!! 

```yaml
dev_dependencies:
  test: any

  build_runner: any
  build_test: any

  # This was missing!!! - and caused the problem
  build_web_compilers: any
```

MikeMitterer:
Would be cool if someone could add this to the documentation - took me x hours to find the problem!

jakemac53:
While documentation is good, I think the most helpful thing would be for package:test to emit a more descriptive error when files are missing in precompiled mode.

We could even specifically mention `build_web_compilers` and `build_test` packages - as both are required for this to work.

Rodsevich:
Same problem here. But with the correct deps.
It should be a verboser message, please

Rodsevich:
My error was on having a transitive import that needed dart:io, so it wasn't being rendered by dart2js.
Would it be worth adding a PR with the same error but with and additional "make sure the script is being generated/processed"? If so I make it

[...]

PR #1144, opened 2020-01-15
Verboser error message

Rodsevich:
Helps with #899

### DIFF
--- a/pkgs/test/lib/dart.js
+++ b/pkgs/test/lib/dart.js
@@ -56,7 +56,7 @@ script.src = link.href + '.browser_test.dart.js';
 
 script.onerror = function(event) {
   var message = "Failed to load script at " + script.src +
-      (event.message ? ": " + event.message : ".");
+      (event.message ? ": " + event.message : ".") + " Make sure the script is being generated/processed.";
   sendLoadException(message);
 }
 


Rodsevich:
@googlebot I signed it!

pkgs/test/lib/dart.js
```diff
@@ -56,7 +56,7 @@ script.src = link.href + '.browser_test.dart.js';
 
 script.onerror = function(event) {
   var message = "Failed to load script at " + script.src +
-      (event.message ? ": " + event.message : ".");
+      (event.message ? ": " + event.message : ".") + " Make sure the script is being generated/processed.";
   sendLoadException(message);
```
jakemac53: It might make more sense to handle this in the test runner, wherever we handle these load exceptions. The reason being we should be able to check if we are in "precompiled" mode at that point, and give a more explicit message about checking your build output (and mention build_web_compilers specifically).

Rodsevich: That's right, but in addition to the plenty more time that that would take me (never digged in build_runner code before) the error makes sense to be in the test package because it appears even without build_runner being called when running bad tests with -p [chrome|firefox] (or at least I think that happened to me while I was fighting the subject)

grouma: I believe that message get's plumbed through here:
https://github.com/dart-lang/test/blob/dcd9b51808d762dae3fbc3a7d96180b7a87a3ba3/pkgs/test_core/lib/src/runner/plugin/platform_helpers.dart#L86

jakemac53:
@Rodsevich did you want to follow up here or should we close the PR?

Rodsevich:
I'm ashamed by the hours of programming time wasted in the future by the guys that will have this problem again.
I don't have the time to be the superhero here, sorry but I can't do much.
Maybe before closing this amazingly simple issue you could mention some more appropiate staff member that gets paid for doing this to make the 2 minutes change here needed, but to me that would be unprofitable hours I even don't have. Do as you consider the best.

jakemac53:
Closing this pull request but leaving the initial issue open.

Issue #36, opened 2015-03-12
Facilitate code coverage in test runner


travissanderson-wf:
I'm sure you guys are thinking about this but I didn't see an open issue to talk about it.  The alpha runner looks great but would like to see the ability to generate test coverage from it by providing a remote debugging port (or some other more elegant solution!).

zoechi:
:+1:

kevmoo:
May be blocked on https://code.google.com/p/dart/issues/detail?id=21791 – we're actively investigating

Sharom:
:+1:

donny-dont:
@nex3 any updates on this? The blocker @kevmoo mentioned is closed.

nex3:
#295 will provide access to the Observatory port for Dartium and content shell tests when run with `--pause-after-load`, which should make it possible to hook up to [the coverage package](https://github.com/dart-lang/coverage). Doing the same on the VM is blocked on #50, which is in turn blocked on dart-lang/sdk#23320.

All of that will require manually hooking up the coverage collector to Observatory. I'd like to support something more automatic, but that's not going to happen for a while.

avoivo:
@nex3 The thing is that i also need to pause the test run after all tests executed in order to collect the complete coverage. Is there any way to do that now?

nex3:
You should be able to talk to observatory and tell it to do that, either by setting some sort of pause-before-exit flag or by adding a breakpoint somewhere.

avoivo:
@nex Thanx for the reply. `--pause-before-exit` sounds good but i cannot find a way to pass that argument to the VM.

Also using 'dart:developer' library does not stop on breakpoint when running the tests.

The way i am running the tests is the following

```dart
pub run test:test --pub-serve=<port> -p dartium --pause-after-load
```

nex3:
> Thanx for the reply. `--pause-before-exit` sounds good but i cannot find a way to pass that argument to the VM.

You can't pass flags directly to the VM from `pub run`, and besides you'd probably want to pause Dartium's VM rather than the VM running the test. You'd have to set this via Observatory, although I'm not sure what the UI for that is.

> Also using 'dart:developer' library does not stop on breakpoint when running the tests.

That might be dart-lang/sdk#25369.

avoivo:
@nex3 ok thanx a lot for the info

eseidel:
In case it's useful, flutter's test wrapper does coverage collection:
[...]

PR #1151, opened 2020-01-22
Flexible Coverage API

grouma:
Towards https://github.com/dart-lang/test/issues/36

Update `deserializeSuite` to take an optional callback for coverage collection. Note that each platform e.g. VM, Browser etc. makes use of this helper. This sets the foundation for per platform custom coverage collection logic.

### DIFF
--- a/pkgs/test_core/lib/src/runner/coverage.dart
+++ b/pkgs/test_core/lib/src/runner/coverage.dart
@@ -14,7 +14,7 @@ Future<void> writeCoverage(
     String coveragePath, LiveSuiteController controller) async {
   var suite = controller.liveSuite.suite;
   var coverage = await controller.liveSuite.suite.gatherCoverage();
-  final outfile = File(p.join('$coveragePath',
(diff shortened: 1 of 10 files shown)


pkgs/test_core/lib/src/runner/coverage.dart
```diff
@@ -5,29 +5,20 @@
...
-      suite.environment.observatoryUrl, false, false, false, {},
-      isolateIds: {isolateId});
-
-  final outfile = File(p.join('$coverage', '${suite.path}.vm.json'))
+/// Collects coverage and outputs to the [coveragePath] path.
+Future<void> writeCoverage(
+    String coveragePath, LiveSuiteController controller) async {
+  var suite = controller.liveSuite.suite;
+  var coverage = await controller.liveSuite.suite.gatherCoverage();
+  final outfile = File(p.join('$coveragePath',
```
natebosch: `coveragePath` is already a string, skip the `'$'`.

Issue #1159, opened 2020-01-25
Would like to configure load exceptions to mark a test as skipped instead of failing


jonahwilliams:
On our CI infra, every couple of runs a web test will flake due to a failed to load exception. Rather than mark the test as failing, I would like to be able to configure it to mark as skipped - or else provide some hook for us to perform additional actions on a load exception.

https://github.com/flutter/flutter/issues/49435

jakemac53:
I would be concerned about simply marking them as skipped, a bug could easily cause all tests to fail to load and you would get a green CI run which would be bad.

Ideally, we would fix the load exceptions. If it is due to a very high number of tests being ran you could split them up into shards.

Another potential option is doing retries, we already have retry support but I don't know for sure that it works for load exceptions. It is possible that somehow bypasses the retry logic.

jakemac53:
It looks like the retry logic is specific to individual tests within a suite, and if an entire suite fails to load that will not be retried. I think fixing that would probably be the best place to start here.

jonahwilliams:
If we could retry the entire suite, I think that would be preferable. I'm a bit concerned we're getting into some sort of bad state where no other tests would pass, but I don't really know

Issue #1104, opened 2019-10-24
`invoker.dart` incorrectly reports which test times out


yjbanov:
In [this test log](https://gist.github.com/yjbanov/debfe93cbd59abf70a70287ed13a1781), notice that [invoker.dart](https://github.com/dart-lang/test/blob/c4e5f76796f32c052e07a558aa19d21cf3d05415/pkgs/test_api/lib/src/backend/invoker.dart#L277) reports timeouts for `flex_test.dart` and `dynamic_intrinsics_test.dart`. However, neither of these two tests time out. If you run them in isolation, they will finish quickly and successfully.

The test that's _actually_ timing out is `paginated_data_table_test.dart`. You can see it logged many times in the log. Notably, the log does not contain a message about this test timing out at all.

My theory is that the implementation of `heartbeat` is flawed, or the test loading system is, in that when you reduce concurrency to 1 (which we have to do, otherwise tests fail to run at all when compiled using DDC) it still attempts to load multiple tests but they are bottlenecked by the lack of concurrency. The heartbeat watcher starts, but the tests are waiting for other tests to finish. In the example test run above, `flex_test.dart` starts the heartbeat, but it's stuck waiting for `paginated_data_table_test.dart` to finish, which counts against `flex_test.dart`'s timeout quota.

jakemac53:
> which we have to do, otherwise tests fail to run at all when compiled using DDC

This is definitely not the case in general - I don't know if its worth investigating or not why you are experiencing that issue? It seems orthogonal in any case but it should speed up your test runs a fair bit to have better concurrency.

jakemac53:
@yjbanov are you able to provide a consistent repro for this? I tried doing a basic repro and couldn't get it to happen.

yjbanov:
If you patch in [this PR](https://github.com/flutter/flutter/pull/43019) and remove the `skip` in `paginated_data_table_test.dart`, that should trigger it on Cirrus very quickly. To reproduce this locally, I also had to artificially lower the timeout value in `heartbeat`.

This should do it:

```
cd packages/flutter
flutter -v test --platform=chrome \
  test/widgets/semantics_test.dart \
  test/rendering/editable_gesture_test.dart \
  test/material/page_transitions_theme_test.dart \
  test/widgets/framework_test.dart \
  test/physics/utils_test.dart \
  test/material/snack_bar_test.dart \
  test/material/paginated_data_table_test.dart \
  test/services/message_codecs_vm_test.dart \
[...]

PR #1160, opened 2020-01-27
Allow retry load suites

jakemac53:
Partial resolution for https://github.com/dart-lang/test/issues/1159.

Retries loading test suites (in loader.dart). Retry configuration is currently based on the suite metadata - and is counted separately from the test retries (does not count as a retry for tests).

Should we configure this separately from normal test retries? It might be reasonable to do retries for loading errors by default for instance (as they are likely transient and caused by test infrastructure). Also you might want to be able to opt into retrying load errors but not test failures.

Note that I also added an artificial delay of 1 second between retries. This seems reasonable to allow for any external factors that might be causing the load error to work themselves out. We might want to also make that configurable or have it do an exponential backoff.

Any other general comments?

cc @jonahwilliams

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -5,9 +5,6 @@
   will be removed in version `2.0.0`.
 * Support coverage collection for the Chrome platform. See `README.md` for usage
   details.
-* Support retrying of entire test suites when they fail to load.
-* Fix the `compiling` message in precompiled mode so it says `loading` instead,
-  which is more accurate.
 
 ## 1.11.1
 
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -5,6 +5,9 @@
   will be removed in version `2.0.0`.
 * Support coverage collection for the Chrome platform. See `README.md` for usage
   details.
+* Support retrying of entire test suites when they fail to load.
+* Fix the `compiling` message in precompiled mode so it says `loading` instead,
+  which is more accurate.
 
 ## 1.11.1
 
--- a/pkgs/test/test/runner/loader_test.dart
+++ b/pkgs/test/test/runner/loader_test.dart
@@ -3,7 +3,9 @@
 // BSD-style license that can be found in the LICENSE file.
 
 @TestOn('vm')
+import 'dart:async';
 
+import 'package:pedantic/pedantic.dart';
 import 'package:path/path.dart' as p;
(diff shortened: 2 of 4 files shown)


jakemac53:
Another thing I noticed when looking at this is that the loader does not currently respect the concurrency configuration - it will pre-emptively load several suites ([up to 8 at a time it looks like](https://github.com/dart-lang/test/blob/master/pkgs/test/lib/src/runner/browser/browser_manager.dart#L54)).

It is pretty easy to make that respect the concurrency setting - but it would likely slow down some tests where the concurrency is set to 1.

jonahwilliams:
I think @yjbanov has mentioned that before. We tried running on a lower concurrency to reduce the chance of hitting oom on CI. If that worked better it would be great

jakemac53:
This seems quite likely to be the cause of these load exceptions honestly, especially if it is actually just a timeout we are hitting (but its hard to tell exactly what the failure is).

I will go ahead and upload that change here as well, I can always revert it. I think anybody running with `-j1` has already accepted the tradeoff of having the tests run slower in order to make them more consistent, so respecting that during loading should be acceptable even if it slows down the tests more.

yjbanov:
We have a fork of `BrowserManager` where we [set the pools size to 1](https://github.com/flutter/flutter/blob/477686ae17fe9c21eef662449c5be1671ee75555/packages/flutter_tools/lib/src/test/flutter_web_platform.dart#L547). However, that does not seem to affect the number of suites loaded at one time. It only limits the loading process concurrency. It will still load multiple suites, just one at a time. This might be the cause of https://github.com/dart-lang/test/issues/1104.

jakemac53:
I believe my changes would have the same behavior - only one test would be _loading_ at a time but we might still have multiple _loaded_ at once.

This could still end up consuming a lot of memory in the browser.

yjbanov:
> This could still end up consuming a lot of memory in the browser.

Multiple DDC-compiled tests loaded in the browser is my current hypothesis for Cirrus flakiness in web tests. We had to disable a few tests and all screenshot tests due to flakiness.

jakemac53:
OK - I am going to revert the last commit I added here, add some tests and land just the retry logic first.

Then I can follow up with a proper change to make the loader/engine coordinate better and fully respect the concurrency so that we limit the number of _loaded_ tests to the concurrency setting. That is going to be a larger change though.

jakemac53:
Ready for review now, PTAL

PR #1161, opened 2020-01-28
Shared loading and running pool

jakemac53:
I will add a test once getting agreement on a fix here. 

This should resolve https://github.com/dart-lang/test/issues/1159

### DIFF
--- a/pkgs/test/test/runner/engine_test.dart
+++ b/pkgs/test/test/runner/engine_test.dart
@@ -280,16 +280,28 @@ void main() {
   group('concurrency', () {
     test('is shared between runner and load suites', () async {
       for (var concurrency = 1; concurrency < 5; concurrency++) {
+        var testsLoaded = 0;
+        var maxLoadConcurrency = 0;
         var testsRunning = 0;
-        var maxActualConcurrency = 0;
+        var maxTestConcurrency = 0;
         var testCount = concurrency * 2;
 
-        Future<void> updateAndCheckConcurrency() async {
-          testsRunning++;
-          maxActualConcurrency = max(maxActualConcurrency, testsRunning);
-          expect(testsRunning, lessThanOrEqualTo(concurrency));
+        Future<void> updateAndCheckConcurrency(
+            {bool isLoadSuite = false}) async {
+          if (isLoadSuite) {
+            testsLoaded++;
+            maxLoadConcurrency = max(maxLoadConcurrency, testsLoaded);
+            expect(testsLoaded, lessThanOrEqualTo(concurrency));
+          } else {
+            testsRunning++;
+            maxTestConcurrency = max(maxTestConcurrency, testsRunning);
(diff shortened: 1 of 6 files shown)


pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -220,7 +213,6 @@ class Engine {
   /// four times [concurrency].
   Engine({int concurrency, int maxSuites, String coverage})
```
natebosch: We should remove the `maxSuites` argument since it has no meaning anymore. Was that used anywhere?

jakemac53: It wasn't used within any of the `test` packages.

It is possible it was used elsewhere but it is somewhat difficult to tell for sure.

We are in a breaking release anyways so I will go ahead and just delete it.

jakemac53:
PTAL - added a test

jakemac53:
Note that bolt was using the `maxSuites` parameter but other than that internal is good to go. Removing that parameter for bolt gives the same behavior now (they also pass `concurrency: 1`) so we can land the change to remove that with the roll.

pkgs/test/test/runner/engine_test.dart
```diff
@@ -275,4 +276,56 @@ void main() {
...
+          if (isLoadSuite) {
+            testsLoaded++;
+            maxLoadConcurrency = max(maxLoadConcurrency, testsLoaded);
+            expect(testsLoaded, lessThanOrEqualTo(concurrency));
+          } else {
+            testsRunning++;
+            maxTestConcurrency = max(maxTestConcurrency, testsRunning);
+            expect(testsRunning, lessThanOrEqualTo(concurrency));
+          }
+          await Future.delayed(Duration(milliseconds: 100));
```
grouma: Comment why this is necessary

pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -265,35 +256,32 @@ class Engine {
...
-
+        var resource = await _runPool.request();
         LiveSuiteController controller;
-        if (suite is LoadSuite) {
-          await _onUnpaused;
-          controller = await _addLoadSuite(suite);
-          if (controller == null) {
-            loadResource.release();
-            return;
+        try {
```
grouma: Why do we now need a try catch clause?

jakemac53: This is just for the finally to ensure that we release the resource from the pool

jakemac53: (we return a couple places from this function, and want to make sure we release it even if there is an error as well)

pkgs/test/test/runner/engine_test.dart
jakemac53: done

PR #1164, opened 2020-02-01
Remove an unnecessary await on event loop

### LINKED ISSUE #865: Remove the unnecessary `await null` from `engine.close()`
With `Dart VM version: 2.0.0-dev.61.0` sync-async is on by default. Without the additional await tests fail in `test/runner/pause_after_load_test.dart`. 

See https://github.com/dart-lang/test/pull/864/files

natebosch:
Closes #865

The tests no longer fail without this.


natebosch:
The mentioned test didn't fail locally, looks like a different one does.

PR #1165, opened 2020-02-03
Fix flakiness

grouma:
Occasionally the test runner would try to connect to Chrome before the test url has been fully loaded. Add a retry which resolves the recent coverage flakiness we have been seing.

### DIFF
--- a/pkgs/test/lib/src/runner/browser/chrome.dart
+++ b/pkgs/test/lib/src/runner/browser/chrome.dart
@@ -83,7 +83,7 @@ class Chrome extends Browser {
           remoteDebuggerCompleter.complete(
               getRemoteDebuggerUrl(Uri.parse('http://localhost:$port')));
 
-          connectionCompleter.complete(_connect(process, port, idToUrl));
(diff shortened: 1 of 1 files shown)


pkgs/test/lib/src/runner/browser/chrome.dart
```diff
@@ -149,15 +149,28 @@ class Chrome extends Browser {
...
   await process.stderr
       .transform(utf8.decoder)
       .transform(LineSplitter())
       .firstWhere((line) => line.startsWith('DevTools listening'));
 
   var chromeConnection = ChromeConnection('localhost', port);
-  var tab = (await chromeConnection.getTabs()).first;
+  ChromeTab tab;
+  var attempt = 0;
+  while (tab == null) {
```
jakemac53: nit: add `attempt < 5` and then do the null check/throw after the while - I think that would be a bit cleaner

grouma: That actually ends up being slightly more verbose. Gonna leave as is.

PR #1166, opened 2020-02-03
Prep to publish

grouma:


robbecker-wf:
@grouma Trying to use test 1.12.0 release and getting
```
Because test >=1.12.0 depends on test_core 0.3.0 and build_test >=0.10.7 depends on test_core
  ^0.2.4, test >=1.12.0 is incompatible with build_test >=0.10.7.
```

jakemac53:
Looks like we need a new version of build_test I can get that out tomorrow

PR #1167, opened 2020-02-04
Add missing constraint

grouma:


pkgs/test/pubspec.yaml
```diff
@@ -11,6 +11,7 @@ dependencies:
   async: ^2.0.0
   boolean_selector: ">=1.0.0 <3.0.0"
   coverage: ^0.13.4
+  http: ^0.12.0
```
natebosch: We should consider dropping the import instead. Does `http` already show up in transitive deps?

grouma: It was already a transitive non dev dep so this shouldn't matter.

PR #1173, opened 2020-02-07
Exit as a failure if no tests ran

### LINKED ISSUE #764: Exit as a failure if no tests are run
This has come up in the past but I can't track down the discussion...

Refiling the issue since this is a problem all over again and solving it in the test package is the best overall fix. We're currently hacking around this internally by scraping for the string "No tests ran." but I'm worried this is fragile and we have to repeat this in all the places we run tests. We're also going to end up having to repeat this hack externally - https://github.com/dart-lang/build/issues/1000

My preference would be for either no option (no tests is always a failure) or for an option where the default is a failure since it's safest for CI environments. If we have an option where the default is to pass that would also work for us and we can spread the convention to always run with that option enabled.

cc @matanlurey @grouma @jakemac53

natebosch:
Closes #764

If there were no passing tests at all exit with a failing exit code.
This could happen if all tests are skipped, or file is run through
the test runner that does not define any tests. This will reduce the
chances of unintentionally configuring CI such that it doesn't actually
check anything.

### DIFF
--- a/pkgs/test_core/lib/src/runner.dart
+++ b/pkgs/test_core/lib/src/runner.dart
@@ -160,9 +160,7 @@ class Runner {
           throw ApplicationException('No tests match $patterns.');
         }
 
-        // Explicitly check "== true" here because [Engine.run] can return `null`
-        // if the engine was closed prematurely.
-        return success == true;
+        return _engine.passed.isNotEmpty && (success ?? false);
       });
 
   /// Emits a warning if the user is trying to run on a platform that's
--- a/pkgs/test_core/lib/src/runner.dart
+++ b/pkgs/test_core/lib/src/runner.dart
@@ -160,7 +160,8 @@ class Runner {
           throw ApplicationException('No tests match $patterns.');
         }
 
-        return _engine.passed.isNotEmpty && (success ?? false);
+        return (success ?? false) &&
+            (_engine.passed.isNotEmpty || _engine.skipped.isNotEmpty);
       });
 
   /// Emits a warning if the user is trying to run on a platform that's
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -14,6 +14,7 @@ analyzer:
     deprecated_member_use_from_same_package: ignore
 linter:
   rules:
+    - avoid_private_typedef_functions
     - await_only_futures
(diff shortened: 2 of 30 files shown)


pkgs/test_core/lib/src/runner.dart
```diff
@@ -160,9 +160,7 @@ class Runner {
           throw ApplicationException('No tests match $patterns.');
         }
 
-        // Explicitly check "== true" here because [Engine.run] can return `null`
-        // if the engine was closed prematurely.
-        return success == true;
+        return _engine.passed.isNotEmpty && (success ?? false);
```
jakemac53: We should emit a message if failing due to no tests running so people understand why we treated it as a failure - otherwise people will file issues because they think it should have passed

natebosch: Both the compact and expanded reporter have a message already. We could print something to `stderr` here I suppose?

jakemac53: Do we not have a general way of logging that works with any reporter?

natebosch:
cc @jonahwilliams - this will change behavior of the flutter test runner. In the case where none of the `_test.dart` files defined any test cases, it would cause a failure exit code. This should be very rare unless the flutter wrapper is splitting into individual suites per file or similar.

jonahwilliams:
Sounds like a reasonable change -  I don't think there will be any issues, but I'll keep an eye out.

pkgs/test_core/lib/src/runner.dart
natebosch: I don't think we do have a nice way to ask the reporters to report a message.

We can either:
- Move the `No test ran` output _out_ of the reporters and write it to stderr here.
- Assume that all reporters will have a clear enough "no tests" indication that the exit code isn't surprising - not sure if we'd really say this is the case for the json reporter today
- Have extra output to both stderr and through the reporter.

I'd lean towards the second option, I don't think this is likely to be confusing and if it does come up as a point of confusion we can address it then. What do you think?

natebosch: @jakemac53 - are you OK with this implementation for now?

PR #1175, opened 2020-02-08
Fix stack trace mapping for SDK locations

natebosch:
These were disabled because non-hermetic URIs were present. They have
been replaced with `org-dartlang-sdk` URIs. Update the SDK URI base so
that the replacement works for these URIs.

The tests in `pub_serve_test.dart` are still skipped for other reasons
so we can't guarantee they are fixed correctly here. Once we unskip
those tests we wouldn't want to hide issues with the stack trace
mapping, so we'll remove that reason for the skip.


natebosch:
@grouma - I updated this to map back to the same strings as before by updating our base URI for the SDK. PTAL

Issue #1177, opened 2020-02-10
META: migrate all transitive deps to the latest package_config


jakemac53:
This package doesn't directly depend on package_config but 3 of its dependencies do:

- [ ] [analyzer](https://github.com/dart-lang/sdk/issues/40569)
- [x] ~[package_resolver](https://github.com/dart-lang/package_resolver/issues/29)~ deprecate instead
- [x] [coverage](https://github.com/dart-lang/coverage/issues/289) - [PR](https://github.com/dart-lang/coverage/pull/292)

~In order for any package that wants to use package_config to run tests (including package_config itself) we need to migrate all of these.~ It looks like we will be able to make a backwards/forwards compatible release, but should still migrate these packages.

jakemac53:
Also we are considering deprecating package_resolver after looking into migrating it - so that would mean migrating the following to use package_config:

- [shelf_packages_handler](https://github.com/dart-lang/shelf_packages_handler/issues/10)
- test
- test_core
- [source_map_stack_trace](https://github.com/dart-lang/source_map_stack_trace/issues/10)

jakemac53:
Closing as this is effectively done - still waiting on the analyzer but we aren't actually blocked on it.

PR #1181, opened 2020-02-13
Drop package resolver from test_core and test, use package_config from test

jakemac53:
Towards https://github.com/dart-lang/test/issues/1177.

This relies on the new source_map_stack_trace here https://github.com/dart-lang/source_map_stack_trace/pull/11 which no longer requires you to pass a `PackageResolver`.

It also relies on the upcoming package_config 1.9.0.

Once package_config is published and this goes green I will merge and publish source_map_stack_trace, and then we can remove the git override here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,8 @@ after_failure:
 jobs:
   include:
     - stage: analyze_and_format
-      name: "SDK: 2.4.0; PKGS: pkgs/test, pkgs/test_api, pkgs/test_core; TASKS: `dartanalyzer --fatal-warnings .`"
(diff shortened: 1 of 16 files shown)


jakemac53:
Note that I did use an extension method and thus had to update the min sdk for the test package to 2.7.0.

I might undo that change and just use a top level method if we want to.... otherwise we can't test any of our packages (or transitive deps) on <2.7.0 so they all effectively have to up their min sdk as well.

jakemac53:
cc @lrhn - I am not submitting anything here until we have actually published package_config so no worries if the api changes from what I have used here.

PR #1182, opened 2020-02-19
Fix usage string for test-randomize-ordering-seed

natebosch:
We changed the behavior to allow passing an explicit `0` and treating
that differently from omitting the argument in #1150


natebosch:
See https://github.com/flutter/flutter/issues/51010

Issue #1183, opened 2020-02-19
StreamMatchers should cancel any StreamQueue that they create once done matching


simolus3:
The following test will print "tearDown - start" but then time out. The test hangs in the `tearDown` method at `await controller.close()`.
```dart
import 'dart:async';
import 'package:test/test.dart';

void main() {
  StreamController<String> controller;

  setUp(() {
    controller = StreamController.broadcast();
  });

  tearDown(() async {
    print('tearDown - start');
    await controller.close(); // this future never completes
    print('tearDown - done');
  });

  test('this test times out', () async {
    final stream = controller.stream;

    final expectation = expectLater(stream, emitsInOrder([isNull, isNotNull, isNull]));

    controller..add(null)..add('foo')..add(null);
    await expectation;
    print('matched expected items, test body done');
  });
}

```

Is this expected? If so, is there anything users can do to prevent it?

jakemac53:
It looks like this is a result of the StreamQueue used by the matcher, this also hangs:

```
test('can close a stream wrapped in a queue', () async {
  var controller = StreamController<String>();
  var queue = StreamQueue(controller.stream);
  var done = queue.next;
  controller..add('foo');
  print(await done);
  await controller.close(); // hangs here
  print('done');
});
```

jakemac53:
This is likely because the `StreamQueue` never consumes the `done` event. I don't know what the solution is here other than just not awaiting the `close` call.

jakemac53:
See also https://github.com/dart-lang/sdk/issues/19095#issuecomment-471722971

After talking with @natebosch we decided the fix would be to cancel the implicit stream queue that is created (only if we create it). This might be breaking for some scenarios so we would have to do some investigation into that or land it with the NNBD changes.

jakemac53:
The recommendation in general is to not await the `close` method - as that can always arbitrarily block if you have a subscriber that is paused.

jakemac53:
Note that I have made the change to cancel any stream queues we create in the matchers and ran that through an internal presubmit without finding any failures. So it should be safe to make that change and land it as non-breaking.

This should make it a lot less likely to run into this type of failure.

PR #1185, opened 2020-02-20
Cancel any StreamQueue that is created as a part of a stream matcher

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1183

Ran through an internal presubmit and everything came back green so this should be safe.

### DIFF
--- a/pkgs/test_api/CHANGELOG.md
+++ b/pkgs/test_api/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.2.15-dev
+
+* Cancel any StreamQueue that is created as a part of a stream matcher once it
+  is done matching.
+  * This fixes a bug where using a matcher on a custom stream controller and
+    then awaiting the `close()` method on that controller would hang.
+
 ## 0.2.14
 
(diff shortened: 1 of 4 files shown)


pkgs/test_api/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.2.15-dev
+
+* Cancel any StreamQueue that is created as a part of a stream matcher once it
```
natebosch: [nit] Can we copy this into the test changelog? That's the one most users will see.

jakemac53: hmm, its a bit complicated since I wasn't planning on publishing this yet? so putting it in the test changelog might not be accurate (we already have one pending change there which could get published before this does)

natebosch: Fair enough. We can consider it a part of the process to rectify changelogs in the commits where we bump the dependency.

jakemac53: Ya I think that would be the best time to do it - not sure the right way to codify that process though.

PR #1189, opened 2020-02-25
Allow a timeout following a completed test

### LINKED ISSUE #1188: Test runner hangs if a `tearDown` should time out following a test failure
```dart
import 'dart:async';

void main() {
  tearDown(() async {
    await Completer<void>().future;
  });
  test('something', () async {
    expect(true, false);
  });
}
```

I think it is because of:

https://github.com/dart-lang/test/blob/594e9e6775c53f5f002818c545f01a0b6e3ec509/pkgs/test_api/lib/src/backend/invoker.dart#L287

The test is already complete because it is known to have failed. Since we don't emit the error for the timeout, we never move on to more tests.

natebosch:
Fixes #1188

A test is marked completed as soon as it fails. If a `tearDown` callback
also take too long, or never completes, we ignore the timeout and never
move on to more tests.

### DIFF
--- a/pkgs/test/test/runner/timeout_test.dart
+++ b/pkgs/test/test/runner/timeout_test.dart
@@ -110,4 +110,54 @@ void main() {
             ['Test timed out after 0.4 seconds.', '-1: Some tests failed.']));
     await test.shouldExit(1);
   });
+
+  test('times out teardown callbacks', () async {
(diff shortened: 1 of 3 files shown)


natebosch:
Trying this out on Travis to see if any existing tests break. I can't find the reason this line was originally written.

PR #1191, opened 2020-02-28
Support package:vm_service 3.0.0+1

bkonyi:


pkgs/test_core/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.3.2-dev
```
jakemac53: you can move this into the previous `-dev` - we don't publish `-dev` releases just use that as a marker for the next release

pkgs/test_core/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: test_core
-version: 0.3.1-dev
+version: 0.3.2-dev
```
jakemac53:
```suggestion
version: 0.3.1-dev
```

pkgs/test_core/pubspec.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
   source_span: ^1.4.0
   stack_trace: ^1.9.0
   stream_channel: ">=1.7.0 <3.0.0"
-  vm_service: '>=1.0.0 <3.0.0'
+  vm_service: '>=3.0.0 <4.0.0'
```
jakemac53: Do we need to increase the lower bound here?

bkonyi: There's the potential for a non-expected `Sentinel` return value in `package:vm_service` <3.0.0 which will cause a type error. It's unlikely we'll bump into it here, but I wanted to be safe rather than sorry. I'm content with leaving the lower bound as-is if that's what you'd prefer.

jakemac53: I think we would prefer to leave it - assuming this was a pre-existing issue? We haven't had reports of it and would prefer to leave this lax so that version solving is more likely to succeed for people.

bkonyi: Fair enough. Done.

jakemac53:
It looks like that version of coverage isn't published yet - do we actually need to require that as well or can we leave it allowing the older version?

bkonyi:
> It looks like that version of coverage isn't published yet - do we actually need to require that as well or can we leave it allowing the older version? (0.13.7 is already allowed by the current constraint)

Right, I always forget how the `^` constraints work. We should be okay to keep the current constraint then.

PR #1192, opened 2020-03-02
release test_api/test_core/test

jakemac53:
cc @bkonyi


pkgs/test/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 1.13.0-dev
+## 1.13.0
```
natebosch: Can you copy in the changes from the `test_api` changelog since this is the one users will see?

Issue #1193, opened 2020-03-03
test: Support latest version of pkg:shelf_packages_handler


kevmoo:
Current constraint doesn't allow v2.0.0

PR #1195, opened 2020-03-03
allow the latest shelf_packages_handler

jakemac53:
Closes https://github.com/dart-lang/test/issues/1193


jakemac53:
This also gets package_resolver completely out of the transitive deps :tada:

PR #1196, opened 2020-03-09
Update to support package:vm_service version 4.0.0

bkonyi:


pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -81,7 +81,7 @@ class VMPlatform extends PlatformPlugin {
       await client.setName(isolateRef.id, path);
       var libraryRef = (await client.getIsolate(isolateRef.id))
           .libraries
-          .firstWhere((library) => library.uri == libraryPath) as LibraryRef;
+          .firstWhere((library) => library.uri == libraryPath);
```
jakemac53: Does removing this cast make us not compatible with old vm service versions or is it still fine?

bkonyi: package:vm_service 3.0.0+1 returns `Future<dynamic>` so we need to be a little more specific here with types I guess... :-\ Any objections to specifying a type for `libraryRef` and ignoring the lint on that line?

natebosch: I think we'd only run analysis against the latest `vm_service`. This looks backwards compatible apart from analysis. I'm with this change.

jakemac53: Sgtm

bkonyi: Yeah, this is failing since we need to update `package:coverage` to use `vm_service` 4.0.0 as it currently restricts it to <4.0.0.

PR #1197, opened 2020-03-17
Update usage output test for new args package

natebosch:
The extra newlines between some options has been removed.


natebosch:
Latest SDK broke `StringLiteralIterator`. I'm going to merge on red and send a separate fix for that.

PR #1199, opened 2020-03-18
migrate test_core to null safety

jakemac53:
Analyzer clean - no tests ran yet.

### DIFF
--- a/pkgs/test_core/lib/src/runner.dart
+++ b/pkgs/test_core/lib/src/runner.dart
@@ -52,7 +52,7 @@ class Runner {
   final Reporter _reporter;
 
   /// The subscription to the stream returned by [_loadSuites].
-  StreamSubscription _suiteSubscription;
+  StreamSubscription? _suiteSubscription;
 
   /// The set of suite paths for which [_warnForUnknownTags] has already been
   /// called.
@@ -64,7 +64,7 @@ class Runner {
   /// The current debug operation, if any.
   ///
   /// This is stored so that we can cancel it when the runner is closed.
-  CancelableOperation _debugOperation;
+  CancelableOperation? _debugOperation;
 
   /// The memoizer for ensuring [close] only runs once.
   final _closeMemo = AsyncMemoizer();
@@ -83,17 +83,17 @@ class Runner {
           final sink =
               (File(filepath)..createSync(recursive: true)).openWrite();
           sinks.add(sink);
-          return allReporters[reporterName].factory(config, engine, sink);
+          return allReporters[reporterName]!.factory(config, engine, sink);
         }
 
         return Runner._(
           engine,
           MultiplexReporter([
             // Standard reporter.
-            allReporters[config.reporter].factory(config, engine, stdout),
+            allReporters[config.reporter]!.factory(config, engine, stdout),
             // File reporters.
             for (var reporter in config.fileReporters.keys)
-              createFileReporter(reporter, config.fileReporters[reporter]),
+              createFileReporter(reporter, config.fileReporters[reporter]!),
           ]),
           sinks,
         );
@@ -129,21 +129,21 @@ class Runner {
         }
 
         if (_config.coverage != null) {
-          await Directory(_config.coverage).create(recursive: true);
+          await Directory(_config.coverage!).create(recursive: true);
         }
 
-        bool success;
+        bool? success;
         if (_config.pauseAfterLoad) {
           success = await _loadThenPause(suites);
         } else {
           _suiteSubscription = suites.listen(_engine.suiteSink.add);
           var results = await Future.wait(<Future>[
-            _suiteSubscription
+            _suiteSubscription!
                 .asFuture()
                 .then((_) => _engine.suiteSink.close()),
             _engine.run()
           ], eagerError: true);
-          success = results.last as bool;
+          success = results.last as bool?;
         }
 
         if (_closed) return false;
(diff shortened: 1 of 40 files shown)


pkgs/test_core/lib/src/runner/configuration/load.dart
```diff
@@ -130,7 +130,7 @@ class _ConfigurationLoader {
               (value) => value is String);
 
           var os = OperatingSystem.find(keyNode.value as String);
-          if (os != null) return os;
+          if (os == OperatingSystem.none) return os;
```
natebosch: Should this be `!=`?

pkgs/test_core/lib/src/runner.dart
```diff
@@ -236,9 +236,9 @@ class Runner {
           });
         }
 
-        if (_debugOperation != null) await _debugOperation.cancel();
+        await _debugOperation?.cancel();
```
kevmoo: Inspired https://github.com/dart-lang/linter/issues/2037

pkgs/test_core/lib/src/runner.dart
```diff
@@ -172,8 +172,8 @@ class Runner {
 
     var unsupportedRuntimes = _config.suiteDefaults.runtimes
         .map(_loader.findRuntime)
-        .where((runtime) =>
-            runtime != null && !testOn.evaluate(currentPlatform(runtime)))
+        .whereType<Runtime>()
+        .where((runtime) => !testOn.evaluate(currentPlatform(runtime)))
```
grouma: How does this work? Is Runtime guaranteed to be non-null here?

pkgs/test_core/lib/src/runner.dart
```diff
@@ -303,7 +303,7 @@ class Runner {
   /// children.
   void _warnForUnknownTags(Suite suite) {
     if (_tagWarningSuites.contains(suite.path)) return;
-    _tagWarningSuites.add(suite.path);
+    _tagWarningSuites.add(suite.path!);
```
grouma: Why can path be null, can we change that?

pkgs/test_core/lib/src/runner/configuration.dart
```diff
@@ -41,61 +41,61 @@ class Configuration {
...
 
   /// The path to dart2js.
   String get dart2jsPath => _dart2jsPath ?? p.join(sdkDir, 'bin', 'dart2js');
-  final String _dart2jsPath;
+  final String? _dart2jsPath;
 
   /// The name of the reporter to use to display results.
   String get reporter => _reporter ?? defaultReporter;
-  final String _reporter;
+  final String? _reporter;
```
grouma: Can these somehow use the `late` pattern so that they may be non-nullable?

pkgs/test_core/lib/src/runner/configuration/load.dart
```diff
@@ -231,7 +231,7 @@ class _ConfigurationLoader {
     var runSkipped = _getBool('run_skipped');
 
     var reporter = _getString('reporter');
-    if (reporter != null && !allReporters.keys.contains(reporter)) {
+    if (!allReporters.keys.contains(reporter)) {
```
grouma: Can reporter not be null anymore?

pkgs/test_core/lib/src/runner/reporter/expanded.dart
```diff
@@ -62,21 +62,21 @@ class ExpandedReporter implements Reporter {
 
   /// The size of `_engine.passed` last time a progress notification was
   /// printed.
-  int _lastProgressPassed;
+  int? _lastProgressPassed;
```
grouma: What's our policy with instances like this here. Should we make them non-nullable and default them to 0?

pkgs/test_core/lib/src/util/io.dart
```diff
@@ -43,7 +43,7 @@ final String sdkDir = p.dirname(p.dirname(Platform.resolvedExecutable));
 final OperatingSystem currentOS = (() {
   var name = Platform.operatingSystem;
   var os = OperatingSystem.findByIoName(name);
-  if (os != null) return os;
+  if (os == OperatingSystem.none) return os;
```
grouma: This doesn't seem right here?

pkgs/test_core/lib/src/runner/reporter/expanded.dart
jakemac53: done for all but one of these, I think that is reasonable

pkgs/test_core/lib/src/runner/configuration.dart
natebosch: That's the pattern I'm leaning towards and it has seemed appropriate so far. The other use is using a variable that needs to be passed in it's initializer, like what happens with stream subscriptions.

jakemac53:
Tests are passing with two warnings, one due to https://github.com/dart-lang/sdk/issues/41114 and one due to https://github.com/dart-lang/sdk/issues/41108.

Also added a workaround for https://github.com/dart-lang/sdk/issues/41111.

pkgs/test_core/lib/src/runner/configuration.dart
jakemac53: Ah right I have used it that way as well

jakemac53: I wonder where best practices for this are being documented? Should we update effective dart?

jakemac53: cc @munificent

jakemac53:
Going to go ahead and merge - @grouma if you have anything else please comment and ill follow up with changes.

pkgs/test_core/lib/src/runner/configuration.dart
munificent: We should update Effective Dart at some point, but I want to make sure we have enough time for the best practices to settle out before we start telling people what to do.

PR #1204, opened 2020-03-20
Expand documentation on timeouts

natebosch:
Make it clear that a timeout does not prevent long running tests.

### DIFF
--- a/pkgs/test/README.md
+++ b/pkgs/test/README.md
@@ -615,9 +615,12 @@ void main() {
 
 ### Timeouts
 
-By default, tests will time out after 30 seconds of inactivity. However, this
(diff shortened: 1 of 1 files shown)


natebosch:
We might also want to add new features to match how some users are wanting to use the timeouts.

https://github.com/dart-lang/test/issues/1205

grouma:
LGTM. I was misinformed and thought that the timeout was intended for overall test runtime. I like the idea of introducing such a timeout. It would definitely be useful for our internal / travis use cases.

natebosch:
> I was misinformed and thought that the timeout was intended for overall test runtime

I think this is about the 5th time I got confused by this.

jakemac53:
lol wow TIL 🤦‍♂

PR #1207, opened 2020-03-26
Check SILENT_OBSERVATORY from Process.environment

### LINKED ISSUE #1206: How to actually set SILENT_OBSERVATORY?
So, there is a warning about observatory output messing up various reporters (jUnit json in my case) which tells to define `DART_VM_OPTIONS=-DSILENT_OBSERVATORY` to workaround this.

`Runner` from `test_core` reads it as `bool.fromEnvironment` that doesn't read process environment but actually reads `dart -D...` options. 

There is no way (at least that I'm aware of) to pass options to VM when executing `pub run test`.

So how do I pass this option actually? Do I have to write my own version of `test_core` to have control over actual `dart` process invocation to pass this option or there is another way?

P.S. Running `DART_VM_OPTIONS=-DSILENT_OBSERVATORY=true pub run test -r json --coverage coverage` seems to supress observatory output but the warning message is still printed so reporter still breaks, ironically.

Anrock:
Should fix #1206

### DIFF
--- a/pkgs/test_core/lib/src/runner.dart
+++ b/pkgs/test_core/lib/src/runner.dart
@@ -31,7 +31,7 @@ import 'runner/reporter.dart';
 import 'runner/reporter/compact.dart';
 import 'runner/reporter/expanded.dart';
 
-final _silentObservatory = const bool.fromEnvironment('SILENT_OBSERVATORY');
+final _silentObservatory = Platform.environment['SILENT_OBSERVATORY'] == true;
 
 /// A class that loads and runs tests based on a [Configuration].
 ///


grouma:
This actually would be a breaking change for our Flutter users.

Let's wait for @natebosch to respond to see if this can be set through `pub run test`. If there isn't a good way, we can always check both the platform environment variables and the VM variables.

natebosch:
This isn't the correct fix for the precompiled case. We _might_ be able to do something like `Platform.environment['DART_VM_OPTIONS'].contains('-DSILENT_OBSERVATORY=true')` or something, but I worry about it being fragile.

Anrock:
Ok, closing as invalid fix.

PR #1209, opened 2020-03-27
Use Never return type for function which throws

natebosch:
Fixes a static error:

```
pkgs/test_api/lib/src/frontend/expect.dart:82:8: Error: A non-null value must be returned since the return type 'Future<dynamic>' doesn't allow null.
 - 'Future' is from 'dart:async'.
Future _expect(actual, matcher,
```

The CFE does not appear to support `@alwaysThrows`, but a return type of
`Never` is the correct way to express this with null safety.

### DIFF
--- a/pkgs/test_api/lib/src/frontend/expect.dart
+++ b/pkgs/test_api/lib/src/frontend/expect.dart
@@ -154,7 +154,7 @@ Future _expect(actual, matcher,
 /// Convenience method for throwing a new [TestFailure] with the provided
 /// [message].
 @alwaysThrows
-Null fail(String message) => throw TestFailure(message);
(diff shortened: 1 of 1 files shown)


pkgs/test_api/lib/src/frontend/expect.dart
```diff
@@ -154,7 +154,7 @@ Future _expect(actual, matcher,
 /// Convenience method for throwing a new [TestFailure] with the provided
 /// [message].
 @alwaysThrows
```
jakemac53: We can actually remove this annotation I think now as well, `Never` effectively handles the same use case.

natebosch: Done

PR #1210, opened 2020-03-31
Handle file URIs during coverage collection

grouma:
Closes https://github.com/dart-lang/test/issues/1214

### DIFF
--- a/pkgs/test/lib/src/runner/browser/chrome.dart
+++ b/pkgs/test/lib/src/runner/browser/chrome.dart
@@ -122,22 +122,16 @@ class Chrome extends Browser {
   Future<Uri> _sourceUriProvider(String sourceUrl, String scriptId) async {
     var script = _idToUrl[scriptId];
     if (script == null) return null;
-    // The script is a URL path so it should contain forward slashes.
-    var pathSegments = script.split('/');
+    var sourceUri = Uri.parse(sourceUrl);
+    if (sourceUri.scheme == 'file') return sourceUri;
+    // If the provided sourceUrl is relative, determine the package path.
+    var uri = Uri.parse(script);
     var path = p.join(
-        p.joinAll(pathSegments.sublist(1, pathSegments.length - 1)), sourceUrl);
-    if (path.contains('/packages/')) {
-      return Uri(scheme: 'package', path: path.split('/packages/').last);
-    } else {
-      try {
-        return Uri.file(p.absolute(path));
-      } catch (e) {
-        print('Unable to parse path: $path');
-        print('Provided sourceUrl: $sourceUrl');
-        print('Corresponding script: $script');
-      }
-      return null;
-    }
+        p.joinAll(uri.pathSegments.sublist(1, uri.pathSegments.length - 1)),
+        sourceUrl);
+    return path.contains('/packages/')
+        ? Uri(scheme: 'package', path: path.split('/packages/').last)
+        : null;
   }
 
   Future<String> _sourceMapProvider(String scriptId) async {
--- a/pkgs/test/lib/src/runner/browser/chrome.dart
+++ b/pkgs/test/lib/src/runner/browser/chrome.dart
@@ -126,9 +126,16 @@ class Chrome extends Browser {
     var path = p.join(
         p.joinAll(uri.pathSegments.sublist(1, uri.pathSegments.length - 1)),
         sourceUrl);
-    return path.contains('/packages/')
-        ? Uri(scheme: 'package', path: path.split('/packages/').last)
-        : Uri.file(p.absolute(path));
+    if (path.contains('/packages/')) {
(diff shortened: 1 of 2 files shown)


pkgs/test/lib/src/runner/browser/chrome.dart
```diff
@@ -126,9 +126,16 @@ class Chrome extends Browser {
...
-        : Uri.file(p.absolute(path));
+    if (path.contains('/packages/')) {
+      return Uri(scheme: 'package', path: path.split('/packages/').last);
+    } else {
+      try {
+        return Uri.file(p.absolute(path));
+      } catch (e) {
+        print('Unable to parse path: $path');
+      }
+      return null;
```
jakemac53: Do we want to return null here or rethrow?

grouma: For the purposes of debugging it shouldn't matter.

jakemac53: kk

pkgs/test/lib/src/runner/browser/chrome.dart
```diff
@@ -122,17 +122,19 @@ class Chrome extends Browser {
   Future<Uri> _sourceUriProvider(String sourceUrl, String scriptId) async {
     var script = _idToUrl[scriptId];
     if (script == null) return null;
-    var uri = Uri.parse(script);
+    // The script is a URL path so it should contain forward slashes.
```
natebosch, line 125: `p.url.split` would be more appropriate.

But if `Uri.parse(script).pathSegments` is somehow including `file:` I don't know how strictly splitting on `/` would do any better...

pkgs/test/lib/src/runner/browser/chrome.dart
```diff
@@ -122,17 +122,19 @@ class Chrome extends Browser {
...
+    var pathSegments = script.split('/');
     var path = p.join(
-        p.joinAll(uri.pathSegments.sublist(1, uri.pathSegments.length - 1)),
-        sourceUrl);
+        p.joinAll(pathSegments.sublist(1, pathSegments.length - 1)), sourceUrl);
```
natebosch, line 128: this could probably be `p.joinAll(pathSegments.skip(1))` or `p.joinAll(uri.pathSegments.skip(1)`

pkgs/test/lib/src/runner/browser/chrome.dart
grouma: My theory is that `Uri.parse` is introducing `file:`. I'll need an update from @donny-dont

natebosch: I can't find any way to get `Uri.parse` to introduce `file:`.

Further, if it _was_ `Uri.parse` adding a scheme, reading `.pathSegments` should _definitely_ not include it. I strongly suspect that it's showing up earlier. You might check eagerly if `_idToUrl` gets assigned something with a `file:` somewhere in it.

grouma: Indeed. It definitely isn't coming from `Uri.parse`. It's getting passed in as `sourceUrl` which comes from the source map here:
https://github.com/dart-lang/coverage/blob/master/lib/src/chrome.dart#L54

That means the compiler is embedding `file:` or `package:source_maps` is adding it. That's odd as I thought these paths were usually relative, hence this munging. Nonetheless I think we can check the `sourceUrl` scheme and just handle the absolute file case directly. Thoughts?

natebosch: Yeah if we can handle it earlier when we read from the source map that would be great.

I'm fine with something relatively local if it fixes it. I'd feel more comfortable if we have a more complete understanding of the conditions that lead up to this - it would potentially be more work but if it's sensible to avoid these from ever showing up, whether in DDC or `package:source_maps`, that could be an improvement for more than just this use case.

grouma: I did a little more digging. I think something may have changed on the Dart2JS side. The source map contains absolute file URIs. I'm not sure if that ever was the case? Anyway, this simple fix resolves the issue.

natebosch: cc @sigmundch - was there some regression here?

sigmundch: No known regression - dart2js embeds source-uris as given to it, and the compiler won't change them. I wish we'd do the same for ddc in the future, so any source-map munging would be external to the compilers.

Some caveats:
 - in both our external sdk and our internal build system (bazel) we provide dart2js with custom URI schemes to make the output agnostic of the file system, in those cases the source-map URIs will likely have custom-uri schemes.
 - internally we also post-process source-map files to make it possible to enable source-map debugging against a code serving server (cs). In that case URIs are turned into absolute http URIs.

PR #1217, opened 2020-04-24
Retain configured file reporters when resolving presets

evanweible-wf:
When a preset is defined and chosen, the file reporter configured via the CLI arg is erroneously discarded. This is because I missed one spot where the `fileReporters` field needed to be copied over that is only used when resolving presets.

Repro:

1. Create a simple project with one test
2. Create a `dart_test.yaml` and define one preset that does _not_ configure any file reporters:

    ```
    presets:
      foo:
        reporter: expanded
    ```

3. Run `pub run test -P foo --file-reporter json:tests.json`
4. Verify that tests complete, but no `tests.json` file is produced

---

I didn't add a test because I couldn't find an existing one that covered all config fields. Maybe it would be useful to add one test that deals with merging configs and resolving presets and exercises every possible config option?


evanweible-wf:
@grouma thanks for getting this merged! Would it be possible to get a release of `test_core` with this in it?

PR #1219, opened 2020-04-29
Prep to publish

grouma:


pkgs/test/CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
-## 1.14.3-dev
+## 1.14.3
 
 * Fix an issue where coverage tests could not run in Chrome headless. 
 * Fix an issue where coverage collection would not work with source
   maps that contained absolute file URIs.
+* Update `test_core` to `0.3.4`.
```
natebosch, line 6: The change in `test_core` is user facing. Should we copy it here?

grouma: Done.

PR #1220, opened 2020-04-30
Updates to support SDK v2.9-

kevmoo:


pkgs/test/test/runner/load_suite_test.dart
```diff
@@ -66,7 +66,7 @@ void main() {
   test('a load test forwards errors and completes LoadSuite.suite to null',
       () async {
     var suite = LoadSuite('name', SuiteConfiguration.empty, suitePlatform, () {
-      fail('error');
+      return fail('error');
```
kevmoo, line 69: Got an analyzer error saying a return value was required...hrm...

pkgs/test_core/lib/src/util/io.dart
```diff
@@ -43,7 +43,7 @@ final String sdkDir = p.dirname(p.dirname(Platform.resolvedExecutable));
 final OperatingSystem currentOS = (() {
   var name = Platform.operatingSystem;
   var os = OperatingSystem.findByIoName(name);
-  if (os == OperatingSystem.none) return os;
+  if (os != OperatingSystem.none) return os;
```
grouma, line 46: Is this right?

pkgs/test_api/lib/src/frontend/future_matchers.dart
```diff
@@ -66,7 +66,7 @@ class _Completes extends AsyncMatcher {
 
       var buffer = StringBuffer();
       buffer.writeln(indent(prettyPrint(value), first: 'emitted '));
-      if (result!.isNotEmpty) buffer.writeln(indent(result, first: '  which '));
+      if (result.isNotEmpty) buffer.writeln(indent(result, first: '  which '));
```
natebosch, line 69: Do CFE and analyzer match with their flow analysis here?

pkgs/test_core/lib/src/util/io.dart
natebosch: It doesn't look right to me.

Why did this need to change? Was there a static error here?

kevmoo: If the os is NOT the `none` os, return it.

I think this was a bug in the original migration.

pkgs/test_api/lib/src/frontend/future_matchers.dart
kevmoo: yup! w/ 2.9

pkgs/test_api/lib/src/frontend/stream_matcher.dart
```diff
@@ -179,7 +179,7 @@ class _StreamMatcher extends AsyncMatcher implements StreamMatcher {
       buffer.writeln(indent(eventsString, first: 'emitted '));
       if (result.isNotEmpty) buffer.writeln(indent(result, first: '  which '));
       return buffer.toString().trimRight();
-    }, onError: (error) {
+    }, onError: (Object error) {
```
kevmoo, line 182: this is weird. You'd think `Object` would be inferred here!

natebosch: The `onError` argument has a type of `Function`. In the _ideal_ world we'd have union types so we could type it as `void Function(Object)|void Function(Object, StackTrace)`. For now we don't get _any_ inference for argument types, and not static checking for argument count.

pkgs/test_core/lib/src/util/io.dart
natebosch: hmm, it does look like there was a bug in the migration, but there is _also_ a bug in the non-migrated code.

From what I can tell, `OperatingSystem.findByIoName` never returned null.

We might need to dig a bit deeper on this one.

natebosch: https://github.com/dart-lang/test/pull/1221

kevmoo: resolved

kevmoo:
PTAL!

pkgs/test/lib/src/runner/browser/chrome.dart
```diff
@@ -121,13 +121,16 @@ class Chrome extends Browser {
...
+    if (sourceUri.scheme == 'file') return sourceUri;
+    // If the provided sourceUrl is relative, determine the package path.
     var uri = Uri.parse(script);
     var path = p.join(
         p.joinAll(uri.pathSegments.sublist(1, uri.pathSegments.length - 1)),
         sourceUrl);
     return path.contains('/packages/')
         ? Uri(scheme: 'package', path: path.split('/packages/').last)
-        : Uri.file(p.absolute(path));
+        : null;
```
grouma, line 133: This doesn't seem right?

kevmoo: It's a merge from **your** commit! e3bf6f6

pkgs/test_api/lib/src/frontend/stream_matcher.dart
kevmoo: yup

pkgs/test/lib/src/runner/browser/chrome.dart
natebosch: We shouldn't be doing merges from master in PRs we we rebase + squash them...

kevmoo: uh...so...land my fixes then do the merge in another PR?

kevmoo: @natebosch –  I was trying to bring in your fix

kevmoo: did the merge then added my fixes on top. PTAL

pkgs/test/test/runner/load_suite_test.dart
jakemac53: file an issue?

PR #1221, opened 2020-04-30
Simplify currentOS initialization

natebosch:
The previous code was working around the possibility of getting a `null`
OS, that that is never returned. Instead `OperatingSystem.none` is used.
Since we can't hit the null branch and exception, the entire initializer
can simplify to the static method call. This will be more clearly
correct after the null safety migration.


natebosch:
https://github.com/dart-lang/test/blob/a0d31b450d8b9bdf3db3a243f913794772662766/pkgs/test_api/lib/src/backend/operating_system.dart#L52-L67

Issue #1228, opened 2020-05-15
Printing to stdout when using runner should be configurable


fodil-a:
Hi,

**When**: I use Runner.run(), 
**I want**: Runner reporters to print into a custom output stream instead of stdout
**Current behavior**: it prints to stdout all the time

fodil-a:
I already pinpointed this issue:

https://github.com/dart-lang/test/blob/7514d98b420f15e577e52606bd85e79ffc0428cd/pkgs/test_core/lib/src/runner.dart#L91-L97

I believe that this behavior is heavily used.
So I think that we should have a different factory, which would take an IOSink as the wanted outputStream.
This outputStream won't be owned by Runner, so Runner won't have to manage it.

natebosch:
It looks like for your specific use case you should be able to use `--file-reporter` argument to ask for the json output to a file while getting the default reporter on stdout.

If you specifically need to have _no_ reporter writing to stdout I think we could create a no-op reporter. Please let us know if that's the case, otherwise I'll close this for now and focus on the immediate issue of interrupting the json output from #1226

PR #1230, opened 2020-05-15
Printing to stdout when using runner should be configurable (#1228)

fodil-a:


fodil-a:
I don't understand why the test doesn't pass on travis.
I tried to reproduce the failure on my computer, and I couldn't.
Any pointers ?

pkgs/test/test/runner/configuration/runner_output_test.dart
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+@TestOn('vm')
```
natebosch, line 7: This needs to be the first non-comment line in the file, above `import 'dart:io';`.

pkgs/test_core/lib/src/runner.dart
```diff
@@ -73,8 +73,14 @@ class Runner {
...
   final List<IOSink> _sinks;
 
-  /// Creates a new runner based on [configuration].
-  factory Runner(Configuration config) => config.asCurrent(() {
+  /// Creates a new runner based on [configuration] with stdout as the standard reporter.
+  factory Runner(Configuration config) =>
+      Runner.withCustomOutputStream(config, stdout);
+
+  /// Creates a new runner based on [configuration] with a custom standard reporter.
+  factory Runner.withCustomOutputStream(
```
natebosch, line 81: How would you use this as a consumer of the test package?

fodil-a: It is used like this:
```dart
  var inn = File('exec.s').openRead();
  var out = File('exec.s').openWrite();
  var process = exec.Runner.withCustomOutputStream(c, out);
  await process.run();
  await out.close();
  inn.listen((event) {
    // processing events
  });
```

pkgs/test/test/runner/configuration/runner_output_test.dart
fodil-a: I pushed a new version, with the @TestOn on the top

pkgs/test_core/lib/src/runner.dart
grouma: The logic contained under `lib/src` is intended to be private to the package. Consumers of this package should not have access to these APIs.

fodil-a: Thanks for your input ! 
Is it fine if I expose class that controls Runner ?

natebosch: The supported way to write a reporter output is with `--file-reporter`. For instance to write the json reporter to a file called `report` `pub run test --file-reporter json:report`.

We don't want to expose this class.

fodil-a: What I want to achieve is communicate with the tests, not parsing a file.
Doing so is quite easy with the Runner.
I can understand why you don't want to expose the Runner, but using `pub run test --file-reporter` is not solving my problem.
Solving this problem would enable me to push the tool I coded, which would be a good contribution to the dart community.

natebosch: An API to run tests from Dart code isn't something we currently support or plan to support in the short term. Unfortunately there are too many risks with exposing an API surface area that we may be forced to change in the future. In the very long term we may be able to evolve `test_core` in that direction, but it's not on our roadmap.

fodil-a: First, I want to thank you for dedicating time explaining the reasons and doing this review.
I thought the same, exposing it IS complicated compared to the gains, so I totally understand you.
Feel free to close the MR, I don't think this will go your way, at least for now.
Since I really want to release my tool, I'll just fork the test repo and use it for my project, assuming this risk for as long as necessary, and will merge back to this repository the day it decides to expose an API.

jakemac53:
Closing this as it looks like that was the decision we reached

Issue #1223, opened 2020-05-04
Short url for json_reporter.md is outdated


simolus3:
This line shows up when running `pub run test --help`:

https://github.com/dart-lang/test/blob/17850e02e6359ce9fc9349aa769521a2e84b22c8/pkgs/test_core/lib/src/runner/configuration/reporters.dart#L42

However, https://goo.gl/gBsV1a points to https://github.com/dart-lang/test/blob/master/doc/json_reporter.md which doesn't exist. I think it should point to https://github.com/dart-lang/test/blob/master/pkgs/test/doc/json_reporter.md instead.

grouma:
I assume @nex3 owns the shortened URl. Is this something you can quickly fix? Otherwise we'll need to land a PR and migrate to a different URL shortening tool.

nex3:
As far as I know those short links are mutable.

PR #1231, opened 2020-05-15
Update URL to point to correct documentation

grouma:
Closes https://github.com/dart-lang/test/issues/1223


grouma:
Merging on red. We'll need to look into the failing test which is impacting HEAD.

PR #1232, opened 2020-05-15
Updates to get tests running with correct weak/strong checks

kevmoo:
...in bleeding edge SDK

### DIFF
--- a/pkgs/test_core/lib/src/runner/plugin/remote_platform_helpers.dart
+++ b/pkgs/test_core/lib/src/runner/plugin/remote_platform_helpers.dart
@@ -30,10 +30,13 @@ import 'package:test_api/src/suite_channel_manager.dart'; // ignore: implementat
 ///
 /// If [beforeLoad] is passed, it's called before the tests have been declared
 /// for this worker.
-StreamChannel serializeSuite(Function Function() getMain,
+StreamChannel<Object> serializeSuite(Function Function() getMain,
         {bool hidePrints = true, Future Function()? beforeLoad}) =>
-    RemoteListener.start(getMain,
-        hidePrints: hidePrints, beforeLoad: beforeLoad);
+    RemoteListener.start(
+      getMain,
+      hidePrints: hidePrints,
+      beforeLoad: beforeLoad,
+    );
 
 /// Returns a channel that communicates with a plugin in the test runner.
 ///
--- a/pkgs/test_api/lib/src/remote_listener.dart
+++ b/pkgs/test_api/lib/src/remote_listener.dart
@@ -44,7 +44,7 @@ class RemoteListener {
   ///
   /// If [beforeLoad] is passed, it's called before the tests have been declared
   /// for this worker.
-  static StreamChannel start(Function Function() getMain,
+  static StreamChannel<Object> start(Function Function() getMain,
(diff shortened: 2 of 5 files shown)


pkgs/test_core/lib/src/runner/plugin/remote_platform_helpers.dart
```diff
@@ -30,10 +30,16 @@ import 'package:test_api/src/suite_channel_manager.dart'; // ignore: implementat
...
-    RemoteListener.start(getMain,
-        hidePrints: hidePrints, beforeLoad: beforeLoad);
+    RemoteListener.start(
+      getMain,
+      hidePrints: hidePrints,
+      beforeLoad: beforeLoad,
+    )
+    // TODO: this cast should not be needed
+    // and yet it fails at SDK dc1d7c87205 without it
+    as StreamChannel<Object>;
```
kevmoo, line 42: This is a head scratcher, but absolutely necessary to get things working!

natebosch: What do you mean by "fails"? At runtime? Statically?

kevmoo:
```
  Failed to load "test/isolate_channel_test.dart":
  Unable to spawn isolate: ../test/pkgs/test_core/lib/src/runner/plugin/remote_platform_helpers.dart:35:20: Error: A value of type 'StreamChannel<dynamic>' can't be assigned to a variable of type 'StreamChannel<Object>'.
   - 'StreamChannel' is from 'package:stream_channel/stream_channel.dart' ('lib/stream_channel.dart').
   - 'Object' is from 'dart:core'.
      RemoteListener.start(getMain,
                     ^
```

kevmoo: statically!

natebosch: Is one of these libraries not opted in, and the other is?

kevmoo: I'm running w/ the bleeding SDK bits that won't let things be mixed – both are opted-in

kevmoo:
@natebosch – LGTY?

natebosch:
> LGTY?

No, I don't think we should be submitting workarounds until we understand them and have issues filed in the SDK.

kevmoo:
> > LGTY?
> 
> No, I don't think we should be submitting workarounds until we understand them and have issues filed in the SDK.

I'll get going on the SDK issue.

pkgs/test_core/lib/src/runner/plugin/remote_platform_helpers.dart
kevmoo: Turns out this **is not** needed. User error...

kevmoo:
> > > LGTY?
> > 
> > 
> > No, I don't think we should be submitting workarounds until we understand them and have issues filed in the SDK.
> 
> I'll get going on the SDK issue.

Not sure what I did wrong, but it turns out the cast is NOT needed.

PTAL, @natebosch

kevmoo:
@jakemac53 – this is much less scary now

Issue #1236, opened 2020-05-18
null safety: test wrapper logic must mirror language version of wrapped _test.dart file


kevmoo:
https://github.com/dart-lang/test/blob/0022fbd1be5faf373986369a050d6b2a088f71b7/pkgs/test_core/lib/src/runner/vm/platform.dart#L128-L143

If the wrapped `_test.dart` file sets a language version, it should be mirrored in the generated data URI so the null-safe opt-in logic is preserved.

jakemac53:
This is done in the null_safety branch now

PR #1240, opened 2020-05-19
Add language version comments to bootstrap files

jakemac53:
- Adds a nullable `languageVersionComment` field to `Metadata`
- Adds a utility to get a language comment that matches the root packages version
- Adds explicit language version comments to all bootstrap files ensuring the versions matches that of the test (note this assumes tests live in the root package)
- Clean up the handling of serving bootstrap dart files in the browser platform which had diverged and was serving the wrong content.

Closes https://github.com/dart-lang/test/issues/1236

I am not sure of the proper way to test this exactly, suggestions welcome. I manually tested the vm stuff and it seems to work correctly in terms of opting in/out of strong mode.

### DIFF
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -707,7 +707,7 @@ void main(List<String> args) async {
   group('nnbd', () {
     final _testContents = '''
 import 'package:test/test.dart';
-import 'weak.dart';
+import 'opted_out.dart';
 
 void main() {
   test("success", () {
@@ -716,12 +716,13 @@ void main() {
 }''';
 
     setUp(() async {
-      await d.file('weak.dart', '''
+      await d.file('opted_out.dart', '''
 // @dart=2.8
 final foo = true;''').create();
     });
 
-    test('strong mode is used if the entrypoint opts in explicitly', () async {
+    test('sound null safety is enabled if the entrypoint opts in explicitly',
+        () async {
       await d.file('test.dart', '''
 // @dart=2.9
 $_testContents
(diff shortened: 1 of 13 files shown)


jakemac53:
Ok I pushed some tests here, I opted for integration tests that rely on the CFE error about having non-opted in libraries in strong mode which seems pretty reliable long term.

jakemac53:
Note that a lot of other tests are failing now, I am going to start looking into that. I think its mostly integration tests getting opted in due to not having a language version default.

pkgs/test/test/runner/runner_test.dart
```diff
@@ -698,4 +703,108 @@ void main(List<String> args) async {
...
+import 'weak.dart';
+
+void main() {
+  test("success", () {
+    expect(foo, true);
+  });
+}''';
+
+    setUp(() async {
+      await d.file('weak.dart', '''
```
natebosch, line 719: [optional] we're changing terminology to "sound null safety" and "no sound null safety". Consider using `unsound.dart` here. It's not likely to confuse us, but there also isn't any reason not to go for consistency when possible.

pkgs/test/test/runner/runner_test.dart
```diff
@@ -698,4 +703,108 @@ void main(List<String> args) async {
...
+// @dart=2.8
+final foo = true;''').create();
+    });
+
+    test('strong mode is used if the entrypoint opts in explicitly', () async {
```
natebosch, line 724: [optional] "sound null safety is used if the entrypoint"

pkgs/test/test/runner/runner_test.dart
```diff
@@ -698,4 +703,108 @@ void main(List<String> args) async {
...
+// @dart=2.9
+$_testContents
+''').create();
+      var test = await runTest(['test.dart']);
+
+      expect(
+          test.stdout,
+          containsInOrder([
+            'Unable to spawn isolate:',
+            'Error: A library can\'t opt out of non-nullable by default, when in nnbd-strong mode.'
```
natebosch, line 735: Is this our message, or is it coming from somewhere else?

cc @kevmoo - how important is it that our user facing messages use "sound" instead of "strong"?

pkgs/test/test/runner/runner_test.dart
jakemac53: Changed it to opted_out.dart, and the test names are changed to use this terminology as well.

pkgs/test/test/runner/runner_test.dart
jakemac53: This is likely originating from the CFE.

PR #1242, opened 2020-05-20
Pass experiment flags to dart2js

jakemac53:
This starts getting web tests compiling through the test runner, if used in combination with https://github.com/dart-lang/test/pull/1240.

However I see some late initialization errors that need to be tracked down still, can follow up with that later.

### DIFF
--- a/pkgs/test_core/lib/src/runner/compiler_pool.dart
+++ b/pkgs/test_core/lib/src/runner/compiler_pool.dart
@@ -152,7 +152,7 @@ final List<String> _enabledExperiments = () {
     if (arg == '--enable-experiment') {
       if (!itr.moveNext()) break;
       experiments.add(itr.current);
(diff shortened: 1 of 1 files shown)


pkgs/test_core/lib/src/runner/compiler_pool.dart
```diff
@@ -139,3 +141,23 @@ class CompilerPool {
...
+/// [Platform.executableArguments].
+final List<String> _enabledExperiments = () {
+  var experiments = <String>[];
+  var itr = Platform.executableArguments.iterator;
+  while (itr.moveNext()) {
+    var arg = itr.current;
+    if (arg == '--enable-experiment') {
+      if (!itr.moveNext()) break;
+      experiments.add(itr.current);
+    } else if (arg.startsWith('--enable-experiment')) {
```
natebosch, line 155: [nit] add the `=` to the `startsWith` argument.

PR #1244, opened 2020-05-20
add explicit types on the lhs to override inference

jakemac53:
Inference from legacy types defaults to the non-nullable type.  Add explicit types on the LHS to make these be non-nullable as they actually are.

Note that explicit casts on the RHS currently trip up the unnecessary cast lint, which I will file a separate issue about.

### DIFF
--- a/pkgs/test_core/lib/backend.dart
+++ b/pkgs/test_core/lib/backend.dart
@@ -1,6 +1,8 @@
 // Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+//
+// @dart=2.8
 
 @Deprecated('package:test_core is not intended for general use. '
     'Please use package:test.')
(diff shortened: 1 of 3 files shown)


jakemac53:
https://github.com/dart-lang/linter/issues/2101 is the linter issue

natebosch:
Can we instead opt these libraries out of the new language version?

jakemac53:
Sure, I can do that instead, it should make us less likely to break later

jakemac53:
PTAL @natebosch ... had to revert a lot of nice things 😭 

Note that the top level `backend.dart` is now also explicitly opted out since it exports a function from an opted out library.

kevmoo:
@jakemac53 have you tried running tests in another package w/ the experiment turned on?

natebosch:
> had to revert a lot of nice things 😭

We'll get them back eventually 😛

jakemac53:
> @jakemac53 have you tried running tests in another package w/ the experiment turned on?

No, only from the test package so far

PR #1245, opened 2020-05-21
Fail with specific error for null runnerSuite

natebosch:
See https://github.com/flutter/flutter/issues/56457

If we get a null `runnerSuite` the attempt to read the `environment`
file will fail, resulting in a confusing error and stack trace. We can't
reproduce this locally so it isn't clear where the null may be coming
from.

Add additional information to the error thrown to try to narrow down the
places we need to search for a root cause.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.14.5
+
+* Add additional information to an exception when we end up with a null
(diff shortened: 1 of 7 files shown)


natebosch:
I'm not sure how to test this, since we don't know how the `null` is flowing through to this point.

jonahwilliams:
Thanks @natebosch !

PR #1246, opened 2020-05-22
Bump test_api dep from test

natebosch:
This was missed when I bumped the dependency on `test_core` which
requires a new `test_api`. Solves an impossible pub get.


natebosch:
I missed this in https://github.com/dart-lang/test/pull/1245

The other 2 packages are published.

PR #1250, opened 2020-05-28
expose the ability to load test configuration files from the test_core package

jakemac53:
* Expose the `Configuration` class and related classes.
* Expose the `load` and `loadFromString` methods for loading a `Configuration`
  from yaml.

Towards https://github.com/dart-lang/build/issues/2684

### DIFF
--- a/pkgs/test_core/lib/backend.dart
+++ b/pkgs/test_core/lib/backend.dart
@@ -11,3 +11,10 @@ export 'package:test_api/backend.dart'
     show Metadata, PlatformSelector, Runtime, SuitePlatform;
 
 export 'src/runner/parse_metadata.dart' show parseMetadata;
+
+export 'src/runner/configuration/load.dart' show load, loadFromString;
+export 'src/runner/suite.dart' show SuiteConfiguration;
+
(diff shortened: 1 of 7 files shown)


pkgs/test_core/lib/backend.dart
```diff
@@ -11,3 +11,10 @@ export 'package:test_api/backend.dart'
     show Metadata, PlatformSelector, Runtime, SuitePlatform;
 
 export 'src/runner/parse_metadata.dart' show parseMetadata;
+
+export 'src/runner/configuration/load.dart' show load, loadFromString;
+export 'src/runner/suite.dart' show SuiteConfiguration;
+
+export 'src/runner/configuration.dart' show Configuration;
```
natebosch, line 18: `configuration.dart` already has a transitive import to `dart:io`, can we add `Configuration.load` and `Configuration.loadFromString` static methods (or factory constructors) to avoid the overly broad top level `load` name?

We can keep the implementation as is for now and add something small to forward them so it's not a large refactor.

jakemac53: Sure, sgtm

PR #1256, opened 2020-06-09
Get tests running on travis

jakemac53:
Note that some are failing still, my plan is to continue pushing fixes here as needed.

Specifically I know the node tests are failing.

Note that the web tests will be failing here also until https://dart-review.googlesource.com/c/sdk/+/150664 is released.

### DIFF
--- a/pkgs/test_api/lib/src/remote_listener.dart
+++ b/pkgs/test_api/lib/src/remote_listener.dart
@@ -67,7 +67,7 @@ class RemoteListener {
 
     SuiteChannelManager().asCurrent(() {
       StackTraceFormatter().asCurrent(() {
-        runZoned(() async {
+        runZonedGuarded(() async {
           Function? main;
           try {
             main = getMain();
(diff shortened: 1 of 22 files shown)


pkgs/test_api/lib/src/remote_listener.dart
```diff
@@ -135,7 +134,7 @@ class RemoteListener {
               // useful errors when calling `test()` and `group()` within a test,
               // and so they can add to the declarer's `tearDownAll()` list.
               zoneValues: {#test.declarer: declarer});
-        }, onError: (error, StackTrace stackTrace) {
+        }, (Object error, StackTrace stackTrace) {
```
natebosch, line 137: [nit] we can remove the argument types here now - one of the benefits of `runZonedGuarded` is we have an _actual_ function type with arguments.
```suggestion
        }, (error, stackTrace) {
```

natebosch:
@googlebot I fixed it.

PR #1258, opened 2020-06-11
fix precompiled and json reporter tests

jakemac53:
After this the only remaining tests that are failing (other than node tests that I can't run locally) are the hybrid tests.

### DIFF
--- a/pkgs/test/test/runner/json_reporter_test.dart
+++ b/pkgs/test/test/runner/json_reporter_test.dart
@@ -543,7 +543,7 @@ import 'package:test/test.dart';
 void customTest(String name, dynamic Function() testFn) => test(name, testFn);
 ''',
           });
-    }, solo: true);
+    });
   });
 
   test(
--- a/pkgs/test/test/runner/json_reporter_test.dart
+++ b/pkgs/test/test/runner/json_reporter_test.dart
@@ -525,7 +525,7 @@ void main() {
               groupJson(2, testCount: 2),
               testStartJson(3, 'success 1',
                   line: 3,
-                  column: 50,
+                  column: 60,
                   url: p.toUri(p.join(d.sandbox, 'common.dart')).toString(),
                   root_column: 7,
                   root_line: 7,
@@ -540,10 +540,10 @@ void main() {
             'common.dart': '''
 import 'package:test/test.dart';
 
-void customTest(String name, Function testFn) => test(name, testFn);
+void customTest(String name, dynamic Function() testFn) => test(name, testFn);
 ''',
           });
-    });
+    }, solo: true);
   });
 
   test(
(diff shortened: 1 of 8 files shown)


pkgs/test/test/runner/json_reporter_test.dart
```diff
@@ -540,10 +540,10 @@ void main() {
             'common.dart': '''
 import 'package:test/test.dart';
 
-void customTest(String name, Function testFn) => test(name, testFn);
+void customTest(String name, dynamic Function() testFn) => test(name, testFn);
 ''',
           });
-    });
+    }, solo: true);
```
natebosch, line 546: why do we need this?

pkgs/test_core/lib/src/runner/reporter/json.dart
```diff
@@ -178,7 +178,8 @@ class JsonReporter implements Reporter {
     // different metadata.
     if (suite is LoadSuite) {
       suite.suite.then((runnerSuite) {
-        _suiteIDs[runnerSuite!] = id;
+        if (runnerSuite == null) return;
```
natebosch, line 181: oops, looks like I never merged this back to the `null_safety` branch.

I did not think this was a situation we wanted to support.

https://github.com/dart-lang/test/pull/1245

Do you have an understanding of how this shows up as null here?

jakemac53: I do not, no

pkgs/test/test/runner/json_reporter_test.dart
jakemac53: whoops! good thing I sent out a pr lol

pkgs/test_core/lib/src/runner/reporter/json.dart
jakemac53: From the docs:

```
/// A future that completes to the loaded suite once the suite's test has been
  /// run and completed successfully.
  ///
  /// This will return `null` if the suite is unavailable for some reason (for
  /// example if an error occurred while loading it).
```

pkgs/test/test/runner/precompiled_test.dart
```diff
@@ -235,6 +236,7 @@ Future<Null> _writePackagesFile() async {
   // TODO: remove try/catch when this issue is resolved:
   // https://github.com/dart-lang/package_config/issues/66
   try {
+    await Directory(p.join(d.sandbox, '.dart_tool')).create();
```
natebosch, line 239: I think this is more idiomatic as

```suggestion
    await d.dir('.dart_tool').create();
```

Although that also means that `savePackageConfig` should probably be doing something similar... We can do it in a followup.

jakemac53: I did file an issue on package_config about this

PR #1266, opened 2020-06-16
Remove dependency on `http_multi_socket`.

### LINKED ISSUE #1264: Remove usage of `package:multi_server_socket`
It's only used by the node platform and can probably be instead something which only listens on either IPv6 or IPv4 depending on the first thing that works.

natebosch:
Closes #1264

Listening on both IPv6 and IPv4 allows for the case where `node` only
supports IPv4 but the OS supports IPv6, however this is not a situation
that should come up in the real world.

Bind to IPv6 and fall back to IPv4 if it doesn't work. The node side
calls `net.connect(port)` which should support either assuming OS
support.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.14.8-dev
+
+* The Node platform will now communicate over only IPv6 if it is available.
+
 ## 1.14.7
 
 * Support the latest `package:coverage`.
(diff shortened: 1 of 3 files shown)


natebosch:
The connect call happens at 

https://github.com/dart-lang/test/blob/f811df70f79a2398c9328e4faf6e5c1db86da81f/pkgs/test/lib/src/runner/node/socket_channel.dart#L36

I don't have a local node to verify, but Travis should catch any issues.

Issue #1255, opened 2020-06-09
`pub run --enable-experiment=non-nullable test` fails with unmigrated package


kevmoo:
w/ pkg:path, fake_async, and stream_transform

LOTs of warnings. Then this error

```
00:01 +0 -1: loading test/fake_async_test.dart [E]
  Failed to load "test/fake_async_test.dart":
  Unable to spawn isolate: ../../.pub-cache/hosted/pub.dartlang.org/stream_channel-2.0.0/lib/isolate_channel.dart:1:1: Error: A library can't opt out of non-nullable by default, when in nnbd-strong mode.
  // Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
  ^
  ../../.pub-cache/hosted/pub.dartlang.org/test_core-0.3.7/lib/src/runner/plugin/remote_platform_helpers.dart:1:1: Error: A library can't opt out of non-nullable by default, when in nnbd-strong mode.
  // Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
  ^
  test/fake_async_test.dart:1:1: Error: A library can't opt out of non-nullable by default, when in nnbd-strong mode.
  // Copyright 2014 Google Inc. All Rights Reserved.
  ^
  ../../.pub-cache/hosted/pub.dartlang.org/stream_channel-2.0.0/lib/src/isolate_channel.dart:1:1: Error: A library can't opt out of non-nullable by default, when in nnbd-strong mode.
  // Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
  ^
  ../../.pub-cache/hosted/pub.dartlang.org/stream_channel-2.0.0/lib/stream_channel.dart:1:1: Error: A library can't opt out of non-nullable by default, when in nnbd-strong mode.
  // Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
  ^
  ../../.pub-cache/hosted/pub.dartlang.org/test_api-0.2.16/lib/src/backend/stack_trace_formatter.dart:1:1: Error: A library can't opt out of non-nullable by default, when in nnbd-strong mode.
  // Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
  ^
  ../../.pub-cache/hosted/pub.dartlang.org/test_api-0.2.16/lib/src/util/stack_trace_mapper.dart:1:1: Error: A library can't opt out of non-nullable by default, when in nnbd-strong mode.
[...]

PR #1267, opened 2020-06-16
Add language version comments to bootstrap files (#1240)

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1255

Backport of the null_safety commit to support this https://github.com/dart-lang/test/commit/5312b230d7100bc7bdbf98b2b9ad500177f74185

- Adds an optional `languageVersionComment` field to `Metadata`
- Adds a utility to get a language comment that matches the root packages version
- Adds explicit language version comments to all bootstrap files ensuring the versions matches that of the test (note this assumes tests live in the root package)
- Clean up the handling of serving bootstrap dart files in the browser platform which had diverged and was serving the wrong content.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.14.8
+
+* Update browser/node bootstrapping logic to ensure the bootstrap library has
+  the same language version as the test.
+
 ## 1.14.7
 
 * Support the latest `package:coverage`.
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 * Update bootstrapping logic to ensure the bootstrap library has
   the same language version as the test.
-* Add `languageVersionComment` on the `MetaData` class. This should only be
-  present for test suites.
 
 ## 1.14.7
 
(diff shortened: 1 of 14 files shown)


pkgs/test/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.14.8
```
natebosch, line 1: This should be `1.15.0`

pkgs/test/test/runner/runner_test.dart
```diff
@@ -696,4 +701,82 @@ void main(List<String> args) async {
...
+bool? foo;''').create();
+    });
+
+    test('nnbd can be enabled in deps', () async {
+      await d.file('test.dart', '''
+// @dart=2.8
+$_testContents''').create();
+      var test = await runTest(['test.dart'],
+          packageConfig: (await Isolate.packageConfig).path,
+          vmArgs: ['--enable-experiment=non-nullable']);
```
natebosch, line 729: Do we need to set the min SDK for this?

jakemac53: we will see!

pkgs/test/CHANGELOG.md
jakemac53: done

pkgs/test/test/runner/runner_test.dart
jakemac53: looks fine

PR #1273, opened 2020-06-17
Ignore a null runner suite in json reporter

natebosch:
The `StateError` had been put in with #1245 in order to facilitate
debugging this problem when it surfaces sometimes for flutter.

In #1258 Jake pointed out that there is a comment which mentions that
the `runnerSuite` may be null if it is unavailable due to a load error.

I'm still not sure exactly the conditions that lead to this behavior,
but it should ideally come with some other more clear error, so ignoring
it here should hopefully make things less confusing.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.15.1-dev
+
+* Avoid a confusing stack trace when there is a problem loading a platform when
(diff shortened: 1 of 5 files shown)


natebosch:
cc @jonahwilliams - this may just solve the problem on the flutter side...

jonahwilliams:
Yay!

PR #1274, opened 2020-06-18
Run tests in declaring zone

natebosch:
Three problems (at least) surfaced by using the Zone from when main was
run rather than the zone when the runner is asking the invoker to run:

1 The heartbeat stuff could get stuck by a FakeAsync zone when the users
  calls `test()`. It wouldn't let the timer run.
2 The communication channel for "remote" tests can't find the channel
  for the specific test that it should be speaking over.
3 We expect an Invoker.guard() to funnel the errors to the correct test
  case. If they surface in the same zone as we ran `main()` it looks
  like a test error.

For 1 I _think_ we can run the timer in the root zone since that should
always have sane async behavior. The invoker zone is necessarily a child
of it anyway.

For 2 With some interface changes we can pass the important zone
variable through rather than setting it on a zone in the remote
listener. There might also be a better overall structure that would make
this less awkward.

For 3 it seems sufficient to always "guard" the tests so that uncaught
errors show up for the right invoker, but there is some conversation
about letting errors (probably `Future.error`s...) cross from `setUpAll`
to `tearDownAll` without getting stuck in an error Zone. It _looks_ like
`Invoker.guard` should already handle some of that.

### DIFF
--- a/pkgs/test_api/lib/src/backend/invoker.dart
+++ b/pkgs/test_api/lib/src/backend/invoker.dart
@@ -39,6 +39,9 @@ class LocalTest extends Test {
   /// Whether the test is run in its own error zone.
   final bool _guarded;
 
+  /// The Zone that the [_body] callback was defined in.
+  final Zone _declaringZone;
+
   /// Creates a new [LocalTest].
   ///
   /// If [guarded] is `true`, the test is run in its own error zone, and any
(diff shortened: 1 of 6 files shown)


mit-mit:
@natebosch can this be closed?

mit-mit:
We could certainly consider that. Is it usual to have a draft like this which is more than a year old?

jakemac53:
I wouldn't say it is super common, no. But I don't personally see an issue with it either? Here is another example https://github.com/dart-lang/build/pull/2779. Also probably stale enough that it could/should be closed.

But in any case an old draft PR is more an indication of something exploratory that was abandoned. They generally tend to hang around longer, and I don't think they should ding the overall score?

natebosch:
We are skipping draft PRs for the staleness metric, but not the total PRs metric. @mit-mit - this is the same category of difference between our issue count and issue staleness, where we do some filtering for staleness.

PR #1279, opened 2020-06-18
Add back multi server for node platform

### LINKED ISSUE #1278: Node tests fail on macOS
Dart Sass is failing when running tests in Node on macOS starting with version 1.15.0.

From [the error](https://travis-ci.org/github/sass/dart-sass/jobs/699863268), it seems like this is likely related to the IPv6 change mentioned in the changelog, but I don't have a Mac available to test on manually.

natebosch:
Fixes #1278

Copy the `_loopback` code from `package:multi_server_socket` but model
it as a `List<ServerSocket>` instead of a single server. Remove the
unnecessary arguments for handling anything other than port 0.

Use a `finally` block to ensure the socket servers are always closed.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Avoid a confusing stack trace when there is a problem loading a platform when
   using the JSON reporter and enabling debugging.
+* Restore behavior of listening for both `IPv6 and `IPv4` sockets for the node
+  platform.
 
 ## 1.15.0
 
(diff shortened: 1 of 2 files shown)


natebosch:
It looks like the error is hidden from us when it happens by https://github.com/dart-lang/test/blob/58383da027c600f78bb619730b927359b00659ca/pkgs/test_api/lib/src/backend/invoker.dart#L355-L357

I can't figure out why an error swallowed there ends up making the runner hang on Travis though, but it does seem to do that.

PR #1286, opened 2020-06-24
Add a `Fake` base class to the test API.

stereotype441:
This can be used by tests to create mock objects that are not expected
to be used; these can be useful when testing non-nullable APIs.

### DIFF
--- a/pkgs/test/test/util/fake_test.dart
+++ b/pkgs/test/test/util/fake_test.dart
@@ -1,0 +1,70 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore: deprecated_member_use
+@TestOn('vm')
+import 'package:test/test.dart';
+
+import '../utils.dart';
+
+void main() {
+  _FakeSample fake;
+  setUp(() {
+    fake = _FakeSample();
+  });
+  test('method invocation', () {
+    expect(
+        () => fake.f(),
+        throwsA(isTestFailure(
+            'Symbol("f") invoked on fake object of type _FakeSample')));
+  });
+  test('getter', () {
+    expect(
+        () => fake.x,
+        throwsA(isTestFailure(
+            'Symbol("x") invoked on fake object of type _FakeSample')));
+  });
+  test('setter', () {
+    expect(
+        () => fake.x = 0,
+        throwsA(isTestFailure(
+            'Symbol("x=") invoked on fake object of type _FakeSample')));
+  });
+  test('operator', () {
+    expect(
+        () => fake + 1,
+        throwsA(isTestFailure(
+            'Symbol("+") invoked on fake object of type _FakeSample')));
+  });
+  test('==', () {
+    expect(
+        () => fake == Object(),
+        throwsA(isTestFailure(
+            'Symbol("==") invoked on fake object of type _FakeSample')));
+  });
+  test('hashCode', () {
+    expect(
+        () => fake.hashCode,
+        throwsA(isTestFailure(
+            'Symbol("hashCode") invoked on fake object of type _FakeSample')));
+  });
+  test('runtimeType', () {
+    expect(
+        () => fake.runtimeType,
+        throwsA(isTestFailure(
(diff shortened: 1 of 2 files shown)


natebosch:
There is already a class called `Fake` in `mockito`.

I'm not sure if this is worth shipping in the test package, it does feel more closely related to mocking than testing in general...

pkgs/test/test/util/fake_test.dart
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
```
jakemac53, line 1:
```suggestion
// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```

pkgs/test/test/util/fake_test.dart
```diff
@@ -0,0 +1,70 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore: deprecated_member_use
```
jakemac53, line 5: Is this cruft? It shouldn't be necessary I don't think here

pkgs/test_core/lib/test_core.dart
```diff
@@ -285,3 +286,33 @@ void setUpAll(dynamic Function() callback) => _declarer.setUpAll(callback);
 /// prohibitively slow.
 void tearDownAll(dynamic Function() callback) =>
     _declarer.tearDownAll(callback);
+
+/// Base class that may be used by a test to create a mock object that should
```
jakemac53, line 290:
```suggestion
/// Base class that may be used by a test to create a fake object that should
```

pkgs/test_core/lib/test_core.dart
```diff
@@ -285,3 +286,33 @@ void setUpAll(dynamic Function() callback) => _declarer.setUpAll(callback);
...
+/// - It may be passed into a non-nullable API.
+/// - It satisfies `is` checks for the type it is mocking.
+/// - It throws an exception for any method invocation, whereas `null` supports
+///   toString(), hashCode, etc.
+class Fake {
+  const Fake();
+
+  @override
+  bool operator ==(Object other) =>
+      noSuchMethod(Invocation.method(#==, [other]));
```
jakemac53, line 305: We will need to fix the analyzer issues here, I think an explicit cast is likely fine.

natebosch:
> One question though is if this would be breaking for users of mockito today? Would they get a name conflict?

Yes they will.

pkgs/test_core/lib/test_core.dart
```diff
@@ -285,3 +286,33 @@ void setUpAll(dynamic Function() callback) => _declarer.setUpAll(callback);
...
+/// never be used.  For example:
+///
+///     class FakeFoo extends Fake implements Foo {}
+///
+/// The behavior of a fake object is similar to `null`, except that:
+/// - It may be passed into a non-nullable API.
+/// - It satisfies `is` checks for the type it is mocking.
+/// - It throws an exception for any method invocation, whereas `null` supports
+///   toString(), hashCode, etc.
+class Fake {
```
natebosch, line 300: If we decide to ship this it should be in `test_api`, and the tests should go there too.

natebosch:
We should check with someone from the `flutter_test` team about this plan too.

pkgs/test/test/util/fake_test.dart
stereotype441: You're right, this was a copy/paste error.  Fixed.

jakemac53:
cc @jonahwilliams (feel free to add somebody else if there is somebody better)

pkgs/test/test/util/fake_test.dart
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@TestOn('vm')
```
jakemac53, line 5: It probably makes sense to remove this actually, this class should be usable on all platforms

jonahwilliams:
I don't imagine this causing issues, fyi @goderbauer and @chunhtai

natebosch:
Superseded by https://github.com/dart-lang/test/pull/1304

PR #1287, opened 2020-06-25
Support Fuchsia in test_core executable.

chingjun:
Fuchsia doesn't have POSIX signals, and compons can't access stdin on
Fuchsia. Avoid accessing anything related to prevent test runner from
crashing for Fuchsia on-device tests.

### DIFF
--- a/pkgs/test_core/lib/src/executable.dart
+++ b/pkgs/test_core/lib/src/executable.dart
@@ -48,11 +48,7 @@ void completeShutdown() {
     signalSubscription = null;
   }
   isShutdown = true;
-
-  // Applications can't access stdin on Fuchsia.
-  if (!Platform.isFuchsia) {
-    stdinLines.cancel(immediate: true);
-  }
+  stdinLines.cancel(immediate: true);
 }
 
 Future<void> _execute(List<String> args) async {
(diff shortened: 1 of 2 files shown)


chingjun:
cc @natebosch

pkgs/test_core/lib/src/executable.dart
```diff
@@ -48,7 +48,11 @@ void completeShutdown() {
     signalSubscription = null;
   }
   isShutdown = true;
-  stdinLines.cancel(immediate: true);
+
+  // Applications can't access stdin on Fuchsia.
+  if (!Platform.isFuchsia) {
```
natebosch, line 53: After some further thought, what do you think about handling this case at the definitio of `stdInLines`.

https://github.com/dart-lang/test/blob/8716882291a62aeb0058aaf544549435439aed7e/pkgs/test_core/lib/src/util/io.dart#L68

We could do something like

```dart
final stdinLines = StreamQueue(
    Platform.isFuchsia ? Stream<String>.empty() : lineSplitter.bind(stdin));
```

As is I think we're relying on this `.cancel` call being the _only_ interaction with `stdin`, which won't help us if we hit one of the other code paths that tries to read it. I anticipate that an empty stream there could break us in other subtle ways, but probably safer anyway.

chingjun: Adding it in the definition of stdinLines is indeed a better idea. Updated. Thanks.

PR #1288, opened 2020-06-25
Stop holding the InvokerZone on the Invoker

natebosch:
The commented purpose and usage of this field are to avoid problems
where a `heartbeat` call is in a `Zone` which interferes with `Timer`
behavior. `Zone.root` should always have known behavior, or at the least
the `_invokerZone` would have always been a descendent of this root zone
anyway, so it shouldn't introduce any problems that weren't present
before.

### DIFF
--- a/pkgs/test_api/lib/src/backend/invoker.dart
+++ b/pkgs/test_api/lib/src/backend/invoker.dart
@@ -158,12 +158,6 @@ class Invoker {
         }
       }));
 
-  /// The zone that the top level of [_test.body] is running in.
-  ///
-  /// Tracking this ensures that [_timeoutTimer] isn't created in a
-  /// timer-mocking zone created by the test.
-  Zone _invokerZone;
-
   /// The timer for tracking timeouts.
   ///
   /// This will be `null` until the test starts running.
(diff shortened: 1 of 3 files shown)


natebosch:
Hmm, I suppose using the root zone makes this harder to test...

I'm not sure the best way to handle that.

grouma:
> Hmm, I suppose using the root zone makes this harder to test...

Could we create a timer-mocking zone in a test?

natebosch:
> Could we create a timer-mocking zone in a test?

That's what we're doing before, but we can't do that if the timer is created with `Zone.root` which is why this test starts to fail.

natebosch:
One option would be to just give this test a longer timeout and let it take 30 seconds to run. It's not a _great_ option, but it is probably the only way to test this behavior without being forced to keep a `Zone` available.

grouma:
The heartbeat is configurable right? If so, do we have to wait the entire 30 seconds? Can we use your suggested approach but with smaller durations?

natebosch:
> The heartbeat is configurable right?

This test is specifically for the default.

https://github.com/dart-lang/test/blob/062a845da04fe20a6bf3e4596ade20a3b1d5f422/pkgs/test_api/test/backend/invoker_test.dart#L428-L449

natebosch:
I'm just going to drop the test in favor of simpler code around zones here.

natebosch:
This should be ready to go - @grouma PTAL

PR #1290, opened 2020-07-01
Fix error when meta add unawaited

natebosch:
We have a definition of `unawaited` which is visible from of of the
`test_api/src/` imports to avoid a transitive dep on `pedeantic`. This
causes a static error onces `meta` has a method with the same name.
`pedantic` will be updated to export this method and `test_api` should
do the same until we migrate all usages to the definition from `meta`.
We can't migrate directly to the definition from `meta` since the
version which adds this has a lower SDK bound of `2.9.0` and we aren't
prepared to set our lower bound that high, especially since that SDK is
not yet published. An explicit show on the import of `meta` is backwards
and forward compatible.

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 0.3.10-dev
+## 0.3.10
+
+* Prepare for `unawaited` from `package:meta`.
 
 ## 0.3.9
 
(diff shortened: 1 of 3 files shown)


natebosch:
cc @bwilkerson

jakemac53:
why does meta have such a high sdk constraint?

PR #1297, opened 2020-07-13
Fix Travis failures

natebosch:


natebosch:
@jakemac53 Any hunch as to why this test wasn't failing before?

```
17:44 +102 -1: test/runner/runner_test.dart: fails gracefully if a test file has a non-function main [E]                                                                                               
  Expected: should do the following in order:
            • eventually emit an event that contains '-1: loading test.dart [E]'
            • eventually emit an event that contains 'A value of type \'int\' can\'t be assigned to a variable of type \'Function\''
    Actual: <Instance of 'StreamQueue<String>'>
     Which: emitted • 00:00 +0 -1: loading test.dart [E]
                    •   Failed to load "test.dart":
                    •   Unable to spawn isolate: test.dart:1:5: Error: Field 'main' should be initialized because its type 'int' doesn't allow null.
                    •   int main;
                    •       ^^^^
                    •   data:application/dart;charset=utf-8,%20%20%20%20%0A%20%20%20%20import%20%22dart:isolate%22;%0A%0A%20%20%20%20import%20%22package:test_core/src/bootstrap/vm.dart%22;%0A%0A%20%20%20%20import%20%22file:///tmp/dart_test_URQAPN/test.dart%22%20as%20test;%0A%0A%20%20%20%20void%20main(_,%20SendPort%20sendPort)%20%7B%0A%20%20%20%20%20%20internalBootstrapVmTest(()%20=%3E%20test.main,%20sendPort);%0A%20%20%20%20%7D%0A%20%20:9:42: Error: A value of type 'int' can't be returned from a function with return type 'Function'.
                    •    - 'Function' is from 'dart:core'.
                    •         internalBootstrapVmTest(() => test.main, sendPort);
                    •                                            ^

jakemac53:
I think the fix you added is correct here, I am not sure how this test was not failing previously :/

PR #1301, opened 2020-07-15
Prep for merge into master

jakemac53:
I kept these as all non-breaking for now.

The main reason is that test_api and test_core do have some dependencies that are not migrated (mockito, build_test), and I don't want to have issues landing in flutter.


natebosch:
I'll revisit https://github.com/dart-lang/test/pull/1297

I'm pretty sure we were either broken more by the SDK or something strange happened...

PR #1302, opened 2020-07-15
Stop using deprecated package:analyzer/analyzer.dart

scheglov:


grouma:
Does this require a change to the analyzer dep?

scheglov:
No, I don't think so.
I did not any changes to test's `pubspec.yaml` files, and these are old APIs.

PR #1303, opened 2020-07-16
Merge the null_safety branch into master

jakemac53:
Note that this effectively means we will no longer be releasing features for the stable branch. Anything additional would be included in the next null safety release.

Cherry picks for bug fixes will be possible but we won't want to backport any new features.

@stereotype441 I don't think the `Fake` class made it in yet? If you want to land that before the null safety version that needs to happen ASAP.


stereotype441:
> @stereotype441 I don't think the `Fake` class made it in yet? If you want to land that before the null safety version that needs to happen ASAP.

Yeah, I haven't had a chance to do it yet.  I think it's ok to wait until after the null safety version, since it's really only needed for clients that are migrating, and clients that are migrating will be pulling down the latest version of the test package.

pkgs/test/mono_pkg.yaml
```diff
@@ -1,18 +1,14 @@
 dart:
-  - dev
+  - be/raw/latest
```
grouma, line 2: Why this change?

jakemac53: Oh this can switch back to dev now, good catch

jakemac53:
> Yeah, I haven't had a chance to do it yet. I think it's ok to wait until after the null safety version, since it's really only needed for clients that are migrating, and clients that are migrating will be pulling down the latest version of the test package.

Sounds good, we should land this relatively soon though (flutter probably wants it for their migration?)

PR #1304, opened 2020-07-20
Move `Fake` class from Mockito to the test API.

stereotype441:
With null safety, making a class derived from Fake will become much
more widely useful, since it will no longer be possible to for a test
method to pass `null` into a non-nullable API.

### DIFF
--- a/pkgs/test_api/CHANGELOG.md
+++ b/pkgs/test_api/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 0.2.20-nullsafety
+
+* Add the `Fake` class, available through `package:test_api/fake.dart`.  This
+  was previously part of the Mockito package, but with null safety it is useful
+  enough that we decided to make it available through `package:test`.  In a
+  future release it will be made available directly through
+  `package:test_api/test_api.dart` (and hence through
+  `package:test_core/test_core.dart` and `package:test/test.dart`).
+
 ## 0.2.19-nullsafety
 
 * Migrate to NNBD.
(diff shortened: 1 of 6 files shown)


pkgs/test_api/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 0.2.20-nullsafety
```
jakemac53, line 1: The previous version is not yet published, so you can merge this into that entry (0.2.19-nullsafety)

jakemac53:
Did we also want to export this from a `package:test/fake.dart` library? We don't really want people to depend on test_api.

pkgs/test_api/CHANGELOG.md
stereotype441: Ok, will do.  Thanks.

stereotype441:
> Did we also want to export this from a `package:test/fake.dart` library? We don't really want people to depend on test_api.

Sure, will do.

pkgs/test_api/lib/src/frontend/fake.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Extend or mixin this class to mark the implementation as a [Fake].
```
natebosch, line 5: Based on comments like https://github.com/dart-lang/language/issues/1068#issuecomment-656923890 I wonder if we should rephrase this to drop the encouragement of using it as a mixin. I also think it could maybe use a different phrasing in general.

```suggestion
/// A stand-in for another object which cannot be used except for specifically overridden methods.
```

stereotype441: Done

stereotype441:
> We should also add a changelog entry into the package:test changelog (can be copy/pasted). That version is also not published so it can be added to the latest version.

Done

PR #1306, opened 2020-07-22
fix dead code warning

jakemac53:


jakemac53:
Also pushing some test fixes here

PR #1307, opened 2020-07-24
Prepare for the 2.10 SDK

natebosch:
Bump the min SDK constraint to `2.10.0-0` so that they pick up the
`2.10` language version which is required to opt in to null safety with
the `2.10` SDKs.


kevmoo:
Also update these to just be `dev`?

https://github.com/dart-lang/test/blob/1c29a6465f269a23d686657e2ecd892519999560/.travis.yml#L17

natebosch:
Those are on the master branch, this is the update for the `null_safety` branch which already uses `dev`.

PR #1314, opened 2020-07-30
Fix declarer inconsistencies with async test setup

jiahaog:
When I have a test like this:

```dart

void main() async {
  setUp(() {});

  await otherSetUp();

  test('a test', () {});
}

Future<void> otherSetUp() {}
```

Running it with `pub test $this_file` works, but running it with `pub run $this_file` throws the following error:

```
Unhandled exception:
Bad state: Can't call test() once tests have begun running.
#0      Declarer._checkNotBuilt (package:test_api/src/backend/declarer.dart:286:5)
#1      Declarer.test (package:test_api/src/backend/declarer.dart:142:5)
#2      test (package:test_core/test_core.dart:138:13)
#3      main (file:///Users/jiahaog/dev/hello_dart/test/hello_dart_test.dart:8:3)
<asynchronous suspension>
#4      _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:301:19)
#5      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:168:12)
```

This is because there is no declarer when running the tests directly, and the test execution is scheduled for the next microtask before the next test can be declared. When running with the test runner, the declarer is set [explicitly](https://github.com/dart-lang/test/blob/3b8d3b90efd24f55f7316cd414716d1cb031761c/pkgs/test_api/lib/src/remote_listener.dart#L128) and microtasks are not involved, so this issue does not occur.

This might just be working as intended, and I'm not sure of the exact semantics of using `scheduleMicrotask` here, but this error was rather unexpected.

### DIFF
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -30,22 +30,6 @@ void main() {
 }
 ''';
 
-final _successWithAsync = '''
-import 'dart:async';
-
-import 'package:test/test.dart';
-
-void main() async {
-  test("success 1", () {});
-
-  await otherSetUp();
-
-  test("success 2", () {});
-}
-
-Future<void> otherSetUp() {}
-''';
-
 final _failure = '''
 import 'dart:async';
 
@@ -379,17 +363,28 @@ $_usage''');
   });
 
   group('runs successful tests with async setup', () {
+    setUp(() async {
+      await d.file('test.dart', '''
+        import 'package:test/test.dart';
+
+        void main() async {
+          test("success 1", () {});
+
+          await () async {};
+
+          test("success 2", () {});
+        }
+      ''').create();
+    });
+
     test('defined in a single file', () async {
-      await d.file('test.dart', _successWithAsync).create();
       var test = await runTest(['test.dart']);
       expect(test.stdout, emitsThrough(contains('+2: All tests passed!')));
       await test.shouldExit(0);
     });
 
     test('directly', () async {
-      await d.file('test.dart', _successWithAsync).create();
       var test = await runDart(['test.dart']);
-
       expect(test.stdout, emitsThrough(contains('All tests passed!')));
       await test.shouldExit(0);
     });
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -30,6 +30,22 @@ void main() {
 }
 ''';
 
+final _successWithAsync = '''
+import 'dart:async';
+
+import 'package:test/test.dart';
+
+void main() async {
+  test("success 1", () {});
+
+  await otherSetUp();
+
+  test("success 2", () {});
+}
+
+Future<void> otherSetUp() {}
+''';
+
 final _failure = '''
 import 'dart:async';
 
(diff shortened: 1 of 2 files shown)


jiahaog:
@natebosch can you take a look at this please

natebosch:
I noticed this too when I was looking at that code.

Your fix helps if there is a microtask in `main`, but it doesn't work if there is something like `await someFile.readAsString()`. I do think do this kind of thing is likely very rare - either you have some async work _before_ you start declaring tests, or you don't have any async work. I don't think I've often seen async work in the middle of declaring tests.

We also don't currently support a `group` callback returning a `Future` which is a similar situation.

Unfortunately I don't think there is a foolproof solution to this. I'd be open to doing something like `pumpEventQueue` to catch a good chunk of the long tail. cc @jakemac53 @grouma WDYT?

grouma:
> Unfortunately I don't think there is a foolproof solution to this. I'd be open to doing something like pumpEventQueue to catch a good chunk of the long tail. 

How do you intend on using `pumpEventQueue`? Would you always call `pumpEventQueue` upon start up? Would that cause some timing issues / slow downs with tests depending on how many pumps?

natebosch:
> How do you intend on using `pumpEventQueue`? Would you always call `pumpEventQueue` upon start up? Would that cause some timing issues / slow downs with tests depending on how many pumps?

This would be used only in the case of running `dart test/some_test.dart` (or compiling directly with `dart2js` etc). The golden path for running tests is still to use the runner and wouldn't be impacted.

The idea is that in the case of running the file directly we'd always have a `pumpEventQueue` before we start running tests in order to give the `main` to complete as much async work as possible. In this case we are being invoke _by_ main, not invoking main ourselves so we otherwise don't have any other signal to know when a returned `Future` is complete for instance.

grouma:
> The idea is that in the case of running the file directly we'd always have a pumpEventQueue before we start running tests in order to give the main to complete as much async work as possible. In this case we are being invoke by main, not invoking main ourselves so we otherwise don't have any other signal to know when a returned Future is complete for instance.

I'm not concerned with adding a call to `pumpEventQueue` for this workflow. Most users go through the actual runner. May want to get feedback from @DanTup as I believe VS code calls the tests files by default. I wouldn't expect any issues though.

DanTup:
If VS Code recognises your file as a test, it goes through the runner (so it can get the JSON output to show in the test tree). There are a few exceptions - like if you open the SDK repo - where we recognise `pub run test` won't work and we just treat it as a normal Dart script.

If I understand the above correctly, I don't think this change can impact anything negatively in VS Code (but it could fix the same described issues in the case where your tests are being run as standard Dart scripts).

jiahaog:
How should we move forward on this? I can change it to use `pumpEventQueue` if you'd like :)

natebosch:
> How should we move forward on this? I can change it to use `pumpEventQueue` if you'd like :)

Yeah, I think that makes sense. I think the simplest way to express it will be to change from

```dart
scheduleMicrotask(() async {
  ...
});
```

to

```dart
() async {
  await pumpEventQueue();
  ...
}();
```

jiahaog:
Done, please take a look!

natebosch:
Thanks!

PR #1318, opened 2020-08-12
Add DetachingFuture utility

natebosch:
Fixes one reason that an instance of `Invoker` is always retained        
through the end of the test. Working towards making it easier to         
diagnose why the test runner sometimes fails to exit on travis for #599.

### DIFF
--- a/pkgs/test_core/lib/src/util/detaching_future.dart
+++ b/pkgs/test_core/lib/src/util/detaching_future.dart
@@ -14,11 +14,11 @@
 /// In the case of a future that resolves to an error the original future is
 /// retained.
 class DetachingFuture<T> {
-  T _value;
-  Future<T> _inProgress;
+  late T _value;
+  Future<T>? _inProgress;
 
-  DetachingFuture(this._inProgress) {
-    _inProgress.then((result) {
+  DetachingFuture(Future<T> inProgress) : _inProgress = inProgress {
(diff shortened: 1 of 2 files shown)


natebosch:
There is still another reason that we "leak" an instance of `Invoker` (and a bunch of other stuff). I haven't been able to track down a satisfying root cause for that - although I have some workarounds that I discovered with guess and check...

This resolves at least one of the reasons that we don't garbage collect the `Invoker` so it's worth submitting. In general it appears that having a variable which is a `Future` at the top level is not a good idea. We have a couple others in `package:test` but so far it doesn't look like they are causing issues.

pkgs/test_core/lib/src/util/detaching_future.dart
```diff
@@ -0,0 +1,28 @@
+/// A handle on a [Future] that detaches from the original evaluation context
```
jakemac53, line 1: copyright

natebosch:
> If [dart-lang/sdk#42457](https://github.com/dart-lang/sdk/issues/42457) was resolved, then I would consider using `runZoned` or something like that as a simpler solution (to ensure it is created in the root zone).

Yeah that would probably look nicer. `final someFuture = Zone.root.run(() async {...});`. If the `awaiter` issue ever gets resolved we can revisit.

PR #1323, opened 2020-08-20
Remove cancel from Reporter interface

natebosch:
This method is never called from outside any reporter so it is
unnecessary to require it. The cancel method on `ExpandedReporter` was
never called before, add a call in `_onDone` to match the other
reporters.


natebosch:
I had hoped that maybe one of those uncanceled `StreamSubscription` may have been the reason we need to sometimes have a hard `exit()` on travis, but it doesn't seem to make a difference.

natebosch:
`Reporter` is not in the public API for `test_core`. I'm only aware of one other internal implementation, so this should only have an impact when it's pulled in. I don't think flutter has an implementation of `Reporter` currently.

Issue #1330, opened 2020-09-01
use the dart.dev/go redirect for the package:test docs


devoncarew:
We should update the (2?) references to `https://bit.ly/2Z7J0OH` to the dart.dev redirect links - `https://dart.dev/go/test-docs/json_reporter.md` (see https://github.com/dart-lang/site-www/pull/2615).

PR #1331, opened 2020-09-01
Clean up --help

grouma:
Clean up `--help` to better align with other commands captured in Dart CLI.

Closes https://github.com/dart-lang/test/issues/1330

### DIFF
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test
-version: 1.16.0-nullsafety.2
+version: 1.16.0-nullsafety.3
 description: A full featured library for writing and running Dart tests.
 homepage: https://github.com/dart-lang/test/blob/master/pkgs/test
 
@@ -34,7 +34,7 @@ dependencies:
   yaml: ^2.0.0
   # Use an exact version until the test_api and test_core package are stable.
   test_api: 0.2.19-nullsafety
-  test_core: 0.3.12-nullsafety.2
+  test_core: 0.3.12-nullsafety.3
 
 dev_dependencies:
   fake_async: ^1.0.0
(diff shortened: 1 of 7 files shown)


pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -27,13 +27,13 @@ final ArgParser _parser = (() {
   parser.addFlag('help',
       abbr: 'h', negatable: false, help: 'Shows this usage information.');
```
natebosch, line 28: Should this be `Show` for consistency?

```suggestion
      abbr: 'h', negatable: false, help: 'Show this usage information.');
```

pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -118,25 +119,24 @@ final ArgParser _parser = (() {
...
   parser.addOption('file-reporter',
-      help: 'The reporter used to write test results to a file.\n'
+      help: 'Set the reporter used to write test results to a file.\n'
           'Should be in the form <reporter>:<filepath>, '
-          'e.g. "json:reports/tests.json"');
+          'Example: "json:reports/tests.json"');
   parser.addFlag('verbose-trace',
-      negatable: false,
-      help: 'Whether to emit stack traces with core library frames.');
+      negatable: false, help: 'Emit stack traces with core library frames.');
```
natebosch, line 134: Do you know why this isn't negatable? This came up on the doc.

grouma: I found the PR that added the flag: https://github.com/dart-lang/test/commit/b61a5362bb37eb7a9ffa6ccb2f7837d79701b4c6

Doesn't have a reason though.

pkgs/test/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: test
-version: 1.16.0-nullsafety.2
+version: 1.16.0-nullsafety.3
```
devoncarew, line 2: I assume this means that users won't see the help text changes until they're using a null safe version of package:test?

grouma: That I don't know. @natebosch

natebosch: That is correct, but using a null safe compatible version of `package:test` does not require that the user migrates to null safety.

Issue #1328, opened 2020-08-25
Make it possible to uniquely identify test cases


natebosch:
There are a few use cases that are easier to tackle when a test case can be uniquely identified.

- Could unblock a mode to retry failed tests - need to be able to specifically target for a retry: https://github.com/dart-lang/test/issues/329
- Could make the test runner more tooling friendly in general: https://github.com/dart-lang/test/issues/908#issuecomment-406250616

Some options to consider:
- If we can give a unique ID to a suite, we could introduce a restriction that test names are unique within a test suite (generally a single test file). Then a suite ID and test name would be unique. https://github.com/dart-lang/test/issues/661 This would be breaking but is something we could feasibly roll out.
- We could use a deterministic ordering (if it isn't already, it may be for sharding purposes) for test suites to load and number test cases as they are declared.

This is going to be challenging to take full advantage of. Since tests are declared by executing Dart code there is no static and consistent set of tests. There are any number of ways a tests could introduce non-determinism in the order that test cases are declared, or even in the number and names of tests. If we consider any pattern of interaction that involves separate invocations of the test runner we might want to build in some safety from the start - for instance we could create a hash of some information when we _assign_ names to tests, and then require that it is passed to us and still matches when we _run_ a named test. (We could use a similar concept when sharding, each shard should ideally arrive at the same hash so that we are sure there is a consistent view of the tests and that cases aren't being dropped due to non-determinism).

cc @jiahaog @DanTup

DanTup:
Maybe dupe of https://github.com/dart-lang/test/issues/837?

In VS Code I worked around the original issue by just matching things based on name. In practice, it works great - with some minor niggles like if you have multiple tests with the same name, we'll run them all when you try to run one. We use the analysis servers Outline data to get tests names so create the CodeLens links though, so I'm not sure unique IDs would solve this since the editor wouldn't have them up-front (though perhaps passing an index number in this case might work - "run the nth test called x").

There are two other issues I have but I don't know if IDs are the solution (perhaps improving the `--name` param is):

1. Can't filter on exact group names (we use regex with a leading `^` and trailing `$` normally to avoid catching too many tests, but we have to leave the `$` off the end for groups, but that can result in matching more than just the intended group).
2. Can't filter on tests that have variables/etc. in the names (the analysis server gives us the exact string, including the variable, but `package:test` only matches on the resolved string passed to the `test` function).

Issue #1310, opened 2020-07-29
Add an API to do some customization and then directly run tests in process


natebosch:
# Background

Most of our runners assume some separation between the process that orchestrates running all the tests, and the tests themselves. This is different Isolates in the VM case, and external processes like browsers in other case.

We also support running a `_test.dart` file directly:

https://github.com/dart-lang/test/blob/1c29a6465f269a23d686657e2ecd892519999560/pkgs/test_core/lib/test_core.dart#L34-L72

This support is hardcoded to the `expanded` reporter and does not take any arguments for things like test filters.

In all cases when we run tests through a runner (internally, flutter_test, build_test, and `package:test` runner) we "wrap" the users library with out own and forward to their `main` in some way where the details vary by platform.

# Goal

We should make it easier to combine the concepts of running without an externally runner doing the orchestration, along with allowing _some_ customization beyond the defaults. The immediate need is to register and pick a custom reporter.

# Plan

Add a function in `package:test_core/backend.dart` like `Future<bool> runTests(FutureOr<void> Function() userMain, {ReporterDetails reporter})`.

Then platforms that want to build a custom "runner" that is baked in to a single execution instead of having a separate orchestration would use it by generating a wrapping library like:

```dart
import 'package:test_core/backend.dart';
import 'package:my_custom_test/my_custom_test.dart' as custom;  // The reporter implementation.

import '$userTestFile' as test; // Generated portion of the file - imports the users _test.dart file

void main() {
  runTests(test.main, reporter: custom.reproter);
}
```

We'd start with `ReporterDetails` for now, and we can hopefully add other named arguments for other customizations that crop up. For instance we might support test filtering by tag with an additional named argument.

natebosch:
We want to explore a bit on https://github.com/dart-lang/test/issues/1311 before making decisions about making more of this stuff public and spreading use.

natebosch:
Another thing to consider here is whether and how we can support coverage.

@grouma - do you know if it is feasible to enable coverage collection from within the isolate?

grouma:
Not sure what you mean by "within the isolate". However, `package:coverage` supports collecting coverage information for a given isolate ID. See [this](https://github.com/dart-lang/coverage/blob/7e18328a0fcb9353d2ad68748b66f5f5cb7c253c/lib/src/collect.dart#L37).

natebosch:
[...]

PR #1332, opened 2020-09-02
Add capabilities for building direct test runners

natebosch:
Towards #1310, #1328

Add `directRunTest` to configure a reporter and run tests directly in
the same isolate.

Add `enumerateTestCases` to collect test names without running them, and
`directRunSingleTest` to run a specific test by its full name. These
APIs ensure the uniqueness of test names to avoid ambiguity. This
restriction may be spread to tests run through the normal test runner as
well.

- Add `fullTestName` option on `Declarer`. When used, only the test (or
  tests if uniqueness is not checked separately) will be considered as a
  test case.
- Add `directRunTest`, `enumerateTestCases`, and `directRunSingleTest`
  APIs. These are kept under a `lib/src/` import for now, and any other
  package that uses these APIs should pin to a specific version of
  `package:test_core`. The details of these APIs might change without a
  major version bump.

### DIFF
--- a/pkgs/test_api/lib/src/backend/declarer.dart
+++ b/pkgs/test_api/lib/src/backend/declarer.dart
@@ -91,6 +91,12 @@ class Declarer {
   /// Whether any tests and/or groups have been flagged as solo.
   bool get _solo => _soloEntries.isNotEmpty;
 
+  /// A full test name to match when running in single test mode.
+  ///
+  /// Groups which are not a strict prefix of `filterToSingleTestName` will be
+  /// skipped.
+  final String? _filterToSingleTestName;
+
   /// The current zone-scoped declarer.
   static Declarer? get current => Zone.current[#test.declarer] as Declarer?;
 
@@ -113,7 +119,8 @@ class Declarer {
       {Metadata? metadata,
       Set<String>? platformVariables,
       bool collectTraces = false,
-      bool noRetry = false})
+      bool noRetry = false,
+      String? filterToSingleTestName})
       : this._(
             null,
             null,
@@ -121,10 +128,19 @@ class Declarer {
             platformVariables ?? const UnmodifiableSetView.empty(),
             collectTraces,
             null,
-            noRetry);
+            noRetry,
+            filterToSingleTestName);
 
-  Declarer._(this._parent, this._name, this._metadata, this._platformVariables,
-      this._collectTraces, this._trace, this._noRetry);
+  Declarer._(
+    this._parent,
+    this._name,
+    this._metadata,
+    this._platformVariables,
+    this._collectTraces,
+    this._trace,
+    this._noRetry,
+    this._filterToSingleTestName,
+  );
 
   /// Runs [body] with this declarer as [Declarer.current].
   ///
@@ -143,6 +159,12 @@ class Declarer {
       bool solo = false}) {
     _checkNotBuilt('test');
 
+    final fullName = _prefix(name);
+    if (_filterToSingleTestName != null &&
+        fullName != _filterToSingleTestName) {
+      return;
+    }
+
     var newMetadata = Metadata.parse(
         testOn: testOn,
         timeout: timeout,
@@ -152,8 +174,7 @@ class Declarer {
         retry: _noRetry ? 0 : retry);
     newMetadata.validatePlatformSelectors(_platformVariables);
     var metadata = _metadata.merge(newMetadata);
-
-    _entries.add(LocalTest(_prefix(name), metadata, () async {
+    _entries.add(LocalTest(fullName, metadata, () async {
       var parents = <Declarer>[];
       for (Declarer? declarer = this;
           declarer != null;
@@ -195,6 +216,11 @@ class Declarer {
       bool solo = false}) {
     _checkNotBuilt('group');
 
+    final fullName = _prefix(name);
+    if (!(_filterToSingleTestName?.startsWith(fullName) ?? true)) {
+      return;
+    }
+
     var newMetadata = Metadata.parse(
         testOn: testOn,
         timeout: timeout,
@@ -207,7 +233,7 @@ class Declarer {
     var trace = _collectTraces ? Trace.current(2) : null;
 
     var declarer = Declarer._(this, _prefix(name), metadata, _platformVariables,
-        _collectTraces, trace, _noRetry);
+        _collectTraces, trace, _noRetry, _filterToSingleTestName);
     declarer.declare(() {
       // Cast to dynamic to avoid the analyzer complaining about us using the
       // result of a void method.
--- a/pkgs/test_api/lib/src/backend/declarer.dart
+++ b/pkgs/test_api/lib/src/backend/declarer.dart
@@ -91,10 +91,16 @@ class Declarer {
   /// Whether any tests and/or groups have been flagged as solo.
   bool get _solo => _soloEntries.isNotEmpty;
 
-  /// A full test name to match when running in single test mode.
+  /// An exact full test name to match.
+  ///
+  /// When non-null only tests with exactly this name will be considered. The
+  /// full test name is the combination of the test case name with all group
+  /// prefixes. All other tests, including their metadata like `solo`, is
+  /// ignored. Uniqueness is not guaranteed so this may match more than one
+  /// test.
   ///
   /// Groups which are not a strict prefix of this name will be ignored.
-  final String? _filterToSingleTestName;
+  final String? _fullTestName;
 
   /// The current zone-scoped declarer.
   static Declarer? get current => Zone.current[#test.declarer] as Declarer?;
@@ -119,7 +125,7 @@ class Declarer {
       Set<String>? platformVariables,
       bool collectTraces = false,
       bool noRetry = false,
-      String? filterToSingleTestName})
+      String? fullTestName})
       : this._(
             null,
             null,
@@ -128,7 +134,7 @@ class Declarer {
             collectTraces,
             null,
             noRetry,
-            filterToSingleTestName);
+            fullTestName);
 
   Declarer._(
     this._parent,
@@ -138,7 +144,7 @@ class Declarer {
     this._collectTraces,
     this._trace,
     this._noRetry,
-    this._filterToSingleTestName,
+    this._fullTestName,
   );
 
   /// Runs [body] with this declarer as [Declarer.current].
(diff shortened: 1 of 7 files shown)


pkgs/test_core/lib/direct_run.dart
```diff
@@ -0,0 +1,41 @@
...
+import 'src/runner/engine.dart';
+import 'src/runner/plugin/environment.dart';
+import 'src/runner/reporter.dart';
+import 'src/runner/reporter/expanded.dart';
+import 'src/runner/runner_suite.dart';
+import 'src/runner/suite.dart';
+import 'src/util/print_sink.dart';
+
+Future<bool> directRunTests(FutureOr<void> Function() testMain,
+    {Reporter Function(Engine)? reporter}) async {
```
jakemac53, line 20: nit: the variable name should indicate this is really a factory, `reporterFactory`, `makeReporter`, `buildReporter`, etc

jiahaog:
Maybe for coverage we could let the user wire that up themselves outside of this function? Another possible approach along the lines of https://github.com/dart-lang/test/issues/1310#issuecomment-676704479 could be to expose a `args` parameter that let users pass the `--coverage` parameter as needed.

natebosch:
> Maybe for coverage we could let the user wire that up themselves outside of this function?

I think we'll probably need to. @grouma do you have an idea of what type of API we'd want to write to make it easy to trigger coverage collection for the main isolate? This differs a little from our normal approach where we are interested in the coverage from an isolate we spawn and control.



> Another possible approach along the lines of [#1310 (comment)](https://github.com/dart-lang/test/issues/1310#issuecomment-676704479) could be to expose a `args` parameter that let users pass the `--coverage` parameter as needed.

I spent some more time thinking about this approach and I don't think it's really feasible. I'll update on the linked issue.

grouma:
> I think we'll probably need to. @grouma do you have an idea of what type of API we'd want to write to make it easy to trigger coverage collection for the main isolate? This differs a little from our normal approach where we are interested in the coverage from an isolate we spawn and control.

I'm not sure if that is even feasible. To collect coverage you need to [provide](https://github.com/dart-lang/coverage/blob/f373b1dae4ec9328a36c75093878bbd672b41b97/lib/src/collect.dart#L37) a service URI. I don't know how the test file would get access to the VM's service URI. I'm also not sure what happens when a VM starts inspecting itself. If a VM can't inspect itself, you may be able to do something super hacky like spawning a detached process to do the inspection. That seems incredibly weird to me though.  Finally, I would expect the user would likely have to pass `dart --observer <some-test>` which also seems quite odd.

jiahaog:
Wouldn't the test file be able to call `dart:developer` to obtain the service URI? I was able to do this in a Flutter app and call `collect` from `package:coverage`, to write the coverage to the file. Could the behavior of the VM inspecting itself be different across other Dart embedders?

grouma:
> Wouldn't the test file be able to import dart:developer to obtain the service URI? I was able to do this in a Flutter app and call collect from package:coverage, to write the coverage to a file. Could the behavior of the VM inspecting itself be different across other Dart embedders?

I was not aware of this workflow. If this works with Flutter then it will almost certainly work with the vanilla VM. In regards to Nates original question then, the API should probably just mirror that of the `--coverage` flag. We can provide an optional parameter so the user can specify the output path for the coverage hitmap file. See [here](https://github.com/dart-lang/test/blob/93c1ad0f0114f7e828886d56511af1058895dccc/pkgs/test_core/lib/src/runner/coverage.dart#L17).

pkgs/test_core/lib/direct_run.dart
natebosch: Done

natebosch:
@jakemac53 @grouma - how do you feel about merging this without tests? I'll keep it in `src/` until it's got tests, but I'd like to get this in rolled so we can start prototyping test runners against it.

pkgs/test_core/lib/src/direct_run.dart
```diff
@@ -0,0 +1,135 @@
...
+
+Future<bool> _directRunTests(FutureOr<void> Function() testMain,
+    {Reporter Function(Engine)? reporterFactory,
+    String? runSingleTestByName}) async {
+  reporterFactory ??= (engine) => ExpandedReporter.watch(engine, PrintSink(),
+      color: Configuration.empty.color, printPath: false, printPlatform: false);
+  final declarer = Declarer(filterToSingleTestName: runSingleTestByName);
+  await declarer.declare(testMain);
+
+  await pumpEventQueue();
```
grouma, line 57: Why do we need this pump?

pkgs/test_core/lib/src/direct_run.dart
```diff
@@ -0,0 +1,135 @@
...
+
+/// An exception thrown when two test cases in the same test suite (same `main`)
+/// have an identical name.
+class DuplicateTestNameException implements Exception {
+  final String name;
+  DuplicateTestNameException(this.name);
+
+  @override
+  String toString() => 'A test with the name "$name" was already declared. '
+      'Test cases much have unique names.';
```
grouma, line 123: [nit] must*

grouma (approved):
I'm fine with landing this as long as we get tests before actually consuming it.

pkgs/test_core/lib/src/direct_run.dart
natebosch: It matches what we did in https://github.com/dart-lang/test/pull/1314

natebosch: oh, but we shouldn't need it in this case since we `await` the returned future...

Good point 😝

pkgs/test_core/lib/src/direct_run.dart
natebosch: Done

pkgs/test_api/lib/src/backend/declarer.dart
```diff
@@ -91,6 +91,11 @@ class Declarer {
   /// Whether any tests and/or groups have been flagged as solo.
   bool get _solo => _soloEntries.isNotEmpty;
 
+  /// A full test name to match when running in single test mode.
+  ///
+  /// Groups which are not a strict prefix of this name will be ignored.
+  final String? _filterToSingleTestName;
```
jakemac53, line 97: This sounds like a boolean to me. I would just call it `_testName`, or possibly `_singleTestName`.

pkgs/test_api/lib/src/backend/declarer.dart
```diff
@@ -143,6 +158,12 @@ class Declarer {
       bool solo = false}) {
     _checkNotBuilt('test');
 
+    final fullName = _prefix(name);
+    if (_filterToSingleTestName != null &&
```
jakemac53, line 162: Do you need the null check here? (is fullName nullable?)

pkgs/test_api/lib/src/backend/declarer.dart
```diff
@@ -195,6 +215,11 @@ class Declarer {
       bool solo = false}) {
     _checkNotBuilt('group');
 
+    final fullName = _prefix(name);
+    if (!(_filterToSingleTestName?.startsWith(fullName) ?? true)) {
```
jakemac53, line 219: Possibly worth a comment?

I also find the `?? true` wrapped in parens and then negated very difficult to parse. I would write a more verbose form.

pkgs/test_core/lib/src/direct_run.dart
```diff
@@ -0,0 +1,130 @@
...
+/// be used.
+Future<bool> directRunSingleTest(
+        FutureOr<void> Function() testMain, String testName,
+        {Reporter Function(Engine)? reporterFactory}) =>
+    _directRunTests(testMain,
+        reporterFactory: reporterFactory, runSingleTestByName: testName);
+
+Future<bool> _directRunTests(FutureOr<void> Function() testMain,
+    {Reporter Function(Engine)? reporterFactory,
+    String? runSingleTestByName}) async {
```
jakemac53, line 50: I would also line up the name here with the field name on Declarer

pkgs/test_core/lib/src/direct_run.dart
```diff
@@ -0,0 +1,130 @@
...
+  return success!;
+}
+
+/// Return the names of all tests declared by [testMain].
+///
+/// Test names declared must be unique. If any test repeats the name of a prior
+/// test a [DuplicateTestNameException] will be thrown.
+///
+/// Skipped tests are ignored.
+Future<Iterable<String>> enumerateTestCases(
```
jakemac53, line 87: Should we type this as `Set` instead of `Iterable` - the contract is guaranteeing uniqueness so I think it makes sense and would avoid some potential copies for users who want a set.

pkgs/test_api/lib/src/backend/declarer.dart
natebosch: What do you think of `_exactTestName` or `_fullTestName`? I realized that `single` isn't quite appropriate here because `Declarer` doesn't do anything to enforce that there is exactly one. That enforcement happens in `directRunSingleTest`.

pkgs/test_api/lib/src/backend/declarer.dart
natebosch: No, `_prefix` returns non-nullable String.

pkgs/test_core/lib/src/direct_run.dart
natebosch: I'm uncertain when I want to used an ordered collection. As implemented the test names are returned in the same order that they'd be run (without randomization turned on). I don't know if that matters or not, or if it's better to return it as a `Set`. I'll try `Set` for now.

pkgs/test_api/lib/src/backend/declarer.dart
jakemac53: `_fullTestName` SGTM

pkgs/test_api/lib/src/backend/declarer.dart
natebosch: I'll need to use `!` for the verbose form since it's a field. That's why I wrote it this way originally. I can expand it with the `!` though.

pkgs/test_core/lib/src/direct_run.dart
natebosch: Done

jakemac53 (approved):

PR #1334, opened 2020-09-09
Update package:vm_service dependency

bkonyi:
Required to roll package:vm_service forward in DevTools


grouma (commented):
You'll want to update the dep from `package:test` to `package:test_core` as well.

bkonyi:
> You'll want to update the dep from `package:test` to `package:test_core` as well.

Done.

grouma (approved):

Issue #962, opened 2018-11-15
Define a stable-ish API for test_core and test_api


jonahwilliams:
This is not a critical issue, but it would be nice to not need to import src files. To define the initial public API, we might look at the imports flutter and fuchsia need to define their own testing platform.

## [flutter_platform.dart](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/test/flutter_platform.dart)
```
import 'package:test_api/src/backend/runtime.dart';
import 'package:test_api/src/backend/suite_platform.dart';

import 'package:test_core/src/runner/platform.dart';
import 'package:test_core/src/runner/hack_register_platform.dart' as hack;
import 'package:test_core/src/runner/runner_suite.dart';
import 'package:test_core/src/runner/suite.dart';
import 'package:test_core/src/runner/plugin/platform_helpers.dart';
import 'package:test_core/src/runner/environment.dart';
```

## [fuchsia_test_helper](https://fuchsia.googlesource.com/topaz/+/master/lib/fuchsia_test_helper/lib/fuchsia_test_helper.dart)
Note: migration from test WIP -but all of these can be sourced from test_api and test_core

```
import 'package:test/src/backend/declarer.dart';
import 'package:test/src/backend/group.dart';
import 'package:test/src/backend/suite_platform.dart';
import 'package:test/src/backend/operating_system.dart';
import 'package:test/src/backend/runtime.dart';
import 'package:test/src/runner/runner_suite.dart';
import 'package:test/src/runner/configuration/suite.dart';
import 'package:test/src/runner/engine.dart';
import 'package:test/src/runner/plugin/environment.dart';
import 'package:test/src/runner/reporter/expanded.dart';
```

PR #1341, opened 2020-09-15
Add a centralized platform import

natebosch:
Towards #962

Make it easier to import the necessary things for writing a custom
platform. This does not expose `executable` so a custom test runner
would need to import at a minimum this and the executable. The symbols
exported here are the ones used by `flutter_tools` in
`flutter_platform.dart` and `test_wrapper.dart`, if other platforms need
other symbols we may need to expand this some.

The import remains a `src` import since these are not yet stable enough
to be fully public. Centralizing and adding `show` clauses will make it
easier to know what is used and for which intent.

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 0.3.12-nullsafety.5-dev
 
+* Add `src/platform.dart` library to consolidate the necessary imports required
+  to write a custom platform.
+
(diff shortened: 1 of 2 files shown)


natebosch:
@jonahwilliams - this now covers `flutter_platform.dart`, `test_wrapper.dart`, and `flutter_web_platform.dart`. Is there anywhere else I should be looking?

jonahwilliams:
I think those should cover it

jakemac53 (approved):

PR #1342, opened 2020-09-17
Fix documentation over custom HTML templates

regenvanwalbeek-wf:
The documentation recommends adding `custom-html-template-path` to a configuration file to use custom HTML templates. However, this will not work - you actually have to add `custom_html_template_path` (See https://github.com/dart-lang/test/pull/1127/files#diff-b9a58c9d38842cb742b063d60ff6fb18R254)


jakemac53 (approved):

jakemac53:
Thanks!

PR #1343, opened 2020-09-17
[WIP] allow analyzer 0.40.x in stable test/test_core packages

jakemac53:
~Will publish these versions as listed here, and then change the base to master, merge with it and submit.~

Will publish after merging into this branch, and then send a separate PR to merge the `pre-nnbd` branch into master (but leave the actual `pre-nnbd` branch intact).


natebosch (approved):

jakemac53:
The remaining failures here are just related to NNBD tests that have an outdated nnbd release version - ignoring those and going to go ahead and publish and merge this into master.

> Don't we just want to maintain a separate `pre-nnbd` branch as long as we care about maintaining things?

We could, I would just land this pr into that branch and then send out a separate PR that merges it into master. @natebosch wdyt? is `pre-nnbd` a fine name for it or would you prefer just `stable` or something else?

Note that I would be tagging this version regardless so we can always get back to this state either way.

natebosch:
I would be happy with a long-lived `pre-nnbd` branch that we delete after we no longer need it once we can publish non-beta package versions.

jakemac53:
Ok, I will go ahead and fix the tests then and land this in the branch. Then I will publish, and send out a separate PR to merge that branch into master, leaving the original `pre-nnbd` branch intact.

PR #1345, opened 2020-09-17
Add Target annotations to annotation classes

natebosch:
This will allow the analyzer to emit diagnostics when the annotations
are used in places where they will not be honored by the test runner.
When used as annotations (as opposed to arguments to `test` or `group`)
the test runner only looks at the library level and all annotations
apply to an entire test suite.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.16.0-nullsafety.5-dev
+
+* Annotate the classes used as annotations to restrict their usage to library
+  level.
+
 ## 1.16.0-nullsafety.4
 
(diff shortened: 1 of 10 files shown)


jakemac53:
Does this still allow annotations on the first directive in the file? That is actually what is supported by the test package.

natebosch:
> Does this still allow annotations on the first directive in the file? That is actually what is supported by the test package.

Yes it does. https://pub.dev/documentation/meta/latest/meta_meta/TargetKind-class.html

> Indicates that an annotation is valid on the first directive in a library, whether that's a `library`, `import`, `export` or `part` directive. This doesn't include the `part of` directive in a part file.

jakemac53 (approved):
Nice

PR #1346, opened 2020-09-21
Drop need for SILENT_OBSERVATORY environment

### LINKED ISSUE #1206: How to actually set SILENT_OBSERVATORY?

natebosch:
Closes #1206

- Pass `silenceOutput` when calling `controlWebServer`.
- Drop the check and warning for missing `SILENT_OBSERVATORY`.
- Drop the test for the warning.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Annotate the classes used as annotations to restrict their usage to library
   level.
+* Stop required a `SILENT_OBSERVATORY` environment variable to run with
+  debugging and the JSON reporter.
 
 ## 1.16.0-nullsafety.4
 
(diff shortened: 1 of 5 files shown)


grouma (approved):

jakemac53 (approved):

SebastianJL:
Thanks a lot for solving this problem :100:

PR #1350, opened 2020-10-02
Update nnbd tests now that 2.11 is in dev

kevmoo:


pkgs/test/test/runner/runner_test.dart
```diff
@@ -759,7 +759,7 @@ final foo = true;''').create();
     test('sound null safety is enabled if the entrypoint opts in explicitly',
         () async {
       await d.file('test.dart', '''
-// @dart=2.10
+// @dart=2.11
```
jakemac53, line 762: These could actually stay at 2.10, it is a bug that it doesn't work

jakemac53 (approved):

pkgs/test/test/runner/runner_test.dart
kevmoo: Uh...so you approve, but it shouldn't be changed?

PR #1356, opened 2020-10-21
Add markTestSkipped API

### LINKED ISSUE #1355: Skip tests during execution
## Context

I have a test suite of end-to-end tests for a backend delivering live Formula 1 data. Depending on the race calendar - whether it's a racing weekend, midweek or a bye week - the backend sends different data and some endpoints (live timings) error.

## Problem

Currently, I have to collect all data for all tests upfront to then decide if a test should be run or skipped.

```dart
// run_all.dart

/// When live == true race is live 
bool live;

Future<void> runAll() async {
  // load all information upfront before declaring the tests
  await _loadServerConfig();

  get_drivers.main();
  lap_times.main();
  // ...
}

Future<void> _loadServerConfig() {
  final status = await api.apiStatus();
  live = status.phase == "live";
}
```

```dart
// lap_times.dart

Future<void> main() async {
  test(
    "GET /race/timings/:player_id",
    () async {
      final players = await api.allPlayers();
      final timings = await api.getLiveTimings(players.first);

	  for (final time in timings) {
        expect(time.current, isNotNull);
        expect(time.lap, isNotNull);
      }
    },
    skip: () {
      // test executed without run_all.dart, always execute
      if (live == null) return false;
	  if (!live) return "race is not live";
	  
      // race live
      return false;
    }(),
  );
}
```
Note: It is not possible to split the upfront logic and move it into the `main()` methods of the other files because once the first call to `test` happened, all tests have to be added synchronously. Any async call between `test` calls results in an error. Moving the upfront logic into the `setup` method also doesn't work because it is executed after the tests are declared, therefore it is not possible to adjust the `skip` flag.

## Expectation

It would be much easier when a test could be skipped after it was started. This would drastically decrease complexity.

```dart
// lap_times.dart

Future<void> main() async {
  test( "GET /race/timings/:player_id", () async {
    final status = await api.apiStatus();
    final live = status.phase == "live";
    if (!live) {
      skip("Race is not live"); // Skip tests here, test should not continue
    }

    final players = await api.allPlayers();
    final timings = await api.getLiveTimings(players.first);

    for (final time in timings) {
      expect(time.current, isNotNull);
      expect(time.lap, isNotNull);
    }
  );
}
```

`skip`, analog to `fail`, throws a specific Exception caught by the test framework to mark the test as skipped.
```dart 
@alwaysThrows
void skip(String message) {
  throw SkipException(message);
}
```

## Other ways to stop test execution

Stopping test execution is currently only possible with
- `return;` which marks the test as passed and
- `throw "skip";` which marks the test as failed.

Both exit conditions do not match was actually happened, which is that the test was skipped and did not reach the necessary assertions for success.

___

The `test` package isn't defined as "unit test" package, therefore I think better support for end-to-end tests could be in the package scope.

natebosch:
Closes #1355

Prepare to publish.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 1.16.0-nullsafety.6-dev
+## 1.16.0-nullsafety.6
+
+* Add `markTestSkipped` API.
 
 ## 1.16.0-nullsafety.5
 
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test
-version: 1.16.0-nullsafety.6-dev
+version: 1.16.0-nullsafety.6
 description: A full featured library for writing and running Dart tests.
 homepage: https://github.com/dart-lang/test/blob/master/pkgs/test
 
@@ -32,8 +32,8 @@ dependencies:
   webkit_inspection_protocol: ">=0.5.0 <0.8.0"
   yaml: ^2.0.0
   # Use an exact version until the test_api and test_core package are stable.
-  test_api: 0.2.19-nullsafety.2
-  test_core: 0.3.12-nullsafety.5
+  test_api: 0.2.19-nullsafety.3
+  test_core: 0.3.12-nullsafety.6
 
 dev_dependencies:
   fake_async: ^1.0.0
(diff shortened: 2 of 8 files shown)


jakemac53 (approved):

passsy:
Thank you so much!

PR #1359, opened 2020-10-28
Update comment referring to closed issue

natebosch:
This had been synchronous in order to respond to syntax errors when the
isolate tried to start running. In dart 2 the isolate will fail to start
entirely and no code will run.


natebosch:
Ah, it looks like maybe we do need it to be sync for one edge case.

If there is a `print` followed by a bunch of synchronous work, we won't see the message until after the synchronous work is done. Keeping this channel sync allows the print to come through synchronously as well.

kevmoo (approved):

jakemac53 (approved):

Issue #1362, opened 2020-10-31
"Null check operator used on null value" when killing test through dartdev tool


jonahwilliams:
To reproduce, start a test using `dart test` and then press Ctrl+C once. After a second or so the exception below will be thrown. This does not happen with `pub run test`.

```
PS C:\Users\Jonah\flutter\packages\flutter_tools> dart test  .\test\general.shard\
Building flutter tool...
Running pub upgrade...
Precompiling executable...
Precompiled test:test.
00:50 +316 ~1: loading .\test\general.shard\asset_bundle_variant_test.dart
Waiting for current test(s) to finish.
Press Control-C again to terminate immediately.
Unhandled exception:
Null check operator used on a null value
#0      CompactReporter._onTestStarted.<anonymous closure> (package:test_core/src/runner/reporter/compact.dart:187:60)
#1      _rootRunUnary (dart:async/zone.dart:1206:13)
#2      _CustomZone.runUnary (dart:async/zone.dart:1100:19)
#3      _CustomZone.runUnaryGuarded (dart:async/zone.dart:1005:7)
#4      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:357:11)
#5      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:285:7)
#6      _SyncStreamControllerDispatch._sendData (dart:async/stream_controller.dart:808:19)
#7      _StreamController._add (dart:async/stream_controller.dart:682:7)
#8      _StreamController.add (dart:async/stream_controller.dart:624:5)
#9      new Stream.periodic.<anonymous closure>.sendEvent (dart:async/stream.dart:358:22)
#10     new Stream.periodic.<anonymous closure>.<anonymous closure>.<anonymous closure> (dart:async/stream.dart:377:22)
#11     Timer._createTimer.<anonymous closure> (dart:async-patch/timer_patch.dart:18:15)
#12     _Timer._runTimers (dart:isolate-patch/timer_impl.dart:397:19)
#13     _Timer._handleMessage (dart:isolate-patch/timer_impl.dart:428:5)
#14     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)
Terminate batch job (Y/N)? y
```

versions:

```
 test: "1.16.0-nullsafety.7"
 test_api:  "0.2.19-nullsafety.4"
 test_core: "0.3.12-nullsafety.7"
```

PR #1364, opened 2020-11-02
fix crash if _lastProgressMessage is null when logging progress

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1362


natebosch (approved):

jakemac53:
Merging on red for now due to other issues

PR #1366, opened 2020-11-03
Bump SDK constraints for pub

natebosch:
Use a 2.12.0 lower bound since pub does not understand allowed
experiments for earlier versions.

Use a 3.0.0 upper bound to avoid a warning in pub and to give some
flexibility in publishing for stable.


jakemac53 (approved):

natebosch:
The red travis is the `runZonedGuarded` analysis errors, not going to wait for a rerun.

Issue #1363, opened 2020-11-02
spawnHybridUri and NNBD opt-out


mraleph:
It seems that using `spawnHybridUri` in web tests does not opt target out if the spawner is opted-out. 

See this test in grpc package: https://github.com/grpc/grpc-dart/blob/master/test/grpc_web_test.dart running it on `dev` channel without manually opting `grpc_web_server.dart` out (spawned here: https://github.com/grpc/grpc-dart/blob/master/test/grpc_web_test.dart#L21) causes the test to fail with 

```
  IsolateSpawnException: Unable to spawn isolate: http://localhost:35703/gxi6aFI%2BPNmQ1A8pOnIQZwHWxkCs8Onb/test/grpc_web_server.dart:80:31: Error: The argument type 'String?' can't be assigned to the parameter type 'String' because 'String?' is nullable and 'String' isn't.
        final port = int.parse(m[1]);
```

/cc @jakemac53

jakemac53:
This looks like a fair bit of reworking is in order. We currently pass the `packageConfig` from the current isolate on the server side in order to spawn the isolate, but the URI we spawn with is an http uri that loads the file from the browser.

I am not sure why/if loading it as an http uri is actually desirable, we could instead pass the relative uri down to the test runner, and let it create an absolute _file_ uri. 

We could also create a custom package config that understands the http uris.

nex3:
This is still broken on the released branch, and it's [breaking our CI](https://travis-ci.com/github/sass/dart-sass/jobs/439795491).

jakemac53:
> This is still broken on the released branch, and it's [breaking our CI](https://travis-ci.com/github/sass/dart-sass/jobs/439795491).

To clarify, does this work if you opt out the file with `// @dart=2.9`?

nex3:
No, it doesn't.

jakemac53:
I think this might be something node specific, we will have to take a look.

mraleph:
FWIW it does not work for gRPC test either (the one which prompted me to file the issue). If I remove opt out from [here](https://github.com/grpc/grpc-dart/blob/master/test/grpc_web_server.dart#L1) then [this spawn](https://github.com/grpc/grpc-dart/blob/master/test/grpc_web_test.dart#L122) fails.

jakemac53:
Thanks for pointing at a repro I will take a look at the grpc scenario

jakemac53:
[...]

PR #1369, opened 2020-11-03
Fix spawnHybridUri for NNBD

jakemac53:
Should fix https://github.com/dart-lang/test/issues/1363

I simplified the logic a bunch here, pushing all resolution of the uri to the server side.

### DIFF
--- a/pkgs/test_core/lib/src/runner/spawn_hybrid.dart
+++ b/pkgs/test_core/lib/src/runner/spawn_hybrid.dart
@@ -6,11 +6,13 @@ import 'dart:async';
 import 'dart:isolate';
 
 import 'package:async/async.dart';
+import 'package:path/path.dart' as p;
 import 'package:stream_channel/isolate_channel.dart';
 import 'package:stream_channel/stream_channel.dart';
 
 import '../util/dart.dart' as dart;
 
+import 'package:test_api/src/backend/suite.dart'; // ignore: implementation_imports
 import 'package:test_api/src/util/remote_exception.dart'; // ignore: implementation_imports
 
 /// Spawns a hybrid isolate from [url] with the given [message], and returns a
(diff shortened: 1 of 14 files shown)


jakemac53:
I will follow up with pubspec/changelog updates I see how the tests do, and after the next release to avoid conflicts.

pkgs/test_core/lib/src/runner/spawn_hybrid.dart
```diff
@@ -19,7 +21,8 @@ import 'package:test_api/src/util/remote_exception.dart'; // ignore: implementat
 /// This connects the main isolate to the hybrid isolate, whereas
 /// `lib/src/frontend/spawn_hybrid.dart` connects the test isolate to the main
 /// isolate.
-StreamChannel spawnHybridUri(String url, Object? message) {
+StreamChannel spawnHybridUri(String url, Object? message, Suite suite) {
```
natebosch, line 24: Is this a breaking change?

jakemac53: This is an internal only method (it is what actually spawns the isolate). Confusing because it has the same name as the public one though.

jakemac53:
There are some additional questions here:

## Should we allow specifying a package config as a new named argument?

This would be necessary if you want to spawn a process that is outside your package.

Do we support that? Does anybody do that?

## How/when should we enable strong mode?

We bootstrap the file here https://github.com/dart-lang/test/blob/ca778785e98df49d8f47d363628eee8ec96e69a3/pkgs/test_core/lib/src/runner/spawn_hybrid.dart#L28 and pass that code as a `data:` uri to `Isolate.spawnUri`.

Probably what makes the most sense is to fully compute the language version of the uri we are given, and then inline a language version marker into the file. To compute this we would need to both use the packag config to determine the language version of the Uri, and then also parse the file for any language markers with the analyzer.

jakemac53:
For now I am focusing on copying language comments into the bootstrap code to match the language version of the uri being spawned, and I am not planning on adding support for a custom package config for now.

natebosch:
> For now I am focusing on copying language comments into the bootstrap code to match the language version of the uri being spawned, and I am not planning on adding support for a custom package config for now.
> 
> This also means we won't have a supported way of _disabling_ sound null safety mode, when the hybrid uri you want to launch is opted in.

I think that limitation is fine. There is always the option of wrapping the entrypoint yourself with a library that opts out.

jakemac53:
https://travis-ci.org/github/dart-lang/test/builds/741463786 is the pending build for the last commit, not sure why github isn't picking it up

natebosch (approved):

PR #1370, opened 2020-11-04
Fix legacy import violations in opted in libraries

jakemac53:
Part of https://github.com/dart-lang/sdk/issues/44063

Had to opt out several libraries and undo their migrations, but we were in a bad state previously where we could have been broken by migrations of several packages.

### DIFF
--- a/pkgs/test_core/lib/src/direct_run.dart
+++ b/pkgs/test_core/lib/src/direct_run.dart
@@ -1,6 +1,8 @@
 // Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+//
+// @dart=2.9
 
 import 'dart:async';
 import 'dart:collection';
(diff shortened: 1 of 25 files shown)


pkgs/test_core/lib/src/direct_run.dart
```diff
@@ -80,7 +83,7 @@ Future<bool> _directRunTests(FutureOr<void> Function() testMain,
       throw MissingTestException(fullTestName);
     }
   }
-  return success!;
+  return success;
```
natebosch, line 86: Is it worth keeping a `/*!*/` comment here?

natebosch (approved):

pkgs/test_core/lib/src/direct_run.dart
jakemac53: I don't think so, they should just get inserted by the migration tool. I didn't actually realize I could... but going back and fixing them would be pretty cumbersome at this point lol.

natebosch: SGTM. We have the git history if we need it.

PR #1371, opened 2020-11-09
Update package_config dependencies to allow versions >2.0.0.

lrhn:
The 2.0.0 versions will remove deprecated backwards compatibility functionality.
The test package no longer needs this functionality, and since the test package
depends on pkg:package_config, it's impossible to build a 2.0.0 version before test
accepts it.

### DIFF
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test
-version: 1.16.0-nullsafety.10
+version: 1.16.0-nullsafety.11
 description: A full featured library for writing and running Dart tests.
 homepage: https://github.com/dart-lang/test/blob/master/pkgs/test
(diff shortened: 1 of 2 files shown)


jakemac53:
You should use dependency overrides on package:test to force pub to allow it.

The package_config package can then release null safety as a breaking change (as well as deleting the old apis), and we can update to require that version shortly after.

natebosch:
+1 on using `dependency_overrides` before publish since is a dev dependency.

harkairt:
we would like to use `bloc_test` with NNBD environment and appearently it is blocked by this pull request. can this be merged? 
thank you! <3

jakemac53:
This looks obsolete to me, going to go ahead and close. This was updated in some other PR along the way since this was sent out.

PR #1381, opened 2020-11-19
[WIP] add root pkg to ease mono_repo usage, start working on node issues

jakemac53:


jakemac53:
actually ive got things running locally now so ill go with that

PR #1383, opened 2020-11-20
Make the RemoteListener channel non-nullable

natebosch:
This channel had been `StremChannel<Object?>` primarily because the
`jsonDocument` transformer could emit `null`. Filter out nulls where
they are decoded to keep the rest of the channel non-nullable.

Refactor the `socketChannel` utility method to resolve casting
difficulties with `StreamChannelTransfomer` APIs. It is easier to
separately deal with the `Stream` and `Sink` portions of the channel
because they have separate needs around casting.

Make `RemoteChannel.start` and `serializeSuite` non-nullable channels.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v2.5.0
+# Created with package:mono_repo v3.0.0
 language: dart
 
 # Custom configuration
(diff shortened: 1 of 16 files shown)


jakemac53 (approved):

jakemac53:
Do we want to close this one now?

natebosch:
This causes some other failures that are hard to track down. I do think it would be a good idea in general to make this channel non-nullable, but I'm not sure it's worth the effort for now.

Issue #1382, opened 2020-11-19
Hybrid testing is choking on Windows paths


nex3:
A recent release of the `test` package seems to have broken hybrid tests on Windows. See [this failing Travis build](https://travis-ci.com/github/sass/dart-sass/jobs/443956550) for an example error:

```
00:00 +0 -1: [VM] test\cli\node\colon_args_test.dart: (setUpAll) [E]
  Invalid argument(s): Illegal character in path
  dart:core                                              new _Uri.file
  package:analyzer/src/string_source.dart 24:53          new StringSource
  package:analyzer/dart/analysis/utilities.dart 80:16    parseString
  package:test_core/src/runner/spawn_hybrid.dart 124:16  _languageVersionCommentFor
  ===== asynchronous gap ===========================
  package:test_core/src/runner/spawn_hybrid.dart 42:17   spawnHybridUri.<fn>
  package:test_core/src/runner/spawn_hybrid.dart 81:4    spawnHybridUri
  package:test_core/src/runner/runner_test.dart 74:13    RunnerTest.load.<fn>.<fn>
```

As far as I can tell, there's nothing unusual about the path on Travis other than being in Windows format.

Note that it's possible that this is an error with the `analyzer` package. I haven't been able to debug deeper because I don't have access to a Windows development machine locally.

jakemac53:
The analyzer wasn't in the loop here at all before and I don't think it accepts windows file paths so it looks like we will need to convert it.

PR #1384, opened 2020-11-20
Fix spawnHybridUri on windows

jakemac53:
Previously we were getting uris like `file:///C:/C:/<stuff>`. I had to do some funky things to get rid of the leading `/` from the `C:` style uris as well.

I also noticed that the uris I get back from `Isolate.resolvePackageUri` are always posix style and not windows style paths, so this keeps the consistency with that where previously we ended up with mixes of `\` and `/` in some places.

This will also probably have to be backported.

Fixes https://github.com/dart-lang/test/issues/1382

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.16.0-nullsafety.12
 
-* Update to the latest `test_core`.
+* Fix `spawnHybridUri` on windows.
 
 ## 1.16.0-nullsafety.11
 
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.16.0-nullsafety.12
+
+* Update to the latest `test_core`.
+
 ## 1.16.0-nullsafety.11
 
 * Set up a stack trace mapper in precompiled mode if source maps exist. If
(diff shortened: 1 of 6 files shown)


pkgs/test/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.16.0-nullsafety.12
+
+* Update to the latest `test_core`.
```
natebosch, line 3: Can you call out the specific fix instead?

pkgs/test/test/io.dart
```diff
@@ -17,7 +17,11 @@ import 'package:test/test.dart';
 /// The path to the root directory of the `test` package.
 final Future<String> packageDir =
     Isolate.resolvePackageUri(Uri(scheme: 'package', path: 'test/'))
-        .then((uri) => p.dirname(uri.path));
+        .then((uri) {
+  var dir = p.dirname(uri.path);
+  if (dir.startsWith('/C:')) dir = dir.substring(1);
```
natebosch, line 22: Can we rely on this _always_ being the `C:` drive?

jakemac53: This is just for testing, so ya for now.

natebosch (approved):

nex3:
Can you backport this to the stable branch as well?

jakemac53:
That is the plan yes, but I likely won't have time until Monday

jakemac53:
Filed https://github.com/dart-lang/test/issues/1387 to track

PR #1390, opened 2020-11-22
Enable GitHub Actions

kevmoo:


jakemac53:
Looks like 2 consistent failures happening here that we will need to diagnose

kevmoo:
> Looks like 2 consistent failures happening here that we will need to diagnose

Indeed!

jakemac53:
The fix to set up the force exit env variable looks to have resolved things, we can  re-upload once we have a real mono_repo version and should be g2g!

jakemac53 (approved):

PR #1393, opened 2020-11-23
Attempt at forwards compatible fix for pre-nnbd branch tests

jakemac53:
Updates the default package config for tests so that it adds an entry for the sandbox dir.

- Sets the default language version to that of the test package.

### DIFF
--- a/pkgs/test/test/io.dart
+++ b/pkgs/test/test/io.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:io';
 import 'dart:isolate';
 
+import 'package:package_config/package_config.dart';
 import 'package:path/path.dart' as p;
 import 'package:test_descriptor/test_descriptor.dart' as d;
 import 'package:test_process/test_process.dart';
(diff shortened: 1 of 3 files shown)


jakemac53:
I am not sure what we want to do about the hybrid tests - we could possibly change the ones that actually run to instead import an opted out library, and expect that the spawning fails due to running in strong mode.

That should be able to pass in both the master and pre-nnbd branches?

natebosch (approved):

natebosch:
> we could possibly change the ones that actually run to instead import an opted out library, and expect that the spawning fails due to running in strong mode.

I'm not sure I understand. We do need some to run. Which test cases specifically were you thinking of changing in this way?

jakemac53:
> We do need some to run.

Yes, agreed that we need to ensure you can communicate back and forth. But all those tests would be ran in opted out mode to be compatible.

> Which test cases specifically were you thinking of changing in this way?

Any hybrid test which runs in opted in mode (anything with a data uri or that runs a non-opted out script in a sandbox dir).

natebosch:
I think the only thing we couldn't catch that way would be for being able to successfully run with opt in, which does exercise some behavior that could be broken in strong mode. I don't know how likely we are to break the code in that way specifically that would be caught.

jakemac53:
Ya my sense is that it would be pretty unlikely. We will have confirmation that it can _attempt_ to run in strong mode, and that it can successfully run in weak mode.

To me, the value of being able to test the `pre-nnbd` branch would outweigh the negatives.

jakemac53:
Alternatively, we could land this as is (and merge it into master with the changelog entry). It still fixes a lot of tests and we could hand analyze the remaining failures each time.

natebosch:
Could we skip the test on the `pre-nnbd` branch? Maybe with a tag, or a hardcoded boolean variable or something? Then when merging in to master we'd be careful to keep that different between the two branches?

jakemac53:
We could skip the tests, but the spawn hybrid apis are where a bunch of fixes have gone - so I would like to have test coverage for it.

You also wouldn't get a merge conflict I don't think - depending on how it was done. So that could make it harder to remember to revert those changes.

PR #1399, opened 2020-11-30
[WIP] use frontend server for VM test platform

jonahwilliams:
A first pass at wiring up the frontend server in the VM platform, similar to how flutter test works for compilation on the fly.

with_fe_server


pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,273 @@
...
+    }
+  }
+
+  Future<void> dispose() async {
+    await _compilerController.close();
+    await _shutdown();
+  }
+
+  Future<String> _languageVersionComment(Uri testUri) async {
+    var localPackageConfig = _packageConfig ??= await loadPackageConfig(File(
```
jonahwilliams, line 67: The existing language version API didn't work for some reason, maybe because of how I was running test from source from another project? It just picked up the test package config and not the project package config

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,273 @@
...
+      if (outputPath == null || compilerOutput!.errorCount > 0) {
+        request.result.complete(null);
+        await _shutdown();
+      } else {
+        final outputFile = File(outputPath);
+        final kernelReadyToRun = await outputFile.copy('${tempFile.path}.dill');
+        final testCache = File(_dillCachePath);
+        if (firstCompile ||
+            !testCache.existsSync() ||
+            (testCache.lengthSync() < outputFile.lengthSync())) {
```
jakemac53, line 128: Worth commenting on this condition - is the assumption that larger file == more relevant for future runs basically?

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,273 @@
...
+
+  Uri mainUri;
+  Metadata metadata;
+  Completer<String> result;
+}
+
+class TestCompiler {
+  TestCompiler(this._dillCachePath)
+      : _outputDillDirectory =
+            Directory.systemTemp.createTempSync('dart_test.') {
```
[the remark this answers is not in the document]
jakemac53: Ah I see you keep a single cached dill, since you can only initialize for one anyways, that sgtm.

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  Future<_CompilerOutput?> _createCompiler(Uri testUri) async {
+    final frontendServer = path.normalize(path.join(Platform.resolvedExecutable,
+        '..', 'snapshots', 'frontend_server.dart.snapshot'));
+    final sdkRoot = Directory(
+            path.relative(path.join(Platform.resolvedExecutable, '..', '..')))
+        .uri;
+    final platformDill = 'lib/_internal/vm_platform_strong.dill';
+    final process = await Process.start(Platform.resolvedExecutable, <String>[
```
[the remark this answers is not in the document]
jonahwilliams: > Does frontend server auto-detect for sound null safety?

Yes

> Does it deal well with switching between the modes between compiles? 

Let me double check that...

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,273 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+import 'dart:math' as math;
+import 'dart:typed_data';
+
+// ignore: import_of_legacy_library_into_null_safe
```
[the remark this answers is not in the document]
jonahwilliams: Fixed

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,273 @@
...
+    _stdoutHandler.reset(expectSources: false);
+  }
+
+  Future<_CompilerOutput?> _createCompiler(Uri testUri) async {
+    final frontendServer = path.normalize(path.join(Platform.resolvedExecutable,
+        '..', 'snapshots', 'frontend_server.dart.snapshot'));
+    final sdkRoot = Directory(
+            path.relative(path.join(Platform.resolvedExecutable, '..', '..')))
+        .uri;
+    final platformDill = 'lib/_internal/vm_platform_strong.dill';
```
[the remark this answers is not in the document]
jonahwilliams: I don't see a dill specifically documented as agnostic, I will follow up on a chat to see if this is indeed the case

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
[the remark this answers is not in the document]
jonahwilliams: The frontend server handles validation of changes in SDK version/source files. There have been bugs in the past, such as when null safety was first introduced but in my experience they usually get fixed pretty quickly

jonahwilliams:
Updated to use the frontend_server_client package. Unfortunately that doesn't yet have the --no-print-incremental-dependencies option so this slows down _slightly_, but not a huge deal. Someone should do the math to figure out when that is safe to use :)

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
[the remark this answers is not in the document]
jonahwilliams: This is an agnostic dill

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,274 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
jonahwilliams: Fixed

jonahwilliams:
I believe the current issue is that the test_on_test.dart (and possibly others). Generate a dart file without a package config, so the compiler cannot resolve anything. This works from source since test will use the package config of the current isolate.

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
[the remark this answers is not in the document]
jonahwilliams: (update: yes)

jonahwilliams:
Looks like I need a new version of the frontend_server client package, but otherwise I think I can get the rest going...

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,154 @@
...
+      _compilationQueue.removeAt(0);
+    }
+  }
+
+  Future<CompileResult?> _createCompiler(Uri testUri) async {
+    final platformDill = 'lib/_internal/vm_platform_strong.dill';
+    final sdkRoot =
+        Directory(p.relative(p.join(Platform.resolvedExecutable, '..', '..')))
+            .uri;
+    var client = _frontendServerClient = await FrontendServerClient.start(
```
jonahwilliams, line 145: I believe I added an option to skip printing incremental dependencies - you should definitely turn that on to get another 20-30% speedup on large numbers of tests

jakemac53: done 👍

jakemac53:
Added you @natebosch to get another set of eyes, probably not appropriate for me to LGTM at this point given I have touched a fair bit of it as well lol

pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -107,6 +107,12 @@ final ArgParser _parser = (() {
       help: "Don't rerun tests that have retry set.",
       defaultsTo: false,
       negatable: false);
+  parser.addFlag('use-data-isolate-strategy',
+      help: 'Use `data:` uri isolates when spawning VM tests instead of the\n'
+          'default strategy. This may be faster when you only ever running\n'
```
grouma, line 112: [nit] when you are* only ever running

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,155 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```
grouma, line 1: [nit] 2021

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,155 @@
...
+
+  final String _dillCachePath;
+  final Directory _outputDillDirectory;
+  final _compilerController = StreamController<_CompilationRequest>();
+  final _compilationQueue = <_CompilationRequest>[];
+
+  late final File _outputDill;
+  FrontendServerClient? _frontendServerClient;
+
+  Future<CompilationResponse> compile(Uri mainDart, Metadata metadata) {
```
grouma, line 49: This could use a doc comment.

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,155 @@
...
+        client.reset();
+      }
+      _compilationQueue.removeAt(0);
+    }
+  }
+
+  Future<CompileResult?> _createCompiler(Uri testUri) async {
+    final platformDill = 'lib/_internal/vm_platform_strong.dill';
+    final sdkRoot =
+        Directory(p.relative(p.join(Platform.resolvedExecutable, '..', '..')))
```
grouma, line 143: Are we sure this works both internal and external. I ran into issues where the paths are different for our internal build environment and I needed to pass that data through. See https://github.com/dart-lang/webdev/pull/1222

grouma (approved):
LGTM besides one question.

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
jakemac53: Well, this PR was created in 2020 (and the commit for this file :P)

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
jakemac53: what would it say?

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
jakemac53: It does not work internally without additional configuration, but we won't use this strategy internally (we will use --use-data-isolate-strategy).

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
jakemac53: (I don't think for instance `Compiles [mainDart] with [_compiler]` would be a useful comment)

natebosch: Describing the failure behavior would be worthwhile. It could also be useful to describe how this doesn't necessarily start the compilation right away.

```dart
/// Enqueues a request to compile [mainDart] and returns the result.
///
/// This request may need to wait for ongoing compilations.
/// If [dispose] was called, the returned future will immediately complete with `null`.
/// The entrypoint [mainDart] is wrapped in a script to`internalBootstrapVmTest`.
```

I'm not sure that `null` is the right behavior for that - maybe we should throw instead? See my other comment as well about this path getting a test.

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,155 @@
...
+  CompilationResponse(
+      {this.compilerOutput, this.errorCount = 0, this.kernelOutputUri});
+}
+
+class TestCompiler {
+  TestCompiler(this._dillCachePath)
+      : _outputDillDirectory =
+            Directory.systemTemp.createTempSync('dart_test.') {
+    _outputDill = File(p.join(_outputDillDirectory.path, 'output.dill'));
+    _compilerController.stream.listen(_onCompilationRequest);
```
natebosch, line 38: I don't follow what the stream is accomplishing for us when we also have a separate queue.

Can we either omit the stream

```dart
Future<CompilationResponse> compile(Uri mainDart, Metadata metadata) {   
    if (_isClosed) {
      return Future.value(null);
    }
    final completer = Completer<CompilationResponse>();
    if(_compilationQueue.isEmpty) Future(_drainQueue);
    _compilationQueue.add(_CompilationRequest(mainDart, completer, metadata));
    return completer.future;
 }
 
 Future<void> _drainQueue() async {
   while (_compilationQueue.isNotEmpty) {
     // ...
 ```
  
  Or use the stream as the queue
  
  ```dart
  void _runCompiles(Stream<_CompilationRequest> requests) async {
    await for (final request in requests) {
      var firstCompile = false;
      CompileResult? compilerOutput;
      final contents =
          await _generateEntrypoint(request.mainUri, request.metadata);
      final tempFile = File(p.join(_outputDillDirectory.path, 'test.dart'))
        ..writeAsStringSync(contents);
      // Rest of _onCompilationRequest implementation...
 ```

Or we can use `package:pool` to handle this in `compile` and skip both the stream and the queue

```dart
  final _compilePool = Pool(1);
  Future<CompilationResponse> compile(Uri mainDart, Metadata metadata) {
    return _compilePool.withResource(() => _compile(mainDart, metadata));
  }
  Future<CompileResult> _compile(Uri mainDart, Metadata metadata) async {
    var firstCompile = false;
    CompileResult? compilerOutput;
    final contents = await _generateEntrypoint(mainUri, metadata);
    final tempFile = File(p.join(_outputDillDirectory.path, 'test.dart'))
      ..writeAsStringSync(contents);

    if (_frontendServerClient == null) {
      compilerOutput = await _createCompiler(tempFile.uri);
      firstCompile = true;
    } else {
      compilerOutput =
          await _frontendServerClient!.compile(<Uri>[tempFile.uri]);
    }
    // The client is guaranteed initialized at this point.
    final client = _frontendServerClient!;
    final outputPath = compilerOutput?.dillOutput;
    if (outputPath == null) {
      return CompilationResponse(
          compilerOutput: compilerOutput?.compilerOutputLines.join('\n'),
          errorCount: compilerOutput?.errorCount ?? 0);
    } else {
      final outputFile = File(outputPath);
      final kernelReadyToRun = await outputFile.copy('${tempFile.path}.dill');
      final testCache = File(_dillCachePath);
      // Keep the cache file up-to-date and use the size of the kernel file
      // as an approximation for how many packages are included. Larger files
      // are prefered, since re-using more packages will reduce the number of
      // files the frontend server needs to load and parse.
      if (firstCompile ||
          !testCache.existsSync() ||
          (testCache.lengthSync() < outputFile.lengthSync())) {
        if (!testCache.parent.existsSync()) {
          testCache.parent.createSync(recursive: true);
        }
        await outputFile.copy(_dillCachePath);
      }
      client.accept();
      client.reset();
      return CompilationResponse(
          compilerOutput: compilerOutput?.compilerOutputLines.join('\n'),
          errorCount: compilerOutput?.errorCount ?? 0,
          kernelOutputUri: kernelReadyToRun.absolute.uri);
    }
  }
```

natebosch (changes requested):

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
jakemac53: Changing the behavior to return an actual `CompilerResponse` with an error saying it was shut down.

I also added some comments as suggested.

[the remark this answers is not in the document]
jakemac53: I changed this to not return null

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
jakemac53: Updated to use package:pool

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -0,0 +1,142 @@
...
+class TestCompiler {
+  final _compilePool = Pool(1);
+  final String _dillCachePath;
+  final Directory _outputDillDirectory;
+
+  late final _outputDill =
+      File(p.join(_outputDillDirectory.path, 'output.dill'));
+  FrontendServerClient? _frontendServerClient;
+
+  TestCompiler(this._dillCachePath)
```
natebosch, line 34: [nit] Can we add a comment either here or on the class level about how instantiating this object doesn't start the process and it instead waits until the first request? That makes it more clear in the usage location that we don't need to worry about waiting.

jakemac53:
There is a bit of an edgecase that still exists it looks like if you ctrl+c while the compiler is running. I am going to clean that up. I am not sure how we can consistently test that, but I am going to think on it and also manually verify the behavior.

natebosch (approved):

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
jakemac53: done

PR #1405, opened 2020-12-30
Remove unnecessary import

srawlins:
The analyzer will start reporting this when dart-lang/sdk#44569 is implemented.


natebosch:
Do you know which symbols we were getting from `dart:collection` that are also exported in one of the other imports or in `dart:core`?

natebosch (approved):

srawlins:
Looks like things like UnmodifiableMapView.

PR #1407, opened 2021-01-13
Roll vm_service to nullsafety

dnfield:
/cc @bkonyi 

Flutter will need this to roll null-safe vm_service because we depend on test_core.

### DIFF
--- a/pkgs/test_core/pubspec.yaml
+++ b/pkgs/test_core/pubspec.yaml
@@ -25,7 +25,7 @@ dependencies:
   source_span: '>=1.8.0-nullsafety <1.8.0'
   stack_trace: '>=1.10.0-nullsafety <1.10.0'
   stream_channel: ">=2.1.0-nullsafety <2.1.0"
-  vm_service: '>=1.0.0 <6.0.0'
+  vm_service: '>=6.0.0-nullsafety-dev <6.0.0'
   yaml: ">=2.0.0 <4.0.0"
   # matcher is tightly constrained by test_api
   matcher: any


pkgs/test_core/pubspec.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
   source_span: '>=1.8.0-nullsafety <1.8.0'
   stack_trace: '>=1.10.0-nullsafety <1.10.0'
   stream_channel: ">=2.1.0-nullsafety <2.1.0"
-  vm_service: '>=1.0.0 <6.0.0'
+  vm_service: '>=6.0.0-nullsafety-dev <6.0.0'
```
jakemac53, line 28: What is up with the `-dev` here? Are we expecting breaking changes coming to this release, because we can't really support it in this package then if we could be broken at any time...

jakemac53:
It looks like `coverage` also needs to be updated - it will probably be easier to start there, with a dependency override in that package for the test packages. That should be lower risk than doing the overrides in this package.

pkgs/test_core/pubspec.yaml
dnfield: @bkonyi

dnfield:
Can we just increase the constraint on coverage rather than making it tight and requiring a dependency_override?

pkgs/test_core/pubspec.yaml
bkonyi: This isn't quite ready to be rolled. Waiting on some initial testing from @grouma and @kenzieschmoll to confirm we don't break anything in DWDS and DevTools, at which point I'll publish a non-dev version.

dnfield: Ok. I'll close this for now then to get it off review queues.

In that case, flutter is still blocked on migrating driver/integration test (/cc @goderbauer @Hixie FYI)

dnfield:
f

Issue #1409, opened 2021-01-15
Migrate to null-safe vm_service


dnfield:
This depends on https://github.com/dart-lang/coverage/issues/321

PR #1410, opened 2021-01-15
Support the latest vm_service

grouma:
Closes https://github.com/dart-lang/test/issues/1409

### DIFF
--- a/pkgs/test_core/pubspec.yaml
+++ b/pkgs/test_core/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test_core
-version: 0.3.12-nullsafety.14
+version: 0.3.12-nullsafety.15
 description: A basic library for writing tests and running them on the VM.
 homepage: https://github.com/dart-lang/test/blob/master/pkgs/test_core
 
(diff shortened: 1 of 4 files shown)


pkgs/test_core/pubspec.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
   source_span: '>=1.8.0-nullsafety <1.8.0'
   stack_trace: '>=1.10.0-nullsafety <1.10.0'
   stream_channel: ">=2.1.0-nullsafety <2.1.0"
-  vm_service: '>=1.0.0 <6.0.0'
+  vm_service: '>=1.0.0 <7.0.0'
```
jakemac53, line 28: Assuming this is a null safety release we need to pin the lower bound to that, we don't want to be in a situation where we allow both the migrated and unmigrated versions.

natebosch: I disagree. Where the changes are non-breaking for un-migrated packages we will have an easier time churning the ecosystem if we keep wide ranges and avoid unnecessarily constrained solves.

jakemac53: I think that is pretty risky though, we could easily opt in a library that imports this package and requires the opted in version without realizing it.

natebosch: We chatted about this and landed on keeping the wide ranges where we can.

There is risk of making a mistake later. The `test` package is already a bit of a linchpin when it comes to this migration and pub constraints - so risk is already high. Since we don't have bandwidth to write the tooling we'd need to really keep ourselves on the rails (like running tests at min versions of all packages) we think it's OK to go with a strategy of "we just need to be careful with the test package". In that spirit - I'll be keeping an eye on our migrations and code changes in the test packages and aware of the fact that new libraries, or newly migrated libraries need particular scrutiny. For the sake of easier pub solving elsewhere we will take on the risk of wide constraints here.

natebosch (approved):

PR #1411, opened 2021-01-20
Remove null safety CI flag

grouma:
Remove `--enable-experiment=non-nullable` flag from CI.


grouma:
Close in favor of https://github.com/dart-lang/test/pull/1412

PR #1414, opened 2021-01-27
Remove raw types from test_api; make make implicit dynamics explicit

srawlins:
I'm happy to revert any of the `Map<dynamic, dynamic>` stuff or `StreamChannel<dynamic>` stuff, but I think it's important to remove other dynamic values. If a function returns `Future<void>` instead of `Function`, then you'll get warned about using the result, etc.


pkgs/test_api/lib/src/frontend/expect_async.dart
```diff
@@ -176,7 +176,7 @@ class _ExpectedFunction<T> {
       _run([a0, a1, a2, a3, a4, a5].where((a) => a != placeholder));
 
   /// Runs the wrapped function with [args] and returns its return value.
-  T _run(Iterable args) {
+  T _run(Iterable<Object?> args) {
```
grouma, line 179: Why is this not dynamic here?

grouma (approved):

grouma:
We should bump the version and update the changelog.

pkgs/test_api/lib/src/frontend/expect_async.dart
srawlins: I think that we do not need to use dynamic dispatch, so `Object?` is a safer type to use. But I can change to `dynamic` if that is preferred for this repository.

jakemac53: We should go with `Object?` unless we really want dynamic in a specific place for some reason

srawlins:
I bumped the version and added a changelog entry with 4 breaking change notes. Let me know if any of those are considered internal API or should otherwise be removed from the changelog.

jakemac53 (changes requested):
We need to hold on this for a followup to the stable release which is going to get pinned in flutter and needs to match exactly what they previously had (be a version number only change).

Please hold until we finish that (should be done today)

natebosch:
@srawlins - is there a reason you didn't enable the analysis options?

@jakemac53 - is this something you think we should try to land?

srawlins:
> is there a reason you didn't enable the analysis options?

I couldn't figure out how the analysis options and the mono repo work. I think I manually copied the analysis options file into each package, but did go so far as to mail that change. I think without manually copying the file, analyzer complained that it couldn't find the pedantic yaml file.

natebosch:
I added the options - `test_core` hadn't been fixed yet so I went ahead and fixed the static analysis there.

I went with `Object?` in a bunch of places instead of `dynamic`. In one additional place this appears to have been breaking, so I might go back to `dynamic` even though I prefer `Object?`. That place is the `suiteChannel` stream becoming a `Stream<Object?>` instead of `Stream<dynamic>` which triggers the need for explicit casts.

https://github.com/srawlins/test/blob/20418d10da89b6b78509d582d0cd9ff699916df6/pkgs/test/lib/src/bootstrap/browser.dart#L18-L19

I think that's probably a sane thing to do. @jakemac53 what do you think?

pkgs/test_api/lib/src/frontend/future_matchers.dart
```diff
@@ -106,8 +106,8 @@ class _DoesNotComplete extends Matcher {
   }
 
   @override
-  Description describeMismatch(
-      item, Description description, Map matchState, bool verbose) {
+  Description describeMismatch(item, Description description,
```
jakemac53, line 109:
```suggestion
  Description describeMismatch(dynamic item, Description description,
```

pkgs/test_api/lib/src/frontend/spawn_hybrid.dart
```diff
@@ -89,7 +89,8 @@ final _transformer = StreamChannelTransformer<dynamic, dynamic>(
 ///
 /// **Note**: If you use this API, be sure to add a dependency on the
 /// **`stream_channel` package, since you're using its API as well!
-StreamChannel spawnHybridUri(uri, {Object? message, bool stayAlive = false}) {
+StreamChannel<dynamic> spawnHybridUri(uri,
```
jakemac53, line 92: This change would be technically statically breaking... if we have other breakages anyways though.... 

```suggestion
StreamChannel<dynamic> spawnHybridUri(Object uri,
```

We could also do `Object?` instead but `null` isn't actually allowed and will fail at runtime...

jakemac53:
> That place is the `suiteChannel` stream becoming a `Stream<Object?>` instead of `Stream<dynamic>` which triggers the need for explicit casts.

Ya I remember trying to change these doing the original migration and giving up due to the complexity of getting all the casts where needed. If you want to try and push it through by all means.

jakemac53:
Lets either decide to try and land this or close it?

It is pretty out of date now...

srawlins:
I'll close. I've moved on. 😬

PR #1416, opened 2021-01-29
pkg:test: support null-safe version of pkg:shelf

kevmoo:


scheglov (approved):

pkgs/test/pubspec.yaml
```diff
@@ -39,9 +39,3 @@ dev_dependencies:
   shelf_test_handler: ^1.0.0
   test_descriptor: ^1.0.0
   test_process: ^1.0.0
-
-dependency_overrides:
```
jakemac53, line 43: For the future we leave these in always and remove locally for publishing. That ensures our testing is always against the latest packages in the repo.

kevmoo:
ack! sorry

PR #1417, opened 2021-02-02
Expand null safe constraints

natebosch:
Change all constraints with an upper bound which prevents the stable
version for null safety into caret constraints which allow them.

Expand constraint on `package_config` to allow `2.x`.

### DIFF
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test
-version: 1.16.0-nullsafety.17
+version: 1.16.0-nullsafety.18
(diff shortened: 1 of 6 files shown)


pkgs/test/pubspec.yaml
```diff
@@ -8,34 +8,41 @@ environment:
 
 dependencies:
   analyzer: '>=0.36.0 <0.42.0'
```
pq, line 10: Can this get widened at the same time?

https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/pubspec.yaml

jakemac53 (approved):

PR #1421, opened 2021-02-03
Prepare to publish test_api for stable null safety

natebosch:
- Drop the `-nullsafety` prerelease suffix from the version.
- Bump to the stable versions for all normal dependencies.
- Bump to the stable versions for all `test_api` dependencies from
  `test` and `test_core`.

Only `test_api` is going stable for now. `test` and `test_core` will go
to stable after the remainder of their dependencies go stable.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.16.0-nullsafety.19
+
+* Use the `test_api` for stable null safety.
+
 ## 1.16.0-nullsafety.18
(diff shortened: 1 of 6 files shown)


pkgs/test/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.16.0-nullsafety.19
```
jakemac53, line 1: Should we push these to stable also?

jakemac53 (approved):

PR #1435, opened 2021-02-18
pkg:test - Support latest web_socket_channel

kevmoo:
Also
* bump test_process dev dependencies
* use repository over homepage

TEMP: use git dependencies everywhere to validate CI before we land


grouma (approved):
LGTM assuming you remove git overrides.

PR #1437, opened 2021-02-23
Allow stdin to be injected

jiahaog:
`stdin` is a global property from `dart:io` that returns a single consumer stream which can only be listened to once. This creates problems when different libraries need to access stdin as part of the same process, without coordination. Listening on `stdin` more than once will cause a `StateError`.

As such, instead of reading `stdin` from `dart:io` and saving a handle to it globally, accept it as a `Stream<List<int>>`, and pass it through to where it is needed. This will allow dependents on this library, such as the Flutter Tools to provide a broadcast stream that can be managed by the respective libraries independently.


jiahaog:
Synced offline - this is no longer needed by Flutter Tools now that #1438 has been merged.

PR #1438, opened 2021-02-23
Don't create stdinLines just to cancel it

kevmoo:
Prepare to release "dot" releases of pkg:test_core and pkg:test

I plan to backport these as well...

### DIFF
--- a/pkgs/test_core/lib/src/util/io.dart
+++ b/pkgs/test_core/lib/src/util/io.dart
@@ -59,11 +59,7 @@ StreamQueue<String> get stdinLines => _stdinLines ??= StreamQueue(
 StreamQueue<String>? _stdinLines;
 
 /// Call cancel on [stdinLines], but only if it's been accessed previously.
-void cancelStdinLines() {
-  if (_stdinLines != null) {
-    _stdinLines!.cancel(immediate: true);
-  }
-}
+void cancelStdinLines() => _stdinLines?.cancel(immediate: true);
 
 /// Whether this is being run as a subprocess in the test package's own tests.
 bool inTestTests = Platform.environment['_DART_TEST_TESTING'] == 'true';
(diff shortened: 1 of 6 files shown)


pkgs/test_core/lib/src/util/io.dart
```diff
@@ -54,9 +53,18 @@ SuitePlatform currentPlatform(Runtime runtime) => SuitePlatform(runtime,
...
-final stdinLines = StreamQueue(
+StreamQueue<String> get stdinLines => _stdinLines ??= StreamQueue(
     Platform.isFuchsia ? Stream<String>.empty() : lineSplitter.bind(stdin));
 
+StreamQueue<String>? _stdinLines;
+
+/// Call cancel on [stdinLines], but only if it's been accessed previously.
+void cancelStdinLines() {
+  if (_stdinLines != null) {
+    _stdinLines!.cancel(immediate: true);
```
natebosch, line 64:
```suggestion
    _stdinLines?.cancel(immediate: true);
```

natebosch (changes requested):

pkgs/test_core/lib/src/util/io.dart
kevmoo: I don't see how we could have a race condition here...

kevmoo: ...I mean, sure. if you want. Be hyper defensive?

kevmoo: Should I set it to null after then?

pkgs/test_core/lib/src/util/io.dart
```diff
@@ -54,9 +53,18 @@ SuitePlatform currentPlatform(Runtime runtime) => SuitePlatform(runtime,
...
+  }
```
natebosch, line 65:
```suggestion
  _stdinLines?.cancel(immediate: true);
```

Sorry, forgot that we can do multiline suggestions now.

Remove the conditional and only use `?.`.

kevmoo:
LGTM @natebosch ?

natebosch (approved):

PR #1442, opened 2021-02-28
update test_core and test to allow latest published deps

jakemac53:
Also removes some unused deps from test_api, and unified quote style in pubspecs.

Waiting to publish until we can update node_preamble (or may just add that here if it lands soon and then remove the dev).

Then we can finish the migration to null safety after that 🎉


natebosch (approved):

jakemac53:
Going to go ahead and publish these, there isn't a strong need to have a version of test that supports a broad range of `node_pramble` versions, so we can bump that with the null safety release.

PR #1451, opened 2021-03-07
fix missing split() call in docs

matthewadams:


matthewadams:
@googlebot I signed it!

pkgs/test/README.md
```diff
@@ -91,7 +91,7 @@ import 'package:test/test.dart';
 
 void main() {
   test('.split() splits the string on the delimiter', () {
-    expect('foo,bar,baz', allOf([
+    expect('foo,bar,baz'.split(), allOf([
```
kevmoo, line 94: `split` is not needed. All of the matchers below apply to String as-is.

matthewadams:
Oh I must’ve misunderstood the example 😊

PR #1453, opened 2021-03-08
prep for full null safety release

jakemac53:


natebosch:
If you don't mind I'd like to make sure that this also lands in the SDK before we publish.

https://dart-review.googlesource.com/c/sdk/+/187580

natebosch (approved):
🚀

kevmoo:
Maybe in a follow-up, but it'd be good to validate this against `2.12.0` in CI

PR #1455, opened 2021-03-08
fix handling of null remote exception messages

jakemac53:
Once this lands internally we can send another pr to publish new versions of test_core and test to allow this version.

### DIFF
--- a/pkgs/test_api/pubspec.yaml
+++ b/pkgs/test_api/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test_api
-version: 0.2.21-dev
+version: 0.3.0-dev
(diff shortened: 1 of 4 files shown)


natebosch (approved):

pkgs/test_api/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: test_api
-version: 0.2.20
+version: 0.2.21-dev
```
kevmoo, line 2: 0.3.0-dev?

jakemac53: good catch

PR #1460, opened 2021-03-12
Add scaffolding, expect, hooks libraries

natebosch:
Shuffles code around to make a more clear distinction between the
different types of use cases that are satisfied by the "frontend"
classes. Exposing an import which does not include `expect` and the
matchers will make it possible to write alternative matching frameworks
without worrying about name conflicts.

- `scaffolding` includes the code to define test cases and high level
  utilities that user level test code might need. A test might import
  `scaffolding` instead of `test` if they want to use a different
  matching framework and don't need `expect`.
- `expect` includes the synchronous and asynchronous matchers, as well
  as `expect` and `expectAsync*`.

Add a new library `test_api/hooks` which has lower level APIs for
interacting with the test runner. `expect` is written against only
`hooks`, and an alternative matching framework could have equivalent
capabilities using the same APIs. The primary capability which was
previously only available in `Invoker` is the ability to report that
some async work has not completed, and must complete before the test can
be considered done.

- Add `hooks.dart` and refactor all of `lib/src/expect/*` to use it.
- Move most code from `src/frontend` to either `src/scaffolding` or
  `src/expect` as appropriate.
- Make two copies of the `indent` utility function and specialize them
  by usage, since `test_core` and `expect` do not overlap in the
  combination of arguments used. Move remaining utilities used only by
  the `expect` implementations into their own utility library.

TODO:
We fold stack frames stemming from the `test*` packages specifically,
which means that frames beneath `expect` get handled for free. An
alternative matching framework defined outside this package would not
have the same advantage, so we may need to find an API to bring that to
parity.


pkgs/test_api/lib/src/scaffolding/utils.dart
```diff
@@ -0,0 +1,43 @@
...
+/// Awaiting this approximates waiting until all asynchronous work (other than
+/// work that's waiting for external resources) completes.
+Future pumpEventQueue({int times = 20}) {
+  if (times == 0) return Future.value();
+  // Use the event loop to allow the microtask queue to finish.
+  return Future(() => pumpEventQueue(times: times - 1));
+}
+
+/// Registers an exception that was caught for the current test.
+void registerException(Object error,
```
natebosch, line 23: cc @jakemac53 @grouma 

I didn't even know this existed until I stumbled across it while working on this change.

I've already excluded some deprecated APIs from `expect.dart` that are otherwise exposed through `test.dart`: https://github.com/dart-lang/test/blob/c5272043c5c6c524c3b3aaf5fa941e1dbb749b8d/pkgs/test_core/lib/test_core.dart#L12-L14

This way the new libraries come with extra API cleanup as we migrate.

We could do the same with `registerException` and avoid putting this in `scaffolding.dart` while retaining it in `test.dart` for backwards compatibility.

I like that both `fail` and `registerException` exist. They behave very differently - `fail` throws a specific type of exception to result in a test "failure", but not by communicating with the test framework directly - the exception needs to be able to bubble up the entire way. `registerException` avoids problems with `try/catch` in the user code because it reports to an error handler directly. It may result in a test "error", or a test "failure" if the exception type is specifically `TestFailure`. I don't think the distinction between "error" and "failure" is actually useful, and our users in practice are accustomed to being able to throw any exception to fail a test. https://github.com/dart-lang/test/issues/1465

Both `fail` and `registerException` have issues with error handling zones. This is a problem we could solve with `registerException`, but can't with `fail`.

Should we consider starting with a new API in `scaffolding.dart` to avoid these issues? For instance we can fairly easily work around the problem of error zones in user code by reporting to the test runner directly instead of reporting to a zone uncaught error handler.

grouma: > Should we consider starting with a new API in scaffolding.dart to avoid these issues?

I think it makes sense. To be clear, this new API would not have the current "error" / "failure" distinction. Or is it required to support the existing `expect` use case?

natebosch: > To be clear, this new API would not have the current "error" / "failure" distinction.

Correct. I'm thinking that we'd drop `registerException` from `scaffolding.dart` and the user facing "API" is effectively "throw any exception". We'd hopefully drop the distinction between "error" and "failure" entirely.

Then in `hooks.dart` add `TestHandle.current.reportFailure(Object, [StackTrace?])` to take the place of `registerException`. This would take the place of `registerException` for the few edges where it is useful, and to report a failure past an error zone boundary.

The downside is there is no user visible API for reporting a failure past an error zone boundary since `hooks.dart` is not really considered user visible.

pkgs/test_api/lib/hooks.dart
```diff
@@ -0,0 +1,75 @@
...
+  void markSkipped(String message) {
+    if (_invoker.closed) throw ClosedException();
+    _invoker.skip(message);
+  }
+
+  /// Indicates that this test should to be considered done until [future]
+  /// completes.
+  ///
+  /// The test may time out before [future] completes.
+  Future<T> mustWaitFor<T>(Future<T> future) {
```
natebosch, line 49: I suspect this may have issues if `future` comes from a different error zone than the test. We can't detect this mistake either. https://github.com/dart-lang/sdk/issues/39825#issuecomment-799862584

I still think it's worth having this API over the increment/decrement calls because with some refactoring we could hopefully make some problems easier to debug (make it possible to know _which_ futures haven't completed yet and investigate them in observatory, as opposed to only seeing that a counter hasn't reached zero).

I think probably anyone who needs to understand this API is going to need to understand some of the details of error zones anyway, so this might not be a big problem. I'm curious if @jakemac53 or @grouma is concerned though.

grouma: > (make it possible to know which futures haven't completed yet and investigate them in observatory, as opposed to only seeing that a counter hasn't reached zero)

Anything that helps towards this goal is great in my opinion. This has wasted so many hours of my life....

natebosch: OK, looks like this isn't a new issue.

https://github.com/dart-lang/test/issues/1468

It turns out my design here actually fixes that issue, which I _think_ is a good thing. I'm a tiny bit concerned about unintended consequences, but my hunch is it will be an improvement.

natebosch: In case it isn't clear, here is a small repro. This test fails currently, and with this PR it will pass.

```dart
import 'dart:async';

void main() {
  test('runs an async outside of the test zone', () async {
    var x = expectAsync0(() {});
    Zone.root.run(x);
  });
}
```

If anyone has concerns about this let me know.

natebosch: There is another case that this impacts, but doesn't fix.

If a test _fails_ in an async callback that is running in the "wrong" zone, we don't still don't show the failure nicely. Consider the following:

```dart
import 'dart:async';

void main() {
  test('fails an async outside of the test zone', () async {
    var x = expectAsync0(() {
      expect(true, false);
    });
    Zone.root.run(x);
  });
}
```

Before this would fail like:

```
00:00 +0 -1: fails an async outside of the test zone [E]
  Instance of 'OutsideTestException'
  package:test_api/hooks.dart 18:26                              TestHandle.current
  package:test_api/src/expect/expect.dart 70:27                  _expect
  package:test_api/src/expect/expect.dart 46:3                   expect
  test/repro_test.dart 8:7                                       main.<fn>.<fn>
  dart:core                                                      Function.apply
  package:test_api/src/expect/expect_async.dart 196:23           _ExpectedFunction._run
  package:test_api/src/expect/expect_async.dart 179:7            _ExpectedFunction.max6
  package:test_api/src/expect/expect_async.dart 145:15           _ExpectedFunction.max0
  dart:async                                                     _RootZone.run
  test/repro_test.dart 10:15                                     main.<fn>
  test/repro_test.dart 6:51                                      main.<fn>
  package:test_api/src/backend/declarer.dart 200:19              Declarer.test.<fn>.<fn>
  dart:async                                                     _asyncErrorWrapperHelper
  package:test_api/src/backend/invoker.dart                      Invoker.waitForOutstandingCallbacks.<fn>
  dart:async                                                     runZoned
  package:test_api/src/backend/invoker.dart 228:5                Invoker.waitForOutstandingCallbacks
  package:test_api/src/backend/invoker.dart 383:17               Invoker._onRun.<fn>.<fn>.<fn>
  dart:async                                                     _asyncThenWrapperHelper
  package:test_api/src/backend/invoker.dart                      Invoker._onRun.<fn>.<fn>.<fn>
  dart:async                                                     runZoned
  package:test_api/src/backend/invoker.dart 370:9                Invoker._onRun.<fn>.<fn>
  package:test_api/src/backend/invoker.dart 415:15               Invoker._guardIfGuarded
  package:test_api/src/backend/invoker.dart 369:7                Invoker._onRun.<fn>
  package:stack_trace                                            Chain.capture
  package:test_api/src/backend/invoker.dart 368:11               Invoker._onRun
  package:test_api/src/backend/live_test_controller.dart 153:11  LiveTestController.run
  package:test_api/src/remote_listener.dart 255:16               RemoteListener._runLiveTest.<fn>
  dart:async                                                     runZoned
  package:test_api/src/remote_listener.dart 254:5                RemoteListener._runLiveTest
  package:test_api/src/remote_listener.dart 207:7                RemoteListener._serializeTest.<fn>
 ```
 
 Whereas now it fails like
 
 ```
 00:01 +0 -1: fails an async outside of the test zone [E]
  Bad state: Can't add or remove outstanding callbacks outside of a test body.
  package:test_api/src/backend/invoker.dart 119:5         Invoker._outstandingCallbacks
  package:test_api/src/backend/invoker.dart 215:5         Invoker.removeOutstandingCallback
  package:test_api/src/frontend/expect_async.dart 208:15  _ExpectedFunction._afterRun
  package:test_api/src/frontend/expect_async.dart 195:7   _ExpectedFunction._run
  package:test_api/src/frontend/expect_async.dart 176:7   _ExpectedFunction.max6
  package:test_api/src/frontend/expect_async.dart 142:15  _ExpectedFunction.max0
  dart:async                                              _RootZone.run
  test/repro_test.dart 10:15                              main.<fn>
  test/repro_test.dart 6:51                               main.<fn>
  package:test_api/src/backend/declarer.dart 200:19       Declarer.test.<fn>.<fn>
  package:test_api/src/backend/declarer.dart 198:7        Declarer.test.<fn>
  package:test_api/src/backend/invoker.dart 231:7         Invoker.waitForOutstandingCallbacks.<fn>
 ```
 
 This feels like a wash to me so I don't think it should be blocking. I'll see if I can solve this case too though.

Comment (natebosch) [reply pkgs/test_api/lib/hooks.dart:49]: > I suspect this may have issues if `future` comes from a different error zone than the test. We can't detect this mistake either.

I can't actually find any situation where this causes a problem. I think we are OK on that part.

File: pkgs/test_api/lib/src/scaffolding/utils.dart (hunk 1 above)
Comment (natebosch) [reply pkgs/test_api/lib/src/scaffolding/utils.dart:23]: For now I'm going to leave `registerException` out of both `scaffolding.dart` and `hooks.dart` and I won't worry yet to design a replacement.

The "official" way to report an error will be to throw an exception, and we'll look at ways that can also sneak through error zones in a followup down the road.

File: pkgs/test_api/lib/hooks.dart [hunk 3]
```diff
@@ -0,0 +1,72 @@
...
+    if (_invoker.closed) throw ClosedException();
+    _invoker.addOutstandingCallback();
+    return future.whenComplete(_invoker.removeOutstandingCallback);
+  }
+
+  /// Converts [stackTrace] to a [Chain] according to the current test's
+  /// configuration.
+  Chain formatStackTrace(StackTrace stackTrace) =>
+      (_stackTraceFormatter ?? _defaultFormatter).formatStackTrace(stackTrace);
```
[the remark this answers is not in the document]
natebosch: Done.

pkgs/test/test/runner/signal_test.dart
```diff
@@ -181,41 +181,6 @@ void main() {
       await signalAndQuit(test);
     });
 
-    test('causes expect() to always throw an error immediately', () async {
```
natebosch, line 184: @jakemac53 - I had to delete this test. This is a user visible behavior difference.

I don't think this is very important. This only impacts behavior after a user sends a `^C`, and then only causes new exceptions (which might interfere with cleanup).

Let me know if you have concerns.

natebosch: This was originally handled by

https://github.com/dart-lang/test/blob/61a8888294937bb7a909f161faaca3052102050c/pkgs/test_api/lib/src/frontend/expect.dart#L93

I don't think that it's a good idea to leak the idea of "closed" into the new API. As I discovered in https://github.com/dart-lang/test/pull/1481 the concept of "closed" is a little fuzzy and brings in details about how the zones are structured in the test runner.

The changed behavior is pretty minor, and should only be hit in unusual cases.

jakemac53: I don't actually understand the test.... it looks like it was expecting to hit the `catch` but I don't see how....?

Exiting the tests early is somewhat nice potentially but I agree it isn't critical or anything.

natebosch: > it looks like it was expecting to hit the `catch`

Yes it was.

> but I don't see how....?

It hit the catch because `expect` used to have this line:

```dart
 if (Invoker.current!.closed) throw ClosedException(); 
 ```
 
 The invoker gets marked as closed on the first signal received. I didn't want to expose a `closed` boolean in `hooks.dart`, so expect implementations can't do this anymore.

> I agree it isn't critical or anything.

I suspect no one will even notice. If it does cause problems we can definitely revisit the decision.

jakemac53: Ah ok ya that behavior seems super weird. The change LGTM

PR #1461, opened 2021-03-13
Split to more focused utility libraries

natebosch:
Move some utility functions out of `test_api/src/utils.dart` into more
focused libraries. Where possible move utilities into `test_core` or
`test` if they are not used `test_api`.

Make two copies of the `indent` utility since the usages in `test_api`
and `test_core` have non-overlapping sets of arguments used and it's
simpler to implement each case individually. Refactor some calls to
`prefixLines` into calls to `indent`.

Make a private copy of `unawaited` in the one library that needs it in
`test_api`. Add a copy in `test_core` for the remaining uses. Use
`pedantic` in tests since it is available as a dev dependency.


grouma (approved):
This might be internally breaking but should be fairly easy to roll nonetheless. This may impact Flutter as well when they decide to up their pinned version;.

natebosch:
> This may impact Flutter as well when they decide to up their pinned version

I don't think they import any of our `util` libraries.

> This might be internally breaking but should be fairly easy to roll nonetheless.

Yeah, that's part of the reason I'm doing this before https://github.com/dart-lang/test/pull/1460 is to deal with moving around sources in a few small steps that we can check for breakage as we go.

PR #1462, opened 2021-03-13
Separate out utilities from io required libraries

natebosch:
Move out utilities which are imported from files which need to avoid
imports to `dart:io` or `dart:isolate`.

### DIFF
--- a/pkgs/test_core/lib/src/util/os.dart
+++ b/pkgs/test_core/lib/src/util/os.dart
@@ -1,0 +1,30 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:path/path.dart' as p;
+import 'package:test_api/src/backend/operating_system.dart'; // ignore: implementation_imports
+
+/// Directories that are specific to OS X.
+///
+/// This is used to try to distinguish OS X and Linux in [currentOSGuess].
+final _macOSDirectories = {
+  '/Applications',
+  '/Library',
+  '/Network',
+  '/System',
+  '/Users',
+};
+
+/// Returns the best guess for the current operating system without using
+/// `dart:io`.
+///
+/// This is useful for running test files directly and skipping tests as
+/// appropriate. The only OS-specific information we have is the current path,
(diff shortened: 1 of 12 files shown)


natebosch:
I should have listened to the warning from @grouma 

Also, we should really think about how we can get the platform import restrictions tested here... I'm almost surprised that just running web tests didn't catch this, and maybe all that is needed is adding more web tests.

pkgs/test_core/lib/src/util/os.dart
```diff
@@ -0,0 +1,30 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:path/path.dart' as p;
+import 'package:test_api/src/backend/operating_system.dart'; // ignore: implementation_imports
+
+/// Directories that are specific to OS X.
+///
+/// This is used to try to distinguish OS X and Linux in [currentOSGuess].
+final _macOSDirectories = {
```
kevmoo, line 11: could be const?

kevmoo (approved):

pkgs/test_core/lib/src/util/os.dart
natebosch: I wondered the same thing but I didn't look too closely at it. I don't know that we can say a priori that `const` would be an optimization, if it might not be read in some runs.

In any case I'm going to intentionally avoid implementation changes with these PRs shuffling code around.

kevmoo: For me it's more about making sure collections like this aren't accidentally mutated. I get your point, though.

grouma:
> should have listened to the warning from @grouma
> 
> Also, we should really think about how we can get the platform import restrictions tested here... I'm almost surprised that just running web tests didn't catch this, and maybe all that is needed is adding more web tests.

What broke? I assume our internal tests? I think it's more of a reflection of how we do internal modular builds which means each `dart_library`, and its sources, must be properly configured for  the various platforms. Building externally we have the luxury of walking the dependency tree and ignoring unrelated files.

natebosch:
> What broke?

We ended up with imports to `dart:io` or `dart:isolate` in places they shouldn't be allowed, so we catch it in bazel with the stricter platform stuff.

PR #1470, opened 2021-03-20
Disable stack trace chains by default - for real

natebosch:
Changing the default in the arg parser did not change the runner
behavior, and the test that changed was a false positive for the
behavior difference.

Change the default for `chainStackTraces` in the `Metadata` class which
is the real owner of the defaults. Enable stack trace chaining by
default when using `--verbose-trace`, otherwise default to off.

Remove the extra failure output about enabling stack trace chaining for
load "test" failures since they often have no stack trace.

The output is still very verbose - this commit does not reduce the
output to only once at the end of the tests, but that is something we
should consider doing.

Enable stack trace chaining in some tests where the output is checked in
the test.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,10 +1,13 @@
-## 1.16.9-dev
+## 1.16.9
+
+* Disable stack trace chaining by default. It can be re-enabled by explicitly
+  passing the `--chain-stack-traces` flag.
 
 ## 1.16.8
 
 * Fix an issue where coverage collection could hang on Chrome. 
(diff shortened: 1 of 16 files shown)


kevmoo:
🤦

kevmoo (approved):

jakemac53 (approved):

mao0613:
Is this feature available now ?

natebosch:
> Is this feature available now ?

Not quite yet. I am working on getting changes landed internally which is a blocker for publishing. Once things have landed and settled I will publish and this will be available.

PR #1471, opened 2021-03-20
Fix example for null safety

kaladron:
The Dart compiler can't detect that the variable is set in setUp before
it's used.  Mark them late to let the compiler know that they'll be set.


jakemac53 (approved):

jakemac53:
thanks!

PR #1479, opened 2021-03-25
Make the `message` argument a Map

### LINKED ISSUE #1478: Change the `message` argument to `PlatformPlugin.load` to a `Map<String, Object?>`
We always cast this to `Map` (and have since the code was introduced) so the argument type `Object` is not adding any power.

https://github.com/dart-lang/test/blob/f5117fc54c7861d9720595e3bbec638f150795b7/pkgs/test_core/lib/src/runner/plugin/platform_helpers.dart#L46-L62

natebosch:
Towards #1478

The argument to `PlatformPlugin.load` is always a `Map`. The intent is
always to pass it to `deserializeSuite`, where it is unconditionally
cast to `Map`. The loose `Object` argument type does not add any
capabilities.

Changing `PlatformPlugin.load` argument type is not likely to be
breaking since it is not intended for use from outside callers. Changing
the `deserializeSuite` argument is breaking for platform
implementations. Change `PlatformPlugin.load` now so implementations
can adjust before the change to `deserializeSuite`.

### DIFF
--- a/pkgs/test/lib/src/runner/browser/browser_manager.dart
+++ b/pkgs/test/lib/src/runner/browser/browser_manager.dart
@@ -202,8 +202,8 @@ class BrowserManager {
   ///
   /// If [mapper] is passed, it's used to map stack traces for errors coming
   /// from this test suite.
-  Future<RunnerSuite> load(
-      String path, Uri url, SuiteConfiguration suiteConfig, Object message,
+  Future<RunnerSuite> load(String path, Uri url, SuiteConfiguration suiteConfig,
+      Map<String, Object?> message,
       {StackTraceMapper? mapper}) async {
     url = url.replace(
         fragment: Uri.encodeFull(jsonEncode({
--- a/pkgs/test/lib/src/runner/browser/platform.dart
+++ b/pkgs/test/lib/src/runner/browser/platform.dart
@@ -208,7 +208,7 @@ class BrowserPlatform extends PlatformPlugin
   /// Throws an [ArgumentError] if `platform.platform` isn't a browser.
   @override
   Future<RunnerSuite?> load(String path, SuitePlatform platform,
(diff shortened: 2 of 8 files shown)


jakemac53 (approved):

natebosch:
This might be mostly breaking for folks using `implicit-casts: false`. If you allow implicit casts, the `Object message` generally is going to flow through OK. I'll check out a sync and see what comes up.

kevmoo:
So cool, @natebosch !

natebosch:
This will trigger some changes in flutter repositories as we do package rolls. The changes are only required for code analyzed without implicit casts but should be done either way.

The fix is to change the type of the `message` argument to `Map<String, Object?>` (or likely omit the `?` since I don't think these locations are migrated).

https://github.com/flutter/engine/blob/e53a01edb8561f0acc7fa8c8b8c7e4a3bb6fd4a7/lib/web_ui/dev/test_platform.dart#L259
https://github.com/flutter/engine/blob/e53a01edb8561f0acc7fa8c8b8c7e4a3bb6fd4a7/lib/web_ui/dev/test_platform.dart#L626
https://github.com/flutter/flutter/blob/bc2c7954db140cf06bfb478e60ba40b8e6a15098/packages/flutter_tools/lib/src/test/flutter_web_platform.dart#L387
https://github.com/flutter/flutter/blob/bc2c7954db140cf06bfb478e60ba40b8e6a15098/packages/flutter_tools/lib/src/test/flutter_platform.dart#L271

@jonahwilliams for a heads up.

I think to reduce risk here it might be worth taking this a bit slower. I'll first change the argument to `PlatformPlugin.load`. Once that rolls through and we update the implementations, the next change to the argument type for `deserializeSuite` will be easy because it won't require further code change.

@jakemac53 WDYT?

jakemac53:
SGTM

PR #1481, opened 2021-03-26
Inline some zone interactions to a single function

natebosch:
Combine the behavior of `Invoker.unclosable`, and `errorsDontStopTest`
into a single method since they were previously used in two places to
wrap the same intent.

Add a `runTearDowns` method on `Invoker`. Use it from `Declarer` where
it is running `tearDownAll` callbacks, and from `Invoker` where it is
running a test's `tearDown` callbacks. This is the only place (outside of
tests internal to this repo) that the `waitForOutstandignCallbacks` zone
will be nested and the only place where an `Invoker` is not closable. The
implementation is intended to provide the following properties:

- The outer "test", which may be either a `test` or a `tearDownAll`,
  will not be considered complete until all outstanding callbacks from
  all tear down callbacks are also complete. This is the reason for a
  call to `addOutstandingCallback` in `runTearDowns` which doesn't get
  removed until all outstanding callbacks from the tear down are also
  complete.
- Tear downs do not need to wait for all outstanding callbacks from
  previous tear downs before they can start. They do wait for the
  `Future` returned from prior tear downs. This is the reason that the
  loop waits for the `Completer` which is completed following the tear
  down callback, and the remainder of the `waitForOutstandingCallbacks`
  work is `unawaited`.
- If an error is thrown within a tear down callback, only that tear
  down's outstanding callbacks will be ignored - other tear down
  callbacks will continue to hold the test open as normal and the test
  will not be considered complete the way it would be if the error had come
  from the test body. This is the reason for the zone nesting, an error
  will make it seem as though outstanding callbacks completed in only
  the zone where the error is raised.
- Give a verbose name and a longer doc comment to the `closable`
  fields. Documents clearly the single intended purpose it has today, and
  should prevent it from being accidentally given an extra purpose without
  updating the doc.
- Remove the future that never completes part. As far as I can tell, it
  isn't an exercised code path today, and if a usage somehow cropped up
  it's very unlikely that a non-completing future would be the right
  behavior.

Make `waitForOutstandingCallbacks` private and rewrite the tests to call
`runTearDowns`.


natebosch:
This does not change the actual number of zones or their nesting - it only moves some of the complex interactions to a single place so that they are hopefully easier to understand, and so it's clear that there aren't other use cases or callers.

pkgs/test_api/test/backend/invoker_test.dart
```diff
@@ -449,49 +449,67 @@ void main() {
...
-      var innerFunctionCompleted = false;
-      await Invoker.current!.waitForOutstandingCallbacks(() {
-        Invoker.current!.waitForOutstandingCallbacks(() async {
-          await pumpEventQueue();
-          innerFunctionCompleted = true;
-        });
-      });
-
-      expect(innerFunctionCompleted, isFalse);
+    test('allows next tear down to run while there are still prior callbacks',
```
grouma, line 486: I don't follow how this is possible. It looks like we don't complete until the teardown future completes in `invoker.dart` line `236`. Care to explain?

natebosch: This test corresponds to this bullet point from the commit message:

> * Tear downs do not need to wait for all outstanding callbacks from
>   previous tear downs before they can start. They do wait for the
>   `Future` returned from prior tear downs. This is the reason that the
>   loop waits for the `Completer` which is completed following the tear
>   down callback, and the remainder of the `waitForOutstandingCallbacks`
>   work is `unawaited`.

To elaborate - the `unawaited` is ignoring the work done by `_waitForOutstandingCallbacks`, whereas the `Completer` is capturing the work encompassed by the future that is returned by the callback. So if you `await` stuff in your `tearDown` callback and you are returning a future, that does cause the other teardowns to wait. However if you `expect(someOtherFuture, completes)` then that `someOtherFuture` is an outstanding callback so the overall test still wants it to complete before it can be considered done, but other teardowns can get started.

jakemac53 (approved):

PR #1482, opened 2021-03-28
Fix detecting argument count of async function

renggli:


jakemac53:
These should also be `Never` and not `Null` now. Note that this is only a deprecated code path that uses these type checks, but I will push something up to migrate them to `Never`.

natebosch (approved):

PR #1484, opened 2021-04-02
Use the root zone to handle expectAsync callbacks

natebosch:
Previously the call to `Invoker.current.removeOutstandingCallbacks`
happened synchronously when the callback was called. After the change to
use `testHandle.mustWaitFor(Future)` the callback would no longer be
called in a `FakeAsync` zone that does not have any further async
progression. Creating the `Completer` in the root zone avoid this issue.

### DIFF
--- a/pkgs/test_api/lib/src/expect/expect_async.dart
+++ b/pkgs/test_api/lib/src/expect/expect_async.dart
@@ -96,7 +96,8 @@ class _ExpectedFunction<T> {
     }
 
     if (isDone != null || minExpected > 0) {
-      var completer = _expectationSatisfied = Completer<void>();
+      var completer = _expectationSatisfied =
+          Zone.root.run<Completer<void>>(() => Completer<void>());
       _test.mustWaitFor(completer.future);
       _complete = false;
     } else {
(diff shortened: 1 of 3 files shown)


natebosch:
@jakemac53 @grouma - this feels like a pretty sharp edge for implementors of matching frameworks.

That _might_ be OK if we add documentation on `mustWaitFor` warning about this? I'm not sure how many use cases it is likely to come up for.

We could also consider, though, using an API that doesn't suffer from this problem. For instance we could do something like

```dart
class CompletionNotifier {
  void signal() {};
}

CompletionNotifier mustWaitFor() {
  // ...
}
```

Then `mustWaitFor` takes the place of `addOutstandingCallback` and `CompletionNotifier.signal` takes the place of `remoteOutstandingCallback`. This still leaves room in the API to keep a list of human readable names for the outstanding work later on with `mustWaitFor({String? name})`.

jakemac53:
This is something I would definitely not expect any normal user to think about, and for that reason I would be in favor of using a different api that is not subject to this issue.

grouma:
+1 on using the suggested API. I don't think documentation is enough given the subtlety of the issue. How did you detect this was a problem?

natebosch:
> How did you detect this was a problem?

Failing tests internally.

> I would be in favor of using a different api that is not subject to this issue.

I'll try an alternative and see how it looks with existing usage.

natebosch:
Proposal in https://github.com/dart-lang/test/pull/1487

PR #1487, opened 2021-04-05
Switch from mustWaitFor to markPending

natebosch:
The `mustWaitFor` API has tricky edges around futures created in a
`FakeAsync` zone. Return to a non-Future based API which abstracts the
trickiness around zones (other than that the `markPending` call must be
in the test zone, which is the case for all APIs) and errors.

### DIFF
--- a/pkgs/test_api/lib/hooks.dart
+++ b/pkgs/test_api/lib/hooks.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:stack_trace/stack_trace.dart';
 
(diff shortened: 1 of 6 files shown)


pkgs/test_api/lib/hooks.dart
```diff
@@ -45,14 +47,13 @@ class TestHandle {
     _invoker.skip(message);
   }
 
-  /// Indicates that this test should not be considered done until [future]
-  /// completes.
+  /// Indicates that this test should not be considered done until the returned
+  /// [OutstandingWork] is marked as complete;
```
grouma, line 51: Is `;` intentional?

grouma (approved):

pkgs/test_api/lib/hooks.dart
natebosch: Typo

PR #1492, opened 2021-04-15
Restrict imports across layers of test_api

natebosch:
Allow splitting `test_api` into multiple targets in bazel. Fix imports
that would cause cycles in the targets, and shuffle some code to make
globbing easier and better reflect the use cases and library coupling.

Add more exports in `backend.dart` of types that are related to how
platforms communicate with the runner, `RemoteException`,
`RemoteListener`, `StackTraceFormatter`, and `StackTraceMapper`. Where
convenient, update imports from `test_api/src/*` to
`test_api/backend.dart` with a `show` clause.
Not all of `src/backend` is exported, including the details that are
currently used from `test_core` and `test` since future work may be able
to keep those details from leaking.

Add `pumpEventQueue` to `hooks.dart`. This allows `expect` to rely only
on `hooks` and avoid a dependency on `scaffolding`. It is an API that
should may be useful in general for async matching frameworks.

Add a test which uses the analyzer to crawl imports and enforce some
aspects of the required structure:
- `backend.dart` and `src/backend/**` may not import any other
  parts of the packages, so `backend` can be a leaf target with a simple
  glob.
- `expect` may not be imported from any of the focused entrypoints
  (`hooks.dart` or `scaffolding.dart`) and may not import any of the
  implementation details other than `hooks.dart`.

Shuffle around code and update a few imports to more narrow libraries to
fit the above criteria.
- Move the annotations like `Timeout` etc to `src/backend/configuration`
  to fix some imports that would be to `src/scaffolding` and re-export
  from `scaffolding.dart`. Not all of these are used, but they are all
  moved for consistency.
- Move `TestFailure` into backend and re-export from `hooks.dart`.
- Move `remote_exception.dart`, `remote_listener.dart`, and
  `stack_trace_mapper.dart` to `src/backend/` since they are now
  exported.
- Move `suite_channel_manager.dart` and `pretty_print.dart` into
  `src/backend/`.
- Move `anchoredHyphenatedIdentifier` into a separate utility library
  under `src/backend/`. Inline the unused private regex and update the
  doc comment avoid referencing it.


pkgs/test_api/lib/src/backend/util/identifier_regex.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// A regular expression matching a full string as a hyphenated identifier.
+///
+/// This is like a standard Dart identifier, except that it can also contain
+/// hyphens.
+final anchoredHyphenatedIdentifier = RegExp(r'^[a-zA-Z_-][a-zA-Z0-9_-]*$');
```
grouma, line 9: Where is this used?

grouma (approved):
LGTM. This may be a little difficult to roll into Flutter. We should validate there aren't any blockers there.

pkgs/test_api/lib/src/backend/util/identifier_regex.dart
natebosch: https://github.com/dart-lang/test/blob/e779df6cc280104d0ce119c202417cd62b20438f/pkgs/test_core/lib/src/runner/parse_metadata.dart#L190

https://github.com/dart-lang/test/blob/e779df6cc280104d0ce119c202417cd62b20438f/pkgs/test_core/lib/src/runner/configuration/load.dart#L540

https://github.com/dart-lang/test/blob/e779df6cc280104d0ce119c202417cd62b20438f/pkgs/test_api/lib/src/backend/metadata.dart#L288

natebosch:
> This may be a little difficult to roll into Flutter. We should validate there aren't any blockers there.

cc @jonahwilliams as a heads up

I don't expect any blockers, there will be a little bit of churn to update imports, but we haven't changed anything that is breaking otherwise.

Issue #1498, opened 2021-04-20
CONTRIBUTING.MD is still in the year 2017


m-terlinde:
Hey guys,

the [CONTRIBUTING.MD](https://github.com/dart-lang/test/blob/fbe5f51042dc919b03953ac4b0647452d0525ded/CONTRIBUTING.md) has a section for the file headers.

The copyright year is still set to 2017. Maybe you want to change it to 2021?

Furthermore, there is an authors file referenced. I was not able to locate it for this repo. Maybe you want to look at this too.

Thanks,
Matthias

PR #1499, opened 2021-04-20
Update year in example copyright blurb

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1498

I would remove the sentence about the AUTHORS file but I don't want to touch the substance of the blurb even though its obviously not providing value right now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,7 @@ new tests, use the existing tests as a guideline for what they should look like.
 Before you send your pull request, make sure all the tests pass! To run all the
 tests, you'll need:
 
-* [NodeJS][https://nodejs.org/]
+* [NodeJS](https://nodejs.org/])
 
 Once you have this, just run `pub run test`.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,17 +37,15 @@ new tests, use the existing tests as a guideline for what they should look like.
 Before you send your pull request, make sure all the tests pass! To run all the
 tests, you'll need:
 
-* [PhantomJS][] version 2.0.0 or higher.
+* [NodeJS]: https://nodejs.org/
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -37,17 +37,15 @@ new tests, use the existing tests as a guideline for what they should look like.
 Before you send your pull request, make sure all the tests pass! To run all the
 tests, you'll need:
 
-* [PhantomJS][] version 2.0.0 or higher.
+* [NodeJS]: https://nodejs.org/
 
-[PhantomJS]: http://phantomjs.org/
-
-Once you have these, just run `pub run test`.
+Once you have this, just run `pub run test`.
```
natebosch, line 42: This isn't true anymore.

We could replace it with `mono_repo presubmit` perhaps...

CONTRIBUTING.md
```diff
@@ -37,17 +37,15 @@ new tests, use the existing tests as a guideline for what they should look like.
 Before you send your pull request, make sure all the tests pass! To run all the
 tests, you'll need:
 
-* [PhantomJS][] version 2.0.0 or higher.
+* [NodeJS](https://nodejs.org/])
```
natebosch, line 40:
```suggestion
* [NodeJS](https://nodejs.org/)
```

CONTRIBUTING.md
jakemac53: lol true, I could remove the entire section around testing... (the expectation I think is more to rely on CI than locally run)

natebosch: Removing the section would probably be fine.

Using CI is a more accurate reflection of what we do in practice.

jakemac53: I updated it to reference mono_repo as a tool, but not a requirement

natebosch (approved):

PR #1503, opened 2021-04-22
Add test for opted in package with opted out test

natebosch:
There is a bug with the frontend server usage where the first test that
runs determines the soundness for all other tests. If the first test is
opted in to null safety, all tests must be and an opted out test will
fail to load. If the first test is opted out of null safety, all tests
will run in unsound mode.

Add an integration test package which is opted in to null safety at the
package level and has one test opted out and one opted in. Both tests
will only pass if they are run with the expected soundness.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -38,16 +38,16 @@ jobs:
       - name: mono_repo self validate
         run: pub global run mono_repo generate --validate
   job_002:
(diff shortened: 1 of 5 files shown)


jakemac53 (approved):

jakemac53:
I have rolled this into https://github.com/dart-lang/test/pull/1504

PR #1504, opened 2021-04-22
Fix mixed mode tests

jakemac53:
Fixes an issue where all tests were compiled in either sound or unsound mode, depending on which type of test was encountered first.

We now create a separate compiler for each language version that we see in tests.

### DIFF
--- a/pkgs/test_core/lib/src/runner/vm/test_compiler.dart
+++ b/pkgs/test_core/lib/src/runner/vm/test_compiler.dart
@@ -78,7 +78,7 @@ class _TestCompilerForLanguageVersion {
       : _dillCachePath =
             '$dillCachePrefix.${base64.encode(utf8.encode(_languageVersionComment.replaceAll(' ', '')))}';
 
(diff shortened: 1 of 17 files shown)


pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -27,49 +28,59 @@ class CompilationResponse {
...
+      File(p.join(_outputDillDirectory.path, 'output.dill'));
+  final _outputDillDirectory =
+      Directory.systemTemp.createTempSync('dart_test.');
+
+  _TestCompilerForLanguageVersion(
+      String dillCachePrefix, this._languageVersionComment)
+      : _dillCachePath =
+            '$dillCachePrefix.${base64.encode(utf8.encode(_languageVersionComment.replaceAll(' ', '')))}';
+
+  Future<String> _generateEntrypoint(Uri testUri) async {
```
natebosch, line 81: [nit] This no longer needs to be `async`

natebosch (approved):

Issue #1506, opened 2021-04-24
test 1.17.1 fails on GitHub Actions / Windows


timsneath:
Compare:
https://github.com/timsneath/winmd/runs/2424326028?check_suite_focus=true (1.17.1)
with
https://github.com/timsneath/winmd/runs/2424365294?check_suite_focus=true (1.16.8)

The only difference between these two passes is the version of the test package. 

/cc @jakemac53

jakemac53:
Note that you can refer to the old behavior with the `--use-data-isolate-strategy` flag until I can track this down

jakemac53:
I also can't see the logs, not sure if I might need some special permissions for that? If you can paste the failure here that would help.

timsneath:
Oh, I'm sorry -- thought the logs would be public.

Here -- this is the error:
https://gist.github.com/timsneath/11fb8891278a60824de2810515afed76#file-1-17-1-txt-L424

The same gist contains a run with test 1.16.8, which is successful. Key line looks like:

```
2021-04-24T00:32:04.5309203Z Invalid argument(s): Illegal character in path
2021-04-24T00:32:04.5310067Z #0      _Uri._checkWindowsPathReservedCharacters (dart:core/uri.dart:1642:11)
2021-04-24T00:32:04.5311118Z #1      _Uri._makeWindowsFileUrl (dart:core/uri.dart:1729:9)
2021-04-24T00:32:04.5311717Z #2      new _Uri.file (dart:core/uri.dart:1609:11)
```

jakemac53:
> Oh, I'm sorry -- thought the logs would be public.

Weirdly I can see them now... I was on my phone maybe the GH app doesn't actually allow you to see logs or something 🤷‍♂️ .

In any case I have a fix coming soon I think here.

timsneath:
Can confirm this fix worked for the above package:
https://github.com/timsneath/winmd/runs/2440189198?check_suite_focus=true (1.17.2)

Thanks!

jakemac53:
Thanks for the reports and good details!

Issue #317, opened 2015-08-12
Unable to run largish test suite due to timeout/hang


lassedamgaard:
I have a suite of 800+ browser based tests spread across some ~80 files. 
Using either content-shell or dartium the test run times out after around 1 minute and 40 seconds. In content-shell the runner just seems to hang. In dartium I get a connection closed error. I'm pretty sure it's not a bad test that causes this because the tests that are not run on dartium can run on content shell. Similarly I can skip some groups longer up the chain to reduce the total number of tests and thus make the last part of the run succeed. But I cannot run everything. 

Using test 0.12.4 and dart 1.10 on OSX and concurrency 1. With higher concurrency the timeout occurs earlier. 

I have no idea how to debug this.

Example run in content shell (last line hangs forever):

```
01:37 +686 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils deploying a revision older than some of the target deployment revisions is a warning
01:37 +687 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils downgrade warning is skipped if commit query fails
01:37 +688 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils downgrade warning is skipped if commit query fails
01:37 +689 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils deploying a new build when a reference build exists is a warning
01:37 +690 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils deploying a different build than the reference build is a warning
01:37 +691 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils deploying a different build than some of multiple reference builds is a warning
01:37 +692 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils deploying a different build than any of multiple reference builds is a warning
01:37 +693 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils deploying a different build than some of multiple reference builds is a warning
01:37 +694 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils deploying a different ref than all multiple reference builds is a warning
01:37 +695 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils reference deployment check proceeds if only some reference builds cannot be retrieved
01:37 +696 ~4: test/test_upgrade_dialog_utils.dart: upgrade_utils reference deployment check is aborted if no reference builds can be retrieved
[...]

PR #1507, opened 2021-04-26
fix sdk root path on windows

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1506

I couldn't repro this locally on my windows box but I am testing the fix on github CI over at [mono_repo #317](https://github.com/google/mono_repo.dart/pull/317).

Basically my assumption here is that the `uri` we were grabbing off of the `Directory` object had a mix of forward/backward slashes.

### DIFF
--- a/pkgs/test_core/lib/src/runner/vm/test_compiler.dart
+++ b/pkgs/test_core/lib/src/runner/vm/test_compiler.dart
@@ -152,13 +152,12 @@ class _TestCompilerForLanguageVersion {
   Future<CompileResult?> _createCompiler(Uri testUri) async {
     final platformDill = 'lib/_internal/vm_platform_strong.dill';
     final sdkRoot =
-        Directory(p.relative(p.join(Platform.resolvedExecutable, '..', '..')))
-            .uri;
+        p.relative(p.dirname(p.dirname(Platform.resolvedExecutable)));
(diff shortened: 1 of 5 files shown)


jakemac53:
This does look to solve the issue for mono_repo, so assuming it doesn't break any of our own tests then this looks like a good fix.

grouma (approved):

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -152,13 +152,12 @@ class _TestCompilerForLanguageVersion {
   Future<CompileResult?> _createCompiler(Uri testUri) async {
     final platformDill = 'lib/_internal/vm_platform_strong.dill';
     final sdkRoot =
-        Directory(p.relative(p.join(Platform.resolvedExecutable, '..', '..')))
-            .uri;
+        p.relative(p.dirname(p.dirname(Platform.resolvedExecutable)));
```
natebosch, line 155: Why `p.relative` here?

jakemac53: It likely isn't necessary, I was just retaining as much of the original code as possible to avoid changing the intent.

It might be removing drive letters or something in come cases? Idk if package:path would actually support that or not.

Issue #1509, opened 2021-04-27
`dart test` doesn't recognize the `--define` or '-D' flag


rxlabz:
even if dart-lang/sdk#44562 is closed, it seems actually not possible to define environment variables fort the `dart test` command ?

natebosch:
I'm not eager to support general `dart` arguments in the test runner - it would start cascading into complexity across all platforms.

I'll leave this issue open in case we decide later to add support for `-D` defines. If we decide to handle this at the test runner level it would imply a breaking change and some work needed for platforms like flutter.

@jonahwilliams - does `flutter test` support `-D` arguments, and in what positions?

jonahwilliams:
It does:

```
flutter test -h

...

    --dart-define=<foo=bar>           Additional key-value pairs that will be available as constants from the
                                      String.fromEnvironment, bool.fromEnvironment, int.fromEnvironment, and
                                      double.fromEnvironment constructors.
                                      Multiple defines can be passed by repeating "--dart-define" multiple times.

```

Note that if you keep the cached dill floating around you need to account for the exact defines used for it.

bkonyi:
If `dart -Dfoo=bar test foo.dart` works then I think all we have to do is define `-D`/`--define` in the test runner argument parser so it's considered a valid flag. The VM looks for VM flags in the command argument list, so as long as the command doesn't complain about an unrecognized option it should just work. It's also not clear that `-D` _can't_ be provided to `dart test` since `dart help test` doesn't delegate to the test runner argument parser.

jakemac53:
I don't think that with the new test loading strategy this would work either way? It may have accidentally worked previously, on the vm only, because we used isolates and they inherited it the defines. But that approach is actually setting the defines for the test runner itself as well which really isn't right.

natebosch:
> I think all we have to do is define `-D`/`--define` in the test runner argument parser so it's considered a valid flag. The VM looks for VM flags in the command argument list, so as long as the command doesn't complain about an unrecognized option it should just work.

This is not the way the VM works today. If I ignore the `-D` in the test runner argument parsing the VM will ignore it if it comes after `test`. `dart -DFOO=bar test` and `dart test -DFOO=bar` do not have the same behavior.



> It's also not clear that `-D` _can't_ be provided to `dart test` since `dart help test` doesn't delegate to the test runner argument parser.

[...]

PR #1511, opened 2021-04-28
Start docs on configuration compilation

natebosch:
Start some docs on the limitations and workarounds for configuration
compilation of the tests.

### DIFF
--- a/pkgs/test/README.md
+++ b/pkgs/test/README.md
@@ -808,12 +808,12 @@ Compilation for browser and node tests can be configured by passing arguments to
 
 Fine grained compilation configuration is not supported for the VM. Any
 configuration which impacts runtime behavior for the entire VM, such as `-D`
-defines and runtime experiments, will influence both the test runner and the
-isolates spawned to run test suites. Experiments which are breaking may cause
-incompatibilities with the test runner. These may be specified with a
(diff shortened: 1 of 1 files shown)


natebosch:
Related:

- #1509 - It's confusing that we don't support `-D` after `dart test` since some other subcommands of `dart` allow the flags in that position.
- #1510 - It's confusing that we don't support `--no-sound-null-safety` since this is a compile time only flag, not a runtime behavior flag, so even with `dart --no-sound-null-safety test` it doesn't impact test behavior.

My hope is that these docs could be what we point to for questions like these about why it doesn't work.

In my draft I avoided being too concrete about which flags are "runtime behavior" vs compile time, because I don't really want to have to maintain a detailed description of VM implementation details here. I don't know if that leaves things too vague to be useful though?

natebosch:
Another issue with confusion about the behavior:

https://github.com/dart-lang/sdk/issues/45835

In this case the `--enable-experiment` used to work because we only had the behavior of `--use-data-isolate-strategy`. After adding support for the frontend server the default behavior changed, and the new default behavior doesn't support experiment args for VM tests at all.

jakemac53 (approved):

Issue #1516, opened 2021-05-12
Race condition between compiling and spawning


jensjoha:
In 
https://github.com/dart-lang/test/blob/4f90091595ef4e82b883507947df204ee9a139fe/pkgs/test_core/lib/src/runner/vm/test_compiler.dart#L97-L151

it compiles with a `Pool` of `1` which I'm guessing means only one at a time. So far so good.
It always compiles to the same file which isn't a problem per say as it's only running one at a time.
It then copies to another file that is also the same every time, and this file (uri) is returned. That's bad. Because when called from outside here:
https://github.com/dart-lang/test/blob/4f90091595ef4e82b883507947df204ee9a139fe/pkgs/test_core/lib/src/runner/vm/platform.dart#L144-L157
which can and does run multiple at the same time a new copy can be mid-way before the VM is finished spawning the isolate (and thus reading the file).

So the process can be this:
Queue spawn 1 (i.e. test 1)
Queue spawn 2 (i.e. test 2)
Compile 1 (i.e. test 1) and copy to `foo.dill` (from start to finish, we can only run 1 at a time).
Spawn 1 (i.e. test 1) asks the VM to spawn an isolate from `foo.dill`.
Compile 2 (i.e. test 2) and copy to `foo.dill`.
Spawn 2 (i.e. test 2) asks the VM to spawn an isolate from `foo.dill`.
Spawn 1 (i.e. test 1) finishes.
Spawn 2 (i.e. test 2) finishes.

So depending on how far the VM comes any of these things might happen:
* Spawn 1 (i.e. test 1) reads `foo.dill` before the file is overwritten (and spawn 2 (i.e. test 2) is read fine too) --- and everything is fine.
* Spawn 1 (i.e. test 1) reads parts of `foo.dill` before the file is overwritten, but then the file is overwritten, reading fails, the VM says it's not a dill file, proceeds to parse it as a dart file (that's not what we want). (This will probably give an error of some sort --- internally we've seen it cause the scanner to go into an infinite loop (which we'll fix)). Spawn 2 (i.e. test 2) will read fine and be happy.
* Spawn 1 (i.e. test 1) reads nothing of `foo.dill` before the file is overwritten. Ends up reading `foo.dill` *after*, i.e. it actually reads test 2. Spawn 2 (i.e. test 2) reads fine and is happy. We end up executing the same test twice. We probably won't be notified, we will probably claim to have run both test 1 and test 2 while we have in fact run test 2 twice.

[...]

PR #1517, opened 2021-05-12
fix race condition between vm test compilation and isolate spawning

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1516

### DIFF
--- a/pkgs/test_core/lib/src/runner/vm/test_compiler.dart
+++ b/pkgs/test_core/lib/src/runner/vm/test_compiler.dart
@@ -73,6 +73,8 @@ class _TestCompilerForLanguageVersion {
       File(p.join(_outputDillDirectory.path, 'output.dill'));
   final _outputDillDirectory =
       Directory.systemTemp.createTempSync('dart_test.');
+  // Used to create unique file names for final kernel files.
+  int compileNumber = 0;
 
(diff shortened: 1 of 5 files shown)


pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -73,6 +73,8 @@ class _TestCompilerForLanguageVersion {
       File(p.join(_outputDillDirectory.path, 'output.dill'));
   final _outputDillDirectory =
       Directory.systemTemp.createTempSync('dart_test.');
+  // Used to create unique file names for final kernel files.
+  int compileNumber = 0;
```
natebosch, line 77:
```suggestion
  int _compileNumber = 0;
```

natebosch (approved):

natebosch:
We're probably still going to have race conditions if you run multiple test runners at the same time.

Would it be sensible to fix that problem even if we don't expect users to do that? Would putting the test runner PID into the filename have negative impacts on caching across test runs?

jakemac53:
> We're probably still going to have race conditions if you run multiple test runners at the same time.

We create a separate temp dir for each instance of this class so it should be fine.

PR #1531, opened 2021-06-07
Drop some unused functions

natebosch:
- Inline `loadFromString` into `load` and drop the extra `Configuration`
  constructor.
- `Engine.liveSuites` had no readers.
- Initialize `LiveSuiteController.liveSuite` with a `late final` to
  avoid an extra getter.
- `LiveSuiteController.isComplete` had no readers.
- `RunnerSuite.channel` had no callers.
- Various test utilities are likely remnants of the package split.

### DIFF
--- a/pkgs/test_core/lib/src/runner/runner_suite.dart
+++ b/pkgs/test_core/lib/src/runner/runner_suite.dart
@@ -44,14 +44,6 @@ class RunnerSuite extends Suite {
   /// The event is `true` when debugging starts and `false` when it ends.
   Stream<bool> get onDebugging => _controller._onDebuggingController.stream;
 
-  /// Returns a channel that communicates with the remote suite.
-  ///
-  /// This connects to a channel created by code in the test worker calling the
-  /// `suiteChannel` argument from a `beforeLoad` callback to `serializeSuite`
-  /// with the same name.
-  /// It can be used used to send and receive any JSON-serializable object.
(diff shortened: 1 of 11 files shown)


pkgs/test_core/lib/src/runner/runner_suite.dart
```diff
@@ -44,14 +44,6 @@ class RunnerSuite extends Suite {
   /// The event is `true` when debugging starts and `false` when it ends.
   Stream<bool> get onDebugging => _controller._onDebuggingController.stream;
 
-  /// Returns a channel that communicates with the remote suite.
-  ///
-  /// This connects to a channel created by code in the test worker calling the
-  /// `suiteChannel` argument from a `beforeLoad` callback to `serializeSuite`
-  /// with the same name.
-  /// It can be used used to send and receive any JSON-serializable object.
-  StreamChannel channel(String name) => _controller.channel(name);
```
grouma, line 53: This may be used by our internal infra. You should double check.

grouma (approved):

pkgs/test_core/lib/src/runner/runner_suite.dart
natebosch: It was an internal check that surfaced these as unused (there were some false positives caused by some of the files that aren't synced). I'm _pretty_ sure that this should be fine when we sync.

PR #1534, opened 2021-06-09
Add back Configuration.loadFromString constructor

natebosch:
This API is not used within the test runner, but is used from
`build_test`.

Rename what was `loadFromString` to `parse` and move the file reading to
the `Configuration.load` constructor.


jakemac53 (approved):

kevmoo:
I've had my first CI failure here. Can haz land?

Issue #1539, opened 2021-06-14
Add ability to turn off shuffling of tests on a per-suite or per group basis.


gspencergoog:
Currently the shuffling of test order as implemented for https://github.com/dart-lang/test/issues/1124 defines the shuffling seed for the entire run.

Flutter would like to transition our framework tests to use randomized ordering for the tests, but since there are a large number of failures from dependent tests when we turn this on, previous attempts to make this transition have failed because the PRs are too huge, and merge conflicts occur too frequently to make landing the PR feasible.

Therefore, we'd like to migrate our tests to shuffled order one file at a time, by turning on the flag globally, disabling shuffle for each of the files that have failures, and then we can then come back at our leisure and fix each file, and turn on shuffle for that file until they're all complete.

I *think* the randomization happens per file or per group (it's not entirely clear to me from looking at the code and docs), but if that's not the case, then this would have to be implemented by running the tests in two groups (one in original order, and one shuffled).

cc @dnfield @jakemac53

dnfield:
One way we could do this without changes it to use tags. We could start tagging tests that we know are safe to run in random order with some special tag, and just increase that (or vice versa, tag all existing tests and start removing the tag to burn down).

I _think_ you can do `@Tag` on a per file basis. https://pub.dev/packages/test#tagging-tests

jakemac53:
Yes tags should be a good way to do this today.

gspencergoog:
OK, it looks like that works, but two things might be confusing:

I used this config file:

```yaml
tags:
  no-shuffle:
    test-randomize-ordering-seed: 0
  shuffle:
    test-randomize-ordering-seed: 123
```

And ran it with `flutter test --test-randomize-ordering-seed=123`, and files that had `@Tags(['no-shuffle'])` in them that would normally fail with that seed would then pass, which is my intent.

But, (confusing thing number one) when I replaced the "no-shuffle" with "shuffle" in one of the tests that would normally fail, and did not specify a seed on the command line, those tagged tests continued to pass, presumably because they still weren't being reordered. Same thing if I change the 0 to 123 for `no-shuffle` in the config file: tagged tests succeed, and untagged ones fail, regardless of the value of the argument. Commenting out the argument in the yaml file makes tagged and untagged tests fail again when the command line arg is given.

[...]

PR #1541, opened 2021-06-16
Move testRandomizeOrderingSeed to global level

natebosch:
Resolves an issue discovered in #1539

This configuration was attached to the `SuiteConfig` likely because that
was the easiest way to communicate it through to the `Engine` where
required. This introduced a bug where a more specific `SuiteConfig` for
a tag configured in `dart_test.yaml` would always override the apparent
configuration for those suites with the default, which is unshuffled.
There was no indication when this happeened.

Remove `testRandomizedOrderingSeed` form the `SuiteCConfiguration` class
entirely instead of trying to track down where the inconsistency slips
in.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.17.8-dev
+
+* Fix a bug where a tag level configuration would cause test suites with that
+  tag to ignore the `--test-randomize-ordering-seed` argument.
+
 ## 1.17.7
 
 * Support the latest `test_core`.
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test
-version: 1.17.7
+version: 1.17.8-dev
 description: >-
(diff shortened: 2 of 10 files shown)


natebosch:
This PR actually moves us further away from the feature requested in https://github.com/dart-lang/test/issues/1539

If we think we do want to support per-suite disabling of randomization it would take a fair amount of additional work around considering the UX and fixing existing implementation issues. My current leaning is towards fixing this confusing bug and closing #1539 with the expectation that users will work around it with the suggestions there around using tags.

jakemac53 (approved):

natebosch:
> This PR actually moves us further away from the feature requested in #1539

Discussed offline - this isn't true. The feature we'd implement if we decide to go forward with #1539 would be to _disable_ at the suite level, not to set the seed at the suite level.

For now we'll wait and see if the workaround in https://github.com/dart-lang/test/issues/1539#issuecomment-861541199 (run twice - once with shuffling and `-x noshuffle` and once without shuffling and `-t noshuffle`) is sufficient before deciding whether to add a suite level option to disable shuffling. We'll fix this bug either way.

Issue #1546, opened 2021-06-24
Add a command line switch to allow zero exit when no tests are found.


gspencergoog:
When running with `-t tag`, it would be useful to be able to say "Yeah, if you don't find any tests with that tag, it's OK, just exit with zero instead of failing".

I want to run all of our tests twice, once with `-x tag`, and once with `-t tag` (and some other different options), but if I run it on a set of test files that don't happen to contain the tag, I'm OK with that, since they all will be run in the other test. Unfortunately, as it is now, the `-t tag` version will always fail when it doesn't find any tests, and I have to parse the test output to look for "No tests ran." and not fail in only that case (which is fragile, obviously).

/cc @natebosch

gspencergoog:
More is here: https://github.com/dart-lang/test/issues/1539#issuecomment-867086151

jakemac53:
Would this satisfy the use case as well? https://github.com/dart-lang/test/issues/1418

tldr; provide a different exit code, so you could just check for that

gspencergoog:
Yeah, that would probably work: anything we can use to distinguish would work, as long as it doesn't collide with other exit codes we might get.

Issue #1418, opened 2021-02-02
Provide different exit code for failed and empty tests


Nikarous:
Hi,

when running tests with no tests defined flutter prints "No tests ran." message and exit with exit code 1. The same as when tests actually fails. This is misleading and I would like to ask you to provide different non-zero exit code for empty tests so user can process response in correct way.

jakemac53:
This was an explicit decision, see the discussion here https://github.com/dart-lang/test/issues/764

jakemac53:
Oh sorry I didn't fully understand the request, it might be reasonable to change the exit code to something other than 1, not sure how breaking that might be. cc @natebosch for thoughts.

natebosch:
I think there is precedent for this. According to https://elinux.org/Test_Result_Codes`pytest` uses exit code 5 for this situation.

It's not a trivial change, we'll have to track extra information or abuse the global `exitCode=` and be careful not to overwrite it which would end up fairly brittle. I'd like to understand more about the value it would bring before we add the complexity and extra tests.

We'll also have to check whether `flutter test` is forwarding the exit code, or if it only distinguishes success/failure.

> so user can process response in correct way.

How would you handle this case differently? What are you writing which invokes the test runner and distinguishes between the error codes?

Nikarous:
Hi, sorry for delayed answer.

We are developing flutter app right now and team still doesn't created all the tests (yeah I know but what will you do). Still they have a placeholder test files prepared. If I as a creator of a pipeline run flutter test on the directory it will fail because some tests are not implemented. I could filter such exit code if it would be different than "standard" fail and just log a warning as this is expected currently.

Nikarous:
Also I really think that tests should not stop at first failure, instead run all tests and collect report.

jakemac53:
> I as a creator of a pipeline run flutter test on the directory it will fail because some tests are not implemented.

I would recommend adding a single placeholder test in these files that passes?

> Also I really think that tests should not stop at first failure, instead run all tests and collect report.

At least in the default framework this is the behavior... I don't know if flutter_test does something different here?

Nikarous:
> I would recommend adding a single placeholder test in these files that passes?

That is what we did at the end.

> At least in the default framework this is the behavior... I don't know if flutter_test does something different here?

It seems so but I will take it to Flutter team then I guess. Thank you for discussion and help!

PR #1549, opened 2021-06-29
add a special exit code, 79, used when no tests were found

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1546

I will update the tests and things if we decide this is a fine route to go. I simply used the next exit code in the sequence of special exit codes we have.

I also added some special logging for tags, similar to what existed for the name patterns. It will list the tag configuration, if one was provided.

~~Both of these end up with a different exit code, 65, which indicates some sort of data error in the input (this was used previously when you provided a name regex and there were no matches). I am not sure if we want to also re-use that exit code for the case where there simply weren't any tests.~~ All of these cases now use exit code 79.

We also may want to specialize for additional configuration like platforms? Not sure how deep that rabbit hole goes.

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -5,6 +5,8 @@
   canceled early.
 * Don't log the --test-randomization-ordering-seed if using the json reporter.
 * Add a new exit code, 79, which is used when no tests were ran.
+  * Previously you would have gotten either exit code 1 or 65 (65 if you had
+    provided a test name regex).
 * When no tests were ran but tags were provided, list the tag configuration.
 
 ## 0.3.29
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Report incomplete tests as errors in the JSON reporter when the run is
   canceled early.
 * Don't log the --test-randomization-ordering-seed if using the json reporter.
+* Add a new exit code, 79, which is used when no tests were ran.
+* When no tests were ran but tags were provided, list the tag configuration.
 
 ## 0.3.29
 
--- a/pkgs/test_core/lib/src/runner.dart
+++ b/pkgs/test_core/lib/src/runner.dart
@@ -18,7 +18,6 @@ import 'package:test_api/src/backend/test.dart'; // ignore: implementation_impor
 import 'package:test_api/src/backend/util/pretty_print.dart'; // ignore: implementation_imports
 import 'package:test_core/src/runner/reporter/multiplex.dart';
 
-import 'runner/application_exception.dart';
 import 'runner/no_tests_found_exception.dart';
 import 'runner/configuration.dart';
 import 'runner/configuration/reporters.dart';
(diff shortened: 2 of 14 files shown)


jakemac53:
cc @gspencergoog - I think for your use case since you are using tags this would actually end up with a different exit code, one that is used by some other platforms for other meanings.

So maybe I should change them all to use this new exit code, but I am a little bit concerned that we might break somebody relying on the old exit code that was specialized when you provided a `-n "some-pattern"` argument.

jakemac53:
Related discussion in https://github.com/dart-lang/test/issues/1418 which was closed by the original author

pkgs/test_core/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 * Report incomplete tests as errors in the JSON reporter when the run is
   canceled early.
 * Don't log the --test-randomization-ordering-seed if using the json reporter.
+* Add a new exit code, 79, which is used when no tests were ran.
```
natebosch, line 7: This isn't quite accurate, it will be a different code if you run a suite with no tests at all.

We could expand this sentence

"which is used when no tests were ran due to tags or other filters matching no tests."

jakemac53: This new exit code is used when:

- No tests were ran and:
  - No tags were configured
  - No name patterns were provided
  
In the other cases the exit code 65 is used (no change for the pattern case, new message/code for the tag case)

jakemac53: I do think its confusing though and am leaning towards making these all the same exception/exit code (79).

natebosch: > am leaning towards making these all the same exception/exit code (79).

I think I'd prefer that as well.

natebosch: Although I think handling the no tests in suite case might be trickier. I think that's exit code 1 today, same as a test failure.

jakemac53: > Although I think handling the no tests in suite case might be trickier. I think that's exit code 1 today, same as a test failure.

Correct, it is exit code 1 today. With this PR it changes to exit code 79.

pkgs/test_core/lib/src/runner.dart
```diff
@@ -135,14 +137,22 @@ class Runner {
...
+                    : '"$pattern"'));
+            throw ApplicationException('No tests match $patterns.');
+          } else if (_config.suiteDefaults.includeTags != BooleanSelector.all ||
+              _config.suiteDefaults.excludeTags != BooleanSelector.none) {
+            throw ApplicationException(
+                'No tests match the requested tag selectors:\n'
+                '  include: "${_config.suiteDefaults.includeTags}"\n'
+                '  exclude: "${_config.suiteDefaults.excludeTags}"');
+          } else {
+            throw NoTestsFoundException('No tests were found.');
```
jakemac53, line 154: Specifically, it hits this branch in the case of no tests ran.

pkgs/test_core/CHANGELOG.md
natebosch: > With this PR it changes to exit code 79.

How?

No tests defined is handled by https://github.com/jepperaskdk/test/blob/76cf9b7744eda85e27e45369fe63d13b7f14f56c/pkgs/test_core/lib/src/runner.dart#L149 which doesn't throw this exception.

natebosch: Oh, nevermind, I see now the removal of `_config.suiteDefaults.patterns.isNotEmpty)`

natebosch (approved):

PR #1551, opened 2021-07-01
Allow disabling test randomization for certain tests

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1539

Adds an `allow_test_randomization` option that can be configured for certain tags, in the dart_test.yml file.

Also restructured the Configuration and SuiteConfiguration classes to make future changes less error prone, by making their parameters required. I also added some specialized constructors for the few cases where we really only want to pass a couple options.


pkgs/test/doc/configuration.md
```diff
@@ -24,46 +24,48 @@ tags:
...
-  * [`define_platforms`](#define_platforms)
-  * [Browser/Node.js Settings](#browser-and-node-js-settings)
-    * [`arguments`](#arguments)
-    * [`executable`](#executable)
-    * [`headless`](#headless)
-* [Configuration Presets](#configuration-presets)
-  * [`presets`](#presets)
-  * [`add_preset`](#add_preset)
-* [Global Configuration](#global-configuration)
+- [Test Configuration](#test-configuration)
```
jakemac53, line 27: Apparently my markdown plugin did this 🤣 . Should I revert? I am tempted to accept the change so I don't have to revert it every time I touch this file... lol.

pkgs/test/doc/configuration.md
```diff
@@ -191,6 +193,21 @@ test_on: vm
...
+### `allow_test_randomization`
+
+This can be used to disable test randomization for certain tests, regardless
+of the `--test-randomize-ordering-seed` configuration.
+
+This is typically useful when a subset of your tests are order dependent, but
+you want to run the other ones with randomized ordering.
+
+```yaml
+# This package uses dart:io.
```
gspencergoog, line 205: Why is this comment here? Copypasta?

jakemac53: Haha yep most definitely 🤣

pkgs/test/doc/configuration.md
natebosch: I'd be fine with a 1 time reformat to this style if it will make it easier in future PRs.

I don't want to submit this in the same commit though. Can you open a separate PR?

pkgs/test/test/utils.dart
```diff
@@ -5,22 +5,32 @@
 import 'dart:async';
 import 'dart:collection';
 
+import 'package:boolean_selector/boolean_selector.dart';
+import 'package:glob/glob.dart';
+import 'package:test_api/scaffolding.dart' // ignore: deprecated_member_use
```
natebosch, line 10: We have an import to `package:test/test.dart` which should expose `Timeout`. Why do we need this import?

natebosch (approved):

pkgs/test/doc/configuration.md
jakemac53: I was able to disable this auto-formatting in my plugin 👍

Issue #1567, opened 2021-08-27
Huge overhead for each test file loaded.


AlexMcConnell:
I first noticed this and commented on an issue in the flutter repo before tracing it back up to here (I think):

https://github.com/flutter/flutter/issues/69429#issuecomment-906627145

Essentially, loading each test file is taking longer for me than running the tests in that file.

16 tests in one file: `00:05 +16: All tests passed!`
16 tests split between 4 files: `00:08 +16: All tests passed!`
16 tests with one test per file: `00:22 +16: All tests passed!`

AlexMcConnell:
The problem has actually gotten worse. The above stats were done in Flutter 2.0.6, though upgrading to Flutter 2.2.3 provided the same stats. But...

Flutter dev - 2.6.0-0.0.pre
16 tests in one file: `00:05 +16: All tests passed!`
16 tests split between 4 files: `00:09 +16: All tests passed!`
16 tests with one test per file: `00:26 +16: All tests passed!`

Flutter master - 2.6.0-1.0.pre.105
16 tests in one file: `00:05 +16: All tests passed!`
16 tests split between 4 files: `00:09 +16: All tests passed!`
16 tests with one test per file: `00:26 +16: All tests passed!`

jakemac53:
Are these flutter tests or regular dart vm tests you are running?

AlexMcConnell:
Flutter widget tests, which I expect have their own special overhead. But the imports aren't drastically different.

svarlet:
I shared some numbers in https://github.com/flutter/flutter/issues/69429#issuecomment-903582893. Though I didn't specify that we have 62 test files (359 tests in total, 14 seconds total). These are vm, non widget, tests. A very limited few of them have file system side effects but they run in an acceptable amount of time (coming from someone who'd like a total timing < 1s).

For people who run their tests very often (every other minute for example), the amount spent waiting is unbearably long. Consequences are loss of focus, running tests less often (if ever), difficulty to teach the benefits of automated testing to juniors, devs pushing broken code, build breaking, developers stuck after pulling broken code, etc...). 

[...]

PR #1569, opened 2021-08-31
reuse incremental dill files when possible

jakemac53:
We were already caching this but weren't actually using it on subsequent runs 🤦‍♂️ . I also optimized it a bit to only copy the dill file once inside `dispose` instead of updating it repeatedly whenever we see a larger file.

This takes our `test_api` test runs from ~9 seconds to ~4 seconds.

Related to https://github.com/dart-lang/test/issues/1567.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.17.12
+
+* Support the latest `test_core`.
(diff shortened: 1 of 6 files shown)


jakemac53:
Note that at this point we are spending a fair bit of time just waiting for isolates to spawn (a total of total: 0:00:03.737934, although that is parallelized to some degree). That is at least somewhat parallelized though.

natebosch (approved):

PR #1585, opened 2021-09-28
Remove support for non-JS browsers

natebosch:
Assume that all browsers will execute JS and remove the conditionals
that existed to support the long-gone Dartium browser.

Small refactor in `_pubServeSuite` to use short-circuits for both
atypical cases of preferring JS stack traces and compile errors. This
makes the similar cases where we drain and ignore the content look
similar. Use `Stream.drain` over `Stream.listen` with a no-op handler.


jakemac53 (approved):

kevmoo:
Ha! Talk about legacy!

PR #1586, opened 2021-09-28
Mark mono_repo outputs as generated

natebosch:
Add a `.gitattributes` file to mark mono_repo outputs as generated files
which allows GitHub diff views to leave them collapsed by default.

Regenerate the file with the latest mono_repo.


kevmoo (approved):
TIL!

kevmoo:
...I also think it removes these files as going towards the languages used in the project!

jakemac53 (approved):

Issue #1311, opened 2020-07-29
Refactor the Reporter API


natebosch:
The API is currently "backwards". The `Reporter` class is designed around how the `Runner` class wants to use it, it doesn't really say anything about what the reporter should _do_, or how to implement one. The details for what the reporter should do are communicated, although not well at all, through the `ReporterFactory` typedef. What really needs to happen is for the reporter to listen on various streams from the `Engine`, but to know the specific ones you really need to just copy existing reporter behavior. This also means that lots of other irrelevant details from `Engine` leak.

We should consider refactoring to a new pattern. Some options that have been discussed:

- Write an interface in terms of the events that the reporter handles. Very roughly something like:
  ```dart
  class Reporter {
    void onTestStarted(...);
    void onTestSuccess(...);
    void onTestFailure(...);
    void onSuiteDone(...);
    ...
  }
  ```
- Remove `Engine` and create an API in terms of one or more Streams that emit values that might be `LiveTest`, or perhaps a new and more limited interface.

Current reporters also read other details from the engine, like the list of "active" tests so some consideration will bee needed about how to expose what is necessary and leave room for enhancement without leaking too many unnecessary details.

jakemac53:
We could also consider just a single Stream of objects, `onTestStateChange`. The objects in the stream would contain all the information for the state change, the Test that changed state, the new state, possibly the old state, etc.

natebosch:
I spent some time digging on this. One bit that has me worried about making a change here is this snippet:

https://github.com/dart-lang/test/blob/62f7408ddfd8e7fcc24bbbe7db3cec6684d45fc9/pkgs/test_core/lib/src/runner/reporter/json.dart#L257

This appears to rely on certain tests _not_ showing up in `Engine.liveTests` which is a combination of `passed`, `skipped`, `failed`, and `active`.

There is some logic that prevents tests which have a `.suite` which is a `LoadSuite` from getting put into `active`, but that is sometimes violated if there are no other tests in `active` when a `LoadSuite` test is added.

https://github.com/dart-lang/test/blob/62f7408ddfd8e7fcc24bbbe7db3cec6684d45fc9/pkgs/test_core/lib/src/runner/engine.dart#L449-L450

[...]

PR #1587, opened 2021-09-30
Stop putting suite loads into active test

natebosch:
Towards #1311

Removes a UX concern from `Engine` which no longer needs to worry about
how a reporter might display information about load suites. Makes the
interaction between the engine and reporters less subtle.

Use the term "suite load" over "load test" to start pushing terminology
away from conflating tests and loading.

### DIFF
--- a/pkgs/test_core/lib/src/runner/engine.dart
+++ b/pkgs/test_core/lib/src/runner/engine.dart
@@ -173,8 +173,7 @@ class Engine {
   List<LiveTest> get active => UnmodifiableListView(_active);
   final _active = QueueList<LiveTest>();
 
-  /// The tests from [LoadSuite]s that are still running, in the order they
-  /// began running.
+  /// The suites that are still loading, in the order they began.
   List<LiveTest> get activeSuiteLoads =>
       UnmodifiableListView(_activeSuiteLoads);
   final _activeSuiteLoads = <LiveTest>{};
--- a/pkgs/test_core/lib/src/runner/engine.dart
+++ b/pkgs/test_core/lib/src/runner/engine.dart
@@ -36,15 +36,10 @@ import 'util/iterable_set.dart';
 ///
 /// The engine has some special logic for [LoadSuite]s and the tests they
 /// contain, referred to as "load tests". Load tests exist to provide visibility
-/// into the process of loading test files, but as long as that process is
-/// proceeding normally users usually don't care about it, so the engine only
-/// surfaces running load tests (that is, includes them in [liveTests] and other
-/// collections) under specific circumstances.
-///
-/// If only load tests are running, exactly one load test will be in [active]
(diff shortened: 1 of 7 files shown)


pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -174,22 +169,22 @@ class Engine {
   Set<LiveTest> get failed => _failedGroup.set;
   final _failedGroup = UnionSetController<LiveTest>(disjoint: true);
 
-  /// The tests that are still running, in the order they begain running.
+  /// The tests that are still running, in the order they began running.
   List<LiveTest> get active => UnmodifiableListView(_active);
   final _active = QueueList<LiveTest>();
 
+  /// The tests from [LoadSuite]s that are still running, in the order they
+  /// began running.
```
jakemac53, line 177: Might as well change the wording here to stop referring to them as "tests" as well.

```suggestion
  /// The [LoadSuite]s that are still loading, in the order they began.
```

pkgs/test_core/lib/src/runner/reporter/compact.dart
```diff
@@ -186,10 +186,12 @@ class CompactReporter implements Reporter {
...
-    // If this is the first test to start, print a progress line so the user
-    // knows what's running. It's possible that the active test may not be
-    // [liveTest] because the engine doesn't always surface load tests.
-    if (_engine.active.length == 1 && _engine.active.first == liveTest) {
+    // If this is the first test or suite load to start, print a progress line
+    // so the user knows what's running.
+    if ((_engine.active.length == 1 && _engine.active.first == liveTest) ||
+        (_engine.liveTests.isEmpty &&
+            _engine.activeSuiteLoads.length == 1 &&
+            _engine.activeSuiteLoads.first == liveTest)) {
```
jakemac53, line 194: I don't think  you want the `liveTests` check here - that would make it not show the loading message for the currently loading test if some test has completed and some are still loading, but none is actively running.

```suggestion
        (_engine.activeSuiteLoads.length == 1 &&
            _engine.activeSuiteLoads.first == liveTest)) {
```

natebosch: Ah, yeah this should have been `_engine.active.isEmpty`, not `_engine.liveTests.isEmpty`.

pkgs/test_core/lib/src/runner/engine.dart
natebosch: Done.

pkgs/test_core/lib/src/runner/reporter/compact.dart
natebosch: We don't want to remove the check entirely or we'd show info about suites that would have never been put into `_enginge.active` before.

jakemac53 (approved):

PR #1599, opened 2021-10-07
Better exception for markTestSkipped outside test

natebosch:
See https://github.com/flutter/flutter/issues/91432

Pull out a `_currentInvoker` utility to share the exception fallback
behavior between `printOnFailure` and `markTestSkipped`.

### DIFF
--- a/pkgs/test_api/lib/src/scaffolding/utils.dart
+++ b/pkgs/test_api/lib/src/scaffolding/utils.dart
@@ -34,13 +34,7 @@ void registerException(Object error,
 /// [print], each individual message passed to [printOnFailure] will be
 /// separated by a blank line.
 void printOnFailure(String message) {
-  var invoker = Invoker.current;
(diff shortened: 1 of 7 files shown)


pkgs/test_api/lib/src/scaffolding/utils.dart
```diff
@@ -34,18 +34,18 @@ void registerException(Object error,
...
+  _currentInvoker.printOnFailure(message);
 }
 
 /// Marks the current test as skipped.
 ///
 /// A skipped test may still fail if any exception is thrown, including uncaught
 /// asynchronous errors. If the entire test should be skipped `return` from the
 /// test body after marking it as skipped.
-void markTestSkipped(String message) => Invoker.current!.skip(message);
+void markTestSkipped(String message) => _currentInvoker..skip(message);
```
lrhn, line 45: Single cascade?

PR #1602, opened 2021-10-10
Update GIFs to use 'dart test' instead of 'pub run test'

bleroux:
Update the two animated GIFs in documentation to remove reference to 'pub run test'


natebosch (approved):

natebosch:
Thanks!

PR #1603, opened 2021-10-15
Implements `test file_path.dart?name="test name"`

### LINKED ISSUE #1598: Add a more precise way to start multiple specific tests at the same time
Hello!   Some context:

I am working on a custom test renderer and I was trying to implement "execute only failed tests from the previous run" among other things.
I was able to get the list of failing tests, with their suites and names/ids. But I haven't found a way to start them all _at the same time_ while skipping the rest

I was innocently thinking that we could do:

```sh
dart test test/foo_test.dart --name "a" test/bar_test.dart --name "b"
```

to run the test "a" from `foo_test` and "b" from `bar_test`, but this isn't the case because of various issues:

- `--name` filters aren't specific to a test file.
- When multiple `--name` queries are defined, they perform an AND filter instead of an OR filter
- There doesn't seem to be a way to have a strict name filter (as in not perform a "contains" comparison but an ==).
  So `--name "hello"` will run both `test('hello', ...)` and `test('hello world', ...)`

As such I think thinking: Would it be possible to add a more specific filter?  

If updating `--name` to behave as described isn't possible due to being a breaking change, what about something like:   
`dart test --test file=test/foo_test.dart name="a" --test file="test/bar_test.dart name="b"`

I'd be more than happy to raise a pull request about this.

rrousselGit:
fixes #1598


pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -174,6 +174,22 @@ String get usage => _parser.usage;
 /// Throws a [FormatException] if [args] are invalid.
 Configuration parse(List<String> args) => _Parser(args).parse();
 
+PathConfiguration _parsePathConfiguration(String option) {
+  final uri = Uri.parse(option);
+
+  final name = uri.queryParameters['name'];
```
jakemac53, line 180: I think for this one specifically we did want to allow multiple - you can support that using `uri.queryParametersAll['name']`

pkgs/test_core/lib/src/runner/configuration.dart
```diff
@@ -30,6 +30,16 @@ import 'suite.dart';
...
 final _currentKey = Object();
 
+class PathConfiguration {
+  const PathConfiguration({
+    required this.filePath,
+    this.testName,
+  });
+
+  final String filePath;
+  final Pattern? testName;
```
jakemac53, line 40: I think we decided that we should allow multiple `name` parameters so this should be a list, maybe just call it `testPatterns`.

pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -174,6 +174,22 @@ String get usage => _parser.usage;
...
+  final fullName = uri.queryParameters['full-name'];
+
+  return PathConfiguration(
+    filePath: uri.path,
+    testName: fullName != null
```
jakemac53, line 185: And then pass the list of them here, which will mean we at least respect all options if passed, even though its probably going to be very rare to pass more than 1.

pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -177,16 +177,14 @@ Configuration parse(List<String> args) => _Parser(args).parse();
 PathConfiguration _parsePathConfiguration(String option) {
   final uri = Uri.parse(option);
 
-  final name = uri.queryParameters['name'];
+  final names = uri.queryParametersAll['name'];
   final fullName = uri.queryParameters['full-name'];
```
jakemac53, line 181: Since we won't combine these (which I think is reasonable) we should probably throw if both are passed.

kevmoo:
CC @jwren @jacob314  – interesting to ponder how to plug this in to test runners in IDES. I know JetBrains Dart plugin often messes up "running only failed" tests. This could likely (greatly) improve that experience!!

jakemac53:
@googlebot I consent

pkgs/test/README.md
```diff
@@ -1,28 +1,32 @@
...
+  - [Platform-Specific Configuration](#platform-specific-configuration)
+  - [Tagging Tests](#tagging-tests)
+  - [Whole-Package Configuration](#whole-package-configuration)
+  - [Compiler flags](#compiler-flags)
+- [Debugging](#debugging)
+- [Browser/VM Hybrid Tests](#browservm-hybrid-tests)
+- [Support for Other Packages](#support-for-other-packages)
+  - [`build_runner`](#build_runner)
+  - [`term_glyph`](#term_glyph)
+- [Further Reading](#further-reading)
```
rrousselGit, line 29:
```suggestion
* [Writing Tests](#writing-tests)
* [Running Tests](#running-tests)
  * [Sharding Tests](#sharding-tests)
  * [Collecting Code Coverage](#collecting-code-coverage)
  * [Restricting Tests to Certain Platforms](#restricting-tests-to-certain-platforms)
  * [Platform Selectors](#platform-selectors)
  * [Running Tests on Node.js](#running-tests-on-nodejs)
* [Asynchronous Tests](#asynchronous-tests)
  * [Stream Matchers](#stream-matchers)
* [Running Tests With Custom HTML](#running-tests-with-custom-html)
  * [Providing a custom HTML template](#providing-a-custom-html-template)
* [Configuring Tests](#configuring-tests)
  * [Skipping Tests](#skipping-tests)
  * [Timeouts](#timeouts)
  * [Platform-Specific Configuration](#platform-specific-configuration)
  * [Whole-Package Configuration](#whole-package-configuration)
* [Tagging Tests](#tagging-tests)
* [Debugging](#debugging)
* [Browser/VM Hybrid Tests](#browservm-hybrid-tests)
* [Support for Other Packages](#support-for-other-packages)
  * [`build_runner`](#build_runner)
  * [`term_glyph`](#term_glyph)
* [Further Reading](#further-reading)
```

pkgs/test/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
-## 1.18.3-dev
+## 1.19.0-dev
 
+* Support query parameters `name` and `full-name` on test paths, which will
```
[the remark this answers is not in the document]
jakemac53: https://api.dart.dev/stable/2.14.4/dart-io/FileSystemEntity/uri.html 🤷‍♂️ . We use uris all over the place in a non-web context.

Issue #1579, opened 2021-09-22
Provide a better way to match up (and run?) tests using the analyzer outline info (literal strings for test/group names)


DanTup:
I'm not sure there's a good solution here, but I thought it was worth starting a discussion.

-- wall of text for some background --

As part of migrating Dart-Code to use the new VS Code test runner, I hit some issues with some functionality like "Run Test at Cursor". In the original custom test runner, when a user invoked this command I would look up the analysis servers Outline node for the current cursor position, and if it was a Test or Group, extract the test name string, naively replace any `${?foo}?` with `.*` and use that as a regex to run the test.

For the new VS Code test runner, the Run Test at Cursor command is handled by VS Code and I can't do this. Instead, I have to give it all of the test `Range`s up-front when discovering the tests. On its own, this is isn't difficult - I walk the analyzers Outline node and produce all of the tests with their ranges.

The problem comes when users start running tests that have dynamic names. I can't just throw the tree away and rebuild from the analyzers Outline info as the dynamic test nodes (and the ability to see their results) would disappear. In the old test runner, I would therefore only use the Outline info once, then always use the test results (with some logic to handle partial runs that resulted in not removing deleted tests from the tree unless you ran the whole file).

So the new solution requires merging both the tree from the analyzer and the results from running tests, including rebuilding nodes from the analyzer as new Outlines arrive. For any test I discover (from the analyzer or test results) I tag where it came from and I only ever remove nodes from the source being rebuilt (eg. new Outlines can only remove nodes tagged with the Outline source, not Results, and vice-versa). Dynamic nodes are now nested under their analyzer nodes so that it's clearer where they came from (and so that if the source test is removed from the file, the results disappear with them):

![Screenshot 2021-09-22 at 12 29 29](https://user-images.githubusercontent.com/1078012/134335836-a3669883-3920-4501-aafc-2fff9aee1f35.png)

This solution is all mostly working, but it has some issues:

- Nesting the test results under the analyzer node requires is "guessing" which test results are actually for that test (and this involves turning the analyzers test name into a regex and then matching it against the node). This is the same regex we build when trying to run-test-at-cursor.. It mostly works, but I'm certain there are edge cases that could easily break it
- The `Range`s of dynamic test nodes are not correctly maintained because when a dynamic test is nested (eg. the name of a parent group is not constant) the test will be nested under the dynamic group and not the original analyzer node

---

[...]

PR #1604, opened 2021-10-15
Allow running tests by line and column

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1579

- Allows querying tests by line & col using test path query strings
- Matches if any frame in the test stack trace matches up (this allows for lots of flexibility in terms of testing styles and parameterized tests).

~This does not currently work on browsers, but I probably could get it to? Not sure how important that ultimately is.~
- If a stack trace formatter is set up for the remote platform, it will use it to format the `trace` sent back for groups/tests that are discovered (enabling browser/node support). 

cc @DanTup

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -6,8 +6,6 @@
   * Global filters (ie: `--name`) are also still respected and must match.
   * The `line` and `col` will match if any frame from the test trace matches
     (the test trace is the current stack trace where `test` is invoked).
-    * If either `line` or `col` are used, stack trace chaining is enabled for
-      that suite, regardless of the global setting.
 * Give a better exception when using `markTestSkipped` outside of a test.
 
 ## 1.18.2
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -6,8 +6,6 @@
   * Global filters (ie: `--name`) are also still respected and must match.
   * The `line` and `col` will match if any frame from the test trace matches
     (the test trace is the current stack trace where `test` is invoked).
-    * If either `line` or `col` are used, stack trace chaining is enabled for
-      that suite, regardless of the global setting.
 * Support the latest `test_api`.
 
 ## 0.4.5


pkgs/test/README.md
```diff
@@ -173,17 +173,31 @@ description (including any group descriptions) match that regular expression
...
 path/to/test.dart`.
 
+### Test Path Queries
+
+Some query parameters are supported on test paths, which allow you to filter the
+tests that will run within just those paths.
+
+- **name**: Works the same as `--name` (simple contains check).
+  - This is the only option that supports more than one entry.
+- **full-name**: Requires an match for the name of the test.
```
natebosch, line 190:
```suggestion
- **full-name**: Requires an exact match for the name of the test.
```

pkgs/test/README.md
```diff
@@ -173,17 +173,31 @@ description (including any group descriptions) match that regular expression
...
+- **line**: Matches any test that originates from this line in the test suite.
+  - Not supported for browser tests.
+- **col**: Matches any test that originates from this column in the test suite.
```
natebosch, line 193: I'm confused who would use this? Is this a specific column? Like the position of the first `t` in `test(`?

pkgs/test_core/lib/src/runner.dart
```diff
@@ -279,6 +280,14 @@ class Runner {
       }
     })).map((loadSuite) {
       return loadSuite.changeSuite((suite) {
+        // TODO: Support this for browser tests
+        if (suite.platform.runtime.isBrowser &&
```
natebosch, line 284: I think we don't support it for node tests either right?

It's probably more accurate to test `isJS` than `isBrowser`.

jakemac53: Good point

jakemac53:
Added support and a test for browser/node tests

pkgs/test/test/runner/line_and_col_test.dart
```diff
@@ -0,0 +1,361 @@
+// Copyright (c) 20121, the Dart project authors.  Please see the AUTHORS file
```
natebosch, line 1:
```suggestion
// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
```

pkgs/test/README.md
natebosch: I still don't understand the precise position this matches.

_Is_ it exactly the first `t` in `test`? Is it 0 based? Does this match the same frame as the line number, or can the line match one frame and the column match a different frame?

I'll read through the test cases tomorrow to find out more, but I do think we should expand the user facing docs.

pkgs/test_core/lib/src/runner.dart
```diff
@@ -298,6 +299,30 @@ class Runner {
             return false;
           }
 
+          // Skip tests that don't start on `line` or `col` if specified.
+          var line = suite.config.line;
+          var col = suite.config.col;
+          var trace = test.trace;
+          if (trace == null && (line != null || col != null)) {
```
natebosch, line 306: When would we be missing a trace? Should we consider it an error to try to run tests with a line and column when we have no trace available?

pkgs/test_core/lib/src/runner.dart
```diff
@@ -298,6 +299,30 @@ class Runner {
...
+            var absoluteSuitePath = p.absolute(suite.path!);
+            if (line != null &&
+                !trace.frames.any((frame) =>
+                    frame.uri.scheme == 'file' &&
+                    frame.uri.toFilePath() == absoluteSuitePath &&
+                    frame.line == line)) {
+              return false;
+            }
+            if (col != null &&
+                !trace.frames.any((frame) =>
```
natebosch, line 318: Synced offline about enforcing line and column match the same frame, not any frame at all.

pkgs/test_core/lib/src/runner.dart
jakemac53: I am not sure if we get load tests and things through this api. I will try making it an error and see what happens.

pkgs/test/CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
...
 
-* Support query parameters `name` and `full-name` on test paths, which will
-  apply the filters to only those test suites.
+* Support query parameters `name`, `full-name`, `line`, and `col` on test paths,
+  which will apply the filters to only those test suites.
   * All specified filters must match for a test to run.
   * Global filters (ie: `--name`) are also still respected and must match.
+  * The `line` and `col` will match if any frame from the test trace matches
+    (the test trace is the current stack trace where `test` is invoked).
+    * If either `line` or `col` are used, stack trace chaining is enabled for
```
natebosch, line 9: Does this mainly matter for web tests? I can't find where in the code this happens - can you point me to it?

jakemac53: It happens for all tests that run remotely (so for everything but the direct dart file invocation case), specifically here https://github.com/dart-lang/test/pull/1604/files#diff-d694d0b45775f9db69469d5db79e8b6f0bbb05b8ddcf110340a1e205c454f5a6R190

natebosch: Ah, that's a different behavior. That isn't related to `--chain-stack-traces` or `Chain.capture`.

I think we can skip this mention from the user facing changelog - it's a minor detail of the communication between the runner and the tests, but it doesn't impact user visible behavior and shouldn't impact perf as much as stack trace chaining would.

jakemac53: Ah ok, ya I was a bit confused by it

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1579

- Allows querying tests by line & col using test path query strings
- Matches if any frame in the test stack trace matches up (this allows for lots of flexibility in terms of testing styles and parameterized tests).
- If a stack trace formatter is set up for the remote platform, it will use it to format the `trace` sent back for groups/tests that are discovered (enabling browser/node support).

PR #1605, opened 2021-10-15
speed up tests by using a precompiled test runner

jakemac53:
Adds a new `precompileTestExecutable` method which you must call before calling `runTest`. It should be invoked in a top level `setUpAll` method, and will clean up its output in `tearDownAll`.

This should significantly speed up tests using `runTest` more than once.


pkgs/test/test/runner/line_and_col_test.dart
```diff
@@ -0,0 +1,321 @@
+// Copyright (c) 20121, the Dart project authors.  Please see the AUTHORS file
```
natebosch, line 1: Revert in this PR.

jakemac53: Oh whoops not sure how this file got in here lol

PR #1607, opened 2021-10-20
Mixin-application uses `with` instead of `extends`

lrhn:
Because it should.

### DIFF
--- a/pkgs/test_core/lib/src/runner/util/iterable_set.dart
+++ b/pkgs/test_core/lib/src/runner/util/iterable_set.dart
@@ -15,7 +15,7 @@ import 'package:collection/collection.dart';
 /// Note also that set operations that are usually expected to be `O(1)` or
 /// `O(log(n))`, such as [contains], may be `O(n)` for many underlying iterable
 /// types. As such, this should only be used for small iterables.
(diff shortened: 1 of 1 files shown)


pkgs/test_core/lib/src/runner/util/iterable_set.dart
```diff
@@ -15,7 +15,7 @@ import 'package:collection/collection.dart';
 /// Note also that set operations that are usually expected to be `O(1)` or
 /// `O(log(n))`, such as [contains], may be `O(n)` for many underlying iterable
 /// types. As such, this should only be used for small iterables.
-class IterableSet<E> extends SetMixin<E> with UnmodifiableSetMixin<E> {
+class IterableSet<E> with SetMixin<E> with UnmodifiableSetMixin<E> {
```
natebosch, line 18:
```suggestion
class IterableSet<E> with SetMixin<E>, UnmodifiableSetMixin<E> {
```

lrhn: Reading code is apparently hard :)

natebosch (approved):

PR #1611, opened 2021-10-21
fix parsing of windows file paths into a URI

jakemac53:
Fixes https://github.com/Dart-Code/Dart-Code/issues/3626


jakemac53:
Ah crap this won't work because now the query params etc aren't allowed :(.

We might just end up needing to manually convert `\` to `/` or something...

PR #1612, opened 2021-10-21
enable windows for the unit_tests in the test package

jakemac53:
We have ran into a few cases where we miss things due to not running _all_ tests on windows (we only run a subset).

Going to try and get this working for all tests although I suspect there will be some issues getting them running, I may just skip some individual tests for now.


natebosch (approved):

kevmoo:
Random aside – there are a LOT of places where we assume STDOUT/ERR/IN are UTF8 – and not using https://api.dart.dev/stable/2.14.4/dart-io/SystemEncoding/decoder.html

...wonder if that matters

Issue #1505, opened 2021-04-23
Every test error gives a --chain-stack-traces warning


timsneath:
Using:
 Dart SDK version: 2.13.0-211.6.beta (beta) (Wed Apr 14 13:59:12 2021 +0200) on "windows_x64"
and 
  test 1.17.1

and running `dart test` on my code, every failure gives a warning about this flag:

```
00:03 +35 -1: loading test\object_tests\structs_test.dart
Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
For example, 'pub run test --chain-stack-traces'.
00:03 +57 -2: test\system_tokens_test.dart: Can find system token 0x0100001d [E]
  Expected: 'System.ValueType'
    Actual: 'Windows.Win32.Gdi.MODIFY_WORLD_TRANSFORM_MODE'
     Which: is different.
            Expected: System.Val ...
              Actual: Windows.Wi ...
                      ^
             Differ at offset 0

  package:test_api                   expect
  test\system_tokens_test.dart 10:5  main.<fn>

00:03 +57 -2: loading test\system_tokens_test.dart
Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
For example, 'pub run test --chain-stack-traces'.
00:08 +179 -2: Some tests failed.
```

This is a regression compared to test 1.16.8, which does not give this warning.

timsneath:
(On the plus side, 1.17 is much faster for my large test suite!)

natebosch:
Yeah, disabling stack trace chaining is part of the perf improvement, but it was a negative UX impact. I'm hoping that some refactoring in the reporter implementation will make it feasible to fix this, but I don't have a timeline.

jiahaog:
Fwiw, there is some confusion internally regarding this message as well - http://shortn/_azHxrPMsQ3

clocksmith:
Running with and without the flag does not seem to produce anything different in the stack traces. The only thing different in the test output for running _without_ `--chain-stack-traces` is the addition of this this message:

```
Consider enabling the flag chain-stack-traces to receive more detailed exceptions.	
For example, 'dart test --chain-stack-traces'.
```

natebosch:
I wonder if it would be possible to detect some of the situations where chaining stack traces would no make a difference and suppress the message in that case. It might be enough to check if the original test suite shows up in the stack trace.

PR #1628, opened 2021-11-02
move log about enabling stack trace chaining to the reporters

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1505

Alternatively we could remove this log entirely, but this is one fairly easy solution.


jakemac53:
I will update the tests if we decide to go with this solution

natebosch (approved):

PR #1629, opened 2021-11-03
Wait for isolate exits from VM platform

natebosch:
Internally the test runner is run with
`--mark-main-isolate-as-system-isolate`, so when the main isolate exits
any paused-at-exit non-system isolates are also closed. This makes it
impossible to keep a debugger attached to, for instance, example the
state of the heap following the tests.

Wait for all started isolates to exit when closing the VM platform so
that the outer "system" isolate doesn't forcibly close them before they
are done.

### DIFF
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -37,7 +37,7 @@ class VMPlatform extends PlatformPlugin {
       p.join(p.current, '.dart_tool', 'test', 'incremental_kernel'));
   final _closeMemo = AsyncMemoizer<void>();
 
-  VMPlatform();
+  final _isolateExits = <Future<Result<void>>>[];
 
   @override
   StreamChannel<dynamic> loadChannel(String path, SuitePlatform platform) =>
@@ -54,6 +54,9 @@ class VMPlatform extends PlatformPlugin {
       isolate =
           await _spawnIsolate(path, receivePort.sendPort, suiteConfig.metadata);
       if (isolate == null) return null;
+      var onExitPort = ReceivePort();
+      _isolateExits.add(Result.capture(onExitPort.first));
+      isolate.addOnExitListener(onExitPort.sendPort);
     } catch (error) {
       receivePort.close();
       rethrow;
@@ -115,7 +118,10 @@ class VMPlatform extends PlatformPlugin {
   }
 
   @override
-  Future close() => _closeMemo.runOnce(() => _compiler.dispose());
+  Future close() => _closeMemo.runOnce(() => Future.wait([
+        for (var exit in _isolateExits) Result.release(exit),
+        _compiler.dispose()
+      ]));
 
   /// Spawns an isolate and passes it [message].
   ///
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -37,7 +37,7 @@ class VMPlatform extends PlatformPlugin {
       p.join(p.current, '.dart_tool', 'test', 'incremental_kernel'));
   final _closeMemo = AsyncMemoizer<void>();
 
-  final _isolateExits = <Future<Result<void>>>[];
+  final _isolateExits = <Future<void>>[];
 
   @override
   StreamChannel<dynamic> loadChannel(String path, SuitePlatform platform) =>
@@ -49,9 +49,9 @@ class VMPlatform extends PlatformPlugin {
     assert(platform.runtime == Runtime.vm);
 
     var receivePort = ReceivePort();
+    var onExitPort = ReceivePort();
     Isolate? isolate;
     try {
-      var onExitPort = ReceivePort();
       isolate = await _spawnIsolate(
           path, receivePort.sendPort, suiteConfig.metadata,
           onExit: onExitPort.sendPort);
@@ -59,9 +59,10 @@ class VMPlatform extends PlatformPlugin {
         onExitPort.close();
         return null;
       }
-      _isolateExits.add(Result.capture(onExitPort.first));
+      _isolateExits.add(onExitPort.first.onError((e, st) {}));
     } catch (error) {
       receivePort.close();
+      onExitPort.close();
       rethrow;
     }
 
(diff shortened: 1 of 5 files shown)


pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -54,6 +54,9 @@ class VMPlatform extends PlatformPlugin {
       isolate =
           await _spawnIsolate(path, receivePort.sendPort, suiteConfig.metadata);
       if (isolate == null) return null;
+      var onExitPort = ReceivePort();
+      _isolateExits.add(Result.capture(onExitPort.first));
+      isolate.addOnExitListener(onExitPort.sendPort);
```
bkonyi, line 59: Are we concerned about races here if the isolate spawns + completes before we can register the listener? Ideally we'll want to pass this port directly to `Isolate.spawnUri`'s `onExit` to ensure we're registered before the isolate runs.

bkonyi (approved):
Overall LGTM with a question about potential race conditions.

pkgs/test_core/lib/src/runner/vm/platform.dart
natebosch: No I don't think we need to be concerned. We control the initial behavior of the spawned Isolate and it won't exit until after it has run tests.

I thought about plumbing it through to the `Isolate.spawn` calls to be safer, but it adds a good deal of noise so I skipped it.

@jakemac53 - let me know if you think I'm overlooking a case where the isolate could immediately exit.

jakemac53: If you have an error in your `main` function it could presumably immediately exit?

natebosch: I don't think it could. The `main` that runs is not the user `main`, it is ours and should not have any immediate errors. If an error happens in the user's main it will still communicate that back through the `SendPort`.

https://github.com/dart-lang/test/blob/9c0ae32d1e6771a43f37e0b2a7f5bd75abe51fd3/pkgs/test_api/lib/src/backend/remote_listener.dart#L135

I can go forward with the earlier `onExit` bind though for safety.

natebosch: If we did have an immediate error I also think the effects likely wouldn't be a problem. It would mean a second error is surfaced, but shouldn't lead to a hang. The `RecievePort` will end and the `.first` call should throw an error.

jakemac53: > I don't think it could. The `main` that runs is not the user `main`, it is ours and should not have any immediate errors. If an error happens in the user's main it will still communicate that back through the `SendPort`.

Oh right good point.

We have had errors in the bootstrapping code before though and this would still help for that case (its notoriously hard to debug).

natebosch:
@bkonyi - this seems to cause some trouble when we do pass `--pause-isolates-on-exit` in order to enable coverage.

Is there some particular API we need to call to resume the exit for these isolates after collecting coverage?

natebosch:
Internally we pass `--pause-isolates-on-exit` to the test runner, and we separately run the coverage collection binary with both `--wait-paused` and `--resume-isolates`. Since we pass `--resume-isolates` I expected these to continue with their exit and allow the VM to shut down.
https://github.com/dart-lang/coverage/blob/a2c1cc0b6d1f2218a896d448b33551d0e56ea501/lib/src/collect.dart#L142-L152

@bkonyi - do you see any issues with that code? Would you expect this to cause all isolates to resume through their exit?

pkgs/test_core/lib/src/runner/vm/platform.dart
natebosch: @bkonyi - passing to the `onExit` listener seems to keep the isolate open in the cases where the isolate failed to spawn. Is there some way to prevent that? I tried also closing the `ReceivePort` in that case, but ti doesn't seem to help.

natebosch: Ah my mistake, I found another error handling path where the `RecievePort` was not getting close.

bkonyi:
> Internally we pass `--pause-isolates-on-exit` to the test runner, and we separately run the coverage collection binary with both `--wait-paused` and `--resume-isolates`. Since we pass `--resume-isolates` I expected these to continue with their exit and allow the VM to shut down. https://github.com/dart-lang/coverage/blob/a2c1cc0b6d1f2218a896d448b33551d0e56ea501/lib/src/collect.dart#L142-L152
> 
> @bkonyi - do you see any issues with that code? Would you expect this to cause all isolates to resume through their exit?

Hm, no that code looks fine to me. Just issuing a regular `resume()` request to the isolate paused at exit should result in that isolate terminating, so it looks like you're doing the right thing there.

bkonyi:
Is this change good to land?

jakemac53 (approved):

natebosch:
> Is this change good to land?

No, we still need to figure out why this breaks coverage runs.

bkonyi:
> > Is this change good to land?
> 
> No, we still need to figure out why this breaks coverage runs.

What's the status of that investigation? b/203694081 is blocked pending this change landing.

natebosch:
@bkonyi - I don't know why these isolates don't end when the coverage binary resumes them.

You can see this in action. From the `coverage` package:

```
dart bin/collect_coverage.dart --uri=http://\[::1\]:7777/ --out=coverage.json --wait-paused --resume-isolates
```

Then in the test runner `pkgs/test/` write a small `test/foo_test.dart` file with a single test.

```
dart --mark-main-isolate-as-system-isolate --pause-isolates-on-exit --enable-vm-service=7777/::1 --disable-service-auth-codes bin/test.dart test/foo_test.dart
```

I dropped some print statements in the coverage collection to verify that it's sending resume commands.

https://github.com/dart-lang/coverage/blob/7e256e00c2e1b54b0e17b3f531fc56656c06f77b/lib/src/collect.dart#L166

After this the Futures from the `onExit` listeners never complete.

natebosch:
As an (apparently separate) problem that made this difficult to debug, it looks like reading `isolate.debugName` can cause a new freeze. Adding the following patch causes this to hang all the time, not just when pausing isolates at exit.

```diff
diff --git a/pkgs/test_core/lib/src/runner/vm/platform.dart b/pkgs/test_core/lib/src/runner/vm/platform.dart
index 3e964d08..13943855 100644
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -59,7 +59,9 @@ class VMPlatform extends PlatformPlugin {
         onExitPort.close();
         return null;
       }
-      _isolateExits.add(onExitPort.first.onError((e, st) {}));
+      _isolateExits.add(onExitPort.first.onError((e, st) {}).whenComplete(() {
+        print('Isolate exit: ${isolate?.debugName}');
+      }));
     } catch (error) {
       receivePort.close();
       onExitPort.close();
```

So trying to add new debugging names and use them to track which isolates are open causes new failure modes.

PR #1631, opened 2021-11-04
Drop PlatformPlugin.loadChannel

natebosch:
This function is no longer used and has no implementations.


jakemac53 (approved):
Did you check internal/flutter as well?

natebosch:
> Did you check internal/flutter as well?

There are internal unimplemented overrides that I will remove when I roll the package.

Flutter does have a real version of this method, but the unnecessary `@override` annotation does not fail the build in that package (internally) so it should be safe for them to remove the annotation (and keep the implementation which they use) when they roll `test_core`. I don't think we need an atomic roll internally.

jakemac53:
> Flutter does have a real version of this method, but the unnecessary `@override` annotation does not fail the build in that package (internally) so it should be safe for them to remove the annotation (and keep the implementation which they use) when they roll `test_core`. I don't think we need an atomic roll internally.

Ok, we should probably pro-actively roll for them and remove it as i think they may just pin test_core if whoever is doing the roll doesn't do the fix instead?

natebosch:
I was wrong about the unnecessary `@override` being safe.

I'm actually not sure how to roll this without publishing before the roll...

Issue #1634, opened 2021-11-25
Add a flag to ignore timeouts to avoid tests accidentally timing out if debugging/stepping


DanTup:
This came up at https://github.com/Dart-Code/Dart-Code/issues/3687. When debugging tests in an IDE, you may accidentally trigger test timeouts because you were paused for some part of the test execution.

For now we're discussing having some option to have `--timeout 1hr` or similar when debugging, however there are some drawbacks to this:

1. It only sets the _default_ timeout, so doesn't override any with explicit timeouts set (that are not Factors)
2. The arbitrary value of 1h (or whatever) seems a little odd given the real goal is to disable them

It could be nice if there was a `--ignore-timeouts` flag or similar, which just entirely disabled timeouts for all tests.

jakemac53:
Ya this is a common issue I have faced, seems reasonable

DanTup:
Awesome, thanks! :)

PR #1635, opened 2021-11-29
add --ignore-timeouts flag

jakemac53:
Closes https://github.com/dart-lang/test/issues/1634

This was a bit awkard to plumb through, I ended up just sticking it on the declarer and accessing that through the invoker.

I also changed the `--pause-after-load` behavior slightly so that it implies this new flag instead of `--timeout=none`. This flag more reliably achieves the desired result (not timeouts) as it won't be overridden by local test config.

Will see about adding a test if this LGTY.

### DIFF
--- a/pkgs/test_api/lib/src/backend/declarer.dart
+++ b/pkgs/test_api/lib/src/backend/declarer.dart
@@ -112,6 +112,9 @@ class Declarer {
   /// `null`.
   final Set<String>? _seenNames;
 
+  /// Whether or not timeouts should be ignored.
+  final bool ignoreTimeouts;
+
   /// Creates a new declarer for the root group.
   ///
   /// This is the implicit group that exists outside of any calls to `group()`.
@@ -139,6 +142,7 @@ class Declarer {
     String? fullTestName,
     // TODO: Change the default https://github.com/dart-lang/test/issues/1571
     bool allowDuplicateTestNames = true,
+    bool ignoreTimeouts = false,
   }) : this._(
             null,
             null,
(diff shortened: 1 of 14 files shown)


natebosch:
> as it won't be overridden by local test config.

What are the current precedence rules?

natebosch (approved):

jakemac53:
> What are the current precedence rules?

The more local the configuration the higher precedence. So arguments passed to test/group are highest precedence, then suite level config (annotations), and then command line/yaml config (not sure the precedence between those).

Basically though what that means is setting the `timeout` to `none` won't consistently do what we want (even if we overrode the suite metadata, a `timeout` argument to `test` or `group` would override it).

jakemac53:
I will look into the test failures here (seems like the declarer isn't set up yet the first time this gets called).

jakemac53:
@natebosch when writing some tests I realized that attaching this to the declarer is actually fairly problematic because the "current" declarer is not always set/consistent (even between the `declare` and `build` calls).

I moved it to the `Suite` class, which is always there. This also allowed me to disable timing out of load suites with this flag, which I chose to do but happy to discuss if we think that is the correct option.

pkgs/test_api/lib/src/backend/declarer.dart
```diff
@@ -112,6 +112,9 @@ class Declarer {
   /// `null`.
   final Set<String>? _seenNames;
 
+  /// Whether or not timeouts should be ignored.
+  final bool ignoreTimeouts;
```
natebosch, line 116: Where is this read now?

natebosch (approved):

pkgs/test_api/lib/src/backend/declarer.dart
jakemac53: oh whoops I meant to drop it here :)

jakemac53:
Unfortunately the addition of the `configuration.dart` import here is causing problems internally (it transitively imports `dart:io`).

Going to have to think a bit on the best way to restructure things.

PR #1636, opened 2021-11-30
Log to stdout from chrome

natebosch:
Log to `stdout` instead of `stderr` so that it works on windows and we
can (hopefully) get more useful output.


natebosch:
This seems to resolve the failures on windows - I think it's worth publishing to see if we hit another failure where it might produce something useful. We could also see if we can hit a repro on this PR to prove it does something useful...

jakemac53 (approved):

PR #1638, opened 2021-11-30
Revert "add --ignore-timeouts flag (#1635)"

jakemac53:
This reverts commit bb4f7e85e0e490cae61bbd13245bbd6dd811263d.

This needs to be reverted because it breaks us internally due to the addition of the `configuration.dart` import to some files that previously didn't transitively import `dart:io`.


jakemac53:
We will have to rethink a bit how this is structured to avoid this.

natebosch (approved):

PR #1643, opened 2021-12-10
Add an extension to pause VM isolates after tests

natebosch:
Enables use cases like checking memory usage after a test has run.

This is primarily aimed at an internal use case, if it turns out to be
more widely useful we can better document it. For now it is considered
experimental.

Add a separate channel for communication between the VM platform and the
test isolate bootstrapping. Use a virtual channel for the existing
communication between the `RemoteListener` and the test control helper.
The only messages that are sent across this channel for now are a signal
from the platform to the isolate to pause, and a signal back indicating
that the pause is complete and the isolate can be killed.

Register a service extension (once globally) when loading a test for the
VM platform. The setting is global, which may need to be changed if we
extend this to support external use cases. Internal tests only ever run
a single test suite, so it makes no difference whether the setting is
global.

### DIFF
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 import 'dart:developer';
 import 'dart:io';
 import 'dart:isolate';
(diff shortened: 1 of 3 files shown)


pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -63,8 +69,15 @@ class VMPlatform extends PlatformPlugin {
 
     VmService? client;
     StreamSubscription<Event>? eventSub;
-    var channel = IsolateChannel.connectReceive(receivePort)
-        .transformStream(StreamTransformer.fromHandlers(handleDone: (sink) {
+    var outerChannel = MultiChannel(IsolateChannel.connectReceive(receivePort));
```
jakemac53, line 72: It would be good do document the flow here in comments since its a bit hard to follow with all the multi channels, and stream queues and things.

jakemac53 (approved):

pkgs/test_core/lib/src/runner/vm/platform.dart
natebosch: I added a description of `channel` and `outerChannel` in a comment.

Longer term I think I can refactor a bit here, but I'll wait on that for now.

PR #1646, opened 2021-12-15
Migrate a ??= getter to a late final field

natebosch:
This field was initialized lazily manually because it needs to refer
to other fields on `this`. Using a `late final` field with an
initializing expression is a more terse way to express this than an
additional private field.


natebosch:
hmm, we're also using a weird pattern with `_knownPresets`...

jakemac53 (approved):
Ya looks like `_knownPresets` will have to be reverted (or change how it works significantly), but the other lgtm.

jakemac53 (approved):

PR #1661, opened 2022-01-08
Canonicalize APIs exported by pkg:test_api

kevmoo:
Also fix a bunch of spelling throughout the monorepo

### DIFF
--- a/pkgs/test_api/lib/scaffolding.dart
+++ b/pkgs/test_api/lib/scaffolding.dart
@@ -2,6 +2,12 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// {@canonicalFor on_platform.OnPlatform}
+/// {@canonicalFor retry.Retry}
+/// {@canonicalFor skip.Skip}
+/// {@canonicalFor tags.Tags}
+/// {@canonicalFor test_on.TestOn}
+/// {@canonicalFor timeout.Timeout}
 @Deprecated('package:test_api is not intended for general use. '
     'Please use package:test.')
 library test_api.scaffolding;


pkgs/test_api/lib/scaffolding.dart
```diff
@@ -2,6 +2,12 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// {@canonicalFor on_platform.OnPlatform}
```
kevmoo, line 5: image

natebosch:
> Also fix a bunch of spelling throughout the monorepo

Can this go out separately? Spelling fixes will be easy to approve without much consideration and not likely to need rollback.

I think we need to pay close attention to the canonicalization choices so I'd prefer if this PR didn't have extra noise.

pkgs/test_api/lib/scaffolding.dart
natebosch: I think we're still in a place where we expect most folks to get these through `package:test/test.dart`, and if they didn't get it from there I think we'd want them to use `package:test/scaffolding.dart`, not `package:test_api/scaffolding.dart`.

kevmoo:
PTAL

pkgs/test_api/lib/scaffolding.dart
kevmoo: We can put notes here saying such.

In the short term, this fixes the dartdoc warnings.

jakemac53: Are there analyzer implications here? Will it start adding this import instead of `package:test/test.dart` or `package:test/scaffolding.dart`? It seems reasonable that it would, but then this isn't the behavior we want.

jakemac53: Can you maybe clarify or point to some docs on exactly what this is/does?

kevmoo: https://github.com/dart-lang/dartdoc/issues/2887

Sadly, there are no docs. Basically, if one member is exposed by more than one library, `dartdoc` guesses where to put it.

This "annotation" makes things explicit

jakemac53: Ok, it sounds like this should be fairly low impact then and we could change it later if we decided to?

kevmoo: yup

jakemac53 (approved):
Lgtm although might want to wait for approval from @natebosch as well.

natebosch (approved):

PR #1662, opened 2022-01-08
Fix a number of spelling issues

kevmoo:


pkgs/test/test/runner/json_reporter_utils.dart
```diff
@@ -5,10 +5,9 @@
 import 'dart:convert';
 
 import 'package:path/path.dart' as p;
-import 'package:test_descriptor/test_descriptor.dart' as d;
```
jakemac53, line 8: Are we no longer putting imports from other packages in a different section?

kevmoo: Just using tool sorting seems like the most stable, canonical thing to do!

jakemac53: I don't see an opinion here https://dart.dev/guides/language/effective-dart/style#ordering so I am fine with this, just curious more than anything? I thought we used to recommend it separating them, but it always felt a bit arbitrary imo.

jakemac53 (approved):

pkgs/test/test/runner/json_reporter_utils.dart
jakemac53: Ahah! https://github.com/dart-lang/site-www/pull/2405 :)

PR #1673, opened 2022-02-28
migrate back to ConstructorName.type from ConstructorName.type2

jakemac53:
The `type2` api is now deprecated, breaking our CI. This requires a bump in analyzer min version, but we don't need to publish right away.


natebosch (approved):

jakemac53:
Oh weird good catch!

Issue #1578, opened 2021-09-14
Add some retry around launching browsers


natebosch:
https://github.com/dart-lang/test/blob/d6329d676c7578c984a5397a6cc4002757c0b21e/pkgs/test/lib/src/runner/browser/browser_manager.dart#L129-L133

We seem to hit timeouts much more often on github actions.

natebosch:
@grouma - does `package:browser_launcher` already have retry?

https://github.com/dart-lang/test/issues/1156

grouma:
I don't think so: https://github.com/dart-lang/browser_launcher/blob/master/lib/src/chrome.dart

PR #1676, opened 2022-03-16
Add some retry around browser launching

natebosch:
Towards #1578

Recursively attempt to launch the browser on the first few timeouts
before throwing the ApplicationException.

### DIFF
--- a/pkgs/test/lib/src/runner/browser/browser_manager.dart
+++ b/pkgs/test/lib/src/runner/browser/browser_manager.dart
@@ -137,7 +137,7 @@ class BrowserManager {
       completer.completeError(error, stackTrace);
     });
 
-    return completer.future.timeout(Duration(seconds: 45), onTimeout: () {
+    return completer.future.timeout(Duration(seconds: 30), onTimeout: () {
       browser.close();
       if (attempt >= _maxRetries) {
         throw ApplicationException(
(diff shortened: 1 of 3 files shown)


natebosch:
I wonder if we should lower the timeout back down too. It had been 30 seconds before and we bumped it to 45 to see if it would reduce flakiness. I don't think it's made any difference - my guess is there is some state we can get in where it will never connect.

pkgs/test/lib/src/runner/browser/browser_manager.dart
```diff
@@ -127,11 +137,14 @@ class BrowserManager {
...
       browser.close();
-      throw ApplicationException(
-          'Timed out waiting for ${runtime.name} to connect.\n'
-          'Browser output: ${utf8.decode(browser.output)}');
+      if (attempt >= _maxRetries) {
+        throw ApplicationException(
+            'Timed out waiting for ${runtime.name} to connect.\n'
+            'Browser output: ${utf8.decode(browser.output)}');
+      }
+      return _start(runtime, url, future, settings, configuration, attempt++);
```
jakemac53, line 147:
```suggestion
      return _start(runtime, url, future, settings, configuration, ++attempt);
```

(this would retry infinitely as written)

natebosch: 😳

natebosch: Fixed

jakemac53 (approved):

Issue #1415, opened 2021-01-27
Add a Github specific reporter and enable it automatically by default based on environment variables


natebosch:
The default compact reporter is too noisy on github which does not support the line overwriting ascii escapes.

We should make a reporter which is less active, and consider adding support for the specific output format which allows the github UI to overlay failures on the code.

kevmoo:
I did make a stab at this...nothing much to show of it, though https://github.com/dart-lang/test/tree/action_fun_null_safe

devoncarew:
I did some experimentation with the github actions output options here:

https://github.com/devoncarew/output_tester/issues/2#issuecomment-1069826373

tldr is that there are some net improvements that can be made to the output, but the workflow commands are slightly less expressive overall than you'd want.

<img width="513" alt="Screen Shot 2022-03-17 at 9 30 11 AM" src="https://user-images.githubusercontent.com/1269969/158848776-1bcb22fc-a521-4a78-8c0d-5b1a87c5422e.png">

as compared to something like:

```
Run dart test

00:00 +0: loading test/output_tester_test.dart                                                                                                                                                         
00:01 +0: loading test/output_tester_test.dart                                                                                                                                                         
00:01 +0: test/output_tester_test.dart: calculate.1                                                                                                                                                    
00:01 +1: test/output_tester_test.dart: calculate.1                                                                                                                                                    
00:01 +1: test/output_tester_test.dart: calculate.2                                                                                                                                                    
00:01 +1 -1: test/output_tester_test.dart: calculate.2 [E]                                                                                                                                             
  Expected: <41>
    Actual: <42>
  
  package:test_api                   expect
  test/output_tester_test.dart 10:5  main.<fn>
  test/output_tester_test.dart [9](https://github.com/devoncarew/output_tester/pull/1/checks#step:8:9):[23](https://github.com/devoncarew/output_tester/pull/1/checks#step:8:23)  main.<fn>
  

00:01 +1 -1: test/output_tester_test.dart: group.a calculate.3                                                                                                                                         
00:01 +2 -1: test/output_tester_test.dart: group.a calculate.3                                                                                                                                         
00:01 +2 -1: test/output_tester_test.dart: group.a calculate.4                                                                                                                                         
00:02 +2 -1: test/output_tester_test.dart: group.a calculate.4                                                                                                                                         
00:02 +3 -1: test/output_tester_test.dart: group.a calculate.4                                                                                                                                         
[...]

PR #1678, opened 2022-03-18
initial implementation of a github actions reporter

devoncarew:
- initial implementation of a github actions reporter
- address https://github.com/dart-lang/test/issues/1415

Creating as a draft PR for now - I want to play with the output some.

### DIFF
--- a/pkgs/test_core/lib/src/runner/reporter/github.dart
+++ b/pkgs/test_core/lib/src/runner/reporter/github.dart
@@ -1,0 +1,188 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: implementation_imports
+
+import 'dart:async';
+
+import 'package:test_api/src/backend/live_test.dart';
+import 'package:test_api/src/backend/message.dart';
+import 'package:test_api/src/backend/state.dart';
+import 'package:test_api/src/backend/util/pretty_print.dart';
+
+import '../engine.dart';
+import '../load_suite.dart';
+import '../reporter.dart';
+
+/// A reporter that prints test output using formatting for Github Actions.
+///
+/// See
+/// https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions
+/// for a description of the output format, and
+/// https://github.com/dart-lang/test/issues/1415 for discussions about this
+/// implementation.
+class GithubReporter implements Reporter {
+  /// The engine used to run the tests.
+  final Engine _engine;
+
+  /// Whether the path to each test's suite should be printed.
+  final bool _printPath;
+
+  /// Whether the reporter is paused.
+  var _paused = false;
+
+  /// The set of all subscriptions to various streams.
+  final _subscriptions = <StreamSubscription>{};
+
+  final StringSink _sink;
+  final _helper = _GithubHelper();
+
+  final Map<LiveTest, List<Message>> _testMessages = {};
+
+  /// Watches the tests run by [engine] and prints their results as JSON.
+  static GithubReporter watch(Engine engine, StringSink sink,
+          {required bool printPath}) =>
+      GithubReporter._(engine, sink, printPath);
+
+  GithubReporter._(this._engine, this._sink, this._printPath) {
+    _subscriptions.add(_engine.onTestStarted.listen(_onTestStarted));
+    _subscriptions.add(_engine.success.asStream().listen(_onDone));
+  }
+
+  @override
+  void pause() {
+    if (_paused) return;
+    _paused = true;
+
+    for (var subscription in _subscriptions) {
+      subscription.pause();
+    }
+  }
+
+  @override
+  void resume() {
+    if (!_paused) return;
+    _paused = false;
+
+    for (var subscription in _subscriptions) {
+      subscription.resume();
+    }
+  }
+
+  void _cancel() {
+    for (var subscription in _subscriptions) {
+      subscription.cancel();
+    }
+    _subscriptions.clear();
+  }
+
+  /// A callback called when the engine begins running [liveTest].
+  void _onTestStarted(LiveTest liveTest) {
+    // Convert the future to a stream so that the subscription can be paused or
+    // canceled.
+    _subscriptions.add(
+        liveTest.onComplete.asStream().listen((_) => _onComplete(liveTest)));
+
+    // Collect messages from tests as they are emitted.
+    _subscriptions.add(liveTest.onMessage.listen((message) {
+      _testMessages.putIfAbsent(liveTest, () => []).add(message);
+    }));
+  }
+
+  /// A callback called when [liveTest] finishes running.
+  void _onComplete(LiveTest test) {
+    final errors = test.errors;
+    final messages = _testMessages[test] ?? [];
+    final skipped = test.state.result == Result.skipped;
+    final failed = errors.isNotEmpty;
+
+    void emitMessages(List<Message> messages) {
+      for (var message in messages) {
+        _sink.writeln(message.text);
+      }
+    }
+
+    void emitErrors(List<AsyncError> errors) {
+      for (var error in errors) {
+        _sink.writeln('${error.error}');
+        _sink.writeln(error.stackTrace.toString().trimRight());
+      }
+    }
+
+    final isLoadSuite = test.suite is LoadSuite;
+    if (isLoadSuite) {
+      // Don't emit any info for 'loadSuite' tests, unless they contain errors.
+      if (errors.isNotEmpty || messages.isNotEmpty) {
+        _sink.writeln('${test.suite.path}:');
+        emitMessages(messages);
+        emitErrors(errors);
+      }
+
+      return;
+    }
+
+    final prefix = failed
+        ? _GithubHelper.failedIcon
+        : skipped
+            ? _GithubHelper.skippedIcon
+            : _GithubHelper.passedIcon;
+    final statusSuffix = failed
+        ? ' (failed)'
+        : skipped
+            ? ' (skipped)'
+            : '';
+
+    var name = test.test.name;
+    if (_printPath && test.suite.path != null) {
+      name = '${test.suite.path}: $name';
+    }
+    _sink.writeln(_helper.startGroup('$prefix $name$statusSuffix'));
+    emitMessages(messages);
+    emitErrors(errors);
+    _sink.writeln(_helper.endGroup);
+  }
+
+  void _onDone(bool? success) {
+    _cancel();
+
+    _sink.writeln();
+
+    final hadFailures = _engine.failed.isNotEmpty;
+    String message =
+        '${_engine.passed.length} ${pluralize('test', _engine.passed.length)} passed';
+    if (_engine.failed.isNotEmpty) {
+      message += ', ${_engine.failed.length} failed';
+    }
+    if (_engine.skipped.isNotEmpty) {
+      message += ', ${_engine.skipped.length} skipped';
+    }
+    message += '.';
+    _sink.writeln(hadFailures ? _helper.error(message) : message);
+  }
+
+  // todo: do we need to bake in awareness about tests that haven't completed
+  // yet?
+
+  // ignore: unused_element
+  String _normalizeTestResult(LiveTest liveTest) {
+    // For backwards-compatibility, report skipped tests as successes.
+    if (liveTest.state.result == Result.skipped) return 'success';
+    // if test is still active, it was probably cancelled
+    if (_engine.active.contains(liveTest)) return 'error';
+    return liveTest.state.result.toString();
+  }
+}
+
+class _GithubHelper {
+  static const String passedIcon = '✅';
+  static const String failedIcon = '❌';
+  static const String skippedIcon = '☑️';
+
+  _GithubHelper();
+
+  String startGroup(String title) => '::group::$title';
+  final String endGroup = '::endgroup::';
+
+  String error(String message) => '::error::$message';
+}
--- a/pkgs/test_core/lib/src/util/io.dart
+++ b/pkgs/test_core/lib/src/util/io.dart
@@ -87,6 +87,12 @@ final _tempDir = Platform.environment.containsKey('_UNITTEST_TEMP_DIR')
 bool get canUseSpecialChars =>
     (!Platform.isWindows || stdout.supportsAnsiEscapes) && !inTestTests;
 
+/// Detect whether we're running in a Github Actions context.
+///
+/// See
+/// https://docs.github.com/en/actions/learn-github-actions/environment-variables.
+bool get githubContext => Platform.environment['GITHUB_ACTIONS'] == 'true';
+
 /// Creates a temporary directory and returns its path.
 String createTempDir() =>
     Directory(_tempDir).createTempSync('dart_test_').resolveSymbolicLinksSync();
--- a/pkgs/test_core/lib/src/runner/configuration/reporters.dart
+++ b/pkgs/test_core/lib/src/runner/configuration/reporters.dart
@@ -11,6 +11,7 @@ import '../engine.dart';
 import '../reporter.dart';
 import '../reporter/compact.dart';
 import '../reporter/expanded.dart';
+import '../reporter/github.dart';
 import '../reporter/json.dart';
 
 /// Constructs a reporter for the provided engine with the provided
@@ -43,6 +44,11 @@ final _allReporters = <String, ReporterDetails>{
           printPath: config.paths.length > 1 ||
               Directory(config.paths.single.testPath).existsSync(),
           printPlatform: config.suiteDefaults.runtimes.length > 1)),
+  'github': ReporterDetails(
+      'Custom output for Github Actions.',
+      (config, engine, sink) => GithubReporter.watch(engine, sink,
+          printPath: config.paths.length > 1 ||
+              Directory(config.paths.single.testPath).existsSync())),
   'json': ReporterDetails(
       'A machine-readable format (see '
       'https://dart.dev/go/test-docs/json_reporter.md).',
@@ -52,9 +58,11 @@ final _allReporters = <String, ReporterDetails>{
 
 final defaultReporter = inTestTests
     ? 'expanded'
-    : canUseSpecialChars
-        ? 'compact'
-        : 'expanded';
+    : githubContext
+        ? 'github'
+        : canUseSpecialChars
+            ? 'compact'
+            : 'expanded';
 
 /// **Do not call this function without express permission from the test package
 /// authors**.


pkgs/test_core/lib/src/runner/reporter/github.dart
```diff
@@ -0,0 +1,196 @@
...
+    _sink.writeln(_helper.endGroup);
+  }
+
+  void _onDone(bool? success) {
+    _cancel();
+
+    _sink.writeln();
+
+    final hadFailures = _engine.failed.isNotEmpty;
+    String message =
```
jakemac53, line 155: Probably doesn't matter in practice but we should use a StringBuffer here

pkgs/test_core/lib/src/runner/reporter/github.dart
```diff
@@ -0,0 +1,196 @@
...
+        _sink.writeln('${error.error}');
+        _sink.writeln(error.stackTrace.toString().trimRight());
+      }
+    }
+
+    final isLoadSuite = test.suite is LoadSuite;
+    if (isLoadSuite) {
+      // Don't emit any info for 'loadSuite' tests, unless they contain errors.
+      if (errors.isNotEmpty || messages.isNotEmpty) {
+        _sink.writeln('${test.suite.path}:');
```
jakemac53, line 120: Should we also be visualizing this as a group? If not why not?

devoncarew: I believe this 'LiveTest' is just plumbing used by package:test during test discovery - it's not an actual user test, and reporting it as such would be odd for the user.

We also special case these LoadSuite test in the other reporters I believe (but don't have the code in front of me right now).

jakemac53: So what this is specifically is a synthetic "test" used to report the success or failure of compiling, loading, and connecting to the test. Basically, it fails if there is a compilation error in the test or some failure connecting to it.

We do want to surface it for sure, and it looks a bit weird imo to not have it show up the same way (I think it makes sense to have the actual errors hidden behind a collapsable group still for instance).

pkgs/test_core/lib/src/runner/reporter/github.dart
```diff
@@ -113,6 +114,9 @@ class GithubReporter implements Reporter {
       }
     }
 
+    // TODO: how to recognize (setUpAll) and (tearDownAll)?
```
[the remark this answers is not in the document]
devoncarew: Do these show up in the other test reporters? I really don't remember them displaying in test results.

I think it's odd - if people have 3 tests defined - to display results for 5 tests. Are the results for these included in the `engine.passed` set of tests? Again, I think it would be odd - if they have 3 tests defined in a project - for the test run to conclude with `All 5 tests passed`.

Perhaps a setup and teardown phase can be visualized a bit differently from a regular test?

▶ test/message_extraction/failed_extraction_test.dart: foo.setUpAll()
▶ test/message_extraction/failed_extraction_test.dart: foo.tearDownAll()

pkgs/test_core/lib/src/runner/reporter/github.dart
devoncarew: So a `Loading test/message_extraction/failed_extraction_test.dart` group, which would generally be empty, but when it had content would be important to display (esp. as likely no other tests would run).

I wonder if we shouldn't hide it normally - as mostly it'll be empty and be noise in the rest of the output.

pkgs/test_core/lib/src/runner/reporter/github.dart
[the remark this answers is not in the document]
jakemac53: > Do these show up in the other test reporters? I really don't remember them displaying in test results.

They do show up today yes (you would usually only see them if they fail, like all other tests in the compact reporter). On github (expanded reporter) you definitely see them though (something like `00:00 +0: (setUpAll)`).

> I think it's odd - if people have 3 tests defined - to display results for 5 tests. 

Yes, it is a bit odd and the current reporters _don't_ count these in the total number of "successes". It gets weirder if they fail, they do count as a failure, and then no other tests run at all so it reports one test failing regardless of how many tests are in the suite (the others disappear basically).

> Perhaps a setup and teardown phase can be visualized a bit differently from a regular test?

I am fine with them being visualized differently, to some extent this already happens (the name is `(setUpAll)` in the example runs here). Not sure if it would also show the group name, but that would be good.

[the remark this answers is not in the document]
devoncarew: From some experimentation, setUp() and tearDown() don't cause new tests to be created; setUpAll() and tearDownAll() do.

They're comparatively rare, so I think what we do here isn't critical. I'll likely update this to display them using groups, and make sure the group doesn't look like a normal, user-authored test.

pkgs/test_core/lib/src/runner/reporter/github.dart
jakemac53: If we only show it at all when it fails, then I would call it `Failed to load test/message_extraction/failed_extraction_test.dart`. But ya that SGTM

pkgs/test_core/lib/src/runner/reporter/github.dart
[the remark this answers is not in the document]
jakemac53: > From some experimentation, setUp() and tearDown() don't cause new tests to be created; setUpAll() and tearDownAll() do.

Ya, because these run _per test_ they are considered just a part of the test. The thing with `setUpAll` is it runs once, before all the tests, and so you can't link it to any one of them. I guess you could link it to _all_ of them though? But I think that just ends up being noisy, would duplicating the same error in each test.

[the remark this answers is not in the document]
devoncarew: Gotcha - thanks for the background, makes sense.

pkgs/test_core/lib/src/runner/reporter/github.dart
devoncarew: Updated to use a StringBuffer

pkgs/test_core/lib/src/runner/reporter/github.dart
```diff
@@ -0,0 +1,198 @@
...
+      hadFailures
+          ? _helper.error(message.toString())
+          : '${_GithubHelper.success} $message',
+    );
+  }
+}
+
+class _GithubHelper {
+  // Char sets avilable at https://www.compart.com/en/unicode/.
+  static const String passed = '✅';
```
kevmoo, line 185: Soon this could be an enum! 🤤

devoncarew:
OK, I think this is ready for review now. Here's the current output:
- everything (tests, loading diagnostics (loadSuite), setUpAll and tearDownAll methods get wrapped in groups
- if there's no content (errors or messages) in the synthetic (non-test) groups, they're not emitted
- test successes have ✅ icons; failures have ❌ icons
- synthetic groups that pass use icons that are different from tests so they're not confused with user authored tests; ⏺
- skipped tests have ❎ icons; I played with many different available icons here - this seemed like the best; many others were false starts - (on github were rendered w/ a red color - indicating a failure, in a different width than the other icons, making the left edge of the output ragged, ...)
- we print a success message at the end with 🎉; if a failure, we use the github `::error::` output

Some pics for the output below; others can be examined from the 'Checks' tab.

The tests for this were adopted from the ones for the other reporters. Some common ones were removed - ones that tested other functionality but that used the reporter to do that (the expanded and compact reporters still have these tests). And tests added for the github specific UI.

One thing likely to address before landing is that because of the way we collect output until test is complete, we can re-sequence the order of stdout and error messages.

Screen Shot 2022-03-20 at 10 23 49 AM

Screen Shot 2022-03-20 at 10 23 13 AM

Screen Shot 2022-03-20 at 10 22 48 AM

Screen Shot 2022-03-17 at 9 30 11 AM

pkgs/test_core/lib/src/runner/reporter/github.dart
devoncarew: I've updated this to show the load suite info if it has any content - messages or errors; it shows as a group, w/ a different icon from a test.

devoncarew:
This reporter also doesn't have the:

```
Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
For example, 'dart test --chain-stack-traces'.
```

message at the end of every run (I'm not quite sure what that is, or if there's not some other way to let users know).

pkgs/test_core/lib/src/runner/reporter/github.dart
natebosch, line 185: [nit] I have a slight preference for `'\u2705'`, but I'm also curious what other folks think.

@grouma @jakemac53 @lrhn ?

pkgs/test_core/lib/src/runner/reporter/github.dart
```diff
@@ -0,0 +1,198 @@
...
+
+    var name = test.test.name;
+    if (!loadSuite) {
+      if (_printPath && test.suite.path != null) {
+        name = '${test.suite.path}: $name';
+      }
+    }
+    _sink.writeln(_helper.startGroup('$prefix $name$statusSuffix'));
+    // TODO: Note that printing messages and errors in this manner could display
+    // them in a different order than they were generated.
```
natebosch, line 149: Do you mean specifically that messages and errors won't be interleaved, not that messages could be out of order relative to other messages?

I think that's probably fine.

I'm curious though if we should try to do anything with tests that print after they are complete - do I understand correctly that they'd get lost?

devoncarew: The messages will be in the right order, and the errors will be in the right order, but this might interleave the two types differently than how they'd been created.

devoncarew: > I'm curious though if we should try to do anything with tests that print after they are complete - do I understand correctly that they'd get lost?

No - we specifically handle that case (before getting the associated test to pass we had been dropping them). We now track when tests are completed; if we get output for a test which has already completed, we just emit the output directly to the sink (w/o any batching). The output will appear between test groups - it may not be clear what test it's from, but at least it won't be lost.

pkgs/test_core/lib/src/runner/reporter/github.dart
kevmoo: `'\u2705'` is NOT super readable 🤷  – why the preference?

kevmoo (changes requested):
Should update the changelog describing the change

Should also add something to the readme explaining how GitHub CI detection is handled – so folks aren't confused.

Looks like it's just auto-magically changed default (👍 ), so it'd be good to document that the user can override the behavior by being explicit w/ the `--reporter` option

pkgs/test_core/lib/src/runner/reporter/github.dart
natebosch: Historically they've not worked well in some editors and they aren't easy to type.
Mostly we just haven't used them in source strings that I recall, and I've only seen the `\u` version used.

devoncarew:
> Should update the changelog describing the change
> 
> Should also add something to the readme explaining how GitHub CI detection is handled – so folks aren't confused.
> 
> Looks like it's just auto-magically changed default (👍 ), so it'd be good to document that the user can override the behavior by being explicit w/ the `--reporter` option

Makes sense - I updated the changelog and added a section about test reporters to the readme.

I also rev'd the version of package:test to a pre-release version.

pkgs/test_core/lib/src/runner/reporter/github.dart
devoncarew: I changed these to use the unicode escaping form.

devoncarew:
In terms of the package:test version, I rev'd this from `1.20.2` to `1.21.0-dev`.

pkgs/test/README.md
```diff
@@ -233,6 +236,19 @@ dart test --test-randomize-ordering-seed=random
 Setting `--test-randomize-ordering-seed=0` will have the same effect as not
 specifying it at all, meaning the test order will remain as-is.
 
+### Selecting a Test Reporter
+
+You can adjust the output format of the test results using the `--reporter=<option>`
+command line flag. The default format is the `compact` output format - a single
+line, continuously updated as tests are run. When running on the GitHub Actions CI
+however, the default changes to the `github` output format - a reporter customized
```
kevmoo, line 244: Pondering if we should be explicit abouth the ENV var we check. So it's clear it's not magic.

kevmoo (approved):

devoncarew:
> In terms of the package:test version, I rev'd this from `1.20.2` to `1.21.0-dev`.

cc @natebosch, @jakemac53 re: the version we should use here

jakemac53:
Minor version bump SGTM.

pkgs/test_core/lib/src/runner/reporter/github.dart
```diff
@@ -0,0 +1,196 @@
...
+    message.write('.');
+    _sink.writeln(
+      hadFailures
+          ? _helper.error(message.toString())
+          : '${_GithubHelper.success} $message',
+    );
+  }
+}
+
+class _GithubHelper {
```
kevmoo, line 181: Why not just make this a `abstract` class w/ everything static?

Planning on having state at some point?

pkgs/test/test/runner/github_reporter_test.dart
```diff
@@ -0,0 +1,346 @@
...
+  });
+
+  test('setUpAll and tearDownAll show when they have content', () {
+    return _expectReport('''
+          group('one', () {
+            setUpAll(() { print('one'); });
+            tearDownAll(() { print('two'); });
+            test('test 1', () {});
+          });''', '''
+          ::group::⏺ one (setUpAll)
```
jakemac53, line 298: Why not use ✅ for these as well? This looks to me like a warning indicator.

I would consider using this icon for skipped tests actually?

pkgs/test_core/lib/src/runner/reporter/github.dart
jakemac53: Or just make these top level private members

pkgs/test_core/lib/src/util/io.dart
```diff
@@ -87,6 +87,12 @@ final _tempDir = Platform.environment.containsKey('_UNITTEST_TEMP_DIR')
 bool get canUseSpecialChars =>
     (!Platform.isWindows || stdout.supportsAnsiEscapes) && !inTestTests;
 
+/// Detect whether we're running in a Github Actions context.
+///
+/// See
+/// https://docs.github.com/en/actions/learn-github-actions/environment-variables.
+bool get githubContext => Platform.environment['GITHUB_ACTIONS'] == 'true';
```
jakemac53, line 94: I could call this maybe `inGithubContext` or something, it doesn't sound like a boolean to me with the current name

pkgs/test_core/lib/src/runner/reporter/github.dart
grouma: Sorry for the late response. I think I prefer `\u2705` as a constant variable with an appropriate name, e.g. `_checkMark`. That way it's readable and will work with all tools / editors.

natebosch: One point against the `'\u'` form is that the Java and JavaScript style guides at google both seem to encourage it when it's "easier to read", which I think does apply here.
https://google.github.io/styleguide/javaguide.html#s2.3.3-non-ascii-characters
https://google.github.io/styleguide/jsguide.html#non-ascii-characters

C++ mentions that no user facing strings should be in source, which isn't really a practice for us, but they do still specifically allow non-ascii in source.
https://google.github.io/styleguide/cppguide.html#Non-ASCII_Characters

I'm very willing to accept that my preference is old-fashioned and choose a more modern style for our source code if that's the consensus - I mostly just wanted to double check that we weren't introducing a new pattern that would become standard without thinking it through.

@munificent - do you think this is something we should cover in Effective Dart?

jakemac53: Fwiw for me having the actual characters is better, I have no idea what `\u2705` is and would have to look it up. I don't really see any downside to using the actual character and you have the advantage of seeing it visually.

munificent: > @munificent - do you think this is something we should cover in Effective Dart?

Seems pretty nit-picky to me, but if you think it's worth it, I'm persuadable.

devoncarew: I think - for readability of this particular piece of code - I'm going to switch back to having the actual char in the string. It's not causing issues w/ tooling, and makes it much easier to read this bit od code.

pkgs/test/test/runner/github_reporter_test.dart
devoncarew: There's some (incomplete) discussion of the icons here: https://github.com/dart-lang/test/pull/1678#issuecomment-1075677817. I did play around w/ a lot of different icons, including just editing the dom of the 'checks' page to see how things would look.

For this icon, I don't want to use the same icon as for a regular test; I think it's valuable for 6 users test ==> 6 test icons in the output. I'd like the additional test plumbing to have distinct icons from the passed tests. Additionally, the icon should be the same width as the other icons - so the LHS isn't ragged - and shouldn't use a color which users already attach some meaning to (red, yellow, ...).

Other similar unicode chars can be found here: https://unicode-table.com/en/blocks/dingbats/.

pkgs/test_core/lib/src/runner/reporter/github.dart
devoncarew: I'd like to keep these grouped with the other github markup code (`startGroup()`, `error()`, ...) as they're related. No real options other than that; `static` members works for me.

pkgs/test_core/lib/src/util/io.dart
devoncarew: 👍 will do

pkgs/test/test/runner/github_reporter_test.dart
jakemac53: What about ☑? That way it is different but still indicates success (it is a test that can fail). And it isn't a color that indicates a warning. My main complaint about the current choice is the orange color indicates a warning to me.

devoncarew: I've been checking the rendering in-line in the test results, just so we're getting the font that github uses there. What I see for the current 'non-test' group char (⏺) is:

Screen Shot 2022-03-29 at 9 45 09 AM

Which - for me - is a gray color, not an orange color. I do agree that orange (or red, or yellow) would indicate a failure or problem of some kind to the user - something we should avoid unless there's an actual issue.

Here's what the ☑ char looks like in-line in test results (note the different size and style):

Screen Shot 2022-03-29 at 9 40 14 AM

jakemac53: Oh weird... ya its orange on a chromebook :(. I am on my mac now and I see it as a greyish blue. These were both looking at the github UI, so the color is still different based on the platform and not the terminal.

That is pretty unfortunate. @munificent was suggesting using an older check mark symbol that is just the check mark (no background) and then coloring the check mark explicitly?

jakemac53: I also think we can definitely launch this with any of the suggested choices, and revist later.

pkgs/test_core/lib/src/runner/reporter/github.dart
munificent: FWIW, I'm generally comfortable with having non-ASCII characters inside Dart comments and string literals. It's entirely well-specified by the language and there's nothing intrinsically better about Unicode codepoints whose value happens to be less than 128. (I, of course, avoid homoglyphs and other nasty ones.)

pkgs/test/test/runner/github_reporter_test.dart
devoncarew: > that is just the check mark (no background) and then coloring the check mark explicitly?

That would be great if we could do it :) From experimentation, we can't use ansi colors in the summary line of a group here.

> I also think we can definitely launch this with any of the suggested choices, and revist later.

Let's do that then; I don't think there's an idea set of icons unfortunately - I've experimented w/ many. I'll change to the green checkmark, and we can iterate as needed.

PR #1680, opened 2022-03-25
Fix an upcoming recommended lint

natebosch:
https://dart-lang.github.io/linter/lints/null_check_on_nullable_type_parameter.html

Eagerly fixing before we start picking up the new lint package.


natebosch:
Another option would be to switch from `!` to `as T`. It doesn't matter much either way - this utility only lives here because it's annoying to move without copying it and it's only used in 2 places.

jakemac53 (approved):

PR #1696, opened 2022-05-04
fix runner test for multiple failures

jakemac53:
Test was previously order dependent, it no longer is


kevmoo (approved):

jakemac53:
Investigating why this didn't fix things on github actions... it does pass for me locally with this change.

jakemac53:
Ok, interestingly it looks like this test only fails on Dev, I can repro locally as well with the latest dev release

jakemac53:
For now I just updated this to convert the stream to a list, and use `containsAll`. But something weird is going on here for sure between stable and dev.

PR #1701, opened 2022-05-16
Fix updated underscore lint

kevmoo:


jakemac53 (approved):
Looks like we need to up the versions to `-dev` as well (no changelog notes needed but should have an entry).

natebosch (approved):

PR #1704, opened 2022-05-23
Restrict TestOn to library level

natebosch:
Closes https://github.com/dart-lang/sdk/issues/49071

Add a `Target` with the `TargetKind.library` so that the analyzer
understands that the annotation is logically applied to the library,
even when it is syntactically applied on an import. This fixes behavior
when sorting imports.
The same configuration was applied to other imports in #1345 and this
looks like it was an accidental omission.

### DIFF
--- a/pkgs/test/test/runner/retry_test.dart
+++ b/pkgs/test/test/runner/retry_test.dart
@@ -2,8 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:test/test.dart';
 @TestOn('vm')
+
+import 'package:test/test.dart';
 import 'package:test_descriptor/test_descriptor.dart' as d;
 
 import '../io.dart';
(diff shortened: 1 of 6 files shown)


pkgs/test/test/runner/retry_test.dart
```diff
@@ -2,8 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:test/test.dart';
```
natebosch, line 5: 😳

pkgs/test/test/runner/retry_test.dart
```diff
@@ -2,8 +2,9 @@
...
 @TestOn('vm')
+
```
lrhn, line 6: You *could* add a `library _;` here. Just to show good sportsmanship :)

lrhn (approved):

pkgs/test/test/runner/retry_test.dart
natebosch: Might be a good idea, but we should probably do the whole repo if we do it.

PR #1711, opened 2022-06-07
Add the ability to load from package URLs

### LINKED ISSUE #1710: spawnHybridUri should accept package Uris
e.g. `Uri(scheme: 'package', path: 'mypackage/src/myfile.dart')`

brianquinlan:
Fixes #1710

### DIFF
--- a/pkgs/test/lib/src/_package_hybrid_test_helper.dart
+++ b/pkgs/test/lib/src/_package_hybrid_test_helper.dart
@@ -1,0 +1,16 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// A private file used for testing `spawnHybridUri` through a package Uri i.e.
+/// package:test/src/_package_hybrid_test_helper.dart`
+
+import "package:stream_channel/stream_channel.dart";
+
+void hybridMain(StreamChannel channel) {
+  channel.sink
+    ..add(1)
+    ..add(2)
+    ..add(3)
+    ..close();
+}
--- a/pkgs/test/lib/src/_package_hybrid_test_helper.dart
+++ b/pkgs/test/lib/src/_package_hybrid_test_helper.dart
@@ -5,7 +5,7 @@
 /// A private file used for testing `spawnHybridUri` through a package Uri i.e.
 /// package:test/src/_package_hybrid_test_helper.dart`
 
-import "package:stream_channel/stream_channel.dart";
+import 'package:stream_channel/stream_channel.dart';
 
 void hybridMain(StreamChannel channel) {
   channel.sink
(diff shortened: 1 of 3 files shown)


pkgs/test/lib/src/_package_hybrid_test_helper.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
```
jakemac53, line 1: We don't want testing code in our `lib` directory, even if under `lib/_src`.

If this is shared across packages we can either create multiple copies or create a shared testing package which is not published in the mono_repo (if one does not already exist).

jakemac53: Although for this one it looks like it could be created as a part of the package being tested? (using the test descriptor apis)

pkgs/test_core/lib/src/runner/spawn_hybrid.dart
```diff
@@ -86,7 +86,7 @@ StreamChannel spawnHybridUri(String url, Object? message, Suite suite) {
 /// scheme.
```
jakemac53, line 86: We also need to update this documentation

jakemac53:
Generally, I think the intention is that we don't actually convert these URIs at all.

What we end up doing with the URI as far as I can tell is just putting it into an import statement. It is intentional that package uris (or any with a scheme) are left alone today, because that _should_ work just fine. Although I don't see them being tested today.

What is the error you see without these changes?

pkgs/test_core/lib/src/runner/spawn_hybrid.dart
brianquinlan: The documentation already says:

"If it's a package: URL, it will be resolved using the current package's dependency constellation."

I can add more detail if that would be helpful but the intent of this PR was to make that statement true ;-)

brianquinlan:
> Generally, I think the intention is that we don't actually convert these URIs at all.
> 
> What we end up doing with the URI as far as I can tell is just putting it into an import statement. It is intentional that package uris (or any with a scheme) are left alone today, because that _should_ work just fine. Although I don't see them being tested today.
> 
> What is the error you see without these changes?

[`readUri`](https://github.com/dart-lang/test/blob/52ecae96726ad1b4ef25ad3fd14860361900948a/pkgs/test_core/lib/src/runner/spawn_hybrid.dart#L153) supports a whitelist of schemes.

So I see an `ArgumentError` if I use the package scheme.

pkgs/test/lib/src/_package_hybrid_test_helper.dart
jakemac53: You could add a single new package in `integration_tests`, which contains both the test and the code to spawn (under lib of that package, it isn't a published package its just for testing purposes).

I would probably just go that route as it would be easiest. You could move this entire test into there, which might make it easier to understand as well (and stop using test descriptor in this package).

natebosch:
> because that _should_ work just fine. Although I don't see them being tested today.

It looks like we broke it with the NNBD migration. Its the code that looks for what language version marker to put in the file. We now want to eagerly read the file.

jakemac53:
Given it was a fair bit of work to move the test I went ahead and did that and created https://github.com/dart-lang/test/pull/1712 which also incorporates these changes. Closing this as it is super-ceded by that one.

Thanks for the fix and bug report!

Issue #829, opened 2018-05-08
Add a quieter reporter that only prints information for failures


natebosch:
In places where the "compact" reporter doesn't work it would be nice to have a reporter that only prints when a test fails and then a summary at the end.

jakemac53:
Note that travis requires some sort of output every 10 minutes, so we would probably need some sort of heartbeat logger that periodically outputs a summary of the status. This would probably be useful anyways and would still be much less noisy than the existing reporters.

jakemac53:
We could rip out the existing heartbeat logging logic from package:build_runner into a package (or add it to the io or cli_utils package). Or we could just re-implement it.

grouma:
I imagine this would be the reporter that is shipped with the `api` package.

dnfield:
This would be nice for Flutter.  The compact reporter doesn't quite manage to print a single line for successes there, and you end up with thousands of lines to sift through.

dnfield:
See #982 which isn't quite all of what this asks for but improves the compact reporter wrt my issue.

natebosch:
> Note that travis requires some sort of output every 10 minutes

I don't think github actions has the same requirement, and I think we'l keep using the github reporter for CI. I think we could get by with a strictly failure output initially and wait for a specific request for heartbeat output before implementing it.

PR #1720, opened 2022-06-15
Show list of failed tests in the final summary

gmpassos:
This is a small change that adds the listing of the failed tests (path + name) in the final summary of the tests.

This is very helpful, since when we have a failed test we need to scroll back many tests outputs to find the one that have failed.

With this extra information in the final summary we have the exact information that we need when a test fails.

### DIFF
--- a/pkgs/test_core/lib/src/runner/reporter/compact.dart
+++ b/pkgs/test_core/lib/src/runner/reporter/compact.dart
@@ -298,6 +298,12 @@ class CompactReporter implements Reporter {
             suffix: ' - did not complete $_bold$_red[E]$_noColor');
         _sink.writeln('');
       }
+      for (var failedTest in _engine.failed) {
+        _progressLine(_description(failedTest),
+            truncate: false,
+            suffix: ' - $_bold${_red}failed$_noColor $_bold$_red[E]$_noColor');
+        _sink.writeln('');
+      }
       _progressLine('Some tests failed.', color: _red);
       _sink.writeln('');
     } else if (_engine.passed.isEmpty) {


jakemac53:
In general, for the compact reporter, we expect the output to really only contain test failures plus a final summary line already, so I am not sure how much this helps there?

I think for the expanded reporter this might make more sense, as it has a lot more noise in the output.

cc @natebosch

natebosch:
I agree that this is not very useful in the compact reporter.

I think it would be OK to add it at the end of the expanded reporter. Or we could make it a separate reporter. https://github.com/dart-lang/test/issues/829

gmpassos:
This is a very useful feature. Some times it takes a significant amount of time to find the failed tests in the output. And when a test fails, this is the main information that you need.

For me the output is still compact, since it's only showing extra information for failed tests (one line for each one).

jakemac53:
@gmpassos is the issue you are having that the test failures themselves are emitting a lot of output? So then you have to scroll up past a lot of stack traces/etc in order to find all the failed tests?

gmpassos:
Is not about a lot of stack traces, but a lot of outputs in the console.

Depending of the library it produces outputs in the console as a normal use case.

jakemac53:
You could run your tests inside of a zone that intercepts all print/stdout/stderr calls with a call to `printOnFailure` or similar? That would eliminate this type of output for passing tests, which seems to be the larger issue.

gmpassos:
I (and any dev) just want to run "dart test" for any kind of test for any kind of library, for any kind of package using the same paradigm to code it.

So, I have pushed a simple pull request to add a very helpful feature for any Dart developer.

This information shouldn't be rocket science to have access to it. Should be only "dart test" since when you have a failed test, you need to know which test you need to fix (ASAP). Developer time is gold, BTW.

kevmoo:
> You could run your tests inside of a zone that intercepts all print/stdout/stderr calls with a call to `printOnFailure` or similar? That would eliminate this type of output for passing tests, which seems to be the larger issue.

@jakemac53 – better to wrap the...JSON reporter things, right? That's what IDEs do. All the info is there in a highly structured format.

Honestly, I had a chat w/ previous engineering ownership about exporting the logging types and supported decode so a user could create a type `Stream` API to support better tooling

natebosch:
The compact reporter is definitely not where I'd want to put this, at least not by default. I don't want to increase the noise for tests that _don't_ have spammy output in order to make it easier for tests that do.

I could imagine a few different approaches but I'm not sure I'll have the bandwidth to push on them.
- Add support for passing configuration to reporters (I don't think there is any configuration today, is there?) and make this an option on both the compact and expanded reporter. Off by default.
- Add support for running multiple reporters to stdout, and implement #829
- Add a separate non-reporter feature like `--print-failure-summary` which can output failures independently of the reporter that was used (It would be invalid to use this flag with the json reporter).

gmpassos:
I vote to be able to pass parameters/configuration to the reporter, specially for the summary output.

Regards.

mosuem:
Closing this as it is stale. Feel free to reopen in case you still want to land it!

PR #1723, opened 2022-06-22
Support the latest pkg:matcher

kevmoo:
Just increased the upper bound because the last matcher release changed
no functionality


pkgs/test_api/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.4.11
```
jakemac53, line 1: If we are publishing this we will also need to publish the other ones, might as well prep them in the same CL.

PR #1724, opened 2022-06-22
check fun

kevmoo:
- checks: String add length, is(Not)Empty, startsWith, endsWith
- checks: Add some TODOs around handling String in `literal` helper

### DIFF
--- a/pkgs/checks/lib/src/extensions/core.dart
+++ b/pkgs/checks/lib/src/extensions/core.dart
@@ -118,4 +118,46 @@ extension StringChecks on Check<String> {
           actual: literal(actual), which: ['Does not contain $pattern']);
     });
   }
+
+  Check<int> get length => has((m) => m.length, 'length');
+
+  void isEmpty() {
+    context.expect(() => const ['is empty'], (actual) {
+      if (actual.isEmpty) return null;
+      return Rejection(actual: literal(actual), which: ['is not empty']);
+    });
+  }
+
+  void isNotEmpty() {
+    context.expect(() => const ['is not empty'], (actual) {
+      if (actual.isNotEmpty) return null;
+      return Rejection(actual: literal(actual), which: ['is empty']);
+    });
+  }
+
+  void startsWith(String other) {
+    context.expect(
+      () => ['starts with ${literal(other)}'],
+      (actual) {
+        if (actual.startsWith(other)) return null;
+        return Rejection(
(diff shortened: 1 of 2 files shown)


kevmoo:
Will land as seperate commits

pkgs/checks/lib/src/extensions/core.dart
```diff
@@ -118,4 +118,34 @@ extension StringChecks on Check<String> {
...
+
+  void isNotEmpty() {
+    context.expect(() => const ['is not empty'], (actual) {
+      if (actual.isNotEmpty) return null;
+      return Rejection(actual: literal(actual), which: ['is empty']);
+    });
+  }
+
+  void startsWith(String other) {
+    context.expect(() => const ['starts with'], (actual) {
```
natebosch, line 139:
```suggestion
    context.expect(() => const ['starts with ${literal(other)}'], (actual) {
```

pkgs/checks/lib/src/extensions/core.dart
```diff
@@ -118,4 +118,34 @@ extension StringChecks on Check<String> {
...
+      if (actual.startsWith(other)) return null;
+      return Rejection(actual: literal(actual), which: ['does not start with']);
```
natebosch, line 141:
```suggestion
      return Rejection(actual: literal(actual), which: ['does not start with ${literal(other)}']);
```

pkgs/checks/lib/src/extensions/core.dart
```diff
@@ -118,4 +118,34 @@ extension StringChecks on Check<String> {
...
+    });
+  }
+
+  void endsWith(String other) {
+    context.expect(() => const ['ends with'], (actual) {
```
natebosch, line 146:
```suggestion
    context.expect(() => const ['ends with ${literal(other)}'], (actual) {
```

pkgs/checks/lib/src/extensions/core.dart
```diff
@@ -118,4 +118,34 @@ extension StringChecks on Check<String> {
...
+      if (actual.endsWith(other)) return null;
+      return Rejection(actual: literal(actual), which: ['does not end with']);
```
natebosch, line 148:
```suggestion
      return Rejection(actual: literal(actual), which: ['does not end with ${literal(other)}']);
```

natebosch (approved):

PR #1727, opened 2022-06-24
checks fun

kevmoo:
- Cleanup mono_repo yaml files
- Update to latest mono_repo
- check: 100% code coverage of core extensions

### DIFF
--- a/pkgs/checks/test/extensions/core_test.dart
+++ b/pkgs/checks/test/extensions/core_test.dart
@@ -1,24 +1,15 @@
 import 'package:checks/checks.dart';
 import 'package:checks/context.dart';
-import 'package:test/test.dart';
-
-Matcher isARejection({Object? which, Object? actual}) {
-  var rejection = isA<Rejection>().having((p0) => p0.which, 'which', which);
-
-  rejection = rejection.having((p0) => p0.actual, 'actual', actual);
-
-  return rejection;
-}
+import 'package:test/scaffolding.dart';
 
 void main() {
   group('TypeChecks', () {
     test('isA', () {
       checkThat(1).isA<int>();
 
-      expect(
+      checkThat(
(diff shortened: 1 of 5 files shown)


kevmoo:
@natebosch – you'll probably say you want separate PRs here...

pkgs/checks/test/extensions/core_test.dart
```diff
@@ -0,0 +1,167 @@
...
+
+  return rejection;
+}
+
+void main() {
+  group('TypeChecks', () {
+    test('isA', () {
+      checkThat(1).isA<int>();
+
+      expect(
```
natebosch, line 18: We should not use `expect` or `Matcher` in these tests.

kevmoo: To bootstrap it seemed a good idea – so we're not leaning on the thing we're testing.

natebosch: I disagree, that sounds like building a bunch of debt.

We had no problem using `expect` to test `expect`. I don't see what we'll have any difficulty here.

kevmoo: fixed!

pkgs/checks/test/extensions/core_test.dart
```diff
@@ -0,0 +1,159 @@
+import 'package:checks/checks.dart';
```
natebosch, line 1: [nit] copyright

natebosch (approved):

pkgs/checks/test/extensions/core_test.dart
kevmoo: Uh...you don't have them anywhere else in this package. Follow-up with a one-off to add em?

PR #1731, opened 2022-06-27
Add copyright headers

natebosch:


kevmoo:
CC @devoncarew – thinking of tooling here. If we're supposed to have headers, we should have trivial CI support to check these!!

kevmoo (approved):

jakemac53 (approved):

PR #1733, opened 2022-06-27
more check checks

kevmoo:
- checks: test MathChecks
- checks: test FunctionChecks
- checks: improve output of toStringEquals test extension


kevmoo:
Will land as three commits. All of the "easy" ones are done now 😁

natebosch (approved):

Issue #1734, opened 2022-06-28
Print a usable reproduction command of failed tests.


sigurdm:
Currently a failing test prints a line like so:
```
08:28 +54 ~9 -1: test/integration.shard/command_output_test.dart: flutter run --machine uses AppRunLogger [E] 
```

Running that single test again from the command-line is kind of cumbersome. One has to

* write `flutter test` or `dart test`
* copy-paste the test filename
* write something like `-n ""<backarrow>`
* copy paste the test name.

It would be awesome if the whole command could be printed. Something like:
```
08:28 +54 ~9 -1: test/integration.shard/command_output_test.dart: flutter run --machine uses AppRunLogger [E] 
To run this test again: `flutter test test/integration.shard/command_output_test.dart --plain-name="flutter run --machine uses AppRunLogger"`
```
With escaping of tricky characters in the name string.

sigurdm:
I would personally use both of these features.

Sometimes you want to drill down into a single failed test.
Sometimes you want to see if you managed to fix all the last failures

jakemac53:
This one is also significantly easier to implement I think

PR #1736, opened 2022-06-28
add a line to re-run the failed test after each failure in the compact reporter

jakemac53:
Closes https://github.com/dart-lang/test/issues/1734

Example output:

Screen Shot 2022-06-28 at 11 00 32 AM

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Make the labels for test loading more readable in the compact and expanded
   reporters, use gray instead of black.
+* Print a command to re-run the failed test after each failure in the compact
+  reporter.
 
 ## 1.21.3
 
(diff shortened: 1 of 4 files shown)


jakemac53:
For now I did this only in the compact reporter, because it only ever runs on the VM (and I have access to `Platform.executable`). The expanded reporter runs on the web etc, and it isn't clear how to replicate the same functionality.

There is also a bit of a question I think in terms of internal, but I believe we always use a different reporter so they won't ever see it.

jakemac53:
I also wonder if I shouldn't move the command down to the next line, which would make it easier to copy/paste?

natebosch (approved):

Hixie:
Is there a way to disable this message? It shows up for `flutter test` test failures but the command it gives doesn't work to actually run the test again.

PR #1743, opened 2022-07-21
Bump min SDK to 2.18.0-165.1.beta.

joshualitt:


joshualitt:
@jakemac53 ptal, I'm also happy to pick `2.18.0-122.0.dev` if we would prefer the exact dev version over a beta release.

jakemac53:
Quite honestly, I have no idea what the point of the beta channel is, or what the advantages/disadvantages of pinning to that would be over a dev release?

joshualitt:
I don't really know either TBH.

According to [this](https://dart.dev/get-dart/archive#beta-channel), `beta` is a preview release, so more stable than `dev`, but less stable than `stable`.

If 'stability' is important, than maybe `beta` makes sense, but `dev` will let us get closer to the actual CL. I imagine the dev release has more potential for bugs though? Neither are recommended for production use.

jakemac53:
This isn't really about stability for us, its just ensuring whatever SDK people are on, it contains whatever apis we are using. It is on them to choose which channel they are on.

I don't really know how the semantic versioning works between dev/beta releases tbh. But we probably want to just be as close to the actual release as possible so I would choose `2.18.0-122.0.dev`. We have historically also always used the dev releases.

PR #1751, opened 2022-08-16
Associate printOnFailure messages with the failing test

greglittlefield-wf:
# Motivation

I recently discovered `printOnFailure`, and I love it! However, I noticed that the error messages it prints don't seem to be associated with the failing test, and show up under "loading \<test filename\>..." headers instead. For example:

```dart

void main() {
  test('1', () {
    printOnFailure('printOnFailure content 1');
    fail('failure');
  });

  test('2', () {
    printOnFailure('printOnFailure content 2');
    fail('failure');
  });
}
```

The output looks like:
```
$ dart test test/foo_test.dart
00:00 +0 -1: 1 [E]
  failure
  package:test_api        fail
  test/foo_test.dart 6:5  main.<fn>

00:00 +0 -1: loading test/foo_test.dart
printOnFailure content 1
00:00 +0 -2: 2 [E]
  failure
  package:test_api         fail
  test/foo_test.dart 11:5  main.<fn>

00:00 +0 -2: loading test/foo_test.dart
printOnFailure content 2
00:01 +0 -2: Some tests failed.
```

This isn't great, but it becomes more of an issue when using tools that consume the JSON output. For example, in the JSON output for the same test, noting that the `"testID"` doesn't match the test:
```
$ dart test test/foo_test.dart -r json
…
{"test":{"id":1,"name":"loading test/foo_test.dart","suiteID":0,"groupIDs":[],"metadata":{"skip":false,"skipReason":null},"line":null,"column":null,"url":null},"type":"testStart","time":5}
…
{"test":{"id":3,"name":"1","suiteID":0,"groupIDs":[2],"metadata":{"skip":false,"skipReason":null},"line":4,"column":3,"url":"file:///Users/greglittlefield/workspaces/over_react6/tools/analyzer_plugin/test/foo_test.dart"},"type":"testStart","time":784}
{"testID":3,"error":"failure","stackTrace":"package:test_api        fail\ntest/foo_test.dart 6:5  main.<fn>\n","isFailure":true,"type":"error","time":846}
{"testID":1,"messageType":"print","message":"printOnFailure content 1","type":"print","time":849}
{"testID":3,"result":"failure","skipped":false,"hidden":false,"type":"testDone","time":849}
…
```

When running tests from JetBrains IDEs, it causes that output to show up separate from the test that called it, and aggregated with output from other failing tests, which makes it harder to use if there's more than one failing test:

# Solution
- Update code processing `printOnFailure` messages to use `_print` instead of `print`, so that those strings get processed like other `print` statements in user tests.
- Update tests to expect these to come through as messages.

These changes seem to get the messages associated correctly!

```
analyzer_plugin $ dart test test/foo_test.dart
00:01 +0 -1: 1 [E]
  failure
  package:test_api        fail
  test/foo_test.dart 6:5  main.<fn>

printOnFailure content 1
00:01 +0 -2: 2 [E]
  failure
  package:test_api         fail
  test/foo_test.dart 11:5  main.<fn>

printOnFailure content 2
00:01 +0 -2: Some tests failed.
```

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## 1.21.5-dev
 
-* Fix `printOnFailure` output to forward through json output instead of being
-  printed directly.
+* Fix `printOnFailure` output to be associated with the correct test.
 * Migrate all dom interactions to static interop.
 
 ## 1.21.4
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 1.21.5-dev
 
+* Fix `printOnFailure` output to forward through json output instead of being
+  printed directly.
 * Migrate all dom interactions to static interop.
 
 ## 1.21.4
--- a/pkgs/test_api/CHANGELOG.md
+++ b/pkgs/test_api/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## 0.4.13-dev
 
-* Fix `printOnFailure` output to forward through json output instead of being
-  printed directly.
+* Fix `printOnFailure` output to be associated with the correct test.
 
 ## 0.4.12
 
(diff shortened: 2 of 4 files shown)


natebosch (approved):
Thank you for finding this!

greglittlefield-wf:
You're welcome! Thank you for reviewing!!

PR #1754, opened 2022-08-19
Widen frontend_server_client constraint

sigurdm:
To unblock: https://github.com/dart-lang/pub/pull/3514

### DIFF
--- a/pkgs/test_core/pubspec.yaml
+++ b/pkgs/test_core/pubspec.yaml
@@ -10,7 +10,7 @@ dependencies:
   analyzer: '>=3.3.0 <5.0.0'
   async: ^2.5.0
   args: ^2.0.0
-  boolean_selector: ^2.1.0
+  boolean_selector: '>=2.1.0 <4.0.0'
   collection: ^1.15.0
   coverage: ^1.0.0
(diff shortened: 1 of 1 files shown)


pkgs/test_core/pubspec.yaml
```diff
@@ -10,7 +10,7 @@ dependencies:
   analyzer: '>=3.3.0 <5.0.0'
   async: ^2.5.0
   args: ^2.0.0
-  boolean_selector: ^2.1.0
+  boolean_selector: '>=2.1.0 <4.0.0'
```
jakemac53, line 13: It looks like 2.1.0 is the latest version here, (also this isn't frontend_server_client?)

jakemac53:
I think this can probably be closed, I did already widen the constraint, although we are figuring out if/how we are going to publish given we have a constraint on a pre-release SDK currently.

natebosch:
A version of `test` is published with a widened constraint.

pkgs/test_core/pubspec.yaml
sigurdm: 🤦 thanks

PR #1757, opened 2022-08-28
Update coverage usage example

f3ath:
- Use `dart pub` instead of `pub`
- Point `--packages` to `.dart_tool/package_config.json` as suggested by the `coverage` package.


jakemac53 (commented):
Looks good but we do also need to update the pubspec version to indicate we are now on the next dev release, and add a changelog entry for that version as well.

The new version should be `1.21.6-dev`.

f3ath:
@jakemac53 updated. Please take another look.

natebosch (approved):

Issue #1776, opened 2022-10-19
Add support for specifying a compiler separately from a platform


natebosch:
As we look to add support for compiling to wasm and running the browser, we should add a `--compiler` flag that allows specifying one or more compilers to use in combination with the specified platforms. Using a separate flag will allow using the wasm compiler with any browser that supports the required wasm features.

We'll add a `wasm` term similar to the `js` term, for use in `@TestOn` configuration.

When no `--compiler` flag is passed, each platform will have a default compiler. When a `--compiler` flag is passed, only the valid combinations of compiler &#x2715; platform will run. So a combination like `--compiler dart2js --platform vm` will run no tests, since the VM platform does not support dart2js as a compiler, but `--compiler jit,dart2js,wasm --platform vm,chrome` would run both dart2js and wasm tests on chrome, as well as from-source tests on the VM.

In the future, we should be able to add support for compiling with DDC (monolithic compile with the frontend server, it's unlikely we'd implement modular compiles), and may consider adding support for AOT compiles on the VM platform.

natebosch:
This require a change in the interface between the test runner, and each `Platform` implementation. In particular, platforms will need to 
- support an argument specifying which compilers to use
- know which compilers are valid for the platform
- have a default compiler when no compilers config is specified

cc @christopherfujino - heads up that this will impact integration with flutter_test, likely sometime in Q1. Does the flutter test setup have any notion of running with different compilation approaches on the same platform today?

jakemac53:
I have been looking into this a bit, there are a few questions we will want to resolve that I have ran into:

- In precompiled mode today we allow VM tests to not actually be precompiled to kernel, and in that case we run them from source. This differs from normal tests which we always compile to kernel by default.
  - I think we should change this to compile them to kernel on the fly, if not precompiled (assuming we are in `--compiler kernel` mode)?
  - Note that build_vm_compilers is no longer supported, so I think we need a fallback. Even though it is a bit weird .
  - We might need to update some stuff internally to pass `--compiler none`, but that should be doable.
  - We could alternatively not do this and require users to pass `--compiler none` in precompiled mode, if they aren't precompiling their vm tests to kernel, but that is fairly breaking I think.
- I am realizing that the behavior of skipping tests for a given platform if no compiler for that platform is specified is potentially a bit weird/cumbersome.
  - For instance, if I want to use wasm for web tests, but also run vm tests, I will have to explicitly specify `--compiler kernel` for the vm tests.
[...]

PR #1777, opened 2022-10-20
Experimental wasm support

jakemac53:
- Adds the experimental platform `experimental-chrome-wasm`. This requires chrome beta and a very recent SDK to run.
  - This platform will be deleted eventually, and replaced with a [--compiler](https://github.com/dart-lang/test/issues/1776) flag.
- Adds `wasm` to the boolean selector variables, and an `isWasm` field to `Runtime` (defaults to false, so should be non-breaking).

Note that the integration test for this will initially be failing, as there is one more change needed from the SDK, but once that lands we can update this to point at the new location for the wasm runtime, update the min SDK, re-run the tests, and land.


jakemac53:
cc @christopherfujino note that this isn't the final API we want here, but if you want to start thinking about trying to get flutter tests working with dart2wasm, then this should provide the basic building blocks required. Just know that the api is highly unstable and subject to change.

jakemac53:
@natebosch not sure how you feel about this runtime showing up in the help text - we could add some logic in to hide any runtime with `experimental` in the name, or add a boolean `hidden` field to runtime, in order to hide it. But also I don't think it hurts to have it visible, given the name clearly indicates the status of it.

pkgs/test/lib/src/runner/wasm/platform.dart
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    var browser = platform.runtime;
+    assert(suiteConfig.runtimes.contains(browser.identifier));
+
+    if (!browser.isBrowser) {
+      throw ArgumentError('$browser is not a browser.');
+    }
+
+    // TODO: Support custom html?
```
joshualitt, line 216: Remove?

pkgs/test_core/lib/src/runner/compiler_pool.dart
```diff
@@ -1,142 +1,57 @@
-// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
```
joshualitt, line 1: It's a bit ambiguous from github, but if this isn't a new file we may not want to update the copyright.

joshualitt (approved):

pkgs/test_core/lib/src/runner/compiler_pool.dart
jakemac53: Oh right, this one isn't new.

pkgs/test/lib/src/runner/wasm/platform.dart
```diff
@@ -0,0 +1,361 @@
...
+      var baseCompiledPath =
+          p.join(dir, '${p.basename(dartPath)}.browser_test.dart');
+      var baseUrl =
+          '${p.toUri(p.relative(dartPath, from: _root)).path}.browser_test.dart';
+      var wasmUrl = '$baseUrl.wasm';
+      var jsRuntimeWrapperUrl = '$baseUrl.js';
+      var jsRuntimeUrl = p.join(p.dirname(dartPath), 'dart2wasm_runtime.mjs');
+      var htmlUrl = '$baseUrl.html';
+
+      // TODO: This may need to be specialized, or it may just work. Not sure.
```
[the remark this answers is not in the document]
jakemac53: removed

pkgs/test/lib/src/runner/wasm/platform.dart
jakemac53: I think this one is valid - we likely will want to support custom  html at some point

natebosch (approved):

natebosch:
> @natebosch not sure how you feel about this runtime showing up in the help text - we could add some logic in to hide any runtime with `experimental` in the name, or add a boolean `hidden` field to runtime, in order to hide it. But also I don't think it hurts to have it visible, given the name clearly indicates the status of it.

I'd be OK with either filtering out names that contain "experimental", or with leaving it in. @joshualitt - is there any concern about the feature "leaking", or having folks notice it and try it out while we are still getting it nailed down in the SDK?

joshualitt:
> > @natebosch not sure how you feel about this runtime showing up in the help text - we could add some logic in to hide any runtime with `experimental` in the name, or add a boolean `hidden` field to runtime, in order to hide it. But also I don't think it hurts to have it visible, given the name clearly indicates the status of it.
> 
> I'd be OK with either filtering out names that contain "experimental", or with leaving it in. @joshualitt - is there any concern about the feature "leaking", or having folks notice it and try it out while we are still getting it nailed down in the SDK?

Great question. I think this should be just fine because the flag is self descriptive and has to be explicitly passed. There is no danger here of someone 'accidentally' switching to Wasm and being very confused.

As to the broader question, as I understand things until we add support for Dart2Wasm to the CLI, we're not officially advertising Dart2Wasm as supported. It's true that adding support to package:test does lower the bar a bit to entry, but only for writing tests, which seems pretty harmless to me. Specifically, there is no danger here of anyone _shipping_ something with Dart2Wasm before we are ready because of this work, so I think it should be completely safe.

pkgs/test_core/lib/src/runner/wasm_compiler_pool.dart
```diff
@@ -0,0 +1,96 @@
...
+      var dartBinPath = Platform.resolvedExecutable;
+      var sdkRoot = p.join(p.dirname(dartBinPath), '../');
+      var jsRuntimePath = p.join(sdkRoot, 'bin', 'dart2wasm_runtime.mjs');
+      File(jsRuntimePath).copy(outJSPath);
+      var platformDill =
+          p.join(sdkRoot, 'lib', '_internal', 'dart2wasm_platform.dill');
+      var dartPrecompiledRuntimePath =
+          p.join(sdkRoot, 'bin', 'dart_precompiled_runtime');
+      var dart2wasmSnapshotPath =
+          p.join(sdkRoot, 'bin/snapshots', 'dart2wasm.snapshot');
```
joshualitt, line 45: I think this needs to be `dart2wasm_product.snapshot`

pkgs/test_core/lib/src/runner/wasm_compiler_pool.dart
```diff
@@ -0,0 +1,96 @@
...
+      var outWasmPath = '$path.wasm';
+      var outJSPath = p.join(p.dirname(path), 'dart2wasm_runtime.mjs');
+      var dartBinPath = Platform.resolvedExecutable;
+      var sdkRoot = p.join(p.dirname(dartBinPath), '../');
+      var jsRuntimePath = p.join(sdkRoot, 'bin', 'dart2wasm_runtime.mjs');
+      File(jsRuntimePath).copy(outJSPath);
+      var platformDill =
+          p.join(sdkRoot, 'lib', '_internal', 'dart2wasm_platform.dill');
+      var dartPrecompiledRuntimePath =
+          p.join(sdkRoot, 'bin', 'dart_precompiled_runtime');
```
joshualitt, line 43: this should just be `dartaotruntime` now.

PR #1778, opened 2022-10-26
Bump to the latest matcher and prepare to publish

natebosch:
Bump test_core SDK constraint to 2.18 since `test_api` and `test`
already have that constraint and it would not be possible to resolve
this `test_core` in a useful way on an older SDK.


kevmoo (approved):
LGTM

Issue #1756, opened 2022-08-25
GitHub reporter does not print failures happening after a test completes


simolus3:
Consider the following test file as a simple repro:

```dart
import 'dart:async';

import 'package:test/test.dart';

void main() {
  test('fail after completion', () {
    Future(() {
      Zone.current.handleUncaughtError('foo', StackTrace.current);
    });
  });

  test('second test so that the first failure is reported', () {});
}
```

Running this test with `--reporter compact` or `--reporter expanded` prints this error message at some point:

```
00:00 +0 -1: fail after completion [E]
  foo
  test/foo_test.dart 8:58  main.<fn>.<fn>
  
  This test failed after it had already completed. Make sure to use [expectAsync]
  or the [completes] matcher when testing async code.
  test/foo_test.dart 8:58  main.<fn>.<fn>
```

When the same test is run with `--reporter github`, it is not obvious which test is responsible for the failure:

```dart

::group::✅ fail after completion
::endgroup::
::group::✅ second test so that the first failure is reported
::endgroup::

::error::1 test passed, 1 failed.
```

PR #1779, opened 2022-10-31
fix an issue with the github reporter

devoncarew:
- fix an issue with the github reporter
- fix https://github.com/dart-lang/test/issues/1756

This addresses an issue with the github reporter. With tests that fail asynchronously after they've completed, the main github reporter UI would show all the tests as green, but would report 'xx' failed tests at the end. So, users would know something failed, but wouldn't know which test.

This changes to re-report tests which fail asynchronously; they'll show as a passing line, but later show as an ❌ failure line (they'll show two failures since the framework also emits a message about how to test async code).

I'm not sure where to put the changelog entry for this / which package to bump.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 1.22.0-dev
 
+* Fix an issue with the github reporter where tests that fail asynchronously
+  after they've completed would show up as succeeded tests.
 * Add the `experimental-chrome-wasm` platform. This is very unstable and will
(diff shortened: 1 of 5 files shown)


jakemac53 (approved):

pkgs/test/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 1.22.0-dev
 
+* Fix an issue with the github reporter where tests that fail asynchronously
```
devoncarew, line 3: 👍 thanks

PR #1787, opened 2022-11-17
Starting on async tests

kevmoo:


pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -80,7 +80,7 @@ extension StreamChecks<T> on Check<StreamQueue<T>> {
   /// Expects that the `Stream` emits any number of events before emitting an
   /// event that satisfies [condition].
   ///
-  /// Returns a `Future` that completes after the stream has emitted an even
+  /// Returns a `Future` that completes after the stream has emitted an event
```
kevmoo, line 83: oops!

kevmoo:
Inspiration to figure out testing failures in async code...could NOT figure it out!

pkgs/checks/test/extensions/async_test.dart
```diff
@@ -0,0 +1,60 @@
...
+import 'package:async/async.dart';
+import 'package:checks/checks.dart';
+import 'package:test/scaffolding.dart';
+
+void main() {
+  group('FutureChecks', () {
+    test('completes', () async {
+      (await checkThat(_futureSuccess()).completes()).equals(42);
+
+      // TODO: validate the failure case
```
natebosch, line 14: Hmm, we might need an async version of `softCheck` for this.

natebosch (approved):

pkgs/checks/test/extensions/async_test.dart
natebosch: Or perhaps we do something like

```dart
test('completes - fails for an error future', () async {
  Future<void> failingTest() async {
    await checkThat(_futureFail()).completes();
  }
  await checkThat(failingTest()).throw();
}
```

kevmoo: All we get is a `TestFailure` exception – can't validate the guts, which I think we want!

natebosch: Yeah, having the `Rejection` would be nicer. I thought that async `softCheck` would be challenging, but it turns out not to be.

https://github.com/dart-lang/test/pull/1789

PR #1789, opened 2022-11-18
Add support for async soft checks

natebosch:
Enables better testing for failing asynchronous expectations. Keep behind a `lib/src/` import for now, until we have a use case to expose it more generally.

Fix fallback for continuing in an async expectation when the expectation failed, but the failure did not throw an exception.


kevmoo (approved):

kevmoo:
wasm flakes it looks like...

jakemac53:
I have a separate issue/pr out for the wasm flakes, should be resolved soon

PR #1797, opened 2022-11-24
Fix missing label and reason after isNotNull

natebosch:
When creating a context which is an alias of another, at the same level
in the context tree, the parent was copied. When this happened on the
root in the tree, the alias would also appear to be a root, but it would
be missing some other important information from the root, like the root
label and reason. Separately, the reason which should have always been
read from the root, was not.

- Rename the first constructor from `_` to `_root` to more clearly
  indicate where should be used.
- Add a doc that only the root constructor should have no parent.
- Treat the aliased context as the parent, instead of copying the
  originals parent. This only matters when the aliased context is the
  root, but it does not hurt for the parent references to have an extra
  hop in other cases.
- Read the reason from the root instead of on the check that failed.
- Add some initial tests for the message in a thrown `TestFailure`. Add
  tests for the `Expected:` label and `Reason:` being retained when
  `isNotNull()` is used on the root context, which means nesting
  `atSameLevel` and creating an aliased root context.

### DIFF
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -297,9 +297,10 @@ class _TestContext<T> implements Context<T>, _ClauseDescription {
         _aliases = original._aliases,
         _fail = original._fail,
         _allowAsync = original._allowAsync,
-        // Properties that are never read from an aliased context because they
-        // are never present in `_clauses`.
+        // Never read from an aliased context because they are never present in
+        // `_clauses`.
         _label = '',
+        // Never read from any context other than the root.
         _reason = null;
 
   _TestContext._child(this._value, this._label, _TestContext<dynamic> parent)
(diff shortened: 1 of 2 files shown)


kevmoo (approved):
LGTM

natebosch:
Why is firefox broken on github actions all of a sudden? I don't see where we set up it up, how did it ever work?

kevmoo:
> Why is firefox broken on github actions all of a sudden? I don't see where we set up it up, how did it ever work?

🤷

kevmoo:
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md – latest ubuntu no longer has firefox as a default!

jakemac53 (approved):

kevmoo:
See https://github.com/google/webdriver.dart/commit/fcce0f0d0ebbda36398e1a6e3f29dfeb4eda43db

PR #1802, opened 2022-11-29
Use double quotes for test names on windows

### LINKED ISSUE #1798: `dart test` recommends a command that doesn't work on Windows
Consider the case where a test fails, when running `dart test`. 

Dart helpfully presents a command to run to replay the failed test. For example:

```
[C:\src\win32\tool\generator] dart test test\utils_test.dart
00:02 +0 -1: iidFromTypeIdentifier [E]
  Expected: '{03362E33-E413-5F29-97D0-48A4780935F9}'
    Actual: '{03362e33-e413-5f29-0000-000000000097}'
     Which: is different.
            Expected: {03362E33-E413-5 ...
              Actual: {03362e33-e413-5 ...
                            ^
             Differ at offset 6

  package:test_api            expect
  test\utils_test.dart 567:5  main.<fn>


To run this test again: c:\src\flutter\bin\cache\dart-sdk\bin\dart.exe test test\utils_test.dart -p vm --plain-name 'iidFromTypeIdentifier'
00:02 +0 -1: Some tests failed.

Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
For example, 'dart test --chain-stack-traces'.
```

However, on Windows, this command has unnecessary quotes, which causes it to fail if you paste it back into the shell:
```
[C:\src\win32\tool\generator] dart test test\utils_test.dart -p vm --plain-name 'iidFromTypeIdentifier' 
No tests ran.
No tests match "'iidFromTypeIdentifier'".

[C:\src\win32\tool\generator]
```

Instead, on Windows, you need to remove the quotes, for example:
```
[C:\src\win32\tool\generator] dart test test\utils_test.dart -p vm --plain-name iidFromTypeIdentifier
00:02 +1: All tests passed!
```

### Repro Environment
```
[C:\src\win32\tool\generator] dart --version
Dart SDK version: 2.19.0-416.0.dev (dev) (Mon Nov 21 18:50:19 2022 -0800) on "windows_x64"
```

natebosch:
Closes #1798

The windows shell does not allow single quotes around arguments. Use
double quotes on windows, and escape double quotes with 3 double quotes
in a row. The escape sequence closes the quoted string, emits an
adjacent double quote, then opens a new quoted string.

If any quotes are escaped the command will not work for users of the
bash shell in VSCode, but there were already issue with the path
separators in that shell so this shouldn't introduce severe new
problems. Test names that don't have double quotes in them should work
in either place.

### DIFF
--- a/pkgs/test/test/runner/compact_reporter_test.dart
+++ b/pkgs/test/test/runner/compact_reporter_test.dart
@@ -429,7 +429,7 @@ void main() {
         chainStackTraces: false);
   });
 
-  group('gives users a way to re-run failed tests', () {
+  group('gives non-windows users a way to re-run failed tests', () {
     final executablePath = p.absolute(Platform.resolvedExecutable);
 
     test('with simple names', () {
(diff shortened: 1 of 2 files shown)


natebosch:
I took the strategy from [this stack overflow answer](https://stackoverflow.com/a/15262019/374798) even though it says its a "not preferable" approach.

I used that approach because anything described in the other approaches looks much more complicated. It's possible that it wouldn't be too complex to take an approach which escapes the quotes and a hardcoded list of characters with special meaning to the windows shell.

jakemac53 (approved):

Issue #1803, opened 2022-11-30
VM tests can fail if they are run after any test which changes the VM working directory


natebosch:
This is especially noticeable when running with `-j 1` since it happens when a test is loaded following a change to the VM's working directory.

We currently pass through the `path` as relative to the working directory (assumed to be the package root). When we create the `.dart` content to compile for a VM test, we import the file against the current working directory.
https://github.com/dart-lang/test/blob/42ef3ed887138cd0f51a1a9db82ad9818da559a2/pkgs/test_core/lib/src/runner/vm/platform.dart#L157

If a test had change the process working directory by assigning to `Directory.current`, then the file we try to import won't exist.

I think what we probably want is to record the current directory when the platform is instantiated, and then resolve test paths against that working directory.

natebosch:
I think we also have a potential bug with matching against the line and column

https://github.com/dart-lang/test/blob/42ef3ed887138cd0f51a1a9db82ad9818da559a2/pkgs/test_core/lib/src/runner.dart#L318

We also do have this bug for the browser and node platforms.

PR #1804, opened 2022-11-30
Record the working directory for VM platform

natebosch:
Towards #1803

A VM test can change the working directory for the entire process. If this happens it will change the result for the absolute path against the relative paths provided to the VM platform plugin.

Record `Directory.current` when the platform plugin is instantiated. Any place where an absolute path is needed attempt to resolve against the original working directory.

Move some top-level methods into the plugin so they can read the cached working directory.


natebosch:
@jakemac53 - if you are OK with the approach I can work on adding tests and extending it to the browser and node platforms.

jakemac53:
Oh actually I realize now this needs to be done on all platforms since any vm test could break browser/node/etc tests.

jakemac53:
Another question - do we instantiate all platforms early on before running any tests? Or do we do it more lazily when a test is run? If the latter we would need to do this earlier before any test is ran and pass it to all the plugins.

natebosch:
> Another question - do we instantiate all platforms early on before running any tests? Or do we do it more lazily when a test is run? If the latter we would need to do this earlier before any test is ran and pass it to all the plugins.

Unfortunately it seems we instantiate them lazily.

I hope to avoid needing to change the API between the runner and test platforms, I wonder if we can get away with more eagerly instantiating them.

cc @christopherfujino - I don't think we have any situations where the flutter test platform is used in the same runner as the VM test platform - do we? I think `flutter test` shouldn't be impacted by this, but you might know better than me.

natebosch:
> I wonder if we can get away with more eagerly instantiating them.

This does not appear very feasible without some significant refactoring.

natebosch:
@jakemac53 - WDYT about moving forward with this workaround for the VM platform, and punting on a full solution for now? This is feeling fairly involved, and as it stands there is really not much workaround for impacted users beyond "Don't change directories without changing it back". If we ship this, it should let impacted users have a workaround of separating runs of VM platform from other platforms.

jakemac53:
I am fine with moving forward with this just for the VM platform.

jakemac53 (approved):

PR #1806, opened 2022-12-02
Merge test selections by path

### LINKED ISSUE #1785: Running the same suite multiple times using ?name=, ?line= runs multiple suites (differently to an OR regex)
I'm trying to implement https://github.com/Dart-Code/Dart-Code/issues/4021, with VS Code using line numbers for running tests instead of names (to improve reliability with dynamic test names we can't compute accurately at analysis time).

In doing so I noticed that `flutter test` isn't supporting the new `?name=foo` `?line=1` syntax (https://github.com/flutter/flutter/issues/114968) and while trying to implement that I noticed a difference when running multiple tests from the same file using this new syntax.

When I run multiple tests "the old way" using a regex, there is a single suite that runs with multiple tests:

```
danny@Dannys-MacBook-Pro dart_sample % dart test -r json "test/line_test.dart" --name "^a$|^b$"
{"protocolVersion":"0.1.1","runnerVersion":"1.22.0","pid":11165,"type":"start","time":0}
{"suite":{"id":0,"platform":"vm","path":"test/line_test.dart"},"type":"suite","time":0}
{"test":{"id":1,"name":"loading test/line_test.dart","suiteID":0,"groupIDs":[],"metadata":{"skip":false,"skipReason":null},"line":null,"column":null,"url":null},"type":"testStart","time":2}
{"count":1,"time":13,"type":"allSuites"}
{"testID":1,"result":"success","skipped":false,"hidden":true,"type":"testDone","time":554}
{"group":{"id":2,"suiteID":0,"parentID":null,"name":"","metadata":{"skip":false,"skipReason":null},"testCount":1,"line":null,"column":null,"url":null},"type":"group","time":557}
{"test":{"id":3,"name":"a","suiteID":0,"groupIDs":[2],"metadata":{"skip":false,"skipReason":null},"line":4,"column":3,"url":"file:///Users/danny/Desktop/dart_sample/test/line_test.dart"},"type":"testStart","time":560}
{"testID":3,"result":"success","skipped":false,"hidden":false,"type":"testDone","time":581}
{"success":true,"type":"done","time":591}
```

However, when I run using the new URI format (which I'll need for lines), it runs the suite twice, each with one test:

```
danny@Dannys-MacBook-Pro dart_sample % dart test -r json "test/line_test.dart?name=a" "test/line_test.dart?name=b"
{"protocolVersion":"0.1.1","runnerVersion":"1.22.0","pid":11196,"type":"start","time":0}
{"suite":{"id":0,"platform":"vm","path":"test/line_test.dart"},"type":"suite","time":0}
{"test":{"id":1,"name":"loading test/line_test.dart","suiteID":0,"groupIDs":[],"metadata":{"skip":false,"skipReason":null},"line":null,"column":null,"url":null},"type":"testStart","time":1}
{"suite":{"id":2,"platform":"vm","path":"test/line_test.dart"},"type":"suite","time":11}
{"test":{"id":3,"name":"loading test/line_test.dart","suiteID":2,"groupIDs":[],"metadata":{"skip":false,"skipReason":null},"line":null,"column":null,"url":null},"type":"testStart","time":11}
{"count":2,"time":11,"type":"allSuites"}
{"testID":1,"result":"success","skipped":false,"hidden":true,"type":"testDone","time":566}
{"group":{"id":4,"suiteID":0,"parentID":null,"name":"","metadata":{"skip":false,"skipReason":null},"testCount":1,"line":null,"column":null,"url":null},"type":"group","time":570}
{"test":{"id":5,"name":"a","suiteID":0,"groupIDs":[4],"metadata":{"skip":false,"skipReason":null},"line":4,"column":3,"url":"file:///Users/danny/Desktop/dart_sample/test/line_test.dart"},"type":"testStart","time":572}
{"testID":3,"result":"success","skipped":false,"hidden":true,"type":"testDone","time":577}
{"group":{"id":6,"suiteID":2,"parentID":null,"name":"","metadata":{"skip":false,"skipReason":null},"testCount":1,"line":null,"column":null,"url":null},"type":"group","time":577}
{"group":{"id":7,"suiteID":2,"parentID":6,"name":"d","metadata":{"skip":false,"skipReason":null},"testCount":1,"line":5,"column":3,"url":"file:///Users/danny/Desktop/dart_sample/test/line_test.dart"},"type":"group","time":577}
{"test":{"id":8,"name":"d b","suiteID":2,"groupIDs":[6,7],"metadata":{"skip":false,"skipReason":null},"line":6,"column":5,"url":"file:///Users/danny/Desktop/dart_sample/test/line_test.dart"},"type":"testStart","time":578}
{"testID":5,"result":"success","skipped":false,"hidden":false,"type":"testDone","time":601}
{"testID":8,"result":"success","skipped":false,"hidden":false,"type":"testDone","time":609}
{"success":true,"type":"done","time":614}
danny@Dannys-MacBook-Pro dart_sample % 
```

It's possible this is WAI but I wanted to query it because VS Code currently has some assumptions that an individual suite would only run once within a session (and I also think we have limitations with `flutter test` that we can only debug a single "suite" when using `--machine --start-paused`).

It is possible (or even sensible?) to "merge" these runs together and apply the name/line/col in a similar way to an OR name regex would be? (or some other way to support this, like doing `dart test foo.dart --line 1 --line 2`)?

@jakemac53

natebosch:
Closes #1785

When passing the same path multiple times on the command line, with
different `name`, `line`, or `col`, options run the test suite only
once. This does not affect cases where the same suite is loaded through
multiple path arguments from the command line, for instance if it was
specified directly and also loaded in a directory.

Remove the single `patterns`, `line`, and `col`, fields from
`SuiteConfiguration` in favor of a set of a new `TestSelection` with
these fields. Within each selection _all_ criteria must match for a test
to run, but test selections can independently allow a test to run.

Drop the `PathConfiguration` class, in favor of a map from the path to a
set of test selections. The global `Configuration` gets the map, and
eventually each suite run gets it's applicable set of selections.
Replace `Configuration.paths` with a map from the path option to a set
of selections. Group test path configurations by path to collect test
selections when parsing command line arguments.

Don't allow the test selections for a `SuiteConfiguration` in any of
the constructors - only accept a value with a new `selectTests` method
so that it is more clear where the data flows.

Also, add a `globalPatterns` field on `Configuration` to work alongside
test selections, but maintain separation between data sources.
Previously the patterns would be copied from the `suiteDefaults` for
each suite, and the suite specific patterns would be merged in.

Check whether any selection uses line or column when deciding whether to
capture test definition traces.

Add a test in `line_and_col_test.dart` that using the same test path and
line multiple times results in a single test ran.


natebosch:
I had a working version of this which kept `PathConfiguration`, but then I broke something when I went further and kept it as a `Map<String, Set<TestSelection>>` the whole way through. I'll see if I can unbreak it.

pkgs/test_core/lib/src/runner.dart
```diff
@@ -299,10 +293,20 @@ class Runner {
             return false;
           }
 
-          // Skip tests that don't start on `line` or `col` if specified.
-          var line = suite.config.line;
-          var col = suite.config.col;
-          if (line != null || col != null) {
+          final testSelections = suite.config.testSelections;
+          if (testSelections!.isEmpty) {
```
jakemac53, line 297: nit:

```suggestion
          final testSelections = suite.config.testSelections!;
          if (testSelections.isEmpty) {
```

It also feels a bit weird that this is nullable, should it have a default (empty) instead?

natebosch: We are not consistent with how we handle `null` and empty collections in our configuration.

jakemac53: The reason this one seems like it should be non-null is because we are requiring it to be set here, it feels a bit brittle (we are relying on a call to `selectTests` before it is valid). We could also handle null here.

natebosch: As written `null` and empty do mean different things, especially when merging.

I think, though, that `{}` can share a meaning with `{TestSelection()}` as written, so I can take a look at moving `{}` to hold the meaning that `null` does currently.

natebosch: I considered writing code to enforce that you can't `selectTests` more than once. This `!` does enforce that `selectTests` was called at least once.

natebosch: Made it non-null.
I added some asserts around having it be set to non-empty only once.

jakemac53 (approved):

PR #1808, opened 2022-12-03
Add nonce to test script

vrana:
This makes the test ready for strict CSP.

### DIFF
--- a/pkgs/test/lib/dart.js
+++ b/pkgs/test/lib/dart.js
@@ -67,6 +67,9 @@ script.onerror = function(event) {
   sendLoadException(message);
 }
 
+var firstScript = document.querySelector('script');
+script.nonce = firstScript.nonce;
+
 var parent = link.parentNode;
 document.currentScript = script;
 parent.replaceChild(script, link);


jakemac53:
This looks to be assuming something about another script existing on the page, which has a `nonce` set? That sounds kind of specific to a certain use case? I am not sure if there is a different workaround we could add instead here, that doesn't rely on that?

natebosch:
@vrana - can you share a bug link or a CL with more details?

vrana:
This is a common technique for strict CSP. All script tags has to be nonced. This just propagates the nonce. I'm doing this for b/248275555.

jakemac53:
> This is a common technique for strict CSP. All script tags has to be nonced. This just propagates the nonce. I'm doing this for b/248275555.

Can you explain more the methodology in this CL? It is just grabbing the first script, what script is that? Is copying nonces like this a standard practice?

I don't have any domain knowledge here and it just doesn't seem right on the surface, so I would like at least a rough explanation here.

kevmoo:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce

natebosch:
> All script tags has to be nonced.

Do you have a CL I can reference for where the script tag on the outer page gets a nonce? Where are you running a test such that these `nonce` fields are non-null?

vrana:
Yes, it's just grabbing the first script. All scripts have to be nonced so the first one has to be too.

natebosch:
Closing in favor of #1987 which looks more flexible.

PR #1809, opened 2022-12-05
Improve return type of runner `main()`.

sigurdm:
Dart allows `void` here, but `Future<void>` seems better.

### DIFF
--- a/pkgs/test/lib/src/executable.dart
+++ b/pkgs/test/lib/src/executable.dart
@@ -11,7 +11,7 @@ import 'runner/browser/platform.dart';
 import 'runner/node/platform.dart';
 import 'runner/wasm/platform.dart';
 
-void main(List<String> args) async {
+Future<void> main(List<String> args) async {
   registerPlatformPlugin([Runtime.nodeJS], () => NodePlatform());
   registerPlatformPlugin([
     Runtime.chrome,


jakemac53:
Generally `void` is the expected return type for `main` methods. Yes it is actually returning a future though, and I can see how the future would be useful if you are bootstrapping this executable entrypoint.

cc @natebosch any reason not to use `Future<void>` here from your perspective?

jakemac53:
Fwiw this entrypoint isn't really intended to be invoked by other programs, being in `src` and all. Are you trying to import it? Do you need a public library?

kevmoo:
I like aligning `Future` w/ async – but that's me

jakemac53:
> I like aligning `Future` w/ async – but that's me

But `void` is sending a more clear signal here that this future isn't intended to be used, which it isn't. There is no contract here that the test runner is totally done running by the time that future completes for instance.

We are only using async in order to be able to use `await` internally to the function.

kevmoo:
yeah...that

natebosch:
> There is no contract here that the test runner is totally done running by the time that future completes for instance.

This is the main sticking point for me.
If there is a useful reason to make the promise that the returned future completes after all tests are done (it might happen to do that today, but we don't make it a guarantee) then I think we could do that. If we don't have a specific reason, I'd prefer to keep the return type as `void` to not make it seem like we have made that promise.

jakemac53:
Unless there is a reason to change this, I would probably just leave it as is then? I don't want to be signaling any sort of guarantees here, and if anything it just encourages use of this file as a library which we don't support.

natebosch:
Closing for now. @sigurdm - let us know if you have a specific use case for calling this function.

sigurdm:
I'd like to call it from a test wrapper we have in the pub repo. https://github.com/dart-lang/pub/pull/3683/files#diff-e5faa0384ab03b74e6dc4109b8e2f9839c454d63d743ce6429adc84b9412ffd9R142 .

I know it is private code and without guarantees that this will work forever - but as long as it is just for my little internal test script I don't mind updating if the main function ever moves, or otherwise changes.

I don't really need to await the future in my case (it is just a tail call), I just noticed and thought it was more consistent to return a Future hence this PR.

I'm OK with closing this if you prefer it as it is.

natebosch:
> from a test wrapper

Consider using a shell script - it might be a little easier to correctly handle signals.

I don't see any reason not to be consistent with the [`main` we have defined in `package:test_core/src/executable.dart`](https://github.com/dart-lang/test/blob/3d6039b3075897adb184094ccc881c63a7479267/pkgs/test_core/lib/src/executable.dart#L37) so I'll merge this.

natebosch (approved):

jakemac53:
> I don't see any reason not to be consistent with the [`main` we have defined in `package:test_core/src/executable.dart`](https://github.com/dart-lang/test/blob/3d6039b3075897adb184094ccc881c63a7479267/pkgs/test_core/lib/src/executable.dart#L37) so I'll merge this.

Well, we do await that internally :). Ideally that function wouldn't even be called `main` unless it is invoked as an executable... which I doubt it is?

jakemac53:
(to be clear I am fine with merging this still though, assuming no guarantees)

Issue #1811, opened 2022-12-08
Prepare for Dart 3


mit-mit:
The source currently contains several references to APIs scheduled for removal in Dart 3 ([details here](https://github.com/dart-lang/sdk/issues/49529)):

* [`CyclicInitializationError`](https://github.com/search?q=repo%3Adart-lang%2Ftest+CyclicInitializationError&type=code)
* [`NullThrownError`](https://github.com/search?q=repo%3Adart-lang%2Ftest+NullThrownError&type=code)

Can we move the min up to 3.0, and start working on a dev release version for this?

jakemac53:
We also reference `CastError`.

jakemac53:
cc @mit-mit what is the timeline here? How soon with this package break in dart 3 (or is it already broken?).

mit-mit:
The plan was to land changes removing the APIs starting Monday, Dec 12th, in Dart main.

The potential breaking is in google3, where there's a copy of the package, which would then fail analysis after a google3 roll of the SDK.

jakemac53:
Ok, so its a P0 then?

mit-mit:
Yes, I believe so

PR #1812, opened 2022-12-08
Prepare for Dart 3.0 api removals

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1811

I also pre-emptively bumped our dep on matcher, which I think should be fine, but please push back if you disagree.


natebosch (approved):

legacy_tests/nnbd_opted_in_with_optout/mono_pkg.yaml
```diff
@@ -0,0 +1,18 @@
+# See https://pub.dev/packages/mono_repo
+
+sdk:
+- pubspec
+- stable # We will want to hardcode 2.19 once 3.x comes out
```
natebosch, line 5: I'm not sure we need to run on stable at all. Running on 2.18 from the pubspec version should be enough.

natebosch (approved):

legacy_tests/nnbd_opted_in_with_optout/mono_pkg.yaml
jakemac53: SGTM, swapped all these to run on 2.18

mit-mit:
Thanks Jake! Was this rolled?

mit-mit:
Doesn't look like it was rolled; doing that in https://dart-review.googlesource.com/c/sdk/+/274542

Issue #1816, opened 2022-12-15
Browser tests can fail due random path (secret) using base64 special characters


gmpassos:
If you try to run a browser test that needs to load a service worker script it may fail if the browser URL has some special characters:

```
Failed to load "/project/test/basic_test.dart": Failed to register a ServiceWorker: The provided scope ('http://localhost:58097/h%2B%2F5%2Bn%2BAjQMt1AWvgdkOqDlm4Pi6AYu6/test/') or scriptURL ('http://localhost:58097/h%2B%2F5%2Bn%2BAjQMt1AWvgdkOqDlm4Pi6AYu6/test/test-worker.js') includes a disallowed escape character.
```

A service worker path can't have `%2F` (`/`) or `%5C` (`\`) characters. Since these are valid base64 characters, the secret
used to define the URL path of a test in the browser can't be a Base64 string.

https://github.com/dart-lang/test/blob/89a8b12bc7998b01e801cbeed82b3f64f49d1ef9/pkgs/test/lib/src/runner/browser/platform.dart#L73

The code above shows that the `_secret ` is actually an URL Encoded Base64 string, what is not a recommended base path for a web content and not compatible with service worker scripts paths.

To fix that it needs to be changed to a secret with only `[a-zA-Z0-9]` characters. Since the `_secret` is already random, change it to only `[a-zA-Z0-9]` won't generate any issues with previous Dart code or IDEs using browser tests.

gmpassos:
@jakemac53 Thanks for the fast fix!

PR #1817, opened 2022-12-16
change the url secrets for browser tests to be alphanumeric characters

jakemac53:
fixes https://github.com/dart-lang/test/issues/1816

### DIFF
--- a/pkgs/test/lib/src/runner/wasm/platform.dart
+++ b/pkgs/test/lib/src/runner/wasm/platform.dart
@@ -71,7 +71,7 @@ class BrowserWasmPlatform extends PlatformPlugin
   ///
   /// This is used to ensure that other users on the same system can't snoop
   /// on data being served through this server.
-  final _secret = Uri.encodeComponent(randomUrlSecret(24));
+  final _secret = randomUrlSecret();
 
   /// The URL for this server.
   Uri get url => _server.url.resolve('$_secret/');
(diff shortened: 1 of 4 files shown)


pkgs/test/lib/src/util/math.dart
```diff
@@ -2,18 +2,19 @@
...
-String randomBase64(int bytes, {int? seed}) {
-  var random = math.Random(seed);
-  var data = Uint8List(bytes);
-  for (var i = 0; i < bytes; i++) {
-    data[i] = random.nextInt(256);
+final _rand = math.Random();
+
+/// Returns a random alphanumeric string ([a-zA-Z0-9]), which is suitable as
+/// a url secret.
+String randomUrlSecret(int length) {
```
natebosch, line 11: How about hardcoding the length, since it's a special purpose utility now? Is 24 the right number of characters to choose? This has a lot less entropy than it used to.

I think we could safely go up to 32 characters since that's how long the base64 encoded version would be. It's still a smaller secret size, but not enough to matter.

pkgs/test/lib/src/runner/wasm/platform.dart
```diff
@@ -71,7 +71,7 @@ class BrowserWasmPlatform extends PlatformPlugin
   ///
   /// This is used to ensure that other users on the same system can't snoop
   /// on data being served through this server.
-  final _secret = Uri.encodeComponent(randomBase64(24));
+  final _secret = Uri.encodeComponent(randomUrlSecret(24));
```
natebosch, line 74: We should not need the `Uri.encodeComponent` anymore.

natebosch (approved):

pkgs/test/lib/src/util/math.dart
jakemac53: sure, done

PR #1818, opened 2022-12-21
Add FailureDetail to allow expressive failures

natebosch:
Rejections do not carry any context about where in the
Check tree a failure occurred, so they are not sufficient to describe the
difference between the expected and actual values like we use in the
messages for test failure exceptions.

Some conditions which do a `softCheck` may want to print more
detailed mismatch information, for example an `Iterable.every` check may
want to print why an element didn't match, in addition to which index
failed.

Change the `fail` callback from taking a formatted String message and
the Rejection, to taking a `CheckFailure` with the rejection and a lazy
details field. The `FailureDetail` has the `expected` and `actual` that
were previously computed in separate methods, as well as information
about how far to indent lines from the `Rejection` so they line up with
the label for the specific check that failed.

Change the `softCheck` methods to return `CheckFailure?` instead of
`Rejection?`. Update conditions which were run against these values to
target `Check<CheckFailure?>` and add an extra `has` to read the
`rejection` field.

Remove the `_reason` field from `_TestContext` and handle it entirely
within `checkThat` since it now constructs the failure message.

Add an optional argument to the `indent` utility to let it indent more
than one level at a time.

### DIFF
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -59,48 +59,57 @@ Check<T> checkThat<T>(T value, {String? because}) => Check._(_TestContext._root(
     value: _Present(value),
     // TODO - switch between "a" and "an"
     label: 'a $T',
-    reason: because,
-    fail: (m, _) => throw TestFailure(m),
+    fail: (f) {
+      final which = f.rejection.which;
+      throw TestFailure([
+        ..._prefixFirst('Expected: ', f.detail.expected),
+        ..._prefixFirst('Actual: ', f.detail.actual),
+        ...indent(['Actual: ${f.rejection.actual}'], f.detail.depth),
+        if (which != null && which.isNotEmpty)
+          ...indent(_prefixFirst('Which: ', which), f.detail.depth),
+        if (because != null) 'Reason: $because',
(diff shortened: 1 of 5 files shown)


kevmoo (approved):

pkgs/checks/lib/src/checks.dart
```diff
@@ -59,48 +59,57 @@ Check<T> checkThat<T>(T value, {String? because}) => Check._(_TestContext._root(
...
     // TODO - switch between "a" and "an"
     label: 'a $T',
-    reason: because,
-    fail: (m, _) => throw TestFailure(m),
+    fail: (f) {
+      final which = f.rejection.which;
+      throw TestFailure([
+        ...prefixFirst('Expected: ', f.detail.expected),
+        ...prefixFirst('Actual: ', f.detail.actual),
+        ...indent(['Actual: ${f.rejection.actual}'], f.detail.depth),
```
kevmoo, line 67: it'd be nice to see a test showing the output here!

natebosch: I'll add a couple cases that show the full output of simple examples. https://github.com/dart-lang/test/pull/1820

PR #1819, opened 2022-12-21
Make the prefixFirst utility public

natebosch:
This will be useful for some condition extensions which want to do
formatting of mismatches from `softCheck`.

### DIFF
--- a/pkgs/checks/lib/src/describe.dart
+++ b/pkgs/checks/lib/src/describe.dart
@@ -13,3 +13,15 @@ String literal(Object? o) {
 }
 
 Iterable<String> indent(Iterable<String> lines) => lines.map((l) => '  $l');
+
+Iterable<String> prefixFirst(String prefix, Iterable<String> lines) sync* {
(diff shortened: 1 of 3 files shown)


kevmoo (approved):

pkgs/checks/lib/src/describe.dart
```diff
@@ -13,3 +13,15 @@ String literal(Object? o) {
 }
 
 Iterable<String> indent(Iterable<String> lines) => lines.map((l) => '  $l');
+
+Iterable<String> prefixFirst(String prefix, Iterable<String> lines) sync* {
```
kevmoo, line 17: at some point, doc comments on public members would be nice 🤷

natebosch: Yeah, I think this entire library will be due for polishing when we get closer to publish. I'm not sure how much churn it will see though, so I'm planning on getting the conditions fleshed out before I do that.

PR #1821, opened 2022-12-21
Add pairwiseCompare condition

natebosch:
Re-implements `pairwiseCompare` from `matcher` which has limited, but
non-zero usage.

The callback gets a `Check` and the expected value, instead of the
actual and expected values, and causes a condition to fail instead of
returning `false` for mismatch.

Rejects iterables that have the wrong number of elements, or that fail
an individual condition for any element.


kevmoo (approved):
Is this really that useful? I guess I've never used it. 🤷

natebosch:
> Is this really that useful? I guess I've never used it.

Yeah I was unsure about adding it at first, but there is a little usage and I didn't see an obvious better way to write some of the tests I looked at without something like this.

PR #1825, opened 2022-12-21
Allow slower node test on windows

natebosch:
This test case is flaky on windows and occasionally times out. It is not
clear if it is slow, or if it sometimes gets stuck entirely. Try
doubling the timeout on windows to see if it stops flaking.


jakemac53 (approved):

natebosch:
These are definitely still flaky 😢

PR #1826, opened 2022-12-21
Add String.equals condition

natebosch:
Adds a specialization of the equals condition which emits more detailed
information about how the values differ. The implementation is mostly
copied from `package:matcher`.

Make the `actual` argument to `isARejection` optional. In most cases the
value will be `literal(actual)` which isn't interesting to test.

Copy `escape` from `package:matcher` into `describe.dart`.

### DIFF
--- a/pkgs/checks/test/extensions/core_test.dart
+++ b/pkgs/checks/test/extensions/core_test.dart
@@ -140,5 +140,56 @@ void main() {
       checkThat(softCheck<String>('bob', (p0) => p0.endsWith('kayleb')))
           .isARejection(actual: "'bob'", which: ["does not end with 'kayleb'"]);
     });
+
+    group('equals', () {
+      test('succeeeds for happy case', () {
+        checkThat('foo').equals('foo');
+      });
+      test('reports extra characters for long string', () {
+        checkThat(softCheck<String>('foobar', (c) => c.equals('foo')))
+            .isARejection(which: [
+          'is too long with unexpected trailing characters:',
+          'bar'
+        ]);
+      });
+      test('reports truncated extra characters for very long string', () {
+        checkThat(softCheck<String>(
+                'foobar baz more stuff', (c) => c.equals('foo')))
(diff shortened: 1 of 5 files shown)


kevmoo (approved):
Only nit: might be good to have a case for an empty string in either case (actual or expected)

natebosch:
> might be good to have a case for an empty string in either case (actual or expected)

Done.

pkgs/checks/test/extensions/core_test.dart
```diff
@@ -145,13 +145,20 @@ void main() {
...
       test('reports extra characters for long string', () {
         checkThat(softCheck<String>('foobar', (c) => c.equals('foo')))
             .isARejection(which: [
           'is too long with unexpected trailing characters:',
           'bar'
         ]);
       });
+      test('reports extra characters for long string against empty', () {
+        checkThat(softCheck<String>('foo', (c) => c.equals(''))).isARejection(
+            which: ['is too long with unexpected trailing characters:', 'foo']);
```
kevmoo, line 160: Right...but we should word this differently. Like we should say 'is not empty' or something

pkgs/checks/test/extensions/core_test.dart
```diff
@@ -167,6 +174,10 @@ void main() {
           'bar'
         ]);
       });
+      test('reports missing characters for empty string', () {
+        checkThat(softCheck<String>('', (c) => c.equals('foo'))).isARejection(
+            which: ['is too short with missing trailing characters:', 'foo']);
```
kevmoo, line 179: ditto: 'is empty, when a string of ... is expected' or something

pkgs/checks/test/extensions/core_test.dart
natebosch: Ah I see what you mean. WDYT now?

kevmoo: *chefs kiss* 👍

PR #1827, opened 2022-12-22
Implement Future.doesNotComplete

natebosch:
Add `Context.lateReject` to allow a condition to report a failure after
it has already returned `null` (instead of a `Rejection`). This will
also be useful for a replacement for `expectAsync` with a limit on
allowed calls.

Add `doesNotComplete` in the Future extension. Always accept the value
synchronously, but reject it late if the Future does complete. Include
the stack trace for futures that result in an error.

### DIFF
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -207,6 +207,18 @@ abstract class Context<T> {
   /// this method will throw.
   Future<Check<R>> nestAsync<R>(
       String label, FutureOr<Extracted<R>> Function(T) extract);
+
+  /// Report that a check failed after the condition has already succeeded.
+  ///
+  /// A condition that some event _never_ happens will not have any point at
+  /// which it can be considered "complete", so a rejection may occur at any
+  /// time.
+  ///
+  /// Late rejections are ignored in [softCheck] and [softCheckAsync], so a
+  /// condition which use this method may not be useful within those method.
+  /// The only useful effect of a late rejections to throw a `TestFailure` when
+  /// used with a [checkThat] check.
+  void lateReject(Rejection rejection);
 }
 
 /// A property extracted from a value being checked, or a rejection.
@@ -418,6 +430,11 @@ class _TestContext<T> implements Context<T>, _ClauseDescription {
     }
     return FailureDetail(expected, foundOverlap, foundDepth);
   }
+
+  @override
+  void lateReject(Rejection rejection) {
+    _fail(_failure(rejection));
+  }
 }
 
 class _SkippedContext<T> implements Context<T> {
@@ -443,6 +460,10 @@ class _SkippedContext<T> implements Context<T> {
   Future<Check<R>> nestAsync<R>(
       String label, FutureOr<Extracted<R>> Function(T p1) extract) async {
     return Check._(_SkippedContext());
+  }
+
+  void lateReject(Rejection rejection) {
+    // Ignore
   }
 }
 
--- a/pkgs/checks/lib/src/extensions/async.dart
+++ b/pkgs/checks/lib/src/extensions/async.dart
@@ -26,6 +26,30 @@ extension FutureChecks<T> on Check<Future<T>> {
     });
   }
 
+  /// Expectst that the `Future` never completes as a value or an error.
+  ///
+  /// Immediately returns and does not cause the test to remain running if it
+  /// ends.
+  /// If the future completes at any time, raises a test failure. This may
+  /// happen after the test has already appeared to succeed.
+  ///
+  /// Not compatible with [softCheck] or [softCheckAsync] since there is no
+  /// concrete end point where this condition has definitely succeeded. Will
+  /// never cause a soft check to fail.
+  void doesNotComplete() {
+    context.expect(() => ['does not complete as value or error'], (actual) {
+      unawaited(actual.then((r) {
+        context.lateReject(
+            Rejection(actual: 'A future that completed to ${literal(r)}'));
+      }, onError: (e, st) {
+        context.lateReject(Rejection(
+            actual: 'A future that completed as an error:',
+            which: ['threw ${literal(e)}', ...st.toString().split('\n')]));
+      }));
+      return null;
+    });
+  }
+
   /// Expects that the `Future` completes as an error.
   ///
   /// Returns a future that completes to a [Check<E>] on the error once the
--- a/pkgs/checks/test/extensions/async_test.dart
+++ b/pkgs/checks/test/extensions/async_test.dart
@@ -2,10 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:async/async.dart';
 import 'package:checks/checks.dart';
 import 'package:checks/src/checks.dart' show softCheckAsync;
 import 'package:test/scaffolding.dart';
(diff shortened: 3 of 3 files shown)


pkgs/checks/lib/src/checks.dart
```diff
@@ -207,6 +207,18 @@ abstract class Context<T> {
...
   Future<Check<R>> nestAsync<R>(
       String label, FutureOr<Extracted<R>> Function(T) extract);
+
+  /// Report that a check failed after the condition has already succeeded.
+  ///
+  /// A condition that some event _never_ happens will not have any point at
+  /// which it can be considered "complete", so a rejection may occur at any
+  /// time.
+  ///
+  /// Late rejections are ignored in [softCheck] and [softCheckAsync], so a
```
natebosch, line 217: I'm not sure if this is worth trying to eagerly block. We currently throw if you try to use an async expectation in `softCheck` instead of `softCheckAsync`, so we could potentially have another type of interaction which isn't allowed from either of those, but is allowed from `checkThat`.

natebosch: I found a pattern that I don't think it too ugly. I'll run with that since I think it will be less confusing to have something which loudly fails rather than silently does nothing.

pkgs/checks/lib/src/checks.dart
```diff
@@ -444,6 +479,12 @@ class _SkippedContext<T> implements Context<T> {
       String label, FutureOr<Extracted<R>> Function(T p1) extract) async {
     return Check._(_SkippedContext());
   }
+
+  @override
+  void expectLate(Iterable<String> Function() clause,
+      void Function(T actual, void Function(Rejection) reject) predicate) {
+    // Ignore
```
kevmoo, line 486: // Ignore what? Ignore noop?

natebosch: Yes. I'll update these to say `// noop`

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -26,6 +26,29 @@ extension FutureChecks<T> on Check<Future<T>> {
...
+  /// never cause a soft check to fail.
+  void doesNotComplete() {
+    context.expectLate(() => ['does not complete as value or error'],
+        (actual, reject) {
+      unawaited(actual.then((r) {
+        reject(Rejection(actual: 'A future that completed to ${literal(r)}'));
+      }, onError: (e, st) {
+        reject(Rejection(
+            actual: 'A future that completed as an error:',
+            which: ['threw ${literal(e)}', ...st.toString().split('\n')]));
```
jakemac53, line 47: We should probably use a LineSplitter here just to be a bit more defensive

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -26,6 +26,29 @@ extension FutureChecks<T> on Check<Future<T>> {
     });
   }
 
+  /// Expectst that the `Future` never completes as a value or an error.
+  ///
+  /// Immediately returns and does not cause the test to remain running if it
+  /// ends.
```
jakemac53, line 32: I am a bit curious about how this doesn't end up making the test process hang? Are we relying on the VM knowing that a certain future can never be completed? What if some I/O related future never completed (maybe that isn't possible, but it seems so?).

Should this possibly take a duration and hold the test open for that long? Granted, generally I would discourage any such test but I don't really see the value in a test like this without something like that. It seems like it is asking for flaky tests...

natebosch: > I am a bit curious about how this doesn't end up making the test process hang?
> What if some I/O related future never completed (maybe that isn't possible, but it seems so?).

I don't think this should cause a test process to hang, but there may be a very narrow pattern where that is possible. This would need to be the only listener on that future.
This implementation won't behave any worse that the current `doesNotComplete`, and that hasn't caused us any issues.

https://github.com/dart-lang/test/blob/8235a25b98be448bca627d58b969de42b8b6633c/pkgs/test_api/lib/src/expect/future_matchers.dart#L101-L104

> Should this possibly take a duration and hold the test open for that long?

That might be a good idea. The current API in `test` does not have that behavior, but we could consider optional arguments. I'd prefer to hold that for a different PR.

pkgs/checks/lib/src/checks.dart
```diff
@@ -173,6 +181,18 @@ abstract class Context<T> {
...
+  /// Report that this check may fail asynchronously at any point in the future.
+  ///
+  /// A condition that some event _never_ happens will not have any point at
+  /// which it can be considered "complete", so a rejection may occur at any
+  /// time.
+  ///
+  /// May not be used from the context for a [Check] created by [softCheck] or
+  /// [softCheckAsync]. The only useful effect of a late rejection is to throw a
+  /// [TestFailure] when used with a [checkThat] check.
+  void expectLate(Iterable<String> Function() clause,
```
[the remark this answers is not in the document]
natebosch: How about `exceptLateFailure`? Or `allowLateFailure`?

pkgs/checks/lib/src/extensions/async.dart
jakemac53: Ok, if we have an existing thing with this behavior, then just porting it directly sgtm

pkgs/checks/lib/src/checks.dart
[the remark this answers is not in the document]
jakemac53: I guess I need a further description of exactly what this is intended to do, to know what a good name would be? It seems to me like the best name would be `expectAsync` but I think I don't understand the difference?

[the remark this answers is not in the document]
jakemac53: will it call `predicate` at any point in the future, or is `predicate` just allowed to invoke the rejection function multiple times?

pkgs/checks/lib/src/extensions/async.dart
natebosch: Done.

pkgs/checks/lib/src/checks.dart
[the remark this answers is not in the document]
natebosch: I reworked the doc and added a section contrasting with `expectAsync`. Does that help?

pkgs/checks/lib/src/checks.dart
```diff
@@ -181,16 +181,22 @@ abstract class Context<T> {
   Future<void> expectAsync<R>(Iterable<String> Function() clause,
       FutureOr<Rejection?> Function(T) predicate);
 
-  /// Report that this check may fail asynchronously at any point in the future.
+  /// Expect that [predicate] will not invoke the passed callback with a
```
jakemac53, line 184: If we are expecting that it will _not_ ever invoke it with a rejection, then shouldn't the name be `expectNoAsyncFailure` or some such?

natebosch: `expectNoAsyncFailure` would work for me.

I also just thought of `expectUnawaited` which might give a better indication of why it would be used.

natebosch: `expectUnawaited` might also convey that the only "useful" side effect is the error which will always be uncaught async.

jakemac53: Actually `expectUnawaited` sounds pretty reasonable to me 👍

jakemac53 (approved):
Either the suggestion `expectUnawaited` or `expectNoAsyncRejection` LGTM - but actually `expectUnawaited` most directly describes the behavior imo

PR #1830, opened 2022-12-27
Add a Map operator[] extension

natebosch:
Allows performing validation on a specific key within a map. Expects
that a value with the given key is present, and returns a check on the
value.

### DIFF
--- a/pkgs/checks/test/extensions/map_test.dart
+++ b/pkgs/checks/test/extensions/map_test.dart
@@ -33,6 +33,12 @@ void main() {
     checkThat(_testMap).values.contains(1);
   });
 
+  test('index operator', () async {
+    checkThat(_testMap)['a'].equals(1);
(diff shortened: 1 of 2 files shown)


natebosch:
I'm iffy on the syntax for this one.

Both this and the comparison operators on `num` sometimes feel nice, and other times feel like too strong of a contrast from the rest of the expectations.

natebosch:
There is also overlap between this and `containsKey`. We could consider changing `containsKey` to have this implementation and allow further checks on the value instead of returning `void` like it does now.

pkgs/checks/test/extensions/map_test.dart
```diff
@@ -33,6 +33,12 @@ void main() {
     checkThat(_testMap).values.contains(1);
   });
 
+  test('index operator', () async {
```
kevmoo, line 36:
```suggestion
  test('operator []', () async {
```

kevmoo (approved):

PR #1835, opened 2022-12-29
Add String.equalsIgnoringWhitespace condition

natebosch:
Copy the `collapseWhitespace` utility from `package:matcher`. Add
`equalsIgnoringWhitespace` and tests. The collapsed version of the
strings needs to also be used for the "display" strings, since they may
have differing indexes.


kevmoo (approved):
There's a (small) part of me that hates this is all getting copied. Fighting the urge to suggest we create `pkg:_test`. 😄

natebosch:
> There's a (small) part of me that hates this is all getting copied.

Yeah, I'm a little wary of some of the more specialized methods.

I'll gather some usage numbers for a few of them in a doc and we can decide if there are any not worth carrying over.

PR #1846, opened 2023-01-12
Add Condition abstraction

natebosch:
Previously, all "condition" arguments were callbacks that take a
`Check`. Now there is a `Condition` class which has an `apply` method
for the same purpose.

Add an `it()` Utility to get an object that is both a `Condition`, and a
`Check`. Expectations run against the check are recorded and played back
against the real check when the condition is applied. Expectations can
be chained with a cascade, so that the outermost object is the one
passed as an argument.

Add a factory constructor which takes a callback because there is
currently no way to express a numeric comparison expectation with a
cascade. This wouldn't be necessary if we had methods like
`.isGreaterThan` instead of `operator >`.

Move `skip` to an extension since there are now multiple implementations
of `Check`.

### DIFF
--- a/pkgs/checks/test/extensions/async_test.dart
+++ b/pkgs/checks/test/extensions/async_test.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 
 import 'package:async/async.dart';
 import 'package:checks/checks.dart';
+import 'package:checks/context.dart';
 import 'package:checks/src/checks.dart' show softCheckAsync;
 import 'package:test/scaffolding.dart';
 import 'package:test_api/hooks.dart';
@@ -15,10 +16,10 @@ void main() {
     test('completes', () async {
       (await checkThat(_futureSuccess()).completes()).equals(42);
 
-      await _rejectionWhichCheck(
+      await _rejectionWhichCheck<Future>(
         _futureFail(),
-        (f) async => await f.completes(),
-        (check) => check.single.equals('Threw UnimplementedError'),
+        it()..completes(),
+        it()..single.equals('Threw UnimplementedError'),
       );
     });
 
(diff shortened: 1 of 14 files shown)


pkgs/checks/test/extensions/async_test.dart
```diff
@@ -15,10 +16,10 @@ void main() {
     test('completes', () async {
       (await checkThat(_futureSuccess()).completes()).equals(42);
 
-      await _rejectionWhichCheck(
+      await _rejectionWhichCheck<Future>(
         _futureFail(),
-        (f) async => await f.completes(),
-        (check) => check.single.equals('Threw UnimplementedError'),
+        it()..completes(),
```
natebosch, line 21: You can't easily `await` anymore, because of the cascade.

In many cases I don't think that should matter, but I'm not sure it will never matter.

natebosch:
I do think this would look better if we replace the operators with named functions.

https://github.com/dart-lang/test/pull/1847/files

natebosch:
After some reflection, I do think this pattern is worth trying for a while.

pkgs/checks/test/extensions/async_test.dart
natebosch: This seems to work OK using the async `.that` support. The replay will work correctly and the expectations are all checked.

kevmoo:
meaning you don't want a review here?

natebosch:
> meaning you don't want a review here?

meaning I do want a review. I think this is worth landing.

I'm happy to chat on GVC about the design or motivation if it helps.

jakemac53 (approved):

jakemac53:
I like how `it` reads/looks for sure, seems good to me

kevmoo (approved):
:shipit:

PR #1852, opened 2023-01-18
Improve output for collections and long strings

natebosch:
Change the `actual` argument for rejections from a single String to an
`Iterable<String>`. This allows for maintaining alignment when a multi
line value is displayed with indentation. Change the `literal` utility
to return an Iterable, and update conditions passing something other
than the return from `literal` to pass a List to `Rejection.actual`.

Add a `postfixLast` utility to mirror `prefixFirst` and append content
to the last line of a value. Combining `prefixFirst` and `postfixLast`
with delimiters like quotes or braces allows wrapping content than may be
be 1 or more lines.

Flesh out the `literal` utility. The implementation is based on
`prettyPrint` from `matcher` with some simplifications and adapted to
return an Iterable instead of a String with newlines. One significant
difference from `prettyPrint` is that type names are not included since
they are often redundant.

In some places where `literal` was used for a signature that must take a
single String (the label in a nested context can only be a single line)
join the result with `r'\n'`.

### DIFF
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -66,7 +66,8 @@ Check<T> checkThat<T>(T value, {String? because}) => Check._(_TestContext._root(
         throw TestFailure([
           ...prefixFirst('Expected: ', f.detail.expected),
           ...prefixFirst('Actual: ', f.detail.actual),
-          ...indent(['Actual: ${f.rejection.actual}'], f.detail.depth),
+          ...indent(
+              prefixFirst('Actual: ', f.rejection.actual), f.detail.depth),
           if (which != null && which.isNotEmpty)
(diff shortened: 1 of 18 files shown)


pkgs/checks/lib/src/checks.dart
```diff
@@ -619,7 +621,7 @@ class Rejection {
   /// message. The message will be indented to the level of the expectation in
   /// the description, and printed following the descriptions of any
   /// expectations that have already passed.
-  final String actual;
+  final Iterable<String> actual;
```
jakemac53, line 624: The doc comment should probably be updated here to describe how multiples are handled

jakemac53 (approved):

pkgs/checks/lib/src/checks.dart
natebosch: Done.

PR #1855, opened 2023-01-19
Avoid empty `::group::` in Github reporter

natebosch:
A `::group::` will have an arrow in the Github UI. This is misleading in
the majority of cases where there are no extra lines, and masks the
cases where there is output from the test.

Only wrap in `::group::` when there is a non-empty set of messages or
errors, since those indicate that there will be additional lines in the
output.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 1.22.3-dev
 
+* Avoid empty expandable groups for tests without extra output in Github
+  reporter.
+
 ## 1.22.2
 
 * Don't run `tearDown` until the test body and outstanding work is complete,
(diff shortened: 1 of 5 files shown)


devoncarew (approved):
lgtm - I think this is an improvement visually

devoncarew:
I think I'd wrapped everything in a group originally for consistency and in order to avoid having a ragged left edge. However, most tests don't emit output, so the left edge of a well-formed test suite will already be aligned:

Screenshot 2023-01-18 at 9 30 14 PM

From spot checking the CI output of this repo, the places where tests emit output:

Screenshot 2023-01-18 at 9 31 03 PM

are the pseudo tests we create for some of the test infrastructure - web compiling tasks and such? I remember having trouble finding the right icons for these peudo tests. Maybe having them generally be the only things in groups helps them stand out enough? Or perhaps this is a reminder to try and find a separate icon for these types of tests.

natebosch:
> Maybe having them generally be the only things in groups helps them stand out enough?

Yeah, I think that's the case. It's pretty easy to see the cases which have extra output (typically unexpected, but sometimes unavoidable).

jakemac53 (approved):

PR #1856, opened 2023-01-19
Fix a test to use an still invalid URI

natebosch:
A test was skipped because a URI which used to result in a `FormatError`
started getting accepted by the SDK. The linked issue is closed as
working as intended. Update the test to use a URI which is invalid
according to the current SDK.


natebosch:
https://github.com/dart-lang/sdk/issues/34988

jakemac53 (approved):

PR #1865, opened 2023-01-23
Overwrite stack traces to expectation call stack

natebosch:
When applying a `Condition` (as opposed to using `softCheck`) the stack
trace includes some uninteresting checks infrastructure, and the line
from the `_test.dart` file typically points to the higher level
expectation which took the Condition as an argument. It is often more
useful to point to the specific expectation called on the Condition
which is failing.

Record `StackTrace.current` for all direct interactions with
`_ReplayContext`. When replaying the interaction on another context,
catch exceptions and rewrite them to the recorded stack trace.

Add an expando to remember which exceptions have already had their
stack traces rewritten. The callback for a check which causes nesting
also runs the interactions for the nested context, but we don't want the
stack to point to the nesting level. Only the inner nested replay
context should rewrite the stack trace.

When replaying onto a replay context (`it()..that(someCondition)`),
directly copy the interactions and original stack traces instead of
calling the method and getting new stack traces. This makes it possible
to use condition syntax inside expectation definitions without
overwriting the stack traces to the utility from the test file, and
should be more efficient.

### DIFF
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -685,66 +685,94 @@ class ConditionCheck<T> implements Check<T>, Condition<T> {
   }
 }
 
-class _ReplayContext<T> implements Context<T>, Condition<T> {
-  final _interactions = <FutureOr<void> Function(Context<T>)>[];
+class _Interaction<T> {
+  final StackTrace stackTrace;
+  FutureOr<void> Function(Context<T>) callback;
+  _Interaction(this.callback, this.stackTrace);
+}
 
+final _rewrittenStackTraces = Expando<bool>();
+
+class _ReplayContext<T> implements Context<T>, Condition<T> {
+  final _interactions = <_Interaction<T>>[];
   @override
   void apply(Check<T> check) {
+    final context = check.context;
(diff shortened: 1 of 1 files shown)


natebosch:
An example where this helps:

In https://github.com/dart-lang/test/pull/1867

```dart
      checkThat(_testIterable).isRejectedBy(it()..every(it()..isLessThan(0)),
                               ^ // Old stack trace root
          hasWhichThat: it()
            ..deepEquals([
              ^ // New stack trace root
              'has an element at index 0 that',
              '  Actual: <0>',
              '  Which: is not less than <0>',
            ]));
```

In a previous state of what became https://github.com/dart-lang/test/pull/1868, this CL also helped with some async checks like the inner conditions of:

```dart
        await checkThat(_countingStream(2)).emitsInOrder([
          it()..emits().that(it()..equals(0)),
          it()..emits().that((it()..equals(1))),
          it()..isDone(),
        ]);
```

However, after I added a transaction for `emitsInOrder` I had to move all the conditions to use `softCheckAsync` instead of applying them directly to the source `Check`, so the failure is no longer getting thrown directly by the inner condition and this PR doesn't help.

I'm going to close this for now.

PR #1868, opened 2023-01-25
Overhaul async matchers

natebosch:
Start a transaction before checking expectations. If the check fails,
reject the transaction. Add tests for the transaction behavior of each
condition.

Add `emitsError` to check that the queue has an error of a specified
type instead of an event.

Add `inOrder`. This API mainly solves the problem of checking
multiple async expectations against the same `Check` instance, since it
is no possible to `await` in between cascades. It also maintains a
transaction for the entirety of the conditions, so that the stream gets
rolled all the way back on failure. The conditions use will use async
expectations, so we cannot synchronously find a description for all the
conditions. Descriptions are collected while checking the stream when we
have room for async work. The `clause` callback is not called until
after the `predicate` callback has completed. During `describeAsync` the
`predicate` callback is never called, so for the describe use case the
clause will fall back to stating the number of conditions.

Add `anyOf` to check multiple possible stream scenarios from the
same point in the queue. Similar to `inOrder` it uses a fallback
for describe, and collects condition descriptions when the predicate
callback is run. On rejection describes the rejection for each
individual condition choice.

Add `mayEmit` and `mayEmitMultiple` to consume events from the queue
without ever failing.

Add `isDone` to check that the queue ends without more events or
errors.

Expose `softCheckAsync` and `describeAsync` for defining expectation
extensions. Previously it was imported from src only for tests.

Fix up some users facing strings so that they read nicely in more
situations. Add a generic bound for `throws` and change the label based
on whether a type was passed. When writing the label for nesting clauses
without sub-clauses, omit the `that:` postfix.

Add `hasDescription` and `hasDescriptionAsync` test utilities. Add tests
for the descriptions of the async conditions.

### DIFF
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -249,11 +249,13 @@ abstract class Context<T> {
   /// more spaces.
   ///
   /// If [atSameLevel] is true then [R] should be a subtype of [T], and a
-  /// returned [Extracted.value] should be the same instance as passed value.
-  /// This may be useful to refine the type for further checks. In this case the
-  /// label is used like a single line "clause" passed to [expect], and
-  /// expectations applied to the return [Check] will behave as if they were
-  /// applied to the Check for this context.
+  /// returned [Extracted.value] should be the same instance as the passed
+  /// value, or an object which is is equivalent but has a type which is more
+  /// convenient to test. In this case expectations applied to the return
+  /// [Check] will behave as if they were applied to the Check for this
+  /// context. The [label] will be used as if it were a single line "clause"
+  /// passed to [expect]. If the label is empty, the clause will be omitted. The
+  /// label should only be left empty if the value extraction cannot fail.
   Check<R> nest<R>(String label, Extracted<R> Function(T) extract,
       {bool atSameLevel = false});
 
@@ -450,7 +452,7 @@ class _TestContext<T> implements Context<T>, _ClauseDescription {
     if (atSameLevel) {
       context = _TestContext._alias(this, value);
       _aliases.add(context);
-      _clauses.add(_StringClause(() => [label]));
+      if (label.isNotEmpty) _clauses.add(_StringClause(() => [label]));
     } else {
       context = _TestContext._child(value, label, this);
       _clauses.add(context);
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -500,11 +500,11 @@ class _TestContext<T> implements Context<T>, _ClauseDescription {
     var foundDepth = thisContextFailed ? 0 : -1;
     var foundOverlap = thisContextFailed ? 0 : -1;
     var successfulOverlap = 0;
-    late final List<String> expected;
+    final expected = <String>[];
     if (_clauses.isEmpty) {
-      expected = [_label];
+      expected.add(_label);
     } else {
-      expected = ['$_label that:'];
+      expected.add('$_label that:');
       for (var clause in _clauses) {
         final details = clause.detail(failingContext);
         expected.addAll(indent(details.expected));
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -128,6 +128,9 @@ Future<CheckFailure?> softCheckAsync<T>(T value, Condition<T> condition) async {
 /// Matches the "Expected: " lines in the output of a failure message if a value
 /// did not meet the last expectation in [condition], without the first labeled
 /// line.
+///
+/// Asynchronous expectations are not allowed in [condition], for async
+/// conditions use [describeAsync].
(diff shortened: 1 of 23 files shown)


pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -131,15 +186,148 @@ extension StreamChecks<T> on Check<StreamQueue<T>> {
...
+
+  /// Expects that the stream statisfies at least one condition from
+  /// [conditions].
+  ///
+  /// If this expectation fails, the source queue will be left in its original
+  /// state.
+  /// If this expectation succeeds, consumes the same events from the source
+  /// queue as the satisfied condition. If multiple conditions are satisfied,
+  /// chooses the condition which consumed the most events.
+  Future<void> anyOf(Iterable<Condition<StreamQueue<T>>> conditions) async {
```
natebosch, line 252: This has the same name as https://github.com/dart-lang/test/blob/7157541a93fd7217355f1e8323de7745751a12bd/pkgs/checks/lib/src/extensions/core.dart#L58 and is now relying on the specificity of the generic to be chosen for stream queues instead of the synchronous one. I was initially avoiding this, but it's one of the advantages of extension methods, so I think it's fine to lean on it a bit. WDYT?

pkgs/checks/lib/src/checks.dart
```diff
@@ -498,18 +498,23 @@ class _TestContext<T> implements Context<T>, _ClauseDescription {
...
-        foundDepth = details.depth + 1;
-        foundOverlap = details._actualOverlap + successfulOverlap + 1;
-      } else {
-        if (foundDepth == -1) {
-          successfulOverlap += details.expected.length;
+    late final List<String> expected;
+    if (_clauses.isEmpty) {
+      expected = [_label];
+    } else {
+      expected = ['$_label that:'];
```
jakemac53, line 505: This would be a bit cleaner I think:

```suggestion
    final expected = <String>[];
    if (_clauses.isEmpty) {
      expected.add(_label);
    } else {
      expected.add('$_label that:');
```

pkgs/checks/lib/src/extensions/async.dart
jakemac53: I think its probably fine, jump to definition etc are very reliable for extension methods

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -156,6 +349,79 @@ extension StreamChecks<T> on Check<StreamQueue<T>> {
...
+        if (softCheck(value, condition) == null) {
+          await actual.next;
+        }
+      } finally {
+        return null;
+      }
+    });
+  }
+
+  /// Optionally consumes events that matche [condition] from the stream.
```
jakemac53, line 375:
```suggestion
  /// Optionally consumes events that match [condition] from the stream.
```

pkgs/checks/test/extensions/async_test.dart
```diff
@@ -6,39 +6,71 @@ import 'dart:async';
...
-      (await checkThat(_futureSuccess()).completes()).equals(42);
-
-      await checkThat(_futureFail()).isRejectedByAsync(
-        it()..completes().that(it()..equals(1)),
-        actual: ['A future that completes as an error'],
-        which: ['Threw <UnimplementedError>'],
-      );
+    group('completes', () {
+      test('succeeds for a future that completes to a value', () async {
+        await checkThat(_futureSuccess()).completes().that(it()..equals(42));
```
jakemac53, line 19: meta comment: the `.that` here reads a bit weird to me here and below, feels more like a `with`, `where`, or `which`?

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -15,17 +15,16 @@ extension FutureChecks<T> on Check<Future<T>> {
...
+        try {
+          return Extracted.value(await actual);
+        } catch (e) {
+          return Extracted.rejection(
+              actual: ['a future that completes as an error'],
+              which: prefixFirst('threw ', literal(e)));
+        }
+      });
 
   /// Expectst that the `Future` never completes as a value or an error.
```
jakemac53, line 29:
```suggestion
  /// Expects that the `Future` never completes as a value or an error.
```

jakemac53 (approved):

pkgs/checks/lib/src/checks.dart

pkgs/checks/lib/src/extensions/async.dart
natebosch: Fixed.

pkgs/checks/lib/src/extensions/async.dart
natebosch: Fixed.

pkgs/checks/test/extensions/async_test.dart
natebosch: I have a doc which had a section considering this. The same suggestion was made an on earlier PR. I think that's probably a good indication that it's worth chagning.

PR #1869, opened 2023-01-25
Use pubspec_overrides.yaml files

natebosch:
`dart pub publish` will ignore these files, so it won't be necessary to manually change the files for each publish. `dart pub get` and `dart pub upgrade` will use the overrides from the files.

### DIFF
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -41,9 +41,3 @@ dev_dependencies:
   lints: '>=1.0.0 <3.0.0'
   test_descriptor: ^2.0.0
   test_process: ^2.0.0
-
-dependency_overrides:
-  test_core:
-    path: ../test_core
-  test_api:
-    path: ../test_api
--- a/pkgs/test/pubspec_overrides.yaml
+++ b/pkgs/test/pubspec_overrides.yaml
@@ -1,0 +1,5 @@
+dependency_overrides:
+  test_core:
+    path: ../test_core
+  test_api:
+    path: ../test_api
--- a/pkgs/test_api/pubspec.yaml
+++ b/pkgs/test_api/pubspec.yaml
@@ -31,9 +31,3 @@ dev_dependencies:
   lints: '>=1.0.0 <3.0.0'
   test: any
   test_core: any
-
-dependency_overrides:
-  test:
-    path: ../test
-  test_core:
-    path: ../test_core
--- a/pkgs/test_api/pubspec_overrides.yaml
+++ b/pkgs/test_api/pubspec_overrides.yaml
@@ -1,0 +1,5 @@
+dependency_overrides:
+  test:
+    path: ../test
+  test_core:
+    path: ../test_core
--- a/pkgs/test_core/pubspec.yaml
+++ b/pkgs/test_core/pubspec.yaml
@@ -34,7 +34,3 @@ dependencies:
 
 dev_dependencies:
   lints: '>=1.0.0 <3.0.0'
-
-dependency_overrides:
-  test_api:
-    path: ../test_api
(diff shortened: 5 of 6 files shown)


devoncarew (approved):
👍 

I'm not sure if there are any 'how to publish' docs to need to be updated?

jakemac53:
Fwiw I typically do the following:

- publish test_api (ignore pub warnings about overrides).
- Remove test_api override in test_core, run `dart pub upgrade`, ensure I get the latest test_api, and then publish (also ignoring pub warnings about overrides).
- Remove all overrides in test, run `dart pub upgrade`, ensure I get the latest test_api and test_core.

We don't afaik document this process anywhere but it is IMO the best strategy to ensure that our version solves are actually going to work out.

Using this overrides file actually makes it slightly more confusing to do that and IMO it also makes it easier to forget the overrides are in place at all? It is easy to skip the warning now in pub about overrides so I sort of like leaving them as is.

I don't feel super strongly though.

natebosch:
> it is IMO the best strategy to ensure that our version solves are actually going to work out.

Pub runs a solve on publish now. If I try to publish `test` from the current state of the repo (it depends on an unpublished `test_core`, with overrides in the separate file)

```
Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
Resolving dependencies... 
Because test depends on test_core 0.4.23 which doesn't match any versions, version solving failed.
```

And if I `pub get` or `pub upgrade`, it works.

```
Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
Resolving dependencies... (1.0s)
Warning: You are using these overridden dependencies:
! test_api 0.4.18 from path ../test_api
! test_core 0.4.23-dev from path ../test_core
Got dependencies!
```

I think this feature does exactly what we want. The new workflow for publish is

- publish `test_api`
- publish `test_core`
- publish `test`

And we retain as much safety as the current process you outlined.

jakemac53:
Interesting, so it does a version solve of just the regular dependencies (ignoring dev dependencies?). Presumably it is just a warning still though?

jakemac53 (approved):

jakemac53:
One thing the new process still doesn't do I think is ensure you are getting the _latest_ test_api/test_core right? Does it give any output regarding the version solve that would highlight that?

jakemac53:
> The pub solve failure is a hard failure. I'm not sure if that would end up causing problem with cyclic package dependencies, but that is the behavior I see.

Interesting, all of our own package cycles are related to dev dependencies afaik, but technically cycles should be allowed

natebosch:
> One thing the new process still doesn't do I think is ensure you are getting the _latest_ test_api/test_core right?

No, typically I maintain that by checking that pub deps are updated in the same PR as versions change.

How does the old process ensure that? Is it that you know to watch out for a line like

```
* test_api 0.4.18 (was 0.4.19 from path ../test_api)
```

And check that the numbers match?

natebosch:
Pub also runs the analyzer on publish now, and would show any static errors due to version mismatch as a warning.

https://github.com/dart-lang/pub/pull/3568

That doesn't help for the cases where we bump versions for coupling reasons that aren't visible statically, but I think we tend to be fairly careful when we make that type of change.

If anything I think I'm inclined to land this and if we hit problems, try to add automation later to check that the non-override constraints get kept up to date with the latest.

jakemac53:
> How does the old process ensure that? Is it that you know to watch out for a line like

When you run `pub upgrade` the output shows you whether there is a newer version available - since you just published test_api or test_core, you would expect to _not_ see that (ie: get the latest, which you just published).

jakemac53:
Fwiw I don't necessarily think that is a deal breaker, my LGTM still stands here. Having it run the analyzer still mitigates the potential issue.

PR #1871, opened 2023-01-26
Retry instead of extend timeout for flaky Node tests

natebosch:
The timeout does not seem to have reduced the flakes. Add a retry instead.


jakemac53 (approved):

natebosch:
Huh, apparently we don't support configuring retry with `onPlatform`... I wonder if that would be difficult.

natebosch:
This also doesn't seem to help 😢 

I'm very curious what causes this. Other tests should run node, so I don't know why only `test/runner/line_and_col_test.dart  ` has a problem with node tests on windows, or why it would be flaky.

PR #1872, opened 2023-01-26
Rename `that` to `which`

natebosch:
Rename both the sync and async versions. "which" reads better in more
situations.

Fix the examples to use Condition instead of a function literal. Minor
wording change for the async version doc.


natebosch:
https://github.com/dart-lang/test/pull/1868#discussion_r1088309627
https://github.com/dart-lang/test/pull/1693#discussion_r883977374

jakemac53 (approved):

PR #1874, opened 2023-01-27
Add String.matches condition

natebosch:
Takes a RegExp and succeeds if the input string matches.

### DIFF
--- a/pkgs/checks/lib/src/extensions/string.dart
+++ b/pkgs/checks/lib/src/extensions/string.dart
@@ -60,17 +60,9 @@ extension StringChecks on Check<String> {
   }
 
   /// Expects that the string matches the regular expression [expected].
-  ///
-  /// [expected] must be a [String] or a [RegExp]. If it is a string it will be
-  /// used as the pattern for a [RegExp].
-  void matches(Pattern expected) {
-    final regex = expected is String
-        ? RegExp(expected)
-        : expected is RegExp
-            ? expected
-            : throw ArgumentError('expected must be a RegExp or a String');
+  void matches(RegExp expected) {
     context.expect(() => prefixFirst('matches ', literal(expected)), (actual) {
-      if (regex.hasMatch(actual)) return null;
+      if (expected.hasMatch(actual)) return null;
       return Rejection(
           which: prefixFirst('does not match ', literal(expected)));
     });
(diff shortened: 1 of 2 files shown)


natebosch:
In `matcher` the [corresponding matcher](https://pub.dev/documentation/matcher/latest/matcher/Matcher/matches.html) takes `Pattern` and allows a `String` argument which gets auto-converted to a RegExp. I'm a little torn on adopting this approach - the _vast_ majority of `Pattern` instances will be `String` or `RegExp`, and it is convenient to do so, but there are other subclasses of `Pattern` so it's not entirely safe.

https://github.com/dart-lang/matcher/blob/7e6a665da079baa3e2a2046ab02e4353ee32556c/lib/src/string_matchers.dart#L147-L152

@kevmoo @jakemac53 WDYT? Should I change to a `Pattern` argument and throw if it's anything other than a `String` or `RegExp`?

natebosch:
> Should I change to a `Pattern` argument and throw if it's anything other than a `String` or `RegExp`?

I implemented this version and added tests for it. Let me know which you prefer - I think I lean now towards taking `Pattern` but I could be pretty easily convinced either way.

kevmoo:
> > Should I change to a `Pattern` argument and throw if it's anything other than a `String` or `RegExp`?
> 
> I implemented this version and added tests for it. Let me know which you prefer - I think I lean now towards taking `Pattern` but I could be pretty easily convinced either way.

Ugh. So we are NOT using the `Pattern`-ness of `String` at all then, right?

I'd almost type it as `Object` to not upset the OO gods

natebosch:
> Ugh. So we are NOT using the `Pattern`-ness of `String` at all then, right?

Right. It's abusing the fact that `Pattern` is _almost_ a union type in practice.

natebosch:
> I'd almost type it as `Object` to not upset the OO gods

I have a strong preference to type it as `RegExp` and push the constructor call into calling code over typing it as `Object`.

kevmoo:
> > I'd almost type it as `Object` to not upset the OO gods
> 
> I have a strong preference to type it as `RegExp` and push the constructor call into calling code over typing it as `Object`.

Right. It's more clear, too. Treating `String` like a `Pattern` vs as the input to the `Regexp` ctor is ...tricky

jakemac53 (approved):

PR #1875, opened 2023-01-27
Rename `Check` to `Subject`

natebosch:
The word "check" is a bit overused in the documentation, and it's a
little confusing to use it as a noun. "Subject" is unambiguously a noun,
and it makes the docs read nicer in some places.

### DIFF
--- a/pkgs/checks/README.md
+++ b/pkgs/checks/README.md
@@ -29,7 +29,8 @@
 
 Replace calls to `expect` with a call to `checkThat` passing the first argument.
 When a direct replacement is available, change the second argument from calling
-a function returning a Matcher, to calling the extension method on the `Check`.
+a function returning a Matcher, to calling the extension method on the
+`Subject`.
 
 When a non-matcher argument is used for the expected value, it would have been
 wrapped with `equals` automatically. See below, `.equals` may not always be the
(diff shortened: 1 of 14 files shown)


natebosch:
I also considered `Target`, but I think I prefer `Subject`. I'm open to opinions and other suggestions.

pkgs/checks/README.md
```diff
@@ -44,6 +45,6 @@ checkThat(actual).deepEquals(expected);
 
 ## Differences in behavior from matcher
 
-- The `equals` Matcher performed a deep equality check on collections.
-  `.equals()` check will only correspond to [operator ==] so some tests may need
-  to replace `.equals()` with `.deepEquals()`. **TODO: implement `deepEquals`**
+-   The `equals` Matcher performed a deep equality check on collections.
```
kevmoo, line 48: Indent diff intentional here?

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -78,7 +78,7 @@ extension FutureChecks<T> on Check<Future<T>> {
 ///
 /// Streams should be wrapped in user test code so that any reuse of the same
 /// Stream, and the full stream lifecycle, is explicit.
-extension StreamChecks<T> on Check<StreamQueue<T>> {
+extension StreamChecks<T> on Subject<StreamQueue<T>> {
```
kevmoo, line 81: ditto

pkgs/checks/lib/src/extensions/function.dart
```diff
@@ -4,19 +4,19 @@
 
 import 'package:checks/context.dart';
 
-extension ThrowsCheck<T> on Check<T Function()> {
+extension ThrowsCheck<T> on Subject<T Function()> {
```
kevmoo, line 7: dinno

kevmoo (commented):
LGTM – although we want to completely ditch `Check` from the package, right?

pkgs/checks/README.md
natebosch: I think I probably autoformatted and this is what the internal formatter does.

jakemac53 (approved):

PR #1876, opened 2023-01-27
Add missing dependency on package:lints

natebosch:
This has been in the analysis options for the entire repo, but not
applying to this package because the package was missing. Add the
dependency, and fix the lints that are newly surfaced.


kevmoo (approved):
:shipit:

PR #1877, opened 2023-01-27
Prepare to publish

natebosch:
Add description and version to pubspec. Remove `publish_to: none`.

Remove the readme section about using a git dependency.

Add readme sections on writing tests and custom expectations.

### DIFF
--- a/pkgs/checks/README.md
+++ b/pkgs/checks/README.md
@@ -1,4 +1,4 @@
-# Checking expectation with `checks`
+# Checking expectations with `checks`
 
 Expectations start with `checkThat`. This utility returns a `Subject`, and
 expectations can be checked against the subject. Expectations are defined as
--- a/pkgs/checks/README.md
+++ b/pkgs/checks/README.md
@@ -1,22 +1,119 @@
+# Checking expectation with `checks`
+
+Expectations start with `checkThat`. This utility returns a `Subject`, and
+expectations can be checked against the subject. Expectations are defined as
+extension methods, and different expectations will be available for subjects
+with different value types.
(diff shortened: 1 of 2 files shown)


pkgs/checks/README.md
```diff
@@ -1,23 +1,120 @@
-# Trying Checks as a Preview
+# Checking expectation with `checks`
```
jakemac53, line 1:
```suggestion
# Checking expectations with `checks`
```

pkgs/checks/README.md
```diff
@@ -1,23 +1,120 @@
...
+  Subject<Bar> get someField => has((a) => a.someField, 'someField');
+}
+```
+
+# Trying Checks as a Preview
 
 1.  Replace the existing `package:test/test.dart` import with
     `package:test/scaffolding.dart`.
 
+1.  Add an import to `package:checks/checks.dart`.
```
jakemac53, line 116: We should mention adding the dependency also probably?

jakemac53 (approved):

jakemac53:
Exciting times! 🚀

pkgs/checks/README.md
natebosch: I was kind of treating that as implied now that it will be on pub. It was helpful before to help with using git constraints, but I don't think we should pub info about pub deps into pub package readmes generally.

jakemac53: It is a required part of the process though, and while its implied for us I think it won't be for everybody?

kevmoo (approved):

PR #1881, opened 2023-01-30
checks: add an example

kevmoo:


kevmoo:
image

pkgs/checks/example/example.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:checks/checks.dart';
```
jakemac53, line 1: nit: copyright

natebosch (approved):

Issue #1862, opened 2023-01-23
Test runner should respect `CHROME_EXECUTABLE`


Zekfad:
I have custom installation of Google Chrome Beta, and using `CHROME_EXECUTABLE` to get it work with Dart.
It works fine with all tools except for tests, here's the log:

<details>
<summary>dart test -p chrome</summary>

```
D:\Users\vorob\Desktop\Projects\fetch_client>dart test -p chrome
00:09 +0: compiling test\client_conformance_test.dart
Info: Compiling with sound null safety.
Compiled 11,042,001 characters Dart to 1,611,431 characters JavaScript in 9.22 seconds using 333.664 MB of memory
Dart file /C:/Users/vorob/AppData/Local/Temp/dart_test_5c16f440/runInBrowser.dart compiled to JavaScript: /C:/Users/vorob/AppData/Local/Temp/dart_test_bc500204/test_970b1edf/client_conformance_test.dart.browser_test.dart.js

00:09 +0 -1: compiling test\client_conformance_test.dart [E]
  Failed to load "test\client_conformance_test.dart":
  Failed to run Chrome: The system cannot find the file specified.

    Command: C:\Users\vorob\AppData\Local\Google\Chrome\Application\chrome.exe --user-data-dir=C:\Users\vorob\AppData\Local\Temp\dart_test_d63e0bf1 http://localhost:9848/rgecBooVKJ7rmWD9HqMkt5sOZVYUpeei/packages/test/src/runner/browser/static/index.html?managerUrl=ws%3A%2F%2Flocalhost%3A9848%2FrgecBooVKJ7rmWD9HqMkt5sOZVYUpeei%2F0&debug=false --enable-logging=stdout --v=1 
--disable-extensions --disable-popup-blocking --bwsi --no-first-run --no-default-browser-check --disable-default-apps --disable-translate --disable-dev-shm-usage --headless --disable-gpu 
--remote-debugging-port=0.
  dart:io-patch/process_patch.dart 401:33                      _ProcessImpl._start
  dart:io-patch/process_patch.dart 38:20                       Process.start
  package:test/src/runner/browser/chrome.dart 80:37            new Chrome.<fn>.tryPort
  package:test/src/runner/browser/chrome.dart 97:40            new Chrome.<fn>
  package:test/src/runner/browser/browser.dart 68:39           new Browser.<fn>
  dart:async/zone.dart 1398:13                                 _rootRun
  dart:async/zone.dart 1300:19                                 _CustomZone.run
[...]

PR #1883, opened 2023-01-31
Add support for CHROME_EXECUTABLE environment variable

jakemac53:
Closes https://github.com/dart-lang/test/issues/1862

### DIFF
--- a/pkgs/test/lib/src/runner/executable_settings.dart
+++ b/pkgs/test/lib/src/runner/executable_settings.dart
@@ -36,6 +36,9 @@ class ExecutableSettings {
 
   /// The path to the executable for the current operating system.
   String get executable {
+    final envVariable = Platform.environment['CHROME_EXECUTABLE'];
+    if (envVariable != null) return envVariable;
+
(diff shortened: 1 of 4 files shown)


natebosch (approved):

pkgs/test/lib/src/runner/executable_settings.dart
```diff
@@ -36,6 +36,9 @@ class ExecutableSettings {
 
   /// The path to the executable for the current operating system.
   String get executable {
+    final envVariable = Platform.environment['CHROME_EXECUTABLE'];
```
natebosch, line 39: We probably should not have put a chrome specific override in a general purpose class...

I'm pretty sure this environment variable is overriding _all_ browsers now.

jakemac53: lol, whoops 😬

natebosch: https://github.com/dart-lang/test/pull/1970

Issue #1891, opened 2023-02-01
"dart test file:///foo.dart" doesn't work without a query string


DanTup:
`dart test` accepts a file URI if there is a querystring:

```
danny@Dannys-MacBook-Pro dart_sample % dart test "file:///Users/danny/Desktop/dart_sample/test/line2_test.dart?name=a"
00:00 +1: All tests passed! 
```

But if there's no querystring, you get an error:

```
danny@Dannys-MacBook-Pro dart_sample % dart test "file:///Users/danny/Desktop/dart_sample/test/line2_test.dart"       
00:00 +0 -1: loading file:///Users/danny/Desktop/dart_sample/test/line2_test.dart [E]                                                                      
  Failed to load "file:///Users/danny/Desktop/dart_sample/test/line2_test.dart": Does not exist.

To run this test again: dart test file:///Users/danny/Desktop/dart_sample/test/line2_test.dart -p vm --plain-name 'loading file:///Users/danny/Desktop/dart_sample/test/line2_test.dart'
00:00 +0 -1: Some tests failed.   
```

I don't know if this is intended, but I'm trying to pass URIs through from `flutter_tools` and wanted to just stick with `file://` URIs for consistency if possible.

DanTup:
(perhaps without a `?` this just looks like a relative path? but maybe a `file://` prefix can be special cased?)

jakemac53:
Seems like a bug to me

jakemac53:
I hadn't realized this was actually not specific to windows also, on windows would we expect to see a drive letter?

DanTup:
Correct, this was on macOS. Yep, on Windows I would expect it include the drive letter (and colon).

I would expect whatever is passed could be parsed with `Uri.parse` and if `.isScheme('file')` then use it as-is, and otherwise treat it as a path (which you can probably do `Uri.file()` on to get also as a Uri, perhaps something like `Uri.base.resolve(Uri.file(input))` if it needs to be absolute?).

jakemac53:
Yeah ok I am working on essentially that change in the linked PR. I want to also make sure we retain the ability to run tests with windows style relative paths also though (test\my_test.dart). Not sure that we can parse that as a Uri but we will see what it does.

DanTup:
[...]

PR #1893, opened 2023-02-01
Support running tests by absolute file: uri

jakemac53:
Fixes https://github.com/dart-lang/test/issues/1891
Fixes https://github.com/dart-lang/test/issues/1899

For any path which looks like it has a scheme (contains "://"), parse it as a uri and grab the path from that. On windows, we also remove any leading `/` from the path because it will be followed by a drive letter, which we want to be the first part of the path (and not preceded by a `/`).

Open to better ideas here too, if you have them :)

### DIFF
--- a/pkgs/test_core/lib/src/runner/configuration/args.dart
+++ b/pkgs/test_core/lib/src/runner/configuration/args.dart
@@ -181,7 +181,7 @@ void _parseTestSelection(
   TestSelection selection;
   String path;
   if (firstQuestion == -1) {
-    path = option;
+    path = Uri.parse(option).path;
     selection = TestSelection();
   } else if (option.substring(0, firstQuestion).contains('\\')) {
     throw FormatException(
--- a/pkgs/test_core/lib/src/runner/configuration/args.dart
+++ b/pkgs/test_core/lib/src/runner/configuration/args.dart
@@ -177,46 +177,35 @@ Configuration parse(List<String> args) => _Parser(args).parse();
 
 void _parseTestSelection(
     String option, Map<String, Set<TestSelection>> selections) {
-  var firstQuestion = option.indexOf('?');
-  TestSelection selection;
-  String path;
-  if (firstQuestion == -1 && !option.contains('://')) {
-    path = option;
-    selection = TestSelection();
-  } else if (firstQuestion != -1 &&
-      option.substring(0, firstQuestion).contains('\\')) {
+  final uri = Uri.parse(option);
+  // Restore the path to how it was given (namely, we want to report paths
+  // with their original separators).
+  var path = Uri.decodeComponent(uri.path);
+  // Strip out the leading slash before the drive letter on windows.
+  if (Platform.isWindows && path.startsWith('/')) {
+    path = path.substring(1);
+  }
+
+  final names = uri.queryParametersAll['name'];
+  final fullName = uri.queryParameters['full-name'];
+  final line = uri.queryParameters['line'];
+  final col = uri.queryParameters['col'];
+
+  if (names != null && names.isNotEmpty && fullName != null) {
     throw FormatException(
-        'When passing test path uris, you must pass the path in URI '
-        'format (use `/` for directory separators instead of `\\`).');
-  } else {
-    final uri = Uri.parse(option);
-
-    final names = uri.queryParametersAll['name'];
-    final fullName = uri.queryParameters['full-name'];
-    final line = uri.queryParameters['line'];
-    final col = uri.queryParameters['col'];
-
-    if (names != null && names.isNotEmpty && fullName != null) {
-      throw FormatException(
-        'Cannot specify both "name=<...>" and "full-name=<...>".',
-      );
-    }
-    path = uri.path;
-    // Strip out the leading slash before the drive letter on windows.
-    if (Platform.isWindows && path.startsWith('/')) {
-      path = path.substring(1);
-    }
-    selection = TestSelection(
-      testPatterns: fullName != null
-          ? {RegExp('^${RegExp.escape(fullName)}\$')}
-          : {
-              if (names != null)
-                for (var name in names) RegExp(name)
-            },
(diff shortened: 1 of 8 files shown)


pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -181,7 +181,15 @@ void _parseTestSelection(
...
-    path = option;
+    if (option.contains('://')) {
+      path = Uri.parse(option).path;
+      // Strip out the leading slash before the drive letter.
+      if (Platform.isWindows && path.startsWith('/')) {
+        path = path.substring(1);
+      }
+    } else {
+      path = option;
+    }
```
natebosch, line 192: Does it work to let it fall through to the other Uri handling path?

```dart
if (firstQuestion == -1 && !option.contains('://')) {
  path = option;
```

natebosch: I guess it might cause a problem with the first `else if`?

Maybe we could do

```dart
if (firstQuestion == -1 && !option.contains('://')) {
  path = option;
  selection = TestSelection();
} else {
  final uri = Uri.tryParse(option);
  if (uri == null) {
    throw FormatException(...);
  }
  // ...
}
```

jakemac53: I unified the URI parsing, not using `tryParse` though, just letting it do its normal exception throwing for now.

jakemac53: Ok I further simplified this now by always parsing it as a URI but then decoding the path component afterwords which restores any back slashes. This means we can support query params with backslash style paths also (even though its weird/invalid, it doesn't hurt?)

jakemac53: Ok I don't understand why but I am now seeing encoded backslashes, at least for the new test I added.

I expanded that test to work on any platform as well as adding a new test which uses a query param with platform specific separators. If it is going to work, we might as well test it.

jakemac53:
ready for another review @natebosch

pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -177,41 +177,35 @@ Configuration parse(List<String> args) => _Parser(args).parse();
...
-    path = option;
-    selection = TestSelection();
-  } else if (option.substring(0, firstQuestion).contains('\\')) {
-    throw FormatException(
-        'When passing test path queries, you must pass the path in URI '
-        'format (use `/` for directory separators instead of `\\`).');
-  } else {
-    final uri = Uri.parse(option);
+  final uri = Uri.parse(option);
+  // Restore the path to how it was given (namely, we want to report paths
```
natebosch, line 181: This comment confuses me.

I don't see how it corresponds to "how it was given", since we lose any knowledge about the original format once it is a URI.

I also don't know what you mean by "original separators". I don't see anything that changes separators below. I would expect `File.fromUri(uri).path` to change to platform separators, but I wouldn't expect `decodeComponent` to make any change to separators.

natebosch (approved):

pkgs/test_core/lib/src/runner/configuration/args.dart
natebosch: FWIW I suspect `File.fromUri(uri).path` might also handle the drive letter thing?

natebosch: Fort the record I don't necessarily think that using the platform separator is an improvement - the changes in the test files look fine to me either way. I'm mostly confused about how the comment lines up with the implementation.

jakemac53: > I don't see how it corresponds to "how it was given", since we lose any knowledge about the original format once it is a URI.

In the case of the test that I added, when that uri is parsed it encodes the backslashes and converts them to %5C. So decoding the component restores those encoded backslashes to their original form.

> I also don't know what you mean by "original separators". I don't see anything that changes separators below. I would expect `File.fromUri(uri).path` to change to platform separators, but I wouldn't expect `decodeComponent` to make any change to separators.

Yes I am not explicitly dealing with separators, just ensuring we decode them if they were encoded.

jakemac53: > FWIW I suspect `File.fromUri(uri).path` might also handle the drive letter thing?

It doesn't appear to, and it also doesn't support fragments or query strings

jakemac53: Note that calling `toFilePath()` would deal with the drive letter issue. But that also fails if there is a fragment or query string, and it would always return a platform specific path, which I think we don't really want.

jakemac53: The encoded backslashes is only happening on github actions, but I think its reason enough to leave the decoding logic in either way, so I restored it.

DanTup:
> On windows, we also remove any leading `/` from the path because it will be followed by a drive letter, which we want to be the first part of the path (and not preceded by a `/`).

You probably know this, but I think if you use `.toFilePath()` instead of `.path` you wouldn't need to do this (I've seen using `.path` on file URIs has been a source of many Windows bugs), although you'll get back the path in Windows form (with backslashes) which is often a good thing but I don't know if it's what you want here.

DanTup:
@jakemac53 it's not urgent, but so I know when to come back to the Flutter side of this, what's the process for it getting into the Dart SDK? If it's just a case of updating the revision in DEPs I'm happy to send a CL for that, but if there's a process/schedule (or you want to get other changes in with it) I'm happy to wait.

jakemac53:
It should just be updating the revision in DEPs, I think it might happen automatically now also but I am not sure. Usually we would wait until we publish as well but we probably could publish soon, cc @natebosch

jakemac53:
We also will want to validate these changes internally first

PR #1894, opened 2023-02-01
Try a postfix phrasing

natebosch:
Make the utility to get a `Subject` an extension method. Since it reads
after the value, change the phrasing to "must" and expectations to
"beSomeDescription".

This causes chained conditions to read poorly

```dart
actualIterable.must.haveLength.beGreaterThan(0);
```

Also needs a solution for passing a "reason" at the top level, but
hopefully which prevents passing a reason anywhere else.

### DIFF
--- a/pkgs/checks/example/example.dart
+++ b/pkgs/checks/example/example.dart
@@ -8,16 +8,16 @@ import 'package:test/scaffolding.dart';
 void main() {
   test('sample test', () {
     final someValue = 5;
-    checkThat(someValue).equals(5);
+    someValue.must.equal(5);
 
     final someList = [1, 2, 3, 4, 5];
-    checkThat(someList).deepEquals([1, 2, 3, 4, 5]);
+    someList.must.deeplyEqual([1, 2, 3, 4, 5]);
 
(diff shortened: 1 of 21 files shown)


pkgs/checks/example/example.dart
```diff
@@ -8,16 +8,16 @@ import 'package:test/scaffolding.dart';
 void main() {
   test('sample test', () {
     final someValue = 5;
-    checkThat(someValue).equals(5);
+    someValue.must.equal(5);
```
kevmoo, line 11: So we rename the package to `pkg:must`? 😄 

`pkg:should` is nice...

kevmoo: It feels a bit weird to have "values" floating around not wrapped in something.

jakemac53: I do agree it feels weird to have this be a getter. I could probably get used to it though.

natebosch:
I think for now we should go with a top level `check()` (or `checkThat`... still unsettled) and if we see folks adding the extension themselves, we can consider adding one (though it would be breaking to roll it out). I don't think I feel comfortable using the extension as the sole UX.

PR #1895, opened 2023-02-02
Experiment with disallowing subject chaining

natebosch:
Remove the return value from `nest` and `nestAsync`, for passing the
nested condition ahead of time.

This interacts really poorly with `has`. We should only consider doing
this if we can find a workaround to allow inference for `has` or if we
expect that the vast majority of users will use codegen and we can hide
it from them.

I do like how this eliminates the `which` calls for some async
expectations.


pkgs/checks/lib/src/extensions/string.dart
```diff
@@ -19,7 +19,8 @@ extension StringChecks on Subject<String> {
     });
   }
 
-  Subject<int> get length => has((m) => m.length, 'length');
+  void hasLengthWhich(Condition<int> lengthCondition) =>
```
natebosch, line 22: We would probably also add

```dart
void hasLength(int expectedLength) {
```

natebosch:
#1896 does this for only async nesting

natebosch:
https://github.com/dart-lang/test/issues/2073#issuecomment-1679773967 shows a use case for assigning a `Subject` to a variable. We'd lose the described advantage over `expect` if we disallow chaining for synchronous expectations.

I don't think we should make this change. I think in practice the distinction between sync and async expectations is not a problem.

PR #1896, opened 2023-02-02
Take Condition for async nesting expectations

natebosch:
Avoid the awkwardness of awaiting before chaining conditions, or
needing a `which` utility.

This introduces a discrepancy between chaining for sync vs async
expectations, but it makes many common async expectations nicer to read.

### DIFF
--- a/pkgs/checks/lib/src/extensions/async.dart
+++ b/pkgs/checks/lib/src/extensions/async.dart
@@ -429,14 +429,15 @@ extension ChainAsync<T> on Future<Check<T>> {
   /// `Future`.
   ///
   /// Extensions written on [Check] cannot be invoked on [Future<Check>]. This
-  /// method allows adding expectations for the value without awaiting it.
+  /// method allows adding expectations for the value without awaiting an
+  /// expression that would need parenthesis.
   ///
   /// ```dart
-  /// await checkThat(someFuture).completes().that((r) => r.equals('expected'));
+  /// await checkThat(someFuture).completes().which(it()..equals('expected'));
   /// // or, with the intermediate `await`:
   /// (await checkThat(someFuture).completes()).equals('expected');
   /// ```
-  Future<void> that(Condition<T> condition) async {
+  Future<void> which(Condition<T> condition) async {
     await condition.applyAsync(await this);
   }
 }
--- a/pkgs/checks/lib/src/extensions/async.dart
+++ b/pkgs/checks/lib/src/extensions/async.dart
@@ -8,14 +8,14 @@ import 'dart:convert';
 import 'package:async/async.dart';
 import 'package:checks/context.dart';
 
-extension FutureChecks<T> on Check<Future<T>> {
+extension FutureChecks<T> on Subject<Future<T>> {
   /// Expects that the `Future` completes to a value without throwing.
   ///
-  /// Returns a future that completes to a [Check<T>] on the result once the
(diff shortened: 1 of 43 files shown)


natebosch:
@lrhn @jakemac53 - What are your thoughts on this change?

Do you think this makes a positive impact in `async_test.dart`? Is it worth having an inconsistency between sync and async nesting (chaining) expectations?

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -15,7 +15,7 @@ extension FutureChecks<T> on Subject<Future<T>> {
   /// future completes.
   ///
   /// Fails if the future completes as an error.
-  Future<Subject<T>> completes() =>
+  Future<void> completes([Condition<T>? completionCondition]) =>
```
lrhn, line 18: I prefer the original here. It's explicit.
I assume that if the future has an error, the returned future has a test-failure error. (Which becomes an uncaught error if you don't do anything, which is fine, because we run tests in error zone.)

The alternative here seems to be, effectively doing a `then` for you, so
```dart
  check(myFuture).completes(conditon);
```
is equivalent to the original
```
  check(myFuture).completes().then(condition.applyAsync);
```

I don't think I'd want to use a `Condition` object directly if I can avoid it.

I see three options here:

Left: Return `Future<Subject<T>>`, and I assume it's a test failure if the future had an error.
Right: Pass in test 

Alternative: Return an "AsynchronousSubject" which accepts all the same tests as a normal `Subject`, but won't apply them until the future completes. (I think I tried that, and it got very complicated very quickly.)

jakemac53 (commented):
The tests to read better to me

pkgs/checks/lib/src/extensions/async.dart
natebosch: > I assume that if the future has an error, the returned future has a test-failure error. (Which becomes an uncaught error if you don't do anything, which is fine, because we run tests in error zone.)

Correct.


> Alternative: Return an "AsynchronousSubject" which accepts all the same tests as a normal `Subject`, but won't apply them until the future completes. (I think I tried that, and it got very complicated very quickly.)

This also wouldn't allow for waiting for the expectation to pass before doing more work or checking more expectations.  Any existing test doing `await expectLater` today would not have a migration pattern if we lost the ability to return a `Future` for async expectations.

natebosch: > Right: Pass in test

The condition is optional.

In both cases `.completes()` behave identically - cause a failure if the future has en error, otherwise succeed.

The the current code, chaining has multiple choices for how to write it. In order of preference for how I would write it:

```dart
await subject.completes().which(it()..furtherCondition());
(await subject.completes()).furtherCondition();
await subject.completes().then((result) => result.furtherCondition());
final result = await subject.completes();
result.furtherCondition();
```

After this PR, this would collapse to one choice

```dart
await subject.completes(it()..furtherCondition));
```

I like
- That there is only 1 option.
- That the 1 option reads nicer than any of the multiple choice options IMO.

I'm less sure that we can confidently say there are _no_ valid use cases for choosing one of the other patterns. I'm also less sure that everyone will have the distaste for `(await subject.completes()).furtherCondition()` that I picked up while working on this package.

natebosch:
I do think we should land this.

If we hear from people that it's too limiting, we can change it back.

jakemac53 (approved):

PR #1897, opened 2023-02-02
Add a top level equals utility

natebosch:
As a shortcut for the very common `it()..equals` pattern. Any other
condition will still need to fall back on `it()..`.


natebosch:
If we consider adding something like this after landing #2697 it would be something like `Condition.equals<T>(T)` with dot shorthand `.equals(t)`.

PR #1898, opened 2023-02-02
Add assume and expectThat utilities

natebosch:
assume marks the test as skipped. This is not useful currently, but
could be useful if it make the test runner not treat exceptions as
failures anymore.

expectThat does not throw an exception when it fails, instead it
collects failures and throws them in a teardown all together.

Add an `_isSkipped` field to `_TestContext` and remove
`_SkippedContext`. The field can be set late which allows us to bail out
of subsequent expectations on a given subject without throwing. This
avoids figuring out how to treat failure messages for an expectations
following another expectation that had failed. Current failure message
formatting relies on the fact that you only see clauses which were
successful.

Neither of these is very easy to test without sharing some testing
utilities from the test runner because it involves making expectations
about the state of the test runner. Another possibility is to run a
subprocess and make expectations about the reporter output.

### DIFF
--- a/pkgs/checks/lib/checks.dart
+++ b/pkgs/checks/lib/checks.dart
@@ -2,7 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-export 'src/checks.dart' show checkThat, Subject, Skip, it;
+export 'src/checks.dart' show checkThat, Subject, Skip, it, assume, expectThat;
 export 'src/extensions/async.dart'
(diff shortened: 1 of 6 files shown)


jakemac53:
Honestly I think I would just prefer a named argument to `check` (`String? skip`)? Would that be possible?

Something about it being its own top level api doesn't quite rub me right, and it isn't very obvious to me looking at it that this skips the expectation.

jakemac53:
Regarding `expectThat` I do think its useful to have this option, but if we move to `check` then it feels like this should be called `expect` which then has a name collision.

mosuem:
Closing this, as it is stale and the files are out of date. Feel free to reopen in case you still want to land it!

PR #1901, opened 2023-02-03
Take callbacks for actual and which

natebosch:
This aligns these arguments with all the other failure formatting
arguments like `clause` and `label`. There may be some performance
benefit in some cases where a rejection from `softCheck` is ignored and
expensive String operations are avoid, but in the typical case this just
introduces closures which will be invoked shortly.

Replace the default empty list for `actual` with a default function.
This is a slight behavior change where passing `() => []` will not get
overwritten with the default, but passing `[]` would have. Only
defaulting for when the argument was not passed at all is slightly
better behavior.

Replace a bunch of `Iterable<String>` with `Iterable<String> Function()`
and invoke them at the moment the strings are needed.


natebosch:
Taking this to the extreme, we could take callbacks for `actual` and `which` too. It would be consistent, but I don't know how much value it buys to go that far.
https://github.com/dart-lang/test/pull/1901/files

I think this PR, taking a callback for `label`, is a good balance.

pkgs/checks/test/test_shared.dart
```diff
@@ -51,16 +57,17 @@ extension RejectionChecks<T> on Subject<T> {
       didRunCallback = true;
       final failure = await softCheckAsync(value, condition);
       if (failure == null) {
-        return Extracted.rejection(which: [
-          'was accepted by the condition checking:',
-          ...await describeAsync(condition)
-        ]);
+        final description = await describeAsync(condition);
```
natebosch, line 60: Here is one place where we are not getting value from the callback potentially not getting invoked.

I can't make the `which` callback async, so we are forced to do this async work outside of the callback.

natebosch:
I don't think I want to go forward with this one. Most rejections will be for failures so the extra complexity isn't buying much value.

PR #1903, opened 2023-02-03
Add support for `--compiler` flag

jakemac53:
Closes https://github.com/dart-lang/test/issues/1776

## Overview

- Adds a new `Compiler` class, similar to the `Runtime` class, but it doesn't support custom ones (at least for now).
  - This is exported from `package:test_api/backend.dart`
- Adds required `defaultCompiler` and `supportedCompilers` fields to `Runtime`. This is breaking for custom runtimes.
  - Platforms will not be asked to load tests for compilers that the associated runtimes don't support.
- Add `compiler` field to `SuitePlatform`. This is how platforms get access to compiler configuration. It means there is no API breaking change for platforms, but they do need to be updated to support the configuration.
  - This is breaking, the `compiler` parameter is required when creating a `SuitePlatform`, although it is nullable, but defaults to the default compiler for the runtime. This is mostly for convenience and up for discussion.
- Adds support in all platforms to respect the compiler option.
  - The vm platform supports the `source` compiler which is roughly equivalent to `--use-data-isolate-strategy`. The `--use-data-isolate-strategy` flag is now just an alias for `--compiler source` and it it deprecated/hidden.
    - We would call this compiler `none` but we already have `OS.none`, so it would conflict in boolean selectors. 
- Adds a `CompilerSelection` class similar to `RuntimeSelection`, used by `SuiteConfig` to track the selected compiler(s).
- Adds new `compilerSelections` field to `SuiteConfig`.
- Adds the compilers to the valid boolean selectors, this makes them usable in `TestOn` etc.

## Still to do

- [ ] ~Consider merging the wasm support into the regular browser platform. Unclear if we want to do this now or wait. We could do it and then just call the compiler `--experimental-dart2wasm`?~
- [x] Write tests
- [x] Update changelogs/pubpsecs
- [x] Write docs
- [x] Probably drop or deprecate --use-data-isolate-strategy


pkgs/test_api/lib/src/backend/compiler.dart
```diff
@@ -0,0 +1,51 @@
...
+  ];
+
+  /// The human-friendly name of the compiler.
+  final String name;
+
+  /// The identifier used to look up the compiler.
+  final String identifier;
+
+  /// Whether this compiler is enabled by default.
+  final bool isDefault;
```
natebosch, line 36: I'm still skimming the change, but I find it confusing that being default is a property of a Compiler, and not a property of a Platform/Compiler relationship.

I could imagine that `dart2js` might remain a default on node, but ddc could become a default for the browser.

pkgs/test_api/lib/src/backend/compiler.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  const Compiler(this.name, this.identifier, {this.isDefault = false});
+
+  /// Converts a JSON-safe representation generated by [serialize] back into a
```
natebosch, line 40: [nit] mention that only the `builtIn` compilers are supported.

We could consider supporting serialization and deserialization to a `Map` for unrecognized compilers, but it's probably not necessary at the moment, since no other part of the system will support pluggable unrecognized compilers.

If and when we get around to looking at proper pluggable platform support, it would be nice to make compiler pluggable too.

pkgs/test_api/lib/src/backend/runtime.dart
```diff
@@ -82,7 +97,14 @@ class Runtime {
   /// That is, returns [parent] if it's non-`null` or [this] if it's `null`.
   Runtime get root => parent ?? this;
 
-  const Runtime(this.name, this.identifier,
+  /// The default compiler to use with this platform.
```
natebosch, line 100:
```suggestion
  /// The default compiler to use with this runtime.
```

pkgs/test_api/lib/src/backend/compiler.dart
jakemac53: Good catch this is definitely just cruft - I need to remove it.

pkgs/test_api/lib/src/backend/compiler.dart
jakemac53: Yeah I figured we would only add support for custom compilers if requested/needed later on.

jakemac53: I went ahead and just removed the public constructor.

jakemac53:
@natebosch I forgot to send it as a PR, but please take a look at https://github.com/dart-lang/test/pull/1903/commits/3748acb6666711b35661c7d406fa120c1490ee3f which updates the pubspecs and changelogs

jakemac53:
Ok, this PR is ready for a final review now. As discussed offline we decided to leave the wasm support as is for now, given that chrome requires extra special flags to run in that mode, it makes sense to be a separate platform still.

pkgs/test/test/runner/compiler_test.dart
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
natebosch, line 1:
```suggestion
// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```

natebosch (approved):

PR #1904, opened 2023-02-03
update legacy tests to ignore language version comment for development on newer sdks

jakemac53:
This gives a diagnostic today if you are on a 3.x SDK. These tests only run on older SDKs and I think its best to just silence the warning.


natebosch (approved):
Is it feasible to tweak the mono repo setup so that these aren't analyzed on new SDKs? or is the concern mostly the local IDE experience?

jakemac53:
The concern is just the local IDE experience

PR #1907, opened 2023-02-04
Rename matches to matchesPattern, accept Pattern

natebosch:
This will treat a `String` argument differently from the equivalent
matcher in `package:matcher`, which would have converted strings to
regex. Document this in the README.

Add tests showing the behavior of passing strings, and explicitly choose
cases that show the argument is not treated as a regular expression.

Expand the doc comment and show examples of both types of arguments.


natebosch:
@lrhn - you had mentioned an `entireStringOnly` flag. Do you think that's important to add now, or can we wait until someone asks for it?

How would you implement that? is it something like `expected.allMatches(actual).any((m) => m.start == 0 && m.end == actual.length)`?

kevmoo (approved):

pkgs/checks/doc/migrating_from_matcher.md
```diff
@@ -66,6 +66,12 @@ check(actualCollection).deepEquals(expected);
     value to check for equality, a matcher for the emitted value, or a matcher
     for the entire queue which would match multiple values. Use `it()..emits()`
     to check the emitted elements.
+-   The [`matches` Matcher][matches] would convert a `String` argument into a
+    `Regex`, so `matches(r'\d')` would match the value `'1'`. With
+    `matchesPattern` a `String` argument is treated as a `Pattern` instead of
```
jakemac53, line 71: A bit confused by this description, it sounds like we are converting a `String` to a `Pattern` or something -if you don't realize that `String` already implements pattern then it isn't obvious that we are just leaving the string alone.

natebosch: I expanded this. Is it more clear now?

jakemac53 (approved):

pkgs/checks/lib/src/extensions/string.dart
```diff
@@ -59,10 +59,18 @@ extension StringChecks on Subject<String> {
...
+  /// Expects that the string matches the pattern [expected].
+  ///
+  /// Fails if [expected] returns an empty result from calling `allMatches` with
+  /// the value.
+  ///
+  /// ```
+  /// check(actual).matchesPattern('abc');
+  /// check(actual).matchesPattern(RegExp(r'\d'));
+  /// ```
+  void matchesPattern(Pattern expected) {
```
lrhn, line 71: Since people just loooove their regexps, consider _also_ providing a shorthand:
```dart
  void matchesRegExp(String source, {bool caseSensitive = true, multiline = false, unicode = false}) { 
      matchesPattern(RegExp(source, caseSensitive: caseSensitive, multiline: multiline, unicode: unicode);
  }
```

If you have a `RegExp` object already, you can just use `matchesPattern`, but it saves a little verbosity to not have to write the `RegExp(...)`.

(Or not, and wait for someone else to provide it.)

lrhn: Could `matchesPattern` return a `Subject<Match>`, so you can keep testing on the result, like:
```dart
  check(someString).matchesPattern(myRegExp)
     ..[0]!.contains("foo")
     ..[1].isNotNull.contains("foo");
```
(Is it better to return a `Subject<Match>?` or a `Subject<Match>` which may be inert/ignoring checks, when there is no match? I guess the latter is safer if you don't know immediately if there is a match.)

lrhn (commented):
LGTM

PR #1910, opened 2023-02-06
Update vm_service to support 11.0.0

CoderDake:


natebosch (approved):

natebosch:
Do we need to publish right away?

CoderDake:
@natebosch, yes it would be great if this could be published as soon as possible :D

PR #1911, opened 2023-02-06
Support targeting compilers via platform selectors

jakemac53:
This updates the `--compiler` flag to support a syntax like this `--compiler [<platform-selector>:]<compiler>`.

This is quite handy because it allows for instance `-c web:dart2js` or `-c chrome:dart2wasm` etc. It ultimately does not increase the complexity a lot either.


pkgs/test_core/lib/src/runner.dart
```diff
@@ -167,21 +166,12 @@ class Runner {
     var testOn = _config.suiteDefaults.metadata.testOn;
     if (testOn == PlatformSelector.all) return;
 
-    var runtimes = _config.suiteDefaults.runtimes
+    var unsupportedRuntimes = _config.suiteDefaults.runtimes
```
jakemac53, line 169: This is actually just a revert of what I had tried to do earlier, it should match the original logic now. What I was trying to do no longer makes sense with the fallback to a default compiler on any runtime that has no compiler specified.

natebosch (approved):

PR #1912, opened 2023-02-06
fix tests in the compiler branch

jakemac53:
This just fixes the existing tests in preparation for adding new tests.

### DIFF
--- a/pkgs/test_core/lib/src/runner/configuration/args.dart
+++ b/pkgs/test_core/lib/src/runner/configuration/args.dart
@@ -71,19 +71,12 @@ final ArgParser _parser = (() {
   // UI to avoid a painful migration.
   parser.addMultiOption('platform',
       abbr: 'p',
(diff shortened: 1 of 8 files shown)


pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -71,27 +71,20 @@ final ArgParser _parser = (() {
...
-      allowed: [
-        'vm',
-        for (var runtime in allRuntimes) runtime.identifier
-      ],
-      allowedHelp: {
-        'vm': Runtime.vm.description,
-        for (var runtime in allRuntimes)
-          runtime.identifier: runtime.description,
-      });
+      help: 'The platform(s) on which to run the tests.\n'
```
jakemac53, line 74: I had to revert my change to use `allowed` and `allowedHelp`, even though the output is much nicer, we don't know about the custom runtimes at this point.

natebosch (approved):

PR #1914, opened 2023-02-07
Add --compiler tests

jakemac53:


natebosch (approved):
Could we add a test case running the same platform with multiple compilers?

natebosch:
> There is one

Ah, I definitely reviewed a partial diff and not the full thing 😛

PR #1918, opened 2023-02-08
Refactor top level and package README files

devoncarew:
- add issue templates to the repo
- refactor the various readmes

This adds some simple issue templates to the readme, meant to help apply the right labels (package:test or package:checks). I'd also considered assigning a `needs-triage` label in these templates as well.

This PR also refactors the readmes a bit. The main goal was to update the top-level readme to help route people to either package:checks or package:test, now that this repo is not just about the test* packages.

### DIFF
--- a/.github/ISSUE_TEMPLATE/checks.md
+++ b/.github/ISSUE_TEMPLATE/checks.md
@@ -1,5 +1,0 @@
----
-name: "package:checks"
-about: "Create a bug or file a feature request against package:checks."
-labels: "package:checks"
----
--- a/.github/ISSUE_TEMPLATE/checks.md
+++ b/.github/ISSUE_TEMPLATE/checks.md
@@ -1,0 +1,5 @@
+---
+name: "package:checks"
+about: "Create a bug or file a feature request against package:checks."
+labels: "package:checks"
+---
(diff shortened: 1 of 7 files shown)


jakemac53 (commented):
Github won't let me comment on the line for some reason, but in the main README, I would either not mention test_api or test_core at all, or I would make it clear they are not for general use. Those are really implementation details more than anything else, only for use by a select few packages and not general users.

devoncarew:
> Github won't let me comment on the line for some reason, but in the main README, I would either not mention test_api or test_core at all, or I would make it clear they are not for general use. Those are really implementation details more than anything else, only for use by a select few packages and not general users.

Yup, I agree that the text should be updated to call out 'the package you want is `test` - the other two are for internal use'. I think we should still list the packages, for completeness.

@natebosch is thinking about the issue templates for getting feedback from users on package:checks - I'll let him weigh in here wrt whether he want this change or to use different templates.

natebosch:
> @natebosch is thinking about the issue templates for getting feedback from users on package:checks - I'll let him weigh in here wrt whether he want this change or to use different templates.

WDYT about https://github.com/dart-lang/test/pull/1919 ?

devoncarew:
Updated to remove the issue template (in preference to those in #1919) and to tweak the wording in the main readme.

natebosch (approved):

PR #1919, opened 2023-02-08
Add issue templates

natebosch:
Add a separate template for reporting a bug or requesting a feature with
the test runner.
Make the language about `package:checks` specific that the package is in
preview and we are taking any type of feedback.


kevmoo (approved):
LGTM

.github/ISSSUE_TEMPLATE/99_other.md
```diff
@@ -0,0 +1,6 @@
+---
```
devoncarew, line 1: I don't think we need this template; unless you disable it, people will see an 'open generic issue' at the bottom of the list of templates.

.github/ISSSUE_TEMPLATE/01_test_runner_bug.md
```diff
@@ -0,0 +1,14 @@
+---
+name: I found a bug in the test runner
+about: >-
+  Report a bug in running tests, integration with a specific platform, or
+  any behavior of 'package:test'.
+title: ''
+labels: bug
```
devoncarew, line 7: Just for clarity - you prefer not to have a `package:test` label applied here?

devoncarew (approved):
lgtm (though I believe we can remove one of the templates)

.github/ISSSUE_TEMPLATE/01_test_runner_bug.md
natebosch: Yes, I'd be fine adding a `needs-triage` label or something if we are worried about the `bug` label potentially causing it to get lost.

My idea is that the user reports the bug, and the author determines which part of the implementation is relevant.

PR #1932, opened 2023-02-13
Add shorthands for Iterable length checks

passsy:
A list length can be checked with:

```dart
check([]).length.equals(2);
```

It also shows some checks in auto-completion that are not useful and might be missleading.

```dart
check([]).length
  ..isCloseTo(0, 0.001) // always int, never double
  ..isInfinite() // would never return if it would be inifinite
  ..isNegative() // is by design always positive
  ..isNaN() // can't be nan
  ..isNull(); // is never null
```

---

This PR creates shorthand APIs, making it easier to check the length of a list.

```dart
check([1]).hasLength(1);
check([1]).hasLengthGreaterThan(0);
check([1]).hasLengthGreaterThanOrEqualTo(0);
check([]).hasLengthLessThanOrEqualTo(0);
```

This is the first kind of this PRs. @natebosch Please let me know if that's a direction you'd like to see `package:checks` move to

### DIFF
--- a/pkgs/checks/lib/src/extensions/iterable.dart
+++ b/pkgs/checks/lib/src/extensions/iterable.dart
@@ -13,6 +13,74 @@ extension IterableChecks<T> on Subject<Iterable<T>> {
   Subject<T> get last => has((l) => l.last, 'last element');
   Subject<T> get single => has((l) => l.single, 'single element');
 
+  /// Checks that the number of values in [Iterable] is exactly [length]
+  void hasLength(int length) {
+    if (length < 0) {
+      throw ArgumentError('length must be greater than 0');
+    }
+    context.expect(() => ['has length <$length>'], (actual) {
+      if (actual.length == length) return null;
+      return Rejection(
+          which: ['does not have length $length but <${actual.length}>']);
+    });
+  }
+
+  /// Checks that the number of values in [Iterable] is greater than [length]
+  void hasLengthGreaterThan(int length) {
+    if (length < 0) {
+      throw ArgumentError('length must be greater than 0');
+    }
+    context.expect(() => ['has length greater than <$length>'], (actual) {
+      if (actual.length > length) return null;
+      return Rejection(which: [
+        'does not have length greater than $length but <${actual.length}>'
+      ]);
+    });
+  }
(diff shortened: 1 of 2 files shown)


pkgs/checks/lib/src/extensions/iterable.dart
```diff
@@ -13,6 +13,74 @@ extension IterableChecks<T> on Subject<Iterable<T>> {
...
+      ]);
+    });
+  }
+
+  /// Checks that the number of values in [Iterable] is less than [length]
+  void hasLengthLessThan(int length) {
+    if (length < 0) {
+      throw ArgumentError('length must be greater than 0');
+    }
+    context.expect(() => ['has length less than <$length>'], (actual) {
```
passsy, line 61: I was following the current style leading to this error:

```
Expected: a List<dynamic> that:
  has length less than <0>
Actual: []
Which: does not have length less than 0 but <0>
```

But, wouldn't it be easier to understand if the error message would use math symbols?

```
Expected: a List<dynamic> that:
  has length < 0
Actual: []
Which: does not have length < 0 but length == 0
```

kevmoo:
Guessing we won't go with this. There is no end to these types of things – that just blow up the number of options. I'll let @natebosch comment, though!

pkgs/checks/lib/src/extensions/iterable.dart
natebosch: > But, wouldn't it be easier to understand if the error message would use comparison symbols?

I'll take a look at the existing extensions, this does look like an easier to read formatting.

natebosch:
My current thinking is to avoid shipping shorthand methods like this - I think it's better for the core package to focus on the basics, with one way to do things. Other packages could publish extra utilities.

I'll close this for now, because I think it's unlikely we'll want to merge it. 

I'll take a note of this and start some internal discussion though, and I'll keep an eye out for other feedback around shorthand utilities like this. If we come around to a different decision about including shorthand extensions I'll reopen this PR. I suspect we'd probably use include the `hasLength`, since the range comparisons are likely used a lot less for length checks (but they are really handy to use in sample code...)

PR #1939, opened 2023-02-16
Add `because` usage in examples

### LINKED ISSUE #1938: Add a `reason` equivalent
In `package:test`, `expect` takes a `reason` argument which can be used to provide more context as to why a check was expected. For example:

```dart
test('sum', () {
  expect(1 + 1, 2, reason: 'Expected 1+1=2 because math');
});
```

If the check fails, the reason is printed alongside the failure. This can double as a mechanism for documenting tests instead of/in addition to doc comments.

```
00:00 +0 -1: sum [E]                                                                                                                                                                                                             
  Expected: <2>
    Actual: <3>
  Expected 1+1=2 because math
```

natebosch:
Closes #1938

Add a description of `because` in the README and an example using it.

Add a `because` argument to one of the examples in `example.dart`.

Call out the rename from `reason` to `because` in the migration guide.

### DIFF
--- a/pkgs/checks/CHANGELOG.md
+++ b/pkgs/checks/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 0.2.2-dev
 
+-   Add example `because` usage and mention the "reason" name in the migration
+    guide.
+
 ## 0.2.1
 
 -   Add a link to file issues with feedback in the README.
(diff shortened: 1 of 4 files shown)


kevmoo (approved):
Nice!

jakemac53 (approved):

PR #1941, opened 2023-02-16
Add exe compiler, supports running tests compiled to native executables

jakemac53:
Only supported by the `vm` platform, usable with `--compiler exe`. I called it this to match `dart compile exe`.

I also added a test which covers basic behavior for the full matrix of supported runtimes and compilers.

### DIFF
--- a/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
+++ b/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
@@ -113,7 +113,9 @@ void main() {
       },
           skip: compiler == Compiler.dart2wasm
               ? 'Wasm tests are experimental and require special setup'
-              : null);
+              : runtime == Runtime.firefox && Platform.isWindows
+                  ? 'https://github.com/dart-lang/test/issues/1942'
+                  : null);
     }
   }
 }
--- a/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
+++ b/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
@@ -14,7 +14,6 @@ import '../io.dart';
 void main() {
   setUpAll(() async {
     await precompileTestExecutable();
-    await d.file('test.dart', _goodTest).create();
   });
 
   for (var runtime in Runtime.builtIn) {
--- a/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
+++ b/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
@@ -1,0 +1,124 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@TestOn('vm')
+import 'dart:io';
+
+import 'package:test/test.dart';
+import 'package:test_api/backend.dart'; // ignore: deprecated_member_use
+import 'package:test_descriptor/test_descriptor.dart' as d;
+
+import '../io.dart';
+
+void main() {
+  setUpAll(() async {
+    await precompileTestExecutable();
+    await d.file('test.dart', _goodTest).create();
+  });
+
+  for (var runtime in Runtime.builtIn) {
+    // Ignore the platforms we can't run on this OS.
+    if (runtime == Runtime.internetExplorer && !Platform.isWindows ||
+        runtime == Runtime.safari && !Platform.isMacOS) {
+      continue;
+    }
+    group('--runtime ${runtime.identifier}', () {
+      for (var compiler in runtime.supportedCompilers) {
+        group('--compiler ${compiler.identifier}', () {
+          final testArgs = [
+            'test.dart',
+            '-p',
+            runtime.identifier,
+            '-c',
+            compiler.identifier
+          ];
+
+          test('can run passing tests', () async {
+            await d.file('test.dart', _goodTest).create();
+            var test = await runTest(testArgs);
+
+            expect(
+                test.stdout, emitsThrough(contains('+1: All tests passed!')));
+            await test.shouldExit(0);
+          });
+
+          test('fails gracefully for invalid code', () async {
+            await d.file('test.dart', _compileErrorTest).create();
+            var test = await runTest(testArgs);
+
+            expect(
+                test.stdout,
+                containsInOrder([
+                  "Error: A value of type 'String' can't be assigned to a variable of type 'int'.",
+                  "int x = 'hello';",
+                ]));
+
+            await test.shouldExit(1);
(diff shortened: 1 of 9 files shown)


pkgs/test_core/lib/src/bootstrap/vm.dart
```diff
@@ -24,3 +27,30 @@ void internalBootstrapVmTest(Function Function() getMain, SendPort sendPort) {
...
+/// a socket.
+void internalBootstrapNativeTest(
+    Function Function() getMain, List<String> args) async {
+  if (args.length != 2) {
+    throw StateError(
+        'Expected exactly two args, a host and a port, but got $args');
+  }
+  var socket = await Socket.connect(args[0], int.parse(args[1]));
+  var platformChannel = MultiChannel<Object?>(StreamChannel(socket, socket)
+      .cast<List<int>>()
```
jakemac53, line 41: There is probably a better way of doing this, so please feel free to suggest something :), same thing exists in the platform code.

pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -48,23 +48,53 @@ class VMPlatform extends PlatformPlugin {
...
+      var serverSocket = await ServerSocket.bind('localhost', 0);
+      Process process;
+      try {
+        process =
+            await _spawnExecutable(path, suiteConfig.metadata, serverSocket);
+      } catch (error) {
+        serverSocket.close();
+        rethrow;
+      }
+      process.stdout.map(stdout.add);
```
jakemac53, line 64: This roughly mirrors the typical isolate behavior, but lmk if you think we should do something different.

pkgs/test/test/runner/compiler_runtime_matrix_test.dart
```diff
@@ -0,0 +1,123 @@
...
+
+  for (var runtime in Runtime.builtIn) {
+    // Ignore the platforms we can't run on this OS.
+    if (runtime == Runtime.internetExplorer && !Platform.isWindows ||
+        runtime == Runtime.safari && !Platform.isMacOS) {
+      continue;
+    }
+    group('--runtime ${runtime.identifier}', () {
+      for (var compiler in runtime.supportedCompilers) {
+        group('--compiler ${compiler.identifier}', () {
```
natebosch, line 27: [optional] I'm not sure if the second level of `group` is adding value.

Could save 2 spaces of indentation in the tests.

```suggestion
    for (var compiler in runtime.supportedCompilers) {
      group('--runtime ${runtime.identifier} --compiler ${compiler.identifier}', () {
```

pkgs/test_core/lib/src/bootstrap/vm.dart
natebosch, line 41: Hmm, is this happening because we aren't getting upwards inference from the `socket` argument?

What happens if you assign it to a variable so inference has no context?

```suggestion
  var socketChannel = StreamChannel(socket, socket);
  var platformChannel = MultiChannel<Object?>(socketChannel
```

Or maybe fill in the generic explicitly instead of use a `.cast`?

```suggestion
  var platformChannel = MultiChannel<Object?>(StreamChannel<List<int>>(socket, socket)
```

pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -48,23 +48,53 @@ class VMPlatform extends PlatformPlugin {
 
     _setupPauseAfterTests();
 
-    var receivePort = ReceivePort();
+    MultiChannel outerChannel;
+    var cleanupFns = <void Function()>[];
```
natebosch, line 52: [nit]
```suggestion
    var cleanupCallbacks = <void Function()>[];
```

pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -48,23 +48,53 @@ class VMPlatform extends PlatformPlugin {
...
+        process =
+            await _spawnExecutable(path, suiteConfig.metadata, serverSocket);
+      } catch (error) {
+        serverSocket.close();
+        rethrow;
+      }
+      process.stdout.map(stdout.add);
+      process.stderr.map(stderr.add);
+      var socket = await serverSocket.first;
+      outerChannel = MultiChannel<Object?>(StreamChannel(socket, socket)
```
natebosch, line 67: Maybe add a utility for this since it is symmetric?

natebosch (approved):

pkgs/test_core/lib/src/runner/vm/platform.dart
natebosch: By default we capture `print` calls in the exe right? So these would only be the actual `stdout.write` calls from the code under test?

I think this would probably be OK, we could also consider treating these more like they were calls to `print` since we _can_, but I don't know how if the inconsistency outweighs the value of making the JSON reporter always work for exe compiled tests.

jakemac53: Yes prints still are caught and handled as special messages.

It actually looks like the stdout/stderr messages aren't making it through right now though interestingly, ill look into that.

jakemac53: Oh, its because I was using `map` instead of `listen` 🤦 . That is fixed now.

I think the consistency is the most important part here, even though it breaks the json reporter, it is helpful for debugging etc.

pkgs/test_core/lib/src/runner/vm/platform.dart
jakemac53: Done

pkgs/test_core/lib/src/bootstrap/vm.dart
jakemac53: The issue is the underlying stream/sink are still typed with the `Uint8List` generic type. The only way I have figured out how to fix it so far is with the cast as I have it.

jakemac53: (there is no static error in any of these cases, but it is a runtime error without the cast on the stream channel itself)

jakemac53: The specific error I get is `type 'Utf8Decoder' is not a subtype of type 'StreamTransformer<Uint8List, String>' of 'streamTransformer'`

natebosch: I wonder if you can cast only the stream or only the sink argument?

It's not  a big deal either way - if the `cast` works I'm fine with keeping it.

jakemac53: Keeping `cast`

PR #1944, opened 2023-02-19
Add Concurrency Testing section

Nidal-Bakir:
Add an explanation about shard testing

### DIFF
--- a/pkgs/test/README.md
+++ b/pkgs/test/README.md
@@ -8,6 +8,7 @@
 * [Writing Tests](#writing-tests)
 * [Running Tests](#running-tests)
   * [Sharding Tests](#sharding-tests)
+  * [Concurrency Testing](#concurrency-testing)
   * [Shuffling Tests](#shuffling-tests)
   * [Selecting a Test Reporter](#selecting-a-test-reporter)
   * [Collecting Code Coverage](#collecting-code-coverage)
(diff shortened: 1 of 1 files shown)


pkgs/test/README.md
```diff
@@ -233,6 +234,25 @@ dart test --total-shards 3 --shard-index 0 path/to/test.dart
...
+Sharding: This refers to the process of splitting up a large test suite into
+smaller subsets (called shards) that can be run independently. Sharding is
+particularly useful for distributed testing, where multiple machines are used
+to run tests simultaneously. By dividing the test suite into smaller subsets,
+you can run tests in parallel across multiple machines, which can significantly
+reduce the overall testing time.
+
+### Concurrency Testing
+
+Tests run concurrently by default on two threads. use `--concurrency` to control
```
jakemac53, line 246: I believe the default is actually to use half of the cores available on your computer

Nidal-Bakir: From dart test --help:

> -j, --concurrency=<threads>           The number of concurrent test suites run.
>                                       (defaults to "2")

But I do not have a way to confirm that.

jakemac53: https://github.com/dart-lang/test/blob/e56c643950fb06a9d249241715730561447ed24f/pkgs/test/test/runner/runner_test.dart#L51 that message is actually specific to your computer

Nidal-Bakir: That makes sense I will fix that, thanks

jakemac53: Oh sorry that is just the test, here is the actual default https://github.com/dart-lang/test/blob/e56c643950fb06a9d249241715730561447ed24f/pkgs/test_core/lib/src/runner/configuration/values.dart#L14

pkgs/test/README.md
```diff
@@ -233,6 +234,25 @@ dart test --total-shards 3 --shard-index 0 path/to/test.dart
...
+to run tests simultaneously. By dividing the test suite into smaller subsets,
+you can run tests in parallel across multiple machines, which can significantly
+reduce the overall testing time.
+
+### Concurrency Testing
+
+Tests run concurrently by default and it will utilize half of your available cores.
+use `--concurrency` to control the number of tests that runs concurrently,
+meaning that multiple tests can be executed at the same time. For example,
+if you wanted to run the tests on 4 threads, you could run the tests as follows:
```
natebosch, line 249:
```suggestion
### Test concurrency

Test suites run concurrently by default, using half of the host's CPU cores. Use
`--concurrency` to control the number of test suites that runs concurrently,
meaning that multiple tests in independent suites or platforms can run at the
same time. For example, if you wanted to run the tests on 4 threads, you could
run the tests as follows:
```

pkgs/test/README.md
```diff
@@ -233,6 +234,25 @@ dart test --total-shards 3 --shard-index 0 path/to/test.dart
...
+
+```bash
+dart test --concurrency=4
+```
+This can speed up the overall testing process, especially if you have a large
+number of tests.
```
natebosch, line 255:
```suggestion
number of test suites.
```

natebosch (approved):
Added some tweaks

Within a test suite we remain single threaded. The concurrency argument lets us run multiple suites concurrently.

natebosch (approved):

natebosch:
Thanks!

Issue #1928, opened 2023-02-09
Missing docs for Futures


feinstein:
Testing Futures and Streams (outside of the obvious cases) can be challenging. I just found one reference in the docs about Futures, it feels like there should be more.

What will happen to `expectLater`, `expectAsync[0-6]` and `expectAsyncUntil`? The migration docs say to keep using them, so aren't there any plans to migrate them?

natebosch:
The migration docs say to keep using `expectAsync[0-6]` and `expectAsyncUntil` for now.

We're still evaluating whether `expectAsync` will remain in `test`, or whether we will add a separate mechanism here.

`expectLater` is now redundant. `await check(someFuture).completes()` takes the place of `await expectLater(someFuture, completes())` - `check` covers both cases, so there is no need for a separation like `expect` and `expectLater`.

I'll improve the docs around `expectLater` now, and we hope to have an answer soon about `expectAsync` - either improved docs or a new API.

PR #1946, opened 2023-02-22
Add async examples to README

natebosch:
Towards #1928

Add a section in the README about async expectations. Give examples of
awaiting the result, and describe that when the result is not awaited
the test runner is still aware of the work. Describe that some
expectations cannot be awaited.

Describe the reason that `StreamQueue` is used for stream expectations
and give examples of using the `withQueue` and a manual wrapping.

Mention that both `expect` and `expectLater` are migrated to the same
`check` call in the migration guide. Give an example of migrated an
awaited expectation.

### DIFF
--- a/pkgs/checks/README.md
+++ b/pkgs/checks/README.md
@@ -118,6 +118,43 @@ check(
   ]));
 ```
 
+## Asynchronous expectations
+
+Expectation extension methods checking asynchronous behavior return a `Future`.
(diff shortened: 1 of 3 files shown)


kevmoo (approved):
I wish we had logic to validate code examples statically.  *sigh*

Emek8830294 (approved):

Emek8830294 (approved):

PR #1950, opened 2023-02-22
Flush stdin before test

natebosch:
On windows we sometimes see the stdout come with no newline between
`hello` and the test runner output for the first running test. Flush
stdout to see if it more reliably includes the newline.


kevmoo (approved):
🤷

Issue #1465, opened 2021-03-15
Drop the distinction between a test "error" and a test "failure"


natebosch:
There is a little bit of both implementation and user facing API complexity around making a distinction between a test "error" and a test "failure". A test "failure" is the result of specifically the `TestFailure` exception being the error. Any other exception type is an "error" instead. I think the intention is to be able to tell when an `expect` fails (throws this exception) vs something that might indicate a bug in the test (any other exception).

As far as I can tell we don't surface this to the user in any way outside of the JSON reporter API.

https://github.com/dart-lang/test/blob/cced2c4865c4054f40868ee4b67cf23d305d7eb8/pkgs/test_core/lib/src/runner/reporter/json.dart#L259

This goes back to the first time the JSON reporter exists and there isn't a history of the specific use case which needed it. If we can't find evidence that our existing uses of the JSON API are relying on it, I'd like to drop it.

CC Some folks who can hopefully help figure out if this is useful:

@evanweible-wf - I think you may use the JSON API. Can you check if this distinction is meaningful in your use case?
@DanTup - Does the VSCode extension read this field?
@jwren - Do you know if the IntelliJ plugin uses this field, or where we could check?

evanweible-wf:
We are at least using it to differentiate between errors and failures in our xunit reports, but I'm asking internally whether that distinction is actually valuable.

evanweible-wf:
Got confirmation that we do end up treating the errors and failures more-or-less the same; we fail our test runs if any errors or failures are reported. So I believe we'd be unaffected if this were removed.

natebosch:
Thank you for confirming!

> we do end up treating the errors and failures more-or-less the same

That matches my understanding - I don't think I've seen users in general care about this distinction and folks seem to `throw` in a test to indicate failure without thinking any differently about it than a failed `expect`.

DanTup:
> @DanTup - Does the VSCode extension read this field?

It doesn't read `isFailure` and both `failure` and `error` are treated exactly the same (they both map onto the same enum value immediately upon recieving the result). I also don't think users care about the difference - if the test isn't green then it needs fixing, and the failure message is the useful thing, not this distinction.

natebosch:
[...]

PR #1952, opened 2023-02-22
Add a hooks_testing library

natebosch:
Previous test for methods like `expect`, `expectLater`, and
`expectAsync` used internal details of the test runner like `LiveTest`
to check behaviors such as holding the test open and specific failure
behavior.

Add an abstraction `TestCaseMonitor` to hide the implementation details
of creating and running a `LocalTest`. Expose only the parts of
`LiveTest` which are used by current tests - the status, result and
errors.
The new class is defined in a new library `hooks_testing` since it is
intended primarily for writing tests for code that uses the `hooks`
library.

Migrate tests under `test_api/test/frontend` for the tests which will be
migrating to `package:matcher` to use the new APIs. Move to a
`utils_new.dart` file which will be moved along with these tests, and
the old `utils.dart` file will remain for testing the scaffolding APIs.
Do not replace the `expectTestsBlock` utility, the usage of this utility
is not any more clear or readable than spelling out the full behavior in
the test.

Add tests for the behavior of holding the test for pending work, and for
unawaited failures.
Add utilities for testing against a `TestMonitor` locally within the
test.

### DIFF
--- a/pkgs/checks/test/context_test.dart
+++ b/pkgs/checks/test/context_test.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 
 import 'package:async/async.dart' hide Result;
 import 'package:checks/checks.dart';
@@ -59,7 +60,7 @@ void main() {
           callback = completer.complete;
           await completer.future;
           return Extracted.value(null);
-        }, null);
+        }, it());
       });
       await pumpEventQueue();
       check(monitor).state.equals(State.running);
@@ -111,7 +112,7 @@ void main() {
         check(null).context.expectUnawaited(() => [''], (actual, reject) {
           final completer = Completer<void>()
             ..future.then((_) {
-              reject(Rejection(which: ['foo']));
+              reject(Rejection(which: ['']));
             });
           callback = completer.complete;
         });
@@ -120,34 +121,16 @@ void main() {
       callback();
       await pumpEventQueue();
       check(monitor)
-        ..state.equals(State.failed)
-        ..errors.deepEquals([
-          it()
-            ..isA<TestFailure>()
-                .has((f) => f.message, 'message')
-                .isNotNull()
-                .endsWith('Which: foo'),
-          it()
-            ..isA<String>()
-                .startsWith('This test failed after it had already completed.')
-        ]);
-    });
-  });
-
-  group('SkipExtension', () {
-    test('marks the test as skipped', () async {
-      final monitor = await TestCaseMonitor.run(() {
-        check(null).skip('skip').isNotNull();
-      });
-      check(monitor).state.equals(State.skipped);
+          .didFail(Result.error)
+          .contains('This test failed after it had already completed.');
     });
   });
 }
 
 extension _MonitorChecks on Subject<TestCaseMonitor> {
   Subject<State> get state => has((m) => m.state, 'state');
-  Subject<Iterable<Object>> get errors => has((m) => m.errors, 'errors');
-  Subject<StreamQueue<Object>> get onError =>
+  Subject<Iterable<AsyncError>> get errors => has((m) => m.errors, 'errors');
+  Subject<StreamQueue<AsyncError>> get onError =>
       has((m) => m.onError, 'onError').withQueue;
 
   /// Expects that the monitored test is completed as success with no errors.
@@ -161,9 +144,25 @@ extension _MonitorChecks on Subject<TestCaseMonitor> {
         (actual, reject) async {
       await for (var error in actual.rest) {
         reject(Rejection(which: [
-          ...prefixFirst('threw late error', literal(error)),
+          ...prefixFirst('threw late error', literal(error.error)),
+          ...(const LineSplitter().convert(
+              TestHandle.current.formatStackTrace(error.stackTrace).toString()))
         ]));
       }
     });
+  }
+
+  /// Expects that the monitored test is completed as [expectedResult] with
+  /// exactly 1 TestFailure.
+  ///
+  /// Returns the message from the failure.
+  Subject<String> didFail(Result expectedResult) {
+    state.equals(State.failed);
+    return errors.single
+        .isA<AsyncError>()
+        .has((a) => a.error, 'error')
+        .isA<TestFailure>()
+        .has((f) => f.message, 'message')
+        .isNotNull();
   }
 }
--- a/pkgs/checks/test/context_test.dart
+++ b/pkgs/checks/test/context_test.dart
@@ -3,7 +3,6 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:convert';
 
 import 'package:async/async.dart' hide Result;
 import 'package:checks/checks.dart';
@@ -39,7 +38,7 @@ void main() {
           callback = completer.complete;
           await completer.future;
           return Extracted.value(null);
-        }, it());
+        }, null);
       });
       await pumpEventQueue();
       check(monitor).state.equals(State.running);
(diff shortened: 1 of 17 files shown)


pkgs/checks/test/context_test.dart
```diff
@@ -0,0 +1,115 @@
...
+
+  /// Expects that the monitored test is completed as success with no errors.
+  ///
+  /// Sets up an unawaited expectation that the test does not emit errors in the
+  /// future in addition to checking there have been no errors yet.
+  void didPass() {
+    errors.isEmpty();
+    status.equals(Status.complete);
+    result.equals(Result.success);
+    onError.context.expectUnawaited(() => ['emits no further errors'],
```
natebosch, line 89: `neverEmits` expects the stream to close.

I wonder if we should change that and have you write `check(stream).allOf([it()..neverEmits(condition), it()..isDone()]);`

pkgs/test_api/lib/hooks_testing.dart
```diff
@@ -0,0 +1,130 @@
...
+  ///
+  /// A test that is still running may have a result of [Result.success] because
+  /// it has not failed _yet_. The result should only be read after the test is
+  /// done.
+  ///
+  /// {@macro result-late-fail}
+  ///
+  /// A failed test my be a [Result.failure] if the test failed with a
+  /// [TestFailure] exception, or a [Result.error] if it failed with any other
+  /// type of exception.
```
natebosch, line 101: I'm not sure I want to expose this... I'm tempted to create a new abstraction for this...

We want to drop the distinction, but have not gotten around to trying it and seeing if anything breaks.
https://github.com/dart-lang/test/issues/1465

I think it might be better to introduce this API with some other model instead of building more on top of it...

pkgs/test_api/lib/hooks_testing.dart
```diff
@@ -0,0 +1,156 @@
...
+
+  /// The errors surfaced by the test.
+  ///
+  /// A test with any errors will have a failing [result].
+  ///
+  /// {@macro result-late-fail}
+  ///
+  /// A test may have more than one error if there were unhandled asynchronous
+  /// errors surfaced after the test is done.
+  Iterable<Object> get errors => _liveTest.errors.map((a) => a.error);
```
natebosch, line 109: @jakemac53 @grouma - the only current usage for looking at the `StackTrace` from the errors is to forward them so they are printed for unexpected error.

https://github.com/dart-lang/test/blob/ba8e43afa183e321a482efdea06e80d07d06c0dc/pkgs/test_api/test/utils.dart#L115

I don't know if that is worthwhile enough to add an indirection for the vast majority of uses cases that don't care about the stack trace. If our own tests in `matcher` or `checks` are failing with unexpected exception the stack trace could end up being helpful though. In the end I landed on not exposing the stack trace, so that we don't need to worry about returning a class from `package:async`.

WDYT?

pkgs/test_api/lib/hooks_testing.dart
```diff
@@ -0,0 +1,156 @@
...
+  var suite = Suite(Group.root([test]), _suitePlatform, ignoreTimeouts: false);
+  return test.load(suite);
+}
+
+/// A dummy suite platform to use for testing suites.
+final _suitePlatform =
+    SuitePlatform(Runtime.vm, compiler: Runtime.vm.defaultCompiler);
+
+/// The running and success state of a test monitored by a [TestCaseMonitor].
+enum State {
```
natebosch, line 130: @grouma @jakemac53 - WDYT about creating this new abstraction to hide some of the details of `Result` and `Status`?

I don't like the idea of introducing definitions for the term `State`, but I also don't like the idea of exposing the messy details when we could have a much simpler model for users.

pkgs/test_api/lib/hooks_testing.dart
grouma: > If our own tests in matcher or checks are failing with unexpected exception the stack trace could end up being helpful though

How so?

pkgs/test_api/lib/hooks_testing.dart
grouma: > but I also don't like the idea of exposing the messy details

What are the messy details? It looks like `State` is just a merge of `Result` and `Status`.

pkgs/test_api/lib/hooks_testing.dart
natebosch: It doesn't make any difference for passing tests. If we have some bug in a test or in library code that causes an unexpected error within the `liveTest` body you'd see a difference in the output.  Currently we'd include a stack trace for the error, and it would show up like any other test failure pointing directly to the cause.

If we strip the `StackTrace` from this view, then when we forward the errors we won't see them.

grouma: That seems valuable imo and worth keeping the `StackTrace` around.

pkgs/test_api/lib/hooks_testing.dart
natebosch: There is more behavior to document if we expose the `Result` and `Status`.
- We'd have to talk about how `Result` doesn't have meaning until `Status` is `Status.complete`.
- We'd have to talk about how users of these APIs will never see a `Status.pending`.
- We'd have to talk about, and carry forward, the "error" vs "failure" distinction which we'd like to move away from. https://github.com/dart-lang/test/issues/1465 

We have an existing `State` class which holds a `Result` and `Status`, but I think it's friendlier to model is as the single enum instead of exposing both.

Overall, it does feel weird to have 2 enums which are not orthogonal. When the value of one field is interesting, the other field can only have 1 value, and vice versa.

grouma: Ah yes. Your proposed changes definitely help and LGTM.

pkgs/test_api/lib/hooks_testing.dart
jakemac53: Yeah I think it would be beneficial to keep around the stack trace. This api shouldn't be used by most people anyways right? So it shouldn't be much of a pain that it is exposing something from package:async.

The biggest risk I see is people may rely on inference to get the type and not have a dependency on async themselves, exposing their package to potential breakage, but that is something I can live with (and is really just a general issue).

natebosch: Fair enough. I restored the stack trace. I have found myself adding more stack trace outputs in `checks` as I go, so that's a pretty good indication it's worthwhile here too.

pkgs/test_api/lib/hooks_testing.dart
```diff
@@ -0,0 +1,156 @@
...
+/// A monitor for the behavior of a callback when it is run as the body of a
+/// test case.
+///
+/// Allows running a callback as the body of a local test case and querying for
+/// the current [state], and [errors], and subscribing to future errors.
+///
+/// Use [run] to run a test body and query for the success or failure.
+///
+/// Use [start] to start a test and query for whether it has finished running.
+class TestCaseMonitor {
```
grouma, line 28: Why do we want this / need this to be package public?

natebosch: The tests that are currently in `test_api` will be moving to `package:matcher`.

This class is also used from `package:checks`.

pkgs/checks/test/context_test.dart
jakemac53: I think `neverEmits` should expect the stream to close, otherwise how long should it wait for? If you want to fail asynchronously then `expectUnawaited` is there for you.

natebosch: Yeah, the case for expecting the stream to close is more common. I don't like having to use the context directly like this, and having the expectations be orthogonal is kind of nice, but there is a strong argument for making the very common case the only implementation we provide. I'll definitely wait for a separate pr if we make any change around this, was just taking a note here that the API did cause a problem at this usage site.

jakemac53 (approved):

PR #1954, opened 2023-02-23
Avoid exceptions for iterable getters

natebosch:
The `has` utility is intended for reading fields that cannot throw. For
`single`, `first`, and `last`, which may throw for unexpected numbers of
elements, implement with `context.next` and test with specific
rejections for the length.


natebosch:
> No tests?

Good call, added tests for the new rejections.

kevmoo (approved):

PR #1956, opened 2023-02-23
Drop the Condition abstraction

natebosch:
Drop the `Condition` interface and replace with callbacks. The `it()`
utility can cause confusion and look like a language feature. Another
potential benefit of function types is they may get syntax or formatting
improvements in the future.

One downside of the change is when inference fails, for instance when
passing a callback in a collection for `deepEquals`, there is no static
help. At runtime there will be errors about `Subject` not having
instance methods matching the extensions that are used.

One upside is some linting setups may be able to statically help avoid
passing a `Future<void> Function(Subject)` where a synchronous condition
is required. This static checking couldn't work for the `it()` syntax.

Tests are mostly migrated to use `it` as the subject name for
convenience of the migration, but tests in the real world (and in the
example code in docs) may be more likely to choose a more meaningful
letter, or a longer name based on the value.

Add special handling in `literal` for `void Function(Subject)` instances
since there is no longer a meaningful `toString()` implementation on the
instances.


pkgs/checks/test/extensions/collection_equality_test.dart
```diff
@@ -51,30 +51,37 @@ void main() {
         'a',
         'b'
       ], [
-        it()
-          ..isA<String>().which(it()
-            ..startsWith('a')
-            ..length.isLessThan(2)),
-        it()..isA<String>().startsWith('b')
+        (Subject<dynamic> it) => it.isA<String>().which((it) => it
```
natebosch, line 54: If you forget to put the argument type, then `it` is `dynamic`, and there is no static warning.

At runtime, the `.isA` call is made on the `Subject` instance because it was not resolved statically to the extension method.

natebosch:
I think we should go forward with this change. I haven't seen enough feedback in support of `Condition` so I think the main person who still prefers it is me.

pkgs/checks/test/extensions/collection_equality_test.dart
jakemac53: That does seem pretty annoying, but I will leave the decision up to you.

pkgs/checks/lib/src/checks.dart
```diff
@@ -97,7 +94,7 @@ Subject<T> check<T>(T value, {String? because}) => Subject._(_TestContext._root(
 ///
 /// Asynchronous expectations are not allowed in [condition] and will cause a
 /// runtime error if they are used.
-CheckFailure? softCheck<T>(T value, Condition<T> condition) {
+CheckFailure? softCheck<T>(T value, void Function(Subject<T>) condition) {
```
jakemac53, line 97: The function type here shows up a lot, I would consider a typedef (or two)

jakemac53 (commented):
Is there any way we can help people when they forget to add the argument type?

I do forsee that as being a consistent issue and the error is probably pretty nasty for most people to debug.

natebosch:
> Is there any way we can help people when they forget to add the argument type?

I suspect we could write a custom lint for it - maybe doing that and including it in the recommended or core set would be sufficient?

The lint would fire anytime you pass a collection literal to a relevant `package:checks` API and it contains any callback that doesn't take `Subject` as an argument. I can't think of a valid use case for a non-subject callback since `operator ==` wouldn't do anything useful with them, so I don't think it should have false positives.

pkgs/checks/lib/src/checks.dart
natebosch: Done

pkgs/checks/README.md
```diff
@@ -114,8 +114,9 @@ check(someValue)
 ### Passing a set of expectations as an argument
 
 Some expectations take arguments which are themselves expectations to apply to
-other values. These expectations take `void Function(Subject)` arguments, which
-check expectations when they called with a `Subject` argument.
+other values. These expectations take `Condition` arguments which have the
+signature void Function(Subject)`. THe conditions check expectations when they
```
jakemac53, line 118:
```suggestion
signature void Function(Subject)`. The conditions check expectations when they
```

pkgs/checks/README.md
```diff
@@ -114,8 +114,9 @@ check(someValue)
...
+called with a `Subject` argument.
```
jakemac53, line 119:
```suggestion
are called with a `Subject` argument.
```

jakemac53 (approved):

PR #1957, opened 2023-02-23
Fix timeouts when async predicate throws

natebosch:
A predicate or extraction callback is not supposed to throw - it should
return a Rejection to indicate a failure - however if an exception is
thrown the test should fail for that exception and immediately be done.

Move the `outstandingWork.complete()` call to a `finally` block so it
is never skipped. Move it after the nested condition is applied to also
fix a potential bug if the `Condition` is a callback that does any async
work outside calls to the context. This isn't a typical pattern, but
could become more typical if we drop the `Condition` abstraction.

### DIFF
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -727,11 +727,14 @@ class _TestContext<T> implements Context<T>, _ClauseDescription {
     }
     _clauses.add(_ExpectationClause(clause));
     final outstandingWork = TestHandle.current.markPending();
-    final rejection = await _value.apply(
-        (actual) async => (await predicate(actual))?._fillActual(actual));
-    outstandingWork.complete();
-    if (rejection == null) return;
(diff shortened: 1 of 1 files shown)


natebosch:
This behavior is not easy to test currently. I have written tests for this specific bug fix, but I will wait to land them until after I land a new library for testing this type of code.

Here is the commit with the tests:

https://github.com/dart-lang/test/pull/1952/commits/7810c06f7a4e7560347a1e71b73e6d11ed3d75ac

natebosch:
This behavior wouldn't have been visible (except if you did something like `try{softCheck(somethingThatThrows);}catch{}`) before https://github.com/dart-lang/test/pull/1815 landed. After that - the timeout after the exception has already failed the test is quite noticeable.

natebosch:
I found these timeouts when I was working on https://github.com/dart-lang/test/pull/1956

If we adopt that pattern then exceptions in the callbacks will be significantly more likely.

PR #1958, opened 2023-02-23
Condition must have dynamic generic in deepEquals

natebosch:
Mention in the doc comment that for deep collection equality checks the
condition must always be `Condition<dynamic>`, and suggest using `isA`
to add expectations for the specific type. This differs from
`Iterable.containsInOrder` which allow a condition on the element type.


natebosch:
I noticed this was not explicitly called out when working on https://github.com/dart-lang/test/pull/1956

kevmoo (approved):

PR #1959, opened 2023-02-23
[checks] Fix isNotEmpty rejection

passsy:
Saw this error which didn't make sense:

```
Expected: a DownloadFolder that:
  has contents that:
    is not empty
Actual: a DownloadFolder that:
  has contents that:
  Actual: []
  Which: is not empty
```

This is a copy-paste oversight from `isEmpty`. The rejection should be inverted (as it is for `isEmpty`)!


natebosch (approved):
Thanks for catching this!

PR #1960, opened 2023-02-24
Add ComparableChecks extension

### LINKED ISSUE #1955: More `isLessThan`, `isGreaterThan`, etc. extension methods would be nice
Previously with `package:matcher` I could compare `Duration`s with `expect(duration1, lessThan(duration2))` since `Duration` provides an `operator <`.

It'd be nice if `package:checks` provided that too.  It'd probably be easiest to just provide such extension methods on `Subject<Comparable>`.

natebosch:
Closes #1955

Replaces the extension on `num` for `isGreaterThan`, `isLessThan`,
`isGreaterOrEqual`, and `isLessOrEqual` using the `Comparable` interface
instead of `operator >`.


kevmoo (approved):
LOVE IT!

pkgs/checks/lib/src/extensions/math.dart
```diff
@@ -86,3 +86,47 @@ extension NumChecks on Subject<num> {
     });
   }
 }
+
+extension ComparableChecks<T> on Subject<Comparable<T>> {
```
kevmoo, line 90: Meh...it's almost worth it's only library. It's more than math. But it's not a biggy. 🤷

kevmoo: Core actually feels morea appropriate

natebosch: Sure, I'll leave it back in `core.dart` for now. It doesn't make a user facing difference, so we can shuffle things later if we want.

PR #1965, opened 2023-03-06
Retry filesystem deletes

natebosch:
We frequently see CI failures with exceptions pointing to various
deletes which fail because of processes still having the file or
directory open. Add a synchronous retry to see if it helps at all.

It may become necessary to make it asynchronous and add a small backoff.

### DIFF
--- a/pkgs/test_core/lib/src/util/io.dart
+++ b/pkgs/test_core/lib/src/util/io.dart
@@ -112,7 +112,7 @@ Future withTempDir(Future Function(String) fn) {
   return Future.sync(() {
     var tempDir = createTempDir();
     return Future.sync(() => fn(tempDir))
-        .whenComplete(() => Directory(tempDir).deleteSync(recursive: true));
+        .whenComplete(() => Directory(tempDir).deleteWithRetry());
   });
 }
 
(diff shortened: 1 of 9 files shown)


natebosch:
This made 3 rounds of the CI without any flakes. I think it's worth landing, and if we see further flakes we can add in async backoff.

pkgs/test_core/lib/src/util/io.dart
```diff
@@ -226,3 +226,16 @@ Future<Uri> getRemoteDebuggerUrl(Uri base) async {
...
 }
+
+extension RetryDelete on FileSystemEntity {
+  void deleteWithRetry() {
+    for (var i = 0; i < 3; i++) {
+      try {
+        deleteSync(recursive: true);
+        return;
+      } on FileSystemException {
+        // Ignore and retry
```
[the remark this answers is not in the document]
natebosch: 🤦 that certainly would explain why we don't see deletion failures during CI

pkgs/test_core/lib/src/util/io.dart
```diff
@@ -226,3 +227,17 @@ Future<Uri> getRemoteDebuggerUrl(Uri base) async {
     return base;
   }
 }
+
+extension RetryDelete on FileSystemEntity {
+  Future<void> deleteWithRetry() async {
+    for (var i = 0; i < 3; i++) {
```
jakemac53, line 233: nit: consider a while(true) instead and use the return/rethrow to exit (increment i inside the loop). That would keep the logic for the number of retries in one place instead of 2, slightly less error prone imo.

jakemac53 (approved):
🚀

pkgs/test_core/lib/src/util/io.dart
natebosch: Done.

PR #1971, opened 2023-03-13
Port dom shim to js types.

joshualitt:


pkgs/test/lib/src/runner/browser/dom.dart
```diff
@@ -2,29 +2,74 @@
...
 
+import 'dart:js_interop';
 import 'dart:js_util' as js_util;
 
 import 'package:js/js.dart';
 
+// Conversion functions to help with the migration.
+extension ObjectToJSAnyExtension on Object? {
+  JSAny? get toJSAnyShallow {
+    if (const bool.fromEnvironment('dart.library.html')) {
```
jakemac53, line 13: is this basically detecting wasm vs js? Maybe add a comment to that effect just to clarify (my initial reaction was this wasn't necessary, because we are always running on the web, but wasm is also web which is easy to forget :P ).

pkgs/test/lib/src/runner/browser/dom.dart
```diff
@@ -2,29 +2,74 @@
...
+      // Cast is necessary on Wasm backends and in the future on JS backends as
+      // well.
+      // ignore: unnecessary_cast
+      return this as JSAny?;
```
jakemac53, line 17: Is there an issue that should be filed? Will this warn in the compilers as well?

pkgs/test/lib/src/runner/browser/dom.dart
```diff
@@ -2,29 +2,74 @@
...
+    } else {
+      return toJSAnyDeep;
```
jakemac53, line 19: this seems a bit weird for a function called `toJsAnyShallow` haha, lets at least put a comment on the function mentioning that it might actually be deep.

pkgs/test/lib/src/runner/browser/dom.dart
```diff
@@ -2,29 +2,74 @@
...
+
+  JSAny? get toJSAnyDeep => js_util.jsify(this) as JSAny?;
+}
+
+extension JSAnyToObjectExtension on JSAny? {
+  Object? get toObjectShallow {
+    if (const bool.fromEnvironment('dart.library.html')) {
+      return this;
+    } else {
+      return toObjectDeep;
```
jakemac53, line 31: same as above, lets comment about this potential deep behavior.

pkgs/test/lib/src/runner/browser/dom.dart
```diff
@@ -2,29 +2,74 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:js_interop';
 import 'dart:js_util' as js_util;
 
 import 'package:js/js.dart';
 
+// Conversion functions to help with the migration.
+extension ObjectToJSAnyExtension on Object? {
```
jakemac53, line 11: I think these extensions could be on the non-nullable version, which would change use cases for nullable variables to be like `foo!.toJsAnyShallow` instead of requiring all cases to do `foo.toJsAnyShallow!` even when `foo` is not null and it is guaranteed to return a value.

jakemac53:
I think we may also need to update the generated javascript by running `dart compile js tool/host.dart -o lib/src/runner/browser/static/host.dart.js` from the `pkgs/test` dir.

pkgs/test/lib/src/runner/browser/dom.dart
```diff
@@ -99,61 +161,99 @@ extension NodeExtension on Node {
 class EventTarget {}
 
 extension EventTargetExtension on EventTarget {
-  void addEventListener(String type, EventListener? listener,
+  @JS('addEventListener')
+  external JSVoid _addEventListener1(JSString type, JSFunction? listener);
+  @JS('addEventListener')
+  external JSVoid _addEventListener2(JSString type, JSFunction? listener,
+      JSBoolean useCapture);
```
natebosch, line 168: Are JS interop definitions allowed to use optional arguments?

pkgs/test/lib/src/runner/browser/dom.dart
joshualitt: This cast will no longer be necessary after https://dart-review.googlesource.com/c/sdk/+/283742 lands. However, that CL is currently blocked on front-end work.

The root problem is that JS types are reified on Wasm backends, but not yet reified on JS backends. On JS backends, JS types are just typedefs for Dart types, hence the cast warning.

It may be that this CL lands after the reification CL, so we may not need this cast after all. If not, we can file an issue to track this if you like. Thoughts?

pkgs/test/lib/src/runner/browser/dom.dart
joshualitt: Yes, JS interop definitions can use optional arguments. And they have JS semantics, that is not-passed is `undefined`. However, this means that these definitions do not have Dart semantics, i.e. default values, and they also have inconsistent semantics for tearoffs. I'm hoping we disallow tearoffs entirely, and instead users who want to tearoff these methods should wrap them with Dart functions as I do in this CL.

We don't have that check just yet though, so this seems like a safer path for now. If you prefer, we can use optional methods, but it may not be worth the possible pain because this shim is going to go away in the near future to be replaced by `package:web` anyways, so it's your call.

Thoughts?

natebosch:
@joshualitt - is this change worth pursuing?

joshualitt:
@natebosch we still want this change. We originally held off on this change because we couldn't bump the SDK dependency. As soon as we can bump the dependency to 3.1, we should try and land this CL.

For the time being, I've returned this PR to draft status and we can revisit in a few months.

Issue #1972, opened 2023-03-18
Failed to run tests on stable Dart SDK


AlexV525:
- https://github.com/dart-lang/sdk/issues/51774
- https://github.com/cfug/dio/actions/runs/4452707510/jobs/7820608359?pr=1738#logs

AlexV525:
Closing as `js: 0.6.8` has been retracted.

PR #1973, opened 2023-03-18
Fix `package:js` version constraint.

joshualitt:
#1972


joshualitt:
We actually need to retract the publish

PR #1974, opened 2023-03-20
Prepare to publish

natebosch:
Remove dependency overrides for matcher.
Drop `-dev` version suffix.


jakemac53:
> Do we need that published soon? It looks like we mainly want to land that so we can get it in the SDK, which we don't need to publish for.

No idea tbh, just wanted to mention it.

PR #1975, opened 2023-03-21
Support native assets

dcharkes:
The Dart VM expects a `native_assets.yaml` next to `packages_config.json` to contain the mapping for `@FfiNative` and `@Native` FFI functions when running  a script or spawning an isolate.

When a program compiles to kernel eagerly (before invoking the Dart VM), it should add that `native_assets.yaml` to the kernel compilation invocation.

`package:test` does manual compilation to kernel, so it needs to provide this mapping.

After https://dart-review.googlesource.com/c/sdk/+/267340, the `native_assets.yaml` will be created by `dartdev` when doing `dart test` or `dart run test`.

I'm not entirely sure how to test this PR.

I can't land the dartdev logic in https://dart-review.googlesource.com/c/sdk/+/267340 without doing dependency overrides for package:test and package:test_core, but I don't want to land dependency overrides pointing to non-merged code in the SDK.

I can't easily add a unit test that compiles a dynamic library and adds a `native_assets.yaml` manually on this repo. I would need to be able to invoke `bin/test.dart` instead of `dart run test:test` from an example project that provides a dylib. But package:test doesn't seem to support that:

```
dacoharkes@dacoharkes-linux:~/src/dart-lang/test/pkgs/test/test/runner/native_assets/test_projects/native_add$ dart ../../../../../bin/test.dart
00:01 +0 -1: loading test/native_add_test.dart [E]                                                                                      
  Failed to load "test/native_add_test.dart":
  Couldn't resolve the package 'native_add' in 'package:native_add/native_add.dart'.
  test/native_add_test.dart:5:8: Error: Not found: 'package:native_add/native_add.dart'
  import 'package:native_add/native_add.dart';
```

I know it works locally with a patched package:test and patched dartdev.

Any suggestions for how to test this?

(For reference: the unit tests in the VM that get a manually provided `native_assets.yaml`: https://dart-review.googlesource.com/c/sdk/+/264842)


pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -159,13 +159,23 @@ class _TestCompilerForLanguageVersion {
     final platformDill = 'lib/_internal/vm_platform_strong.dill';
     final sdkRoot =
         p.relative(p.dirname(p.dirname(Platform.resolvedExecutable)));
+    final packageConfigUriAwaited = await packageConfigUri;
+
+    // If we have native assets for the host os in JIT mode, they are here.
+    Uri? nativeAssetsYaml =
+        packageConfigUriAwaited.resolve('native_assets.yaml');
```
jakemac53, line 166: Should this be a configurable path? Or just wait to allow that until/if it is needed later?

The package config isn't always necessarily what you think it is... tools could copy it to a new location and modify it before invoking us etc.

dcharkes: The use case I'm thinking about is `dart test` which is equal to `dart run test:test`. In this case `dart run`/`dart test` will run themselves in JIT mode use the `.dart_tools/package_config.json` and `.dart_tools/native_assets.yaml`.

(`dart test/foo_test.dart` already works after `.dart_tools/package_config.json` and `.dart_tools/native_assets.yaml` are present. Basically running a test that way skips the whole runner infrastructure.)

> tools could [...]

I hope these tools would then still run `dart pub get` + `dart test`/`dart run test` which would then result in the same layout. Are you aware of tools that do different things?

jakemac53: > I hope these tools would then still run dart pub get + dart test/dart run test which would then result in the same layout. Are you aware of tools that do different things?

I am not sure if it is still the case but for a long time flutter used to mess with the package config _after_ pub get, to inject some references to generated packages. It did cause problems for build_runner etc or other tools that expect the `.dart_tool/package_config.json` file to be the source of truth, or the `pubspec.yaml` or `pubspec.lock` files to agree with the `package_config.json` file (which was the case with build_runner).

Also as soon as you step outside of the pub ecosystem all these assumptions no longer apply.

dcharkes: Yes, for Flutter I have implementations for this in `flutter_tools`. (It does cross-compiling, Dart doesn't. It bundles libraries differently per target OS as well, so it also transforms the paths where libraries are found for the target bundles.)

jakemac53: Just a random thought here, but would it make sense to add a new (optional) field to the `package_config.json` which lists the location of the `native_assets.yaml` file? That way we need to changes to this package at all, and internally it should also be more straightforward.

jakemac53:
> I can't easily add a unit test that compiles a dynamic library and adds a `native_assets.yaml` manually on this repo. I would need to be able to invoke `bin/test.dart` instead of `dart run test:test` from an example project that provides a dylib. But package:test doesn't seem to support that:

Why do you need to invoke `bin/test.dart`?

My guess is that you haven't ran a `dart pub get` in that example directory, it seems like it is loading the package config from package:test and not from the new package you created.

Ultimately though I would add this test to the `integration_tests` directory, that is where we store this style of test, and it won't accidentally pick up package configs from the package it is nested under.

dcharkes:
> Why do you need to invoke `bin/test.dart`?

`dart test` will do the native assets mapping once we land support in the Dart SDK, but I can't test `dart test` in the Dart SDK without this fix in `package:test`.

Before landing native assets mapping in the Dart SDK for `dart test`, `dart test` will just ignore native assets, so no native assets will be provided. The only workaround is to not run `dart test` but `dart path/to/a/script.dart`, as the VM (not `dartdev`) is configured to pick up `native_assets.yaml`. The `path/to/a/script.dart` should not be the `test/my_test.dart` because then we don't exercise the `package:test` runner. I want to invoke te runner so that we exercise the code path that is modified in this PR, but not via `dart test` or `dart run` because that requires us to land the Dart SDK changes first.

I need to either land this PR first and roll it into the SDK and add a test there. And later a test here once an SDK has been released with `dart test` changes also add a test here so we test with `dart test`.

> integration_tests

Here is what the test would look like once `dart test` is providing the mapping in the Dart SDK:

https://github.com/dart-lang/test/compare/native-assets-test

jakemac53:
> `dart test` will do the native assets mapping once we land support in the Dart SDK, but I can't test `dart test` in the Dart SDK without this fix in `package:test`.

package:test itself does not require any native assets, so this shouldn't matter right?

> I need to land this PR first and roll it into the SDK and add a test there. And later a test here once an SDK has been released with `dart test` changes also add a test here so we test with `dart test`. (Or somehow do `bin/test.dart` as described above.)

Ok so maybe this explains things for me, is the issue that the `native_assets.yaml` file won't exist at all in any current sdks? Or the functionality to read it doesn't exist?

jakemac53:
I am just going to check out your branch and see if I can get things working locally :)

jakemac53:
I tried building the latest sdk and running `dart pub get`, `dart run test/native_add_test.dart`, and `dart compile exe native_add_test.dart` and none of them created the native_assets.yaml file?

I wanted to try just checking in this file into the integration test, at least until the SDK can actually build it as a part of `dart test`?

dcharkes:
> I am hesitant to land anything that is enabled by default but not tested though (on CI). I wouldn't want a new SDK to drop that creates this file but then the way we use it is somehow broken and people get broken.

I the CL that would start creating this file, would also land a unit test that exercises `dart test`. The tests are here: https://dart-review.googlesource.com/c/sdk/+/267340/40/pkg/dartdev/test/native_assets/test_test.dart. They are currently skipped. When this lands, I can unskip them in that CL.

> Could we guard this code behind an environment variable so that you can enable it explicitly in the SDK tests? I am not sure if our CI setup there supports setting an environment variable for a test.

That could work, as I'm invoking `dart test` manually with `Process.run` in the unit tests.

I think a better solution might be to gate all native assets logic in the SDK behind an experimental flag, that way the `native_assets.yaml` file will not be produced if the user doesn't opt in.

jakemac53:
Experiments are supported at least to some extent, IIRC you need to pass them before the `test` command, so like `dart --enable-experiment=some-experiment test`?

jakemac53:
Would the file potentially still exist but with stale contents though if somebody ran with the experiment once and then without it? We could parse the VM arguments to see if this experiment is enabled though and guard the logic behind that.

jakemac53:
See https://github.com/dart-lang/test/blob/master/pkgs/test_core/lib/src/util/dart.dart#L75

dcharkes:
> Experiments are supported at least to some extent, IIRC you need to pass them before the `test` command, so like `dart --enable-experiment=some-experiment test`?

Good catch. `dart compile` [takes experiment flags after the compile](https://dart-review.googlesource.com/c/sdk/+/169902), due to being able to compile for a different experiment set than the `dart` invocation. `dart run` can [forward experiment flags](https://dart-review.googlesource.com/c/sdk/+/245202) to the script being invoked. However, for `dart run` / `dart test` behavior itself we should use the current `dart` experiments.

>Would the file potentially still exist but with stale contents though if somebody ran with the experiment once and then without it? We could parse the VM arguments to see if this experiment is enabled though and guard the logic behind that.
> See https://github.com/dart-lang/test/blob/master/pkgs/test_core/lib/src/util/dart.dart#L75

Perfect! That was indeed an issue.

https://dart-review.googlesource.com/c/sdk/+/267340/42 with https://github.com/dart-lang/test/tree/native-assets-test
```
$ dart-lang/test/integration_tests/native_add$ dart test
[fail]
$ dart-lang/test/integration_tests/native_add$ dart --enable-experiment=native-assets test
[success]
$ dart-lang/test/integration_tests/native_add$ dart test
[fail]
```

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
jonasfj: > I believe the pub authors don't like us modifying package_config.json after pub created that file.

Yeah, ideally it's best not to do this. It messes up timestamps. We can add support for it -- but I would suggest just writing it next to `package_config.json` in `.dart_tool/native_assets.yaml`.

jakemac53:
I think this PR is pretty much ready to go but I think we want to land https://github.com/dart-lang/test/pull/1974 first, cc @natebosch . Then we will want to update the pubspec/changelog here.

jakemac53:
The versioning here is slightly complicated so I will push up a pubspec/changelog change

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -159,13 +159,25 @@ class _TestCompilerForLanguageVersion {
...
+
     var client = _frontendServerClient = await FrontendServerClient.start(
       testUri.toString(),
       _outputDill.path,
       platformDill,
       enabledExperiments: enabledExperiments,
       sdkRoot: sdkRoot,
-      packagesJson: (await packageConfigUri).toFilePath(),
+      packagesJson: packageConfigUriAwaited.toFilePath(),
+      nativeAssets: nativeAssetsYaml?.toFilePath(),
```
natebosch, line 180: It's unlikely this argument existed in the `2.1.0` of `frontend_server_client`. We need to bump the lower bound of the constraint to whatever version adds support for this argument.

dcharkes: We need to bump it to `3.2.0`https://pub.dev/documentation/frontend_server_client/latest/frontend_server_client/FrontendServerClient/start.html

jakemac53: Good catch @natebosch !

dcharkes:
Ready to merge from my side if you're happy with it.

jakemac53:
@natebosch good on your end?

jakemac53 (approved):

natebosch (approved):

PR #1978, opened 2023-03-28
Deprecate the export of package:matcher APIs

natebosch:
Tests should import the matching framework, either `package:matcher/expect.dart`, or `package:checks/checks.dart`, or some third party alternative.

Blocked by https://github.com/dart-lang/sdk/issues/48997

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 1.24.1-dev
 
+* Deprecate the export of `package:matcher` APIs. Add an import to
+  `package:matcher/expect.dart`.
+
 ## 1.24.0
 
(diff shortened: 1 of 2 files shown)


natebosch:
@keertip @bwilkerson - Is there a way to handle a class moved to a different import?

This seems to be working well to use `replacedBy` for the top level functions and variables, but  I cannot find any allowed configuration which handles the moved classes.

`A change of type 'replacedBy' can't be used for an element of kind 'class'`

bwilkerson:
It appears that we don't currently support using `replacedBy` with class-like declarations.

I know that at one point we discussed adding a `movedTo` transform that would support moving top-level declarations to a different library, but it looks like we haven't implemented that yet.

PR #1980, opened 2023-03-30
More smoothly handle missing compiler info

natebosch:
Towards #1977

The constructor signature uses a nullable argument to smooth over the
compatibility across versions which allowed us to land in monolithic
repositories without needing to roll `test_api` and `flutter_test`
atomically. This extends the same backwards compatibility to the json
API used between the runner and tests to smooth over compatibility with
`flutter test` which may try to run test compiled against a newer API
than it uses to communicate with them.

### DIFF
--- a/pkgs/test_api/lib/src/backend/suite_platform.dart
+++ b/pkgs/test_api/lib/src/backend/suite_platform.dart
@@ -52,7 +52,9 @@ class SuitePlatform {
   factory SuitePlatform.deserialize(Object serialized) {
     var map = serialized as Map;
     return SuitePlatform(Runtime.deserialize(map['runtime'] as Object),
-        compiler: Compiler.deserialize(map['compiler'] as Object),
+        compiler: map.containsKey('compiler')
+            ? Compiler.deserialize(map['compiler'] as Object)
+            : null,
         os: OperatingSystem.find(map['os'] as String),
         inGoogle: map['inGoogle'] as bool);
   }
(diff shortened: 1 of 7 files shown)


pkgs/test_api/lib/src/backend/suite_platform.dart
```diff
@@ -52,7 +52,9 @@ class SuitePlatform {
   factory SuitePlatform.deserialize(Object serialized) {
     var map = serialized as Map;
     return SuitePlatform(Runtime.deserialize(map['runtime'] as Object),
-        compiler: Compiler.deserialize(map['compiler'] as Object),
+        compiler: map.containsKey('compiler')
```
jakemac53, line 55: why not just `as Object?`

natebosch: I would have to change the definition of `Compiler.deserialize` for that.

jakemac53 (approved):
🤮

natebosch:
I'm not 100% sure this will be fully compatible, there may be other errors at runtime.

jakemac53:
I get some weird errors like:

```
tmp/flutter_tools.BZVXDA/flutter_test_listener.SFDWOY/listener.dart:18:21: Error: Undefined name 'main'.
  await Future(test.main);
```

But it does ultimately work

jakemac53:
I kind of suspect that is just an issue that has always been present?

PR #1987, opened 2023-04-03
Make tests compatible with Strict CSP

vrana:
This propagates the nonce from any script on the page to the script created by dart.js.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.24.2-dev
+
+* Copy an existing nonce from a script on the test HTML page to the script
+  created by the test runner host javascript. This only impacts environments
+  testing with custom HTML that includes a nonce.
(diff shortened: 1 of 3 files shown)


jakemac53:
Is this different from https://github.com/dart-lang/test/pull/1808?

natebosch:
It looks like #1808 assumes every script must have the `nonce`, while this one allows it on any script element. This one has a matching internal CL that passed presubmit, and I think its more likely to work with custom HTML in case someone has a html file with scripts that have a mix of nonce and not.

natebosch:
Thanks @vrana for sharing the extra info about where we are adding the nonce in our DDC handling. I think we should be able to land this and get it merged into google3

natebosch (approved):

vrana:
Thanks. Yes, there could be e.g. extensions adding un-nonced scripts breaking the other PR.

PR #1988, opened 2023-04-04
regenerate mono_repo

devoncarew:
- regenerate mono_repo
- move back to using the latest stable version of the `setup-dart` action (from using a version from head)
- likely obsolete some of the recent dependabot PRs

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.5.0
+# Created with package:mono_repo v6.5.3
 name: Dart CI
 on:
   push:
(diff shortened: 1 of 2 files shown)


natebosch (approved):

devoncarew:
It looks like there was one test failure:

```
❌ test\runner\compiler_runtime_matrix_test.dart: --runtime vm --compiler exe forwards stdout/stderr (failed)
  
  Process `dart bin/test.dart` exited with exitCode 0. Output:
      hello00:00 +0: [VM, Exe] success
      
  [e] world
      00:00 +1: All tests passed!
  Expected: should eventually emit an event that 'hello'
    Actual: <Instance of 'StreamQueue<String>'>
     Which: emitted * hello00:00 +0: [VM, Exe] success
                    * 
                    * 00:00 +1: All tests passed!
                    x Stream closed.
              which never did emit an event that 'hello' because it:
                    * emitted an event that is different. Both strings start the same, but the actual value also has the following trailing characters: 00:00 +0:  ...
                    * emitted an event that is different. Both strings start the same, but the actual value is missing the following trailing characters: hello
                    * emitted an event that is different.
                      Expected: hello
                        Actual: 00:00 +1:  ...
                                ^
                       Differ at offset 0
  
  package:matcher/src/expect/expect.dart 1[49](https://github.com/dart-lang/test/actions/runs/4610316630/jobs/8148667796?pr=1988#step:5:52):31  fail
  package:matcher/src/expect/expect.dart 125:9   _expect.<fn>
```

PR #1992, opened 2023-04-09
Add support for Microsoft Edge

### LINKED ISSUE #1142: Support for Edge
We are adding Edge support to our test harness in Flutter for Web: [PR](https://github.com/flutter/engine/pull/15641)

Is it possible to add Edge as a supported [Runtime](https://github.com/dart-lang/test/blob/master/pkgs/test_api/lib/src/backend/runtime.dart)?

I added Edge Browser to Flutter for Web test runner, we can also add it to dart test [browsers](https://github.com/dart-lang/test/tree/master/pkgs/test/lib/src/runner/browser).

simolus3:
This adds support for the (Chromium-based) Microsoft Edge browser as a test platform. Since the browser is so similar to Chrome in general, I've moved most of the code supplying the startup arguments to `chromium.dart`.  So most of the functionality is shared between Chrome and MS Edge. I wasn't sure if the debug protocol used by Chrome is supported by Edge as-is, so I opted to not support that for now.

Closes #1142

### DIFF
--- a/pkgs/test/test/runner/browser/microsoft_edge_test.dart
+++ b/pkgs/test/test/runner/browser/microsoft_edge_test.dart
@@ -1,0 +1,76 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@TestOn('vm')
+@Tags(['chrome'])
+
+import 'package:test/src/runner/browser/microsoft_edge.dart';
+import 'package:test/src/runner/executable_settings.dart';
+import 'package:test/test.dart';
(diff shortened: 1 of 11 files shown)


pkgs/test/test/runner/browser/microsoft_edge_test.dart
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@TestOn('vm')
+@Tags(['chrome'])
```
jakemac53, line 6: We should probably have a new `edge` tag?

natebosch (approved):

PR #1994, opened 2023-04-13
Remove deprecation of test_api top level libraries

### LINKED ISSUE #1990: `group` and `test` definitions are in a deprecated library (making them deprecated)
See discussions at https://github.com/Dart-Code/Dart-Code/issues/4480.

Currently these declarations are in `scaffolding.dart` which is marked `@Deprecated`, meaning these declarations are technically deprecated too.

The analysis server doesn't currently consider them deprecated, but apparently should. Fixing that will result in all calls to `group`/`test` generating diagnostics, so this should be addressed first.

Some more info on the intended use/interpretation at https://github.com/Dart-Code/Dart-Code/issues/4480#issuecomment-1497990136

natebosch:
Closes #1990

The `scaffolding` library should not be deprecated to avoid marking the useful APIs as deprecated during autocomplete.

The `backend` library is no longer deprecated because it _is_ intended for use in `flutter_test`, and the deprecation isn't likely to be blocking any uses.

Export `registerException` from the scaffolding library. This had originally been held back from the focused libraries because it isn't a great API, but it's too entrenched to be worth deprecating at this point, so it may as well be fully supported.

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 0.5.2
 
+* Use the version `0.5.2` of `packge:test_api`.
+
 ## 0.5.1
 
 * Start adding experimental support for native_assets.yaml, when
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 0.5.2
+
 ## 0.5.1
 
 * Start adding experimental support for native_assets.yaml, when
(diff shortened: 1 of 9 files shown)


pkgs/test_core/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+## 0.5.2
+
```
kevmoo, line 2: need an entry here?

kevmoo (approved):

pkgs/test_core/CHANGELOG.md
natebosch: Yeah, probably shouldn't leave it blank...

jakemac53 (approved):

PR #1997, opened 2023-04-19
Dart 3 compatibility: change `extends Iterator` to `implements Iterator`

oprypin:
Otherwise in Dart language 3 there will be an error:

    The class 'Iterator' can't be extended outside of its library because it's an interface class. [INVALID_USE_OF_TYPE_OUTSIDE_LIBRARY]

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 0.5.3-wip
+
 ## 0.5.2
 
 * Use the version `0.5.2` of `packge:test_api`.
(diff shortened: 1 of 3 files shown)


jakemac53:
Note that this is only necessary when opting this package into Dart 3. You can avoid having to make these types of changes by respecting the declared language version in google3.

(but, this change is also fine to make, it just would make your life easier and be a long term benefit)

We also need to up the version here to `0.5.3-wip` and add an (empty) changelog entry for this version. If you would like me to do that part let me know and I can push a commit here.

oprypin:
>  If you would like me to do that part let me know and I can push a commit here.

Could you please push such a commit, yes

jakemac53 (approved):

PR #2003, opened 2023-05-01
Only run edge tests on windows

natebosch:
Add condition for the new edge runtime in the compiler matrix test.


jakemac53 (approved):

natebosch:
We should figure out why this test has _passed_ on linux before.

https://github.com/dart-lang/test/actions/runs/4694459349/jobs/8322625395#step:6:29

natebosch:
Hmm, edge is available for linux (benefit of building on chromium I guess). It's possible that github actions CI has it available.

PR #2004, opened 2023-05-02
Remove the matcher export from test_api

natebosch:
`package:test` exports directly from `package:matcher`. Resolves a
cyclic dependency between `matcher` and `test_api`.


natebosch:
We will need https://github.com/flutter/flutter/pull/125854 merged and synced before we can land this internally

jakemac53 (approved):

PR #2005, opened 2023-05-02
Prepare to publish checks

natebosch:
Expand the dependency range on `test_api` to allow the next major
version.

### DIFF
--- a/pkgs/checks/pubspec.yaml
+++ b/pkgs/checks/pubspec.yaml
@@ -1,5 +1,5 @@
 name: checks
-version: 0.2.2-dev
+version: 0.2.2
 description: >-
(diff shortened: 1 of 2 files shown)


kevmoo (approved):
LGTM

pkgs/checks/pubspec.yaml
```diff
@@ -11,7 +11,7 @@ environment:
 dependencies:
   async: ^2.8.0
   meta: ^1.9.0
-  test_api: ^0.5.0
+  test_api: ">=0.5.0 <0.6.0"
```
jakemac53, line 14: This is the same as the previous constraint right?

jakemac53 (approved):
See comment though :)

Issue #2006, opened 2023-05-04
Deserialization of timeout in suite metadata fails in dart2wasm


eyebrowsoffire:
The following line:

https://github.com/dart-lang/test/blob/9484592acc10f52efccdce03df73229fa2b10509/pkgs/test_api/lib/src/backend/metadata.dart#L285

Fails on dart2wasm, because the underlying type is double and cannot be simply cast to an int.

PR #2007, opened 2023-05-04
Don't do direct casts to int to prevent dart2wasm breakages.

eyebrowsoffire:
This fixes https://github.com/dart-lang/test/issues/2006

### DIFF
--- a/integration_tests/wasm/mono_pkg.yaml
+++ b/integration_tests/wasm/mono_pkg.yaml
@@ -18,6 +18,6 @@ stages:
       - which google-chrome-beta
       os:
         - linux
-    - test:
+    - test: --timeout=60s --retry=2
       os:
       - linux
--- a/integration_tests/wasm/mono_pkg.yaml
+++ b/integration_tests/wasm/mono_pkg.yaml
@@ -18,6 +18,6 @@ stages:
       - which google-chrome-beta
       os:
         - linux
-    - test: --timeout=60s --retry=2
+    - test: --timeout=60s
       os:
       - linux
--- a/integration_tests/wasm/mono_pkg.yaml
+++ b/integration_tests/wasm/mono_pkg.yaml
@@ -18,6 +18,7 @@ stages:
       - which google-chrome-beta
       os:
         - linux
+    # The config here is a regression test for https://github.com/dart-lang/test/issues/2006
     - test: --timeout=60s
       os:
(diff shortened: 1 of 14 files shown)


natebosch:
Are there tests in this repo that will fail without this change so that we don't regress?

jakemac53 (approved):

jakemac53:
> Are there tests in this repo that will fail without this change so that we don't regress?

It looks like if we made a dart2wasm test that uses a timeout we could get that

eyebrowsoffire:
I'll write a test

jakemac53:
See https://github.com/dart-lang/test/tree/master/integration_tests/wasm which is probably the easiest place to put this test for now

integration_tests/wasm/mono_pkg.yaml
```diff
@@ -18,6 +18,6 @@ stages:
       - which google-chrome-beta
       os:
         - linux
-    - test:
+    - test: --timeout=60s --retry=2
```
jakemac53, line 21:
```suggestion
    # The config here is a regression test for https://github.com/dart-lang/test/issues/2006
    - test: --timeout=60s --retry=2
```

jakemac53, line 21: Note that you actually can't configure the number of retries here (its just a flag to enable or disable retries). The number of retries can be configured in dart_test.yml though.

integration_tests/wasm/test/hello_world_test.dart
```diff
@@ -1,7 +1,9 @@
 // Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+
 @TestOn('wasm')
+@Retry(2)
```
jakemac53, line 6:
```suggestion
# This retry is a regression test for https://github.com/dart-lang/test/issues/2006
@Retry(2)
```

jakemac53 (approved):

jakemac53:
@eyebrowsoffire did you need a publish of this soon? @natebosch thoughts on when we should publish?

eyebrowsoffire:
We have a workaround in Flutter for now, so there's no huge rush on a publish, but let me know there is a new version and we'll bump to it.

PR #2012, opened 2023-05-10
Add error for empty groups

### LINKED ISSUE #1961: Warn when tests are accidentally disabled due to nesting within setUp() method
Hello,

We recently discovered that several tests were accidentally nested inside a `setUp()` method. 

While the tests compile, they were not actually run by the test runner, and we had no way to detect this. We only happened to notice coincidentally.

It would be helpful if the test runner could warn that tests are disabled when this happens.

Thanks for your consideration.

Rachel

natebosch:
Closes #1961

After declaring a group check for declared tests, if there were none add
a synthetic test that fails with an error about the empty group.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 1.24.3-wip
 
 * Fix compatibility with wasm number semantics.
+* Consider empty `group` to be test failures.
 
 ## 1.24.2
 
--- a/pkgs/test_api/CHANGELOG.md
+++ b/pkgs/test_api/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.5.3-wip
 
 * Fix compatibility with wasm number semantics.
+* Consider empty `group` to be test failures.
 
 ## 0.5.2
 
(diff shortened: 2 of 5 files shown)


natebosch:
This does cause some failures in existing tests.

For instance: https://github.com/flutter/packages/blob/dfcb21c087d3b7781365e915ad2864b4856c3a41/packages/flutter_markdown/test/emphasis_test.dart#L4393

@jakemac53 - Do you think we can get away with fixing the google3 impacted tests and releasing this as a non-breaking change?

jakemac53:
If we don't have evidence of the programmatic tests being an issue today I think it is probably fine though.

jakemac53:
Would it be reasonable to have an `allowEmpty` parameter on `group` to help mitigate the programmatic cases?

natebosch:
cc @christopherfujino - This is a technically breaking change but we plan on rolling this out without a major version bump. This will also impact the behavior of the `group` API exported through `flutter_test`. It will be possible for the flutter team to decide to not make the breaking change by always passing `allowEmpty: true` when forwarding `group` calls. It will also be possible to roll it out as breaking and expose the `allowEmpty` argument through the forwarding API.

jakemac53:
> @jakemac53 - WDYT about allowing completely empty groups, but making an error for the case where there is `setUp` or `tearDown` with no `test`.

I would be fine with that

jakemac53:
> no `setUp`, but the second `group` has no tests. It should have used `defineTests(config)();` because that utility was apparently designed to return the body argument to `group`, and the returned callback must be invoked.

Wow, that is a nasty API 🤣 . I don't know if I care enough about that to make the TODO case be more noisy.

PR #2028, opened 2023-06-01
Make State and Result enums

natebosch:
These are used in the other test runner implementation packages, but the
usage should not be impacted. Implement the `parse` methods by iterating
over the values instead of a manual `switch` statement.

### DIFF
--- a/pkgs/test_api/lib/src/backend/state.dart
+++ b/pkgs/test_api/lib/src/backend/state.dart
@@ -44,12 +44,12 @@ class State {
 }
 
 /// Where the test is in its process of running.
-final class Status {
+enum Status {
   /// The test has not yet begun running.
-  static const pending = Status._('pending');
+  pending('pending'),
 
   /// The test is currently running.
-  static const running = Status._('running');
+  running('running'),
 
   /// The test has finished running.
   ///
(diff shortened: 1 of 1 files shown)


jakemac53 (approved):

pkgs/test_api/lib/src/backend/state.dart
```diff
@@ -58,53 +58,43 @@ final class Status {
   /// first error or when all [expectAsync] callbacks have been called and any
   /// returned [Future] has completed, but it's possible for further processing
   /// to happen, which may cause further errors.
-  static const complete = Status._('complete');
+  complete('complete');
 
   /// The name of the status.
   final String name;
```
kevmoo, line 64: I mean, `name` is `name` here. You don't need the extra field, I don't think!

pkgs/test_api/lib/src/backend/state.dart
```diff
@@ -58,53 +58,43 @@ final class Status {
...
 }
 
 /// The outcome of the test, as far as it's known.
-final class Result {
+enum Result {
   /// The test has not yet failed in any way.
   ///
   /// Note that this doesn't mean that the test won't fail in the future.
-  static const success = Result._('success');
+  success('success'),
```
kevmoo, line 80: ditto!

pkgs/test_api/lib/src/backend/state.dart
```diff
@@ -58,53 +58,43 @@ final class Status {
...
-        return Status.pending;
-      case 'running':
-        return Status.running;
-      case 'complete':
-        return Status.complete;
-      default:
-        throw ArgumentError('Invalid status name "$name".');
-    }
-  }
+  factory Status.parse(String name) =>
```
kevmoo, line 66: https://api.dart.dev/stable/3.0.2/dart-core/EnumByName/byName.html

kevmoo (approved):

PR #2030, opened 2023-06-05
Remove test_api copies of scaffolding functions

natebosch:
There are definitions in `test_core` which override the `_declarer` to
work even when the test script is run directly instead of through the
test runner. The definitions from `test_api` should not be used anywhere
currently.


natebosch:
This is currently incompatible with `flutter_test` which uses a `hide` for these members. If we change that to a `show` listing the stuff that will stay it should work better.

PR #2034, opened 2023-06-13
Allow null descriptions

natebosch:
The change from an implicit `dynamic` to `Object` is statically breaking
for any usage passing a variable or any expression with type `dynamic`
or `Object?`.


jakemac53 (approved):

natebosch:
> error: pubspec version (0.5.4-wip) and changelog (0.5.4) don't agree

@devoncarew - was there a change in how `firehose` detects the version in the CHANGELOG? It is reading it without the `-wip` suffix, but those characters _are_ present in the file.

kevmoo:
Ugh. I get it. But ugh...

devoncarew:
I'll adjust the regex that firehose is using an re-publish.

natebosch:
> I'll adjust the regex that firehose is using and re-publish.

Ah, I had started on that before your comment.

PR #2039, opened 2023-06-15
Minor cleanup in json_reporter_utils

natebosch:
- Decode the output once instead of repeatedly decoding it for every
  `expect` call.
- Use `=>` for single expression return methods.
- Use conditional collection elements where applicable.


kevmoo (approved):
LGTM

You know the PRs that will make me happy. 😄

PR #2059, opened 2023-07-06
Drop usage of window.testRunner

natebosch:
The comment mentions content_shell and Dartium which are no longer used.
I can only find a few references to `window.testRunner`on the web, and
it does not exist in current versions of Chrome. As far as I can tell,
this is always null and does not do anything useful.


kevmoo (approved):
LGTM

kevmoo:
dartfmt needed...

jakemac53 (approved):

PR #2060, opened 2023-07-06
Read message href from message source

natebosch:
The linked issue (https://github.com/dart-lang/sdk/issues/22554) impacts
`dart:html` but does not impact usage through `@JS()` interop.

- Add a `source` field to `MessageEvent`. It is typed as
  `MessageEventSource` which is an empty interface. A source will be
  either a `WindowProxy`, a `MessagePort` or a `ServiceWorker`. For our
  use case it will always be a `WindowProxy`. The proxy has the same API
  as `Window` and static JS interop does not do nominal type checking,
  it is safe to cast it directly to `Window`.
- Stop manually adding the iframe `href` in sent messages.

### DIFF
--- a/pkgs/test/lib/dart.js
+++ b/pkgs/test/lib/dart.js
@@ -12,7 +12,6 @@ window.onload = function() {
 // This mimics a MultiChannel-formatted message.
 var sendLoadException = function(message) {
   window.parent.postMessage({
-    "href": window.location.href,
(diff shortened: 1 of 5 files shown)


kevmoo:
@natebosch – put some of that context in the source as comments maybe maybe?

natebosch:
I have some more changes I want to make nearby, I'll roll this in with the other changes.

PR #2061, opened 2023-07-07
Remove unused href field from error report

natebosch:
Originally the href was added manually to the data to work around a
`dart:html` limitation. A different workaround for the host side was
implemented in #2066, but the frame side was not changed in the same
commit to allow some cross-compatibility. Remove the final manual href
field now that no host implementations would read it.


pkgs/test/lib/src/runner/browser/post_message_channel.dart
```diff
@@ -2,58 +2,28 @@
...
-      port.start();
-      var portSubscription =
-          dom.Subscription(port, 'message', allowInterop((dom.Event event) {
-        controller.local.sink.add((event as dom.MessageEvent).data);
-      }));
-
-      controller.local.stream.listen((data) {
-        port.postMessage({'data': data});
-      }, onDone: () {
-        port.postMessage({'event': 'done'});
```
natebosch, line 45: `bolt` was using this

It also has a dependency on the current interaction structure because it re-implements `host.dart`.

natebosch:
Flutter has an old copy of this code.

https://github.com/flutter/flutter/blob/2b6492426e5c9a17f3597f40323ce902d4540396/packages/flutter_tools/lib/src/web/bootstrap.dart#L248

If we make this change, when we roll `package:test` into the flutter repo we'll also need to adjust that.

cc @christopherfujino

pkgs/test/lib/src/runner/browser/post_message_channel.dart
natebosch: I have removed the handling of this message internally. We can safely revert https://github.com/dart-lang/test/pull/939 but will need further changes to land this PR.

natebosch:
This host <-> frame communication has been copied or reimplemented in a few places. Internal google code has a separate host implementations, and `flutter_test` has an implementation of the frame side.

I cannot land the change in flutter until the package is published to pub. I will not publish a change like this to pub until we have internal validation.

To resolve the deadlock I'm going to take a try at first implementing a backwards/forwards compatible `host.dart`.

Issue #2065, opened 2023-07-19
Simplify the communication between the browser host and iframe


natebosch:
Currently there are 3 hops during startup.

- [iframe (postMessageChannel) posts a message to parent](https://github.com/dart-lang/test/blob/8bc188f70c6de2ed21dc0aab1e7e4ef375d73757/pkgs/test/lib/src/runner/browser/post_message_channel.dart#L54-L56)
- [parent posts message back to iframe](https://github.com/dart-lang/test/blob/8bc188f70c6de2ed21dc0aab1e7e4ef375d73757/pkgs/test/tool/host.dart#L224-L229)
- Communication can run over the private channel wired up to the other test infrastructure by forwarding [from the frame](https://github.com/dart-lang/test/blob/8bc188f70c6de2ed21dc0aab1e7e4ef375d73757/pkgs/test/lib/src/runner/browser/post_message_channel.dart#L42-L47) and [from the host](https://github.com/dart-lang/test/blob/8bc188f70c6de2ed21dc0aab1e7e4ef375d73757/pkgs/test/tool/host.dart#L238-L248)

This communication all happens an an opaque way, especially on remote CI infrastructure, and _may_ be involved with some flakes that we see in a few places where browser tests never start. We should simplify the communication, and add some logging to see if we can get any insight into the cause of the flakes.

A better pattern would be

- iframe posts the `MessageChannelPort` to the parent and wires it up to the channel.
- host receives the port and wires it up to the channel. Start forwarding messages when the channel is ready instead immediately forwarding to a callback that [waits for the channel to be ready](https://github.com/dart-lang/test/blob/8bc188f70c6de2ed21dc0aab1e7e4ef375d73757/pkgs/test/tool/host.dart#L245).

Separately, we have some unnecessary legacy behavior with extra messages that turned out to be a no-op round trip sending a signal back to the calling code to continue. We can clean that up in the same pass - I've already removed the usage in source but have not yet released that version internally.

#2061  does the full refactoring, and would be safe to land if this was the only test runner. As mentioned in https://github.com/dart-lang/test/pull/2061#issuecomment-1641134897 we will need to do this in multiple steps with backwards compatibility along the way because there is another copy of both sides of the implementation, and we cannot roll them all atomically.

Somewhat interesting side note - Even though the flutter use case doesn't pin `package:test` as a `dependency`, they reach into the pub solve within the flutter SDK to find the version of `test` pinned through a `dev_dependency` and use that version of the compiled host js.

[...]

PR #2067, opened 2023-07-19
Handle initial message entirely within conditional

natebosch:
Towards #2065

Refactor the current implementation to reduce complexity and keep the
code for the current behavior within the condition body. This will allow
the new behavior to be implemented in another branch of the conditional.

- Remove the `readyCompleter`. Don't start listening on the server
  channel until the frame communication is ready instead of eagerly
  listening and holding the messages callbacks awaiting the ready
  future.
- Keep only the long-term message channel in the stored subscriptions.
  We only expect a single message on the window channel from each frame,
  so we can cancel the subscription after we receive it.


pkgs/test/tool/host.dart
```diff
@@ -216,36 +207,34 @@ StreamChannel<dynamic> _connectToIframe(String url, int id) {
...
 
     message.stopPropagation();
+    windowSubscription.cancel();
 
     if (message.data['ready'] == true) {
       // This message indicates that the iframe is actively listening for
       // events, so the message channel's second port can now be transferred.
-      channel.port2.start();
+      var channel = dom.createMessageChannel();
+      _domSubscriptions[id] = dom.Subscription(channel.port1, 'message',
```
jakemac53, line 216: Should we assert that there isn't one already just in case?

jakemac53 (approved):

pkgs/test/tool/host.dart
```diff
@@ -216,36 +201,35 @@ StreamChannel<dynamic> _connectToIframe(String url, int id) {
...
+          allowInterop((dom.Event event) {
+        controller.local.sink.add((event as dom.MessageEvent).data['data']);
+      }));
+
+  _subscriptions[id] =
```
jakemac53, line 215: This formatting is also a bit weird, might want to file an issue if the formatter did this?

natebosch: forgot to reformat.

pkgs/test/tool/host.dart
natebosch: I don't think we ever compile and run this script with assertions enabled since we manually compile it. We could potentially explore recompiling with asserts during CI at some point in the future so it won't hurt to add them.

jakemac53: We could do a regular check/throw also I suppose? But I am not sure we would actually want to crash. It does indicate an unexpected situation and would result in an uncancelled listener though.

natebosch: I'll probably leave this code path using assert for now.

We can discuss more whether to us non-assert checks for the new code path when I have that implementation ready.

PR #2068, opened 2023-07-28
Allow frame to send the message port

natebosch:
Towards #2065

Handle the new communication pattern from the host side. Allows the
frame to send a single `'port'` message with a read `MessageChannel` in
stead of the `{'ready': true}` message signalling that the host should
send a message port. This will remove one hop in the communication
pattern.

Retains handling of the current pattern and does not update the frame
side yet. There are multiple implementations of both the host and the
frame behavior. All host implementations will be updated to allow either
pattern before the frame implementations are updated.

Change from a `if/else` chain to a `switch`. Use pattern matching to
destructure the `'data'` field in the case of exceptions.

### DIFF
--- a/pkgs/test/tool/host.dart
+++ b/pkgs/test/tool/host.dart
@@ -195,7 +195,6 @@ MultiChannel<dynamic> _connectToServer() {
 /// Returns a [StreamChannel] which will be connected to the frame once the
 /// message channel port is active.
 StreamChannel<dynamic> _connectToIframe(String url, int id) {
-
   var iframe = dom.createHTMLIFrameElement();
   _iframes[id] = iframe;
   var controller = StreamChannelController(sync: true);
--- a/pkgs/test/tool/host.dart
+++ b/pkgs/test/tool/host.dart
@@ -178,7 +178,7 @@ MultiChannel<dynamic> _connectToServer() {
 ///
 /// [id] identifies the suite loaded in this iframe.
 ///
-/// While the iframe is starting, adds a listener for `window.onMessage` which
+/// Before the frame is attached, adds a listener for `window.onMessage` which
 /// filters to only the messages coming from this frame (by it's URL) and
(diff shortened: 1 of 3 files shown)


jakemac53 (approved):

pkgs/test/tool/host.dart
```diff
@@ -173,10 +173,27 @@ MultiChannel<dynamic> _connectToServer() {
...
 }
 
-/// Creates an iframe with `src` [url] and establishes a connection to it using
-/// a [MessageChannel].
+/// Creates an iframe with `src` [url] and expects a message back to connect a
+/// message channel with the suite running in the frame.
 ///
 /// [id] identifies the suite loaded in this iframe.
+///
+/// While the iframe is starting, adds a listener for `window.onMessage` which
```
jakemac53, line 181: I wonder if this is partly the issue regarding flakyness? If the message posts before our listener is set up we might not see it?

natebosch: It looks easy enough in this implementation to ensure the listener is added before the frame.

jakemac53 (approved):

pkgs/test/tool/host.dart
natebosch: > I wonder if this is partly the issue regarding flakyness?

Hard to say, but apparently it wasn't the sole issue cause it still flakes on this PR 😝 

https://github.com/dart-lang/test/actions/runs/5717430441/job/15491160863?pr=2068

```
      00:00 +0: [Microsoft Edge, Dart2Js] loading test.dart
      Info: Compiling with sound null safety.
      Compiled 10,952,326 characters Dart to 1,225,100 characters JavaScript in 7.32 seconds using 327.988 MB of memory
      Dart file ../dart_test_FFBBZA/runInBrowser.dart compiled to JavaScript: ../dart_test_VZICMC/test_AFKCDJ/test.dart.browser_test.dart.js
```

PR #2072, opened 2023-08-02
Send a MessagePort to host on frame startup

natebosch:
Towards #2065

All host implementations are updated to allow either the legacy format,
or this new format.

Simplify the communication and avoid a listener on the frame window
message events. Upon startup replace the `{'ready': true}` message with
a `'port'` message and the `MessagePort` for a channel created on the
frame side. Omit the manual 'href' field, no host implementations would
read it.


jakemac53 (approved):
Can you check this with package:build (or by just passing -O2 to dart2js in this package)?

natebosch:
> Can you check this with package:build (or by just passing -O2 to dart2js in this package)?

I recompiled `host.js` with `-O2` and `dart test -p chrome` still works. I also overrode the dep to this version in `build/_test` and ran `dart run build_runner test -- -p chrome` and that worked too.

Issue #2074, opened 2023-08-07
build_web_compilers dart2js tests are failing with the latest package:test


jakemac53:
If you run `dart run build_runner test --release` with test_core version 0.5.5 it will fail to bootstrap and time out - with version 0.5.4 it will succeed.

jakemac53:
This appears to only happen if running minified dart2js (which is the default in build_web_compilers for release mode).

The specific error I get is this:

```
Uncaught TypeError: Instance of 'minified:EventTarget': type 'minified:EventTarget' is not a subtype of type 'minified:a'
```

jakemac53:
cc @sigmundch

PR #2075, opened 2023-08-07
fix failing casts when talking to minified apps from the host

jakemac53:
Fixes https://github.com/dart-lang/test/issues/2074

This is more a workaround than a true fix, but when talking from a minified dart2js test app to a non-minified host we would get cast failures on these getters. Changing it to an explicit getProperty call and adding the pragma to trust the casts silences the failure.


jakemac53:
cc @sigmundch

natebosch (approved):
Should we minify the host app?

Would that have surfaced issues in the test runner communicating with a non-minified test?

sigmundch:
Indeed, minifying the host would have caused failures in an unminified suite. It could also cause issues in a minified suite if they were to pick different names for some types.

FYI the underlying cause is described here: https://github.com/dart-lang/sdk/issues/53154

jakemac53:
I don't think we should minify the host because it will make errors even harder to diagnose

PR #2080, opened 2023-08-11
Timeout browser suite loads

natebosch:
The current timeout applies to starting the process and connecting back
from the host. Add a timeout which applies on each suite as it is
loaded.

Change the chrome argument to log to `stderr` instead of `stdout`. The
latter seems to have no effect.

Eagerly decode and split lines from the browser output instead of
storing the bytes to avoid mixing lines between the two outputs. In most
cases this shouldn't matter, I have not seen output on `stdout`.

Use `--v=0` instead of `--v=1` since the 0 level includes console
output, and the extra browser level output with higher verbosity does
not look useful.


jakemac53 (approved):

natebosch:
@jakemac53 - I switched this from configurable per suite to hardcoded at 30 seconds. Do you have any concerns about that? This timeout _does_ include waiting for `main` to complete for the test suite.

The reason I hardcoded it is because we have some tests where the configured timeout seems to be too low for a suite and the load was timing out.

natebosch:
I think it may work to leave it configurable so there is an escape hatch for slow `main`, but have a lower bound at 30 seconds if the suite sets the timeout lower. WDYT?

jakemac53:
A lower bound but configurable is probably fine.

PR #2083, opened 2023-08-16
Add some console logging to browser test startup

natebosch:
Add support for console log and warn level in the dom library.
Switch an existing `print` to a `console.warn` for an error. Use
`console.log` consistently otherwise. This is less important in the host
(where a `print` will anyways log to the console) than it is in the
remote listener (where a `print` could be intercepted and forwarded to
the test runner).

Refactor an `if`/`else` chain over a map field to a `switch` with
destructuring for all the patterns that can be handled. Add a default
case with a console warning instead of silently treating potentially
non-matching commands as `'closeSuite'`. This also changes some
potential cast failures into falling through to the warning case which
should be easier to debug.

Add a message from the frame when it is connecting back to the host.

### DIFF
--- a/pkgs/test/lib/src/runner/browser/dom.dart
+++ b/pkgs/test/lib/src/runner/browser/dom.dart
@@ -14,6 +14,7 @@ extension WindowExtension on Window {
   @pragma('dart2js:as:trust')
   Window get parent => js_util.getProperty<dynamic>(this, 'parent') as Window;
   external Location get location;
+  Console get console => js_util.getProperty(this, 'console') as Console;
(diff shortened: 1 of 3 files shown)


natebosch:
From my testing internally so far (does not use this browser platform or `host.dart`, but does use this `postMessageChannel`) it seems like the Dart executing in the frame stops at a point which is past this infrastructure, but not yet declaring tests. If the external flakes are similar, these prints will still show up and we'll need more to further narrow down the problem. I think it's worth adding at least these initial messages so that we can confirm the problem isn't this early externally either.

brianquinlan (approved):

Issue #2088, opened 2023-09-02
[ERROR] Deferred library load doesn't work inside browser tests


gmpassos:
When `dart2js` is used by the package [test][pkg_test_github] it's not
able to serve the `...part.js` file.

You can reproduce the bug running the test of this repository:
https://github.com/gmpassos/deferred_lib_issue

```shell
dart test -p chrome
```

Error output:
```text
...

00:08 +0 -1: test/main_test.dart: View Test [E]                                                                                                                                                       
  DeferredLoadException: 'Loading http://localhost:50656/Sa1ttbh3Sd59FoFnteBXDihHx9rZDIUn/test/main_test.dart.browser_test.dart.js_1.part.js?dart2jsRetry=3 failed: Instance of 'Event'
  Context: js-failure-wrapper
  event log:
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"downloadFailure","l":"view_deferred","s":"http://localhost:50656/Sa1ttbh3Sd59FoFnteBXDihHx9rZDIUn/test/main_test.dart.browser_test.dart.js","i":13}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"download","l":"view_deferred","i":12}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"startLoad","l":"view_deferred","i":11}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"retry2","l":"view_deferred","i":10}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"download","l":"view_deferred","i":9}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"startLoad","l":"view_deferred","i":8}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"retry1","l":"view_deferred","i":7}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"download","l":"view_deferred","i":6}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"startLoad","l":"view_deferred","i":5}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"retry0","l":"view_deferred","i":4}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"download","l":"view_deferred","i":3}
  {"p":"main_test.dart.browser_test.dart.js_1.part.js","e":"startLoad","l":"view_deferred","i":2}
[...]

PR #2089, opened 2023-09-06
Support deferred loading in browser tests

jakemac53:
Fixes https://github.com/dart-lang/test/issues/2088.

- Updates the CompilerPool to be generic so that compilers can return custom values from each compile.
- Updates the Dart2JsCompilerPool to return a list of absolute file paths for the compiled files.
- Updates the browser platform to serve all compiled files instead of hard-coding the expected ones.
- Add a test

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## 0.5.7-wip
 
+* Update the CompilerPool to be generic, so that custom compiler pools can
+  return their own data for each compile step.
+* Update `withTempDir` to be generic so that you can get return values out of
+  the provided callback.
+
+
 ## 0.5.6
 
 * Add support for discontinuing after the first failing test with `--fail-fast`.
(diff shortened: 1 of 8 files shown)


jakemac53:
Looks like I broke the stack trace mapper :P, will look into that

pkgs/test_core/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## 0.5.7-wip
 
+* Update the CompilerPool to be generic, so that custom compiler pools can
```
natebosch, line 3: This isn't breaking?

natebosch (approved):

pkgs/test_core/CHANGELOG.md
jakemac53: It can default to dynamic safely right? The return types were always just `Future<void>` before.

jakemac53:
Fixing the stack trace mapper up properly seems like it might be quite difficult. So I might revert this and instead just pass the argument to dart2js to tell it not to split the files.

But I do think its better to have more "real world" behavior in the tests.

jakemac53:
I don't have sufficient time to continue diagnosing this and trying to come up with a solution, especially considering that we have a much easier option, so I am going to close this. I will open a new PR with the easier fix later.

PR #2090, opened 2023-09-06
Fix running browser tests that use deferred loading

jakemac53:
Fixes https://github.com/dart-lang/test/issues/2088

Pass `--disable-program-split` so that no splitting actually happens.

I [first tried fixing this](https://github.com/dart-lang/test/pull/2089) more "properly" by serving all the output files but this was a lot more complex and supporting stack trace mapping in particular was looking very challenging.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Simplify the initialization of the per-suite message channel within browser
   tests. See https://github.com/dart-lang/test/issues/2065
 * Add a timeout to browser test suite loads.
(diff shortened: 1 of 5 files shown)


natebosch (approved):

sigmundch (approved):

gmpassos:
This needs to be documented somewhere:

When a test is compiled to JS, the parameter `--disable-program-split` is passed to `dart2js`, preventing it from splitting the program to avoid issues when serving the split parts in browser tests. Deferred libraries are the ones that require the compiler to split the program, and they shouldn't be affected by `--disable-program-split`.

PR #2093, opened 2023-09-18
Add a silent reporter

natebosch:
Resolve a flaky test by eliminating the reporter output as a
potential race condition.

Add a `hidden` flag to `ReporterDetails` since the reporter is not worth
showing in the CLI usage help. Request the silent reporter for a test
which is flaky due to interleaving of forwarded `stdout` with the
reporter's output. During flakes the `hello` and the newline can come as
separate events in the stream, so sometimes the newline is written
following unexpected reporter content.

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 * Pass --disable-program-split to dart2js to fix tests which use deferred
   loading.
+* Add a 'silent' reporter option. Keep it hidden in the CLI args help since it
+  is not useful in the general case, but can be useful for tests of the test
+  runner.
 
 ## 0.5.6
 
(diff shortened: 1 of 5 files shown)


jakemac53 (approved):

pkgs/test_core/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 * Pass --disable-program-split to dart2js to fix tests which use deferred
   loading.
+* Add a 'silent' reporter option. Keep it hidden in the CLI args help since it
```
jakemac53, line 5: Do we even want to mention it here?

jakemac53: sgtm, if we just add it here but not in `test`

Issue #1614, opened 2021-10-21
Internet Explorer tests time out on windows


jakemac53:

natebosch:
@sigmundch - has dart2js dropped support for IE?

I wonder if we should just remove support for IE from the test runner. Would anyone even notice?

gmpassos:
I don't think that anyone should incentivize the use IE, ate least since 2016.

I hope this won't break Edge support.

jakemac53:
We probably could just drop IE support. Edge should continue to work fine.

sigmundch:
Agree, feel free to drop IE coverage. We dropped support for IE two years ago now, since Dart 2.14. See https://github.com/dart-lang/sdk/issues/46545

PR #2098, opened 2023-09-20
Try unskipping IE tests

natebosch:


natebosch:
Towards #1614 

I think we could install IE on github actions, but is it worth it? Should we drop support instead?

PR #2099, opened 2023-09-21
Drop support for legacy iframe communication

### LINKED ISSUE #2065: Simplify the communication between the browser host and iframe

natebosch:
Closes #2065

All known iframe side implementations have been updated to send the
`'port'` message with an immediately available channel over the map and
the extra window message.

Move some still relevant `assert` into the implementation for the
current approach.


jakemac53 (approved):
Are we going to run into the issue with missing `flutter_test` dependencies again?

natebosch:
> Are we going to run into the issue with missing `flutter_test` dependencies again?

We shouldn't. The two sides of this particular dependency are managed by dependencies in the `flutter` checkout, not the users project.

https://github.com/flutter/flutter/pull/131881 updated `flutter_tools` to send the `'port'` message, and the compiled `host.dart.js` is used from the flutter repo pub solve.
https://github.com/flutter/flutter/blob/4d8e390c1be6237ab45df9c57b87e50e499056db/packages/flutter_tools/lib/src/test/flutter_web_platform.dart#L156-L157

This would only be a problem if the iframe code could be an old version - but that is controlled by the test platform, not the pub dependency.

PR #2101, opened 2023-09-25
Filter test by line when kernel is compiled with --filesystem-scheme.

chingjun:
Filtering by line number is hardcoded to assume that test cases are specified in `file://` uri. If the kernel is compiled with `--filesystem-scheme`, filtering by line did not work.

---

<details>

</details>

### DIFF
--- a/pkgs/test_core/lib/src/runner.dart
+++ b/pkgs/test_core/lib/src/runner.dart
@@ -323,8 +323,24 @@ class Runner {
             var absoluteSuitePath = File(path).absolute.uri.toFilePath();
 
             bool matchLineAndCol(Frame frame) {
-              if (frame.uri.scheme != 'file' ||
-                  frame.uri.toFilePath() != absoluteSuitePath) {
+              if (frame.uri.scheme != 'file') {
+                // In this case, this is either a package: URI, or the kernel is
+                // compiled using --filesystem-scheme.
+
+                if (frame.uri.scheme == 'package') {
+                  // It is unlikely that the test case is specified in a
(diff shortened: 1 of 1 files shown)


pkgs/test_core/lib/src/runner.dart
```diff
@@ -323,8 +323,24 @@ class Runner {
             var absoluteSuitePath = File(path).absolute.uri.toFilePath();
 
             bool matchLineAndCol(Frame frame) {
-              if (frame.uri.scheme != 'file' ||
-                  frame.uri.toFilePath() != absoluteSuitePath) {
+              if (frame.uri.scheme != 'file') {
```
jakemac53, line 326: I think this would be better structured as a switch statement at this point:

```dart
              switch (frame.uri.scheme) {
                case 'file':
                  if (frame.uri.toFilePath() != absoluteSuitePath) {
                    return false;
                  }
                case 'package':
                  // It is unlikely that the test case is specified in a
                  // package: URI. Ignore this case.
                  return false;
                default:
                  // Now we can assume that the kernel is compiled using
                  // --filesystem-scheme.
                  // In this case, because we don't know the --filesystem-root, as
                  // long as the file path matches we assume it is the same file.
                  if (!absoluteSuitePath.endsWith(frame.uri.path)) {
                    return false;
                  }
              }
```

jakemac53 (approved):

jakemac53:
@chingjun is this something you would like published externally soon? Or do you only need it rolled internally?

jakemac53:
I can do it 👍

PR #2102, opened 2023-09-26
Update the documentation for the `skip` parameter in the `test`

gmpassos:
It's not documented that `test(skip:)` will overwrite the behavior of the `group(skip:)`

- [v] I’ve reviewed the contributor guide and applied the relevant portions to this PR.


pkgs/test_core/lib/scaffolding.dart
```diff
@@ -94,7 +94,11 @@ Declarer get _declarer {
 ///
 /// If [skip] is a String or `true`, the test is skipped. If it's a String, it
 /// should explain why the test is skipped; this reason will be printed instead
-/// of running the test.
+/// of running the test. If a [test] is nested within a [group], the
```
jakemac53, line 97: It is probably worth mentioning the behavior with regards to a suite level `@Skip` as well? I presume neither group/test configuration will take precedence because we never even compile/run the test in that case. But you would have to confirm.

jakemac53: I am not seeing any for this specific behavior, for the group/test behavior you could add a test after [this one](https://github.com/dart-lang/test/blob/master/pkgs/test/test/runner/engine_test.dart#L236).

For the `@Skip` handling it has to pretty much be an integration tests, could be next to this one https://github.com/dart-lang/test/blob/master/pkgs/test/test/runner/runner_test.dart#L514.

kevmoo:
This PR needs to be rebased

gmpassos:
> This PR needs to be rebased

Resolved

natebosch:
Hmm, I pushed a commit which added tests, but the PR didn't update like it should.

I'm not sure if there is some way to poke the PR to get it to pick up the new commit?

https://github.com/gmpassos/test/commit/62855be3834a4b0f96c5c162f1d4ec1cf97aa70d

gmpassos:
I think that the PR is OK now. It's showing the changed test files.

gmpassos:
Why use `1group1` and `1test1` in the documentation? Is this recognized in a different way by `dartdoc`?

natebosch:
> Why use `1group1` and `1test1` in the documentation? Is this recognized in a different way by `dartdoc`?

Unintentional typos, I ran a shortcut to change the surrounding characters but typed `1` instead of  `` ` `` for the surrounding character.

natebosch:
@jakemac53 - can you take another look with the added tests?

jakemac53 (approved):

PR #2103, opened 2023-09-28
Add `onDone` argument to `deserializeSuite`

natebosch:
This will allow the flutter web platform to wait for the browser to
shut down between test suites.


natebosch:
@yjbanov - this API only looks useful if the platform can also enforce `-j 1` behavior. Do you already have a mechanism where you enforce that argument for `flutter test` runs?

How feasible would it be to make the flutter test web platform tolerant of test suite concurrency?

natebosch:
@yjbanov - is this still an API that we should be considering?

natebosch:
@polina-c - This might be something we can for the leak tracking

PR #2110, opened 2023-10-02
Update package:vm_service to 12.0.0

bkonyi:
This will require a release to unblock rolling this new version of vm_service through to Flutter.

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Add a 'silent' reporter option. Keep it hidden in the CLI args help since it
   is not useful in the general case, but can be useful for tests of the test
   runner.
+* Update to `package:vm_service` `12.0.0`
 
 ## 0.5.6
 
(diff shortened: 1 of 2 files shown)


bkonyi:
Also, I think there's a cyclical dependency between the various `test` packages and `coverage` which might make this tricky.

jakemac53:
> Also, I think there's a cyclical dependency between the various `test` packages and `coverage` which might make this tricky.

This only really gets tricky if we actually _require_ the latest vm_service, but it doesn't seem like we do. We won't actually be able to fetch the latest one until both release support though.

The tricky part is making sure we actually do support the latest vm_service. I would actually start with the `coverage` package, because it has less code/tests, and add a `dependency_override` to force the latest vm_service. Run the tests/analyzer locally on that package to make sure it works (this will also serve to ensure package:test works to some degree).

Then once that is released, the tests here should be able to pick up the latest vm_service too. The only really sensible way to run the tests in this package is via CI because of all the setup required, so that will be the easiest path.

jakemac53:
Another option could be adding a dependency_override to all the packages in this repo, to force the vm_service to be the latest version, so our tests run with that. And then we would revert that commit before landing.

PR #2114, opened 2023-10-05
Remove support for Internet Explorer

### LINKED ISSUE #1614: Internet Explorer tests time out on windows

natebosch:
Closes #1614

Use should be replaced by the edge browser.

It has been 2 years since we removed support for internet explorer in
dart2js. We have not been testing internet explorer integration, and it
may already be broken by dart2js changes.

Remove every reference to `ie` or internet explorer in the repo.

Treat this as non-breaking for `package:test` even though there is a
small risk of breaking some users still using internet explorer and it
happens to still be working.


natebosch:
There is a usage of the Runtime constant in flutter engine.
https://github.com/flutter/engine/blob/0ff05464f42ffea289e0475d7fc0125b83c49f91/lib/web_ui/dev/edge.dart#L27

I'll open a PR to move them to `edge` for that use case.

It's unfortunate to need a major version bump for this - I wonder if we can get away with eagerly fixing the flutter usage and then treating it as non-breaking. WDYT @jakemac53 ?

Or maybe we can drop the support in `package:test` and just leave the `Runtime` value around in `test_api`?

jakemac53:
I would be fine with leaving it around but just dropping support. Can we mark it deprecated?

natebosch:
Currently this is set up to land before the publish in https://github.com/dart-lang/test/pull/2156, but we shouldn't consider it a blocker for publishing.

jakemac53 (approved):

jakemac53:
Lets go ahead and merge this prior to the publish

PR #2117, opened 2023-10-13
Make compiler optional in calls to currentPlatform

natebosch:
Allow backwards compatibility with current callers. This might make it
easier to land https://github.com/flutter/engine/pull/46592

### DIFF
--- a/pkgs/test_core/lib/src/util/io.dart
+++ b/pkgs/test_core/lib/src/util/io.dart
@@ -48,7 +48,7 @@ final currentOS = OperatingSystem.findByIoName(Platform.operatingSystem);
 /// [inGoogle] determined automatically.
 ///
 /// If [runtime] is a browser, this will set [os] to [OperatingSystem.none].
-SuitePlatform currentPlatform(Runtime runtime, Compiler? compiler) =>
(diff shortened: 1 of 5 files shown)


pkgs/test_core/lib/src/util/io.dart
```diff
@@ -48,7 +48,7 @@ final currentOS = OperatingSystem.findByIoName(Platform.operatingSystem);
 /// [inGoogle] determined automatically.
 ///
 /// If [runtime] is a browser, this will set [os] to [OperatingSystem.none].
-SuitePlatform currentPlatform(Runtime runtime, Compiler? compiler) =>
```
jakemac53, line 51: I did intentionally make this non-optional to avoid accidentally not passing it when you do have a compiler - I wanted it to be an explicit choice to just pass `null`, which is really only there for testing purposes or callers who don't support user configuration of compilers.

If we need to make it optional for some period of time we could do that but I would like to make it non-optional eventually, so we should file a bug/todo.

natebosch: Done.

jakemac53 (approved):

PR #2128, opened 2023-10-20
Tweak dart2js source map rewriting

natebosch:
- Remove a TODO referencing a closed issue. The issue was deduped to
  another which was closed without changes. Manual testing in a browser
  shows that the rewriting is still necessary.
- Extract a variable for the source map path as a Uri instead of
  computing it within the loop.
- Directly return URIs which are already `file://`. This would be a
  slight behavior change for the `runInBrowser.dart` synthetic source if
  it was ever put in the source map file with an absolute `file://`
  URI, but with the current (and apparently planned) dart2js output they
  do not have the file scheme. This is not a behavior change for any
  other cases where the source is a `file` URI, since it is no necessary
  to call `resolveUri` with an absolute path, and all `file` URI are
  absolute.

A sample run with a `Stopwatch` shows this reduces the non-IO overhead
by about 75% (out of 165ms in my sample with multiple suites).

### DIFF
--- a/pkgs/test_core/lib/src/runner/dart2js_compiler_pool.dart
+++ b/pkgs/test_core/lib/src/runner/dart2js_compiler_pool.dart
@@ -107,7 +107,7 @@ class Dart2JsCompilerPool extends CompilerPool {
           switch (Uri.parse('$root$source')) {
             Uri(hasScheme: true) && final uri => uri.toString(),
             Uri(:final path) when path.endsWith('/runInBrowser.dart') => '',
-            final uri => mapUri.resolveUri(uri),
+            final uri => mapUri.resolveUri(uri).toString(),
           }
       ],
     );
--- a/pkgs/test_core/lib/src/runner/dart2js_compiler_pool.dart
+++ b/pkgs/test_core/lib/src/runner/dart2js_compiler_pool.dart
@@ -93,18 +93,18 @@ class Dart2JsCompilerPool extends CompilerPool {
     });
   }
 
-  // TODO(nweiz): Remove this when sdk#17544 is fixed.
   /// Fix up the source map at [mapPath] so that it points to absolute file:
   /// URIs that are resolvable by the browser.
   void _fixSourceMap(String mapPath) {
     var map = jsonDecode(File(mapPath).readAsStringSync());
     var root = map['sourceRoot'] as String;
 
-    map['sources'] = map['sources'].map((Object? source) {
+    final mapUri = p.toUri(mapPath);
+    map['sources'] = (map['sources'] as List).map((Object? source) {
       var url = Uri.parse('$root$source');
-      if (url.scheme != '' && url.scheme != 'file') return source;
+      if (url.hasScheme) return url.toString();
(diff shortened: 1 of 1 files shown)


natebosch:
@jakemac53 - I find `switch` to be pretty readable - it looks somewhat declarative to me. Is this too esoteric compared to the `.map().toList()` though?

```diff
-    map['sources'] = (map['sources'] as List).map((Object? source) {
-      var url = Uri.parse('$root$source');
-      if (url.hasScheme) return url.toString();
-      if (url.path.endsWith('/runInBrowser.dart')) return '';
-      return mapUri.resolveUri(url).toString();
-    }).toList();
+    map.update(
+      'sources',
+      (sources) => [
+        for (var source in sources as List)
+          switch (Uri.parse('$root$source')) {
+            Uri(hasScheme: true) && final uri => uri.toString(),
+            Uri(:var path) when path.endsWith('/runInBrowser.dart') => '',
+            final uri => mapUri.resolveUri(uri).toString(),
+          }
+      ],
+    );
```

pkgs/test_core/lib/src/runner/dart2js_compiler_pool.dart
```diff
@@ -93,19 +93,24 @@ class Dart2JsCompilerPool extends CompilerPool {
...
-      var url = Uri.parse('$root$source');
-      if (url.scheme != '' && url.scheme != 'file') return source;
-      if (url.path.endsWith('/runInBrowser.dart')) return '';
-      return p.toUri(mapPath).resolveUri(url).toString();
-    }).toList();
+    final mapUri = p.toUri(mapPath);
+    map.update(
+      'sources',
+      (sources) => [
+        for (var source in sources as List<Object?>)
```
jakemac53, line 106: nit: I think I would prefer to just call `cast` on the map as opposed to doing the individual casts here

jakemac53:
> @jakemac53 - I find `switch` to be pretty readable - it looks somewhat declarative to me. Is this too esoteric compared to the `.map().toList()` though?

I like the `switch` variant, a bit weird how you can bind a variable to the original expression with `final uri` like that, haven't seen that before, but I did intuit the correct behavior 🤷‍♂️

jakemac53 (approved):

pkgs/test_core/lib/src/runner/dart2js_compiler_pool.dart
natebosch: A cast on the map would not do what we need, we can only cast the generic `V` to  `Object?`, not `List<Object?>` like we need.

jakemac53: I mean a literal call to `.cast<String, List<Object?>>()`, which will have the same effect (lazily casting each value as its pulled out), but it encodes that assumption a bit earlier and more directly imo.

Ultimately, I am fine with either.

natebosch: WDYT about the latest commit?

```dart
    map.cast<String, List<Object?>>().update(
          'sources',
          (sources) => [
            for (var source in sources)
```

PR #2140, opened 2023-11-22
Update to latest version of lints

kevmoo:
Remove a <R> type parameter from checks that seems unused

### DIFF
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -433,7 +433,7 @@ abstract final class Context<T> {
   ///   }
   /// }
   /// ```
-  Future<void> expectAsync<R>(Iterable<String> Function() clause,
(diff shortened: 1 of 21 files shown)


pkgs/checks/lib/src/checks.dart
```diff
@@ -433,7 +433,7 @@ abstract final class Context<T> {
   ///   }
   /// }
   /// ```
-  Future<void> expectAsync<R>(Iterable<String> Function() clause,
```
kevmoo, line 436: @natebosch – `<R>` seems superfluous! See overrides below, too!

natebosch: Huh, I think this is a leftover from a _very_ early design where `nest` didn't exist. I'm surprised it survived so long without getting noticed.

natebosch (approved):

Issue #2143, opened 2023-11-28
Move dart2wasm support to be "normal"


kevmoo:
At the moment there is an `experimental-chrome-wasm` flag, etc.

Ideally, Wasm support should just be like dart2js support. Wasm tests should "just pass" on (latest) Chrome and Firefox.

We need to test the following packages with wasm

- [ ] `sse` - https://github.com/dart-lang/sse/issues/97
- [ ] `http` - https://github.com/dart-lang/http/issues/1056
- [ ] `web_socket_channel` - https://github.com/dart-lang/web_socket_channel/issues/297

CC @jakemac53 @srujzs

PR #2144, opened 2023-11-30
Promote dart2wasm compiler support to the stable browser platform.

jakemac53:
Closes https://github.com/dart-lang/test/issues/2143

I ended up going on quite the cleanup adventure here 🤣 . But I think the result is much cleaner around how it handles --precompiled and the various compilers in the browser platform. We may want to adopt a similar cleanup in other platforms too.

- Deletes the experimental wasm platform, moves it into the `chrome` platform (works on chrome stable now).
- Refactors heavily how the browser platform works, abstracting away the logic for precompiled support, dart2js support, and dart2wasm support into separate classes which implement a new `CompilerSupport` interface.
- Drops entirely support for `--pub-serve` mode. We could bring this back if necessary as another kind of `CompilerSupport` but I don't think it has any users (let me know if you feel different).
- Each `CompilerSupport` instance now gets its own server, browser manager, secret, etc. This resolves complications around knowing which kinds of compiler a given request is asking for, which the previous model couldn't handle well. You will get multiple browser instances if multiple compilers are used.
- ~Adds the "precompiled" compiler option. This isn't intended to be passed directly, it is inferred when passing the `--precompiled` argument. No other compiler options are allowed if it is used (although we could probably if desired?).~
- Dropped `isJS` and `isWasm` from `Runtime`, these no longer make sense. Added extensions to `Compiler` instead for convenience.

Likely there is some additional consolidation that could happen here, as there is a fair bit of duplicate logic across some of the implementations. But I am not sure if it is really worth it or not.


jakemac53:
TODO: Currently `isWasm` is tied to the Runtime, but it really shouldn't be. I need to fix how that works so the `wasm` selector works again.

kevmoo:
pkg:http at https://github.com/dart-lang/http/commit/064f5103a54743922198e35cad648247c8e69452

```
To run this test again: /Users/kevmoo/github/flutter/bin/cache/dart-sdk/bin/dart test test/html/client_conformance_test.dart -p chrome --plain-name 'close multiple close after request'
00:14 +96 ~13 -48: [Chrome, Dart2WASM] test/html/client_conformance_test.dart: close request after close [E]
  FormatException: Invalid radix-10 number (at character 1)
  50096.0
  ^

  unparsed  unparsed      at int._parseRadix (http://localhost:50014/3LkfHuzHF7HV3rWe9Lg3IlLmMJ2Lng3F/test/html/client_conformance_test.dart.browser_test.dart.wasm:wasm-function[1729]:0x74221)
  unparsed  unparsed      at int._parse (http://localhost:50014/3LkfHuzHF7HV3rWe9Lg3IlLmMJ2Lng3F/test/html/client_conformance_test.dart.browser_test.dart.wasm:wasm-function[1726]:0x74095)
  unparsed  unparsed      at int.parse (http://localhost:50014/3LkfHuzHF7HV3rWe9Lg3IlLmMJ2Lng3F/test/html/client_conformance_test.dart.browser_test.dart.wasm:wasm-function[1281]:0x656e0)
  unparsed  unparsed      at _Uri._makeHttpUri (http://localhost:50014/3LkfHuzHF7HV3rWe9Lg3IlLmMJ2Lng3F/test/html/client_conformance_test.dart.browser_test.dart.wasm:wasm-function[3784]:0xac09f)
  unparsed  unparsed      at _Uri.http (http://localhost:50014/3LkfHuzHF7HV3rWe9Lg3IlLmMJ2Lng3F/test/html/client_conformance_test.dart.browser_test.dart.wasm:wasm-function[3143]:0x9e103)
  unparsed  unparsed      at testClose closure at file:///Users/kevmoo/github/http/pkgs/http_client_conformance_tests/lib/src/close_tests.dart:45:33 inner (http://localhost:50014/3LkfHuzHF7HV3rWe9Lg3IlLmMJ2Lng3F/test/html/client_conformance_test.dart.browser_test.dart.wasm:wasm-function[5349]:0xccc85)
  unparsed  unparsed      at testClose closure at file:///Users/kevmoo/github/http/pkgs/http_client_conformance_tests/lib/src/close_tests.dart:45:33 (http://localhost:50014/3LkfHuzHF7HV3rWe9Lg3IlLmMJ2Lng3F/test/html/client_conformance_test.dart.browser_test.dart.wasm:wasm-function[5328]:0xcc5c0)
```

kevmoo:
pkg:web_socket_channel at https://github.com/dart-lang/web_socket_channel/commit/df096a9718e41f6883e683f177376a0cfd1b604c

```
00:03 +0 -1: [Chrome, Dart2WASM] test/html_test.dart: (setUpAll) [E]
  Type 'double' is not a subtype of type 'int' in type cast
  unparsed  unparsed      at _awaitHelper closure at org-dartlang-sdk:///lib/_internal/wasm/lib/async_patch.dart:82:16 (http://localhost:51661/GPLvykRlNyzVTKHYuiQNTOoWqFnLrW8R/test/html_test.dart.browser_test.dart.wasm:wasm-function[3706]:0xa11ae)
  unparsed  unparsed      at closure wrapper at org-dartlang-sdk:///lib/_internal/wasm/lib/async_patch.dart:82:16 trampoline (http://localhost:51661/GPLvykRlNyzVTKHYuiQNTOoWqFnLrW8R/test/html_test.dart.browser_test.dart.wasm:wasm-function[3709]:0xa1295)
  unparsed  unparsed      at _rootRunUnary (http://localhost:51661/GPLvykRlNyzVTKHYuiQNTOoWqFnLrW8R/test/html_test.dart.browser_test.dart.wasm:wasm-function[1020]:0x57b41)
  unparsed  unparsed      at _rootRunUnary tear-off trampoline (http://localhost:51661/GPLvykRlNyzVTKHYuiQNTOoWqFnLrW8R/test/html_test.dart.browser_test.dart.wasm:wasm-function[4122]:0xaa720)
  unparsed  unparsed      at _CustomZone.runUnary (http://localhost:51661/GPLvykRlNyzVTKHYuiQNTOoWqFnLrW8R/test/html_test.dart.browser_test.dart.wasm:wasm-function[3990]:0xa8766)
  unparsed  unparsed      at _FutureListener.handleValue (http://localhost:51661/GPLvykRlNyzVTKHYuiQNTOoWqFnLrW8R/test/html_test.dart.browser_test.dart.wasm:wasm-function[1014]:0x578df)
  unparsed  unparsed      at _Future._propagateToListeners closure at org-dartlang-sdk:///lib/async/future_impl.dart:836:33 (http://localhost:51661/GPLvykRlNyzVTKHYuiQNTOoWqFnLrW8R/test/html_test.dart.browser_test.dart.wasm:wasm-function[991]:0x5736e)
```

Both feel like weird double or int issues...obviously...

jakemac53:
Another open question here: How should you specify that you want to target precompiled `wasm` tests?

Maybe I should drop the precompiled compiler notion and allow special `--compiler` flags still in combination with it. Then the "PrecompiledSupport" class might accept in a compiler as configuration, but it would be chosen not based on the compiler but based on whether a precompiled dir was given. We wouldn't be able to validate necessarily that the correct compiler was actually used reliably, but we don't today anyways.

jakemac53:
I went ahead and dropped `Runtime.isJS` and `Runtime.isWasm`, as well as `Compiler.precompiled`. I think that is indeed the best path forward, but feel free to push back and I can revert it.

I am a bit concerned about the wire format breaking here (for Runtime in particular), this has caused problems in the past with flutter due to https://github.com/flutter/flutter/issues/91107. We might be able to make it more backwards compatible though?

pkgs/test/lib/src/runner/browser/compilers/compiler_support.dart
```diff
@@ -0,0 +1,31 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:test_api/backend.dart' show StackTraceMapper, SuitePlatform;
+import 'package:test_core/src/runner/suite.dart'; // ignore: implementation_imports
+import 'package:web_socket_channel/web_socket_channel.dart'; // ignore: implementation_imports
+
+/// The shared interface for all compiler support libraries.
+abstract class CompilerSupport {
```
natebosch, line 12:
```suggestion
interface class CompilerSupport {
```

to match the doc

pkgs/test/lib/src/runner/browser/compilers/compiler_support.dart
```diff
@@ -0,0 +1,31 @@
...
+  /// The URL at which this compiler serves its tests.
+  ///
+  /// Each compiler serves its tests under a different directory.
+  Uri get serverUrl;
+
+  /// Compiles [dartPath] using [suiteConfig] for [platform].
```
natebosch, line 18: Can you add a description of the format for `dartPath`? I assume it's something like "The relative path under [serverUrl] for a specific `.dart` test suite"?

pkgs/test/lib/src/runner/browser/compilers/compiler_support.dart
```diff
@@ -0,0 +1,31 @@
...
+  Future<void> compileSuite(
+      String dartPath, SuiteConfiguration suiteConfig, SuitePlatform platform);
+
+  /// Retrieves a stack trace mapper for [dartPath] if available.
```
natebosch, line 22: Why do different paths have different stack trace mappers? I would have expected 1 per platform, not 1 per path.

pkgs/test_core/lib/src/runner/configuration/args.dart
```diff
@@ -413,7 +419,9 @@ extension _RuntimeDescription on Runtime {
     var message = StringBuffer('[$identifier]: ');
     message.write('${defaultCompiler.identifier} (default)');
     for (var compiler in supportedCompilers) {
-      if (compiler == defaultCompiler) continue;
+      if (compiler == defaultCompiler) {
+        continue;
+      }
```
natebosch, line 424: FWIW I prefer the current single line version.

natebosch (approved):

natebosch:
> * Dropped `isJS` and `isWasm` from `Runtime`

We should double check if this impacts flutter usage. Have you tried this in google3 yet? We might want to create a CL for it  and run through a presubmit before landing on github.

> I made this backwards compatible (I think?) and pinged the flutter issue.

The json protocol is backwards compatible - do we have any Dart API compatibility concerns?

natebosch:
I think this all looks good assuming it doesn't cause any issues rolling to flutter or google3 - I'm nervous about dropping support for running tests from an arbitrary server, but I do prefer removing it and seeing if anyone complains over assuming it's working untested.

jakemac53:
> I think this all looks good assuming it doesn't cause any issues rolling to flutter or google3

Yes, I definitely plan on validating these changes internally, in the Dart SDK, and in flutter before publishing.

> I'm nervous about dropping support for running tests from an arbitrary server, but I do prefer removing it and seeing if anyone complains over assuming it's working untested.

Yeah, I feel like we have kicked the can down the road a lot on this, and I want to see if anybody even notices. We definitely can bring back support if needed, and I will sign up for doing that.

pkgs/test/lib/src/runner/browser/compilers/compiler_support.dart
jakemac53: Each stack trace mapper corresponds to a single source map

pkgs/test_core/lib/src/runner/configuration/args.dart
jakemac53: Fixed, I had made this more complicated originally so it didn't fit on one line, but now it does

pkgs/test_core/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
-## 0.5.10-wip
+## 0.6.0-wip
 
 * Handle paths with leading `/` when spawning test isolates.
+* Added `dart2wasm` as a supported compiler for the `chrome` runtime.
```
kevmoo, line 4: And not firefox?

jakemac53: Does it work? If it works, I can add it.

kevmoo: with stable FF, yep. Haven't tried it with test yet...

jakemac53: I will just add it and see what happens :)

jakemac53: aaand I hit a race condition with the auto submit bot so its landed, 🤣

jakemac53: but, it works!

Issue #2142, opened 2023-11-28
VM platform fails to compile tests under `lib`


koutaro-masaki:
I want to follow the principle of colocation and place test code next to the code being tested. However, when the code being tested uses a combination of relative and absolute imports within a dependency, and also involves class inheritance, the inheritance hierarchy breaks during test execution, resulting in an error. Is there a way to avoid this?

While importing the code being tested with absolute imports from the test code prevents errors, I would prefer not to do this if possible, as it might not clearly indicate that the code being tested is imported from the colocated test. Alternatively, it is possible to avoid errors by enforcing the use of either absolute or relative imports exclusively within the library code, but this is not practical.

This is the minimal reproducing code.
https://github.com/koutaro-masaki/dart_test_issue_sample

```
00:01 +0 -1: loading lib/bar_test.dart [E]                                                                                                                                                                                             
  Failed to load "lib/bar_test.dart":
  lib/bar.dart:6:12: Error: The argument type 'Bar' can't be assigned to the parameter type 'Foo'.
   - 'Bar' is from 'lib/bar.dart'.
   - 'Foo' is from 'package:dart_test_issue_sample/foo.dart' ('lib/foo.dart').
    return f(Bar());
             ^
```

jakemac53:
> I want to follow the principle of colocation and place test code next to the code being tested.

Following this strictly will not work well in dart, because it doesn't follow our [package layout conventions](https://dart.dev/tools/pub/package-layout).

And in fact, doing this for a package dependency would have concrete negative effects (dependencies which should be dev only dependencies for the package would become real dependencies).

What many packages instead do is mirror the structure of their `lib` directory under their `test` directory, which still allows you to easily find the tests for any given file based on convention, so `lib/bar.dart` would be tested at `test/bar_test.dart` for instance.

jakemac53:
It is worth noting though that typically a Uri like `lib/bar.dart` should have been converted to the corresponding `package:` uri, to avoid the issue you are seeing. I am not sure exactly why that isn't happening, maybe frontend_server (which we use for compilation) does not automatically do this conversion like `dart compile <path>` does, and we should be doing it before calling frontend_server.

koutaro-masaki:
We believe that mirroring poses challenges in terms of operational and maintenance aspects of the codebase, and this is precisely why it requires colocation. 
https://kentcdodds.com/blog/colocation#tests

[...]

PR #2147, opened 2023-12-05
Add regression test for relative imports in tests under `lib`

jakemac53:
Validates the issue listed in https://github.com/dart-lang/test/issues/2142.

Skipped for now since it fails in all configurations.


integration_tests/regression/test/issue_2142_test.dart
```diff
@@ -0,0 +1,15 @@
+@Skip('https://github.com/dart-lang/test/issues/2142')
+library;
+
+import 'package:test/test.dart';
+
+// ignore: avoid_relative_lib_imports
+import '../lib/issue_2142/import.dart';
```
jakemac53, line 7: Note that this one probably would be broken either way?

natebosch: Yes, I think we should not attempt to support `../lib`. I'd be happy to check for this and block it if it's easy.

natebosch:
Oops, I meant to push a new branch and open a PR to pull into this branch, I hadn't intended to push on this branch. Sorry @jakemac53 !

jakemac53:
> Oops, I meant to push a new branch and open a PR to pull into this branch, I hadn't intended to push on this branch. Sorry @jakemac53 !

No worries, should I revert or want to land that together with the test?

natebosch:
> No worries, should I revert or want to land that together with the test?

 I reverted it on this branch. I think we should go ahead and land the test and then handle the fix in a separate PR.

integration_tests/regression/lib/issue_2142/test.dart
```diff
@@ -0,0 +1,13 @@
+@Skip('https://github.com/dart-lang/test/issues/2142')
```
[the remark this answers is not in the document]
jakemac53: Presumably (but I have no idea). I will add one.

natebosch (approved):

PR #2148, opened 2023-12-05
Use package: URIs for imports when possible

natebosch:
When a compiler is reading imports it's possible for a relative file
import and absolute package import to the same library to get treated as
separate libraries. The VM works around this in the case of a `main`
under `lib/` by checking when the entrypoint, and only the entrypoint,
could be referenced with a `package:` URI so that further relative URIs
resolved from that point are consistently treated as `package:` URIs.

Use the same workaround for bootstrap files. Add an `absoluteUri`
utility in `test_core` which prefers `package:` URIs when feasible, and
falls back to absolute file URIs otherwise. When this initial import
uses a `package:` URI it will make the subsequent relative imports
correctly resolve as `package:` URIs.

Add `async` and `await` as necessary since we're currently using the
async APIs to read the package config.


jakemac53 (approved):

pkgs/test_core/lib/src/util/package_config.dart
```diff
@@ -18,3 +20,15 @@ final Future<Uri> packageConfigUri = () async {
   }
   return uri;
 }();
+
+final _originalWorkingDirectory = Directory.current.uri;
```
jakemac53, line 24: I am seeing some breakage internally, I think the issue is that for precompiled tests we need to use the precompiled dir instead of the current dir?

jakemac53: Ok, so the issue is actually not that, the this logic is working correctly to convert a URI to a package: URI, but the test they are trying to launch actually lives in another package. It is getting converted to a package: uri (as it should) and then failing to load for some reason.

PR #2153, opened 2023-12-08
Update frontend_server_client constraint to allow version 4.0.0

derekxu16:
Updating to `package:frontend_server_client v4.0.0` is only a breaking change for dependents that set the `debug` argument of `FrontendServerClient.start` to true, and`package:test_core` never does that.

### DIFF
--- a/pkgs/test_core/pubspec.yaml
+++ b/pkgs/test_core/pubspec.yaml
@@ -13,7 +13,7 @@ dependencies:
   boolean_selector: ^2.1.0
   collection: ^1.15.0
   coverage: ^1.0.0
-  frontend_server_client: '>=3.2.0 <4.0.0'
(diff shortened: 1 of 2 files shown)


pkgs/test_core/pubspec.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
   boolean_selector: ^2.1.0
   collection: ^1.15.0
   coverage: ^1.0.0
-  frontend_server_client: '>=3.2.0 <4.0.0'
+  frontend_server_client: '>=3.2.0 <=4.0.0'
```
jakemac53, line 16:
```suggestion
  frontend_server_client: '>=3.2.0 <5.0.0'
```

jakemac53 (approved):

PR #2154, opened 2023-12-11
Declare addTearDownLibrary.

polina-c:
Prerequisite for https://github.com/flutter/flutter/pull/138057.
TODO: add test coverage.

### DIFF
--- a/pkgs/test_api/lib/src/backend/declarer.dart
+++ b/pkgs/test_api/lib/src/backend/declarer.dart
@@ -309,6 +309,13 @@ class Declarer {
   void addTearDownAll(dynamic Function() callback) =>
       _tearDownAlls.add(callback);
 
+  /// Registers a function to be called after all tests in the test library are completed.
+  void addTearDownLibrary(dynamic Function() callback) {
+    var declarer = this;
+    while (declarer._parent != null) declarer = declarer._parent;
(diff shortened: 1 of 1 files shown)


natebosch:
Why do we need this over `setUpAll` and `addTearDown`?

natebosch:
> > Why do we need this over `setUpAll` and `addTearDown`?
> 
> Because we need it to tear down all tests in the library, not all tests in the group. I do not see other way to traverse to root group. Am i missing something?

I thought  the plan was to intercept the call to `group` and inject the `setUpAll` call?

If we want to avoid that because it's too brittle we can try something like this. I'd go with the name `addTearDownSuite`. @jakemac53 any concerns about such an API?

natebosch:
Ah this isn't user facing either. Should be pretty safe then. Let's  rename to "suite" and add a changelog entry.

polina-c:
> Ah this isn't user facing either. Should be pretty safe then. Let's rename to "suite" and add a changelog entry.

I will be able to use it only with added method `addTearDownSuite` to flutter_test/lib/src/test.compat.dart, that will make it user facing.

Or we can go with how it is currently implemented in https://github.com/flutter/flutter/pull/138057 (intercept the call to group and other user calls), that is little hacky, but not user facing. 

I am ok with both ways.

PR #2157, opened 2023-12-13
Update the package:vm_service constraint to allow version 14.x

derekxu16:


jakemac53:
It looks like CI is not actually pickup up the latest version here - are we confident it will work?

derekxu16:
> It looks like CI is not actually pickup up the latest version here - are we confident it will work?

Ah, that's because `package:coverage` needs to have its `package:vm_service` constraint updated as well. I'm in the process of doing that. I am confident that this will work.

jakemac53 (approved):

PR #2158, opened 2023-12-13
handle missing package config files

jakemac53:
Should fix an internal breakage with the latest release

### DIFF
--- a/pkgs/test_core/lib/src/util/package_config.dart
+++ b/pkgs/test_core/lib/src/util/package_config.dart
@@ -29,6 +29,11 @@ Future<Uri> absoluteUri(String path) async {
   final uri = p.toUri(path);
   final absoluteUri =
       uri.isAbsolute ? uri : _originalWorkingDirectory.resolveUri(uri);
-  final packageConfig = await currentPackageConfig;
-  return packageConfig.toPackageUri(absoluteUri) ?? absoluteUri;
+  try {
+    final packageConfig = await currentPackageConfig;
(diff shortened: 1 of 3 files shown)


natebosch (approved):

kevmoo (approved):

pkgs/test_core/lib/src/util/package_config.dart
```diff
@@ -29,6 +29,11 @@ Future<Uri> absoluteUri(String path) async {
   final uri = p.toUri(path);
   final absoluteUri =
       uri.isAbsolute ? uri : _originalWorkingDirectory.resolveUri(uri);
-  final packageConfig = await currentPackageConfig;
-  return packageConfig.toPackageUri(absoluteUri) ?? absoluteUri;
+  try {
+    final packageConfig = await currentPackageConfig;
+    return packageConfig.toPackageUri(absoluteUri) ?? absoluteUri;
+  } on StateError catch (_) {
```
kevmoo, line 35:
```suggestion
  } on StateError {
```

Doesn't this work? 🤷 

Not a big deal AT ALL

jakemac53: I think that is fine, but I don't want to wait for CI again as I want to get this merged internally to unblock some folks. I will follow up with a fix :)

jakemac53: https://github.com/dart-lang/test/pull/2159

PR #2170, opened 2024-01-09
fix a bug where test html files were not created in precompiled mode

jakemac53:
I think this should fix the failures we have been seeing on the bots for package:build.

Moved the logic for the test html wrapper into the base class for all modes of browser tests, so that it can be shared with the precompiled mode.  The dart2wasm mode overrides that behavior since it does things a bit different.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,4 +1,7 @@
-## 1.25.1-wip
+## 1.25.1
+
+* Fix a bug where in precompiled mode, html files for tests were no longer
+  created.
 
 ## 1.25.0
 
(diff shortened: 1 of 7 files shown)


jakemac53:
Note that I will merge this internally before publishing, but I do plan to land it here first to simplify testing internally

natebosch (approved):
Consider whether this would work to implement the shared part in a `mixin` so we can keep the parent class an interface.

jakemac53:
> Consider whether this would work to implement the shared part in a `mixin` so we can keep the parent class an interface.

Actually, having separate mixins for the wasm vs dart2js bootstrapping would allow for PrecompiledDart2Wasm and PrecompiledDart2Js pretty reasonably as well, so I think I will do that.

jakemac53:
@natebosch PTAL, I did create two new mixins, for the wasm/dart2js bootstrapping, but left the fields on the base class to support those mixins. I could make those abstract getters for a pure interface, but then I would have to add them to all the concrete subclasses, so I think it makes sense to define them in the base class.

natebosch (approved):

PR #2172, opened 2024-01-15
Remove reference to old Stream Matchers section from README.md

maBarabas:
Remove reference to Stream Matchers section removed in #1969


natebosch:
Sorry for the slow response here - I think we had another PR which removed this link.

PR #2184, opened 2024-02-13
Register a service extension to return the test target library

kenzieschmoll:
This adds a service extension on the test runner isolate that will allow other tools (e.g. DevTools) to get the test target root from the test runner isolate. See https://github.com/flutter/devtools/pull/7206 for an example of how this service extension will be used.

Fixes https://github.com/flutter/flutter/issues/143170.

---

<details>

</details>

### DIFF
--- a/pkgs/test_core/lib/src/executable.dart
+++ b/pkgs/test_core/lib/src/executable.dart
@@ -150,6 +150,7 @@ Future<void> _execute(List<String> args) async {
   try {
     runner = Runner(configuration);
     _setupTestTargetLibraryExtension(configuration);
+    print('registering extension');
     exitCode = (await runner.run()) ? 0 : 1;
   } on ApplicationException catch (error) {
     stderr.writeln(error.message);
--- a/pkgs/test_core/lib/src/executable.dart
+++ b/pkgs/test_core/lib/src/executable.dart
@@ -204,9 +204,10 @@ void _setupTestTargetLibraryExtension(Configuration configuration) {
   // [configuration.testSelections] must be part of the same Dart package.
   final testTarget = configuration.testSelections.keys.first;
   final testTargetAbsolutePath = File(testTarget).absolute.path;
+  final packageRootFileUri = Uri.file(testTargetAbsolutePath).toString();
   registerExtension('ext.test.testTargetLibrary', (method, parameters) async {
     return ServiceExtensionResponse.result(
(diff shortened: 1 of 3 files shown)


pkgs/test_core/lib/src/executable.dart
```diff
@@ -194,3 +197,16 @@ Usage: dart test [files or directories...]
 ${Configuration.usage}
 ''');
 }
+
+void _setupTestTargetLibraryExtension(Configuration configuration) {
+  // TODO(kenz): do we need to get a common prefix for all targets, or is
+  // using the first target sufficient? This is assuming all test targets in
+  // [configuration.testSelections] must be part of the same Dart package.
+  final testTarget = configuration.testSelections.keys.first;
+  final testTargetAbsolutePath = File(testTarget).absolute.path;
```
kenzieschmoll, line 206: @jakemac53 I'm returning a single test target here, assuming that all test targets will be part of the same Dart package. What I'd really like to do is return the package root, but I wanted to get your feedback on what we can and cannot assume here. 

Is a test target's path guaranteed to have `test` on it? If so, could we assume the parent of the `test` directory is the package root?

Otherwise, we could walk up the path until we find a directory that also contains the `.dart_tool/package_config.json` file.

jakemac53: I am pretty sure you can just use the current directory, but I need to confirm that. I don't believe we support running from anything other than the root directory of the package

jakemac53: Tests are allowed to be anywhere so there are no guarantees with that, and I am not sure if there are always even any test selections, unless we plug in a default, which we might.

jakemac53: Fwiw, I did confirm that `dart test` at least only allows running from the package root directory, so this can always just return the current working dir.

natebosch: > (or maybe make that public)

This sounds like a good idea. I think we still have ways the platforms can interact poorly with eachother, and storing a process global variable with the initial working directory is a good way to solve that problem.

@kenzieschmoll Can you elaborate on how devtools will use this? I hope we can find a more concrete name than "target library" which doesn't mean anything specific to me.

kenzieschmoll:
Closing in favor of a different solution. See https://github.com/flutter/flutter/issues/143170 for context.

PR #2190, opened 2024-02-29
Add failure only reporter

### LINKED ISSUE #829: Add a quieter reporter that only prints information for failures

natebosch:
Closes #829

Copies the expanded reporter and test and removes output unrelated to
failed tests.

### DIFF
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -35,7 +35,7 @@ dependencies:
 
   # Use an exact version until the test_api and test_core package are stable.
   test_api: 0.7.0
-  test_core: 0.6.0
+  test_core: 0.6.1
 
   typed_data: ^1.3.0
   web_socket_channel: ^2.0.0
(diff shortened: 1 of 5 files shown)


natebosch:
Some details worth getting agreement on:
- Naming. I'm going with `-r failures-only` to be specific, but if anyone thinks we should do `--quiet` or anything else we should pick a good name to start with.
- This currently doesn't print skips since a skip doesn't cause a failure. We still show the number skipped at the end of the test.
- We do print the final line with the number of passing/failing/skipped tests. So there is _some_ output even for fully passing suites. We could instead be _totally_ silent for passing runs, but I think the main goal is to reduce noise, and the summary line carries a lot of signal.

jakemac53 (approved):

natebosch:
We also print the time since starting the test run today - looking at this again I'm leaning towards removing that. In the compact reporter it makes sense because we keep updating the same display of the time on the current line in the typical case. In the expanded reporter it can be used to see the delta between a test start and test failure. Absent other information, knowing the time since test start for an arbitrary failure is not useful.

Knowing the full runtime could be useful, but can also be easily measured externally (as long as it's not a problem picking up the test runner overhead).

jakemac53 (approved):

PR #2198, opened 2024-03-22
Try removing isolate.kill()

natebosch:
At one point isolates were killed when the loader was closed after tests
are down. Later the kill was moved to when the test suite is done. There
was no specific comment about the `isolate.kill()` changes.
https://codereview.chromium.org//1206033004

Before that the `isolate.kill` on close goes back to the original
implementation of the loader.
https://codereview.chromium.org//920703006

No internal tests seem to be relying on this behavior, and nothing times
out when it is removed.

Check whether there are any tests on CI that validate this behavior.


natebosch:
After some internal discussion and thinking more about the intent in https://codereview.chromium.org//1206033004 this `isolate.kill` is not here for correctness, but to reduce concurrent resource use when running multiple VM tests suites in sequence. It probably isn't tested because that's a hard test to write.

We should consider writing an integration test that uses the VM service to count how many isolates are active or something along those lines.

PR #2203, opened 2024-04-03
support enabling experiments when using the dart2wasm compiler

jakemac53:
We don't appear to have any tests for experiments right now, nor any really good way of adding them. I did manually test though 🤷‍♂️


natebosch (approved):

jakemac53:
FYI @kevmoo `dart --enable-experiment=macros test -c dart2wasm` works now too after this :)

PR #2226, opened 2024-05-13
Add test coverage for test suite import prefix value

kenzieschmoll:
Addresses a TODO from https://github.com/flutter/flutter/issues/143170#issuecomment-2108312624.  The matching Flutter Tools PR is here: https://github.com/flutter/flutter/pull/148244


pkgs/test/lib/src/runner/browser/compilers/dart2js.dart
```diff
@@ -115,7 +116,7 @@ class Dart2JsSupport extends CompilerSupport with JsHtmlWrapper {
         import 'package:test/src/bootstrap/browser.dart';
         import 'package:test/src/runner/browser/dom.dart' as dom;
 
-        import '${await absoluteUri(dartPath)}' as test;
+        import '${await absoluteUri(dartPath)}' as $testLibraryImportPrefix;
```
jakemac53, line 119: I would call this `testSuiteImportPrefix`, `testLibrary` sounds like an import of package:test

pkgs/test/test/runner/prefix_test.dart
```diff
@@ -0,0 +1,19 @@
...
+import 'package:test/test.dart';
+import 'package:test_core/src/util/prefix.dart';
+
+void main() {
+  test('testLibraryImportPrefix value is "test"', () {
+    expect(
+      testLibraryImportPrefix,
+      'test',
+      reason:
+          'testLibraryImportPrefix must be equal to the String "test". Some '
```
jakemac53, line 14: We should call out exactly which tools, so that if we need to change it we know what to fix.

jakemac53 (commented):
While this could work, this is definitely not intended to be a public part of our API, it is an implementation detail only.

So from that perspective I would definitely prefer any other option. If there is no other solution that works, we could consider it.

This may not work for other integrations either - such as build_test, flutter, google3, or any other custom integrations with package:test.

jakemac53:
cc @natebosch

kenzieschmoll:
The matching Flutter Tools PR is here: https://github.com/flutter/flutter/pull/148244

jakemac53:
Note that this also currently precludes us from adding a `show main` or something like that, which tbh we probably should be adding currently.

kenzieschmoll:
> Note that this also currently precludes us from adding a `show main` or something like that, which tbh we probably should be adding currently.

How so? `import 'some_uri' as test show main;` should still work fine.

jakemac53:
> How so? `import 'some_uri' as test show main;` should still work fine.

I believe the code I saw was looking for `as test;` specifically (with the semicolon). I think it also probably would get broken by any changes to whitespace.

natebosch:
I left a comment on the other issue, but we don't currently restrict test suites to those under the package root, so I don't think you can guarantee that this URI will be useful.

kenzieschmoll:
> > How so? `import 'some_uri' as test show main;` should still work fine.
> 
> I believe the code I saw was looking for `as test;` specifically (with the semicolon). I think it also probably would get broken by any changes to whitespace.

We are checking the prefix directly, so this should handle whitespace changes like adding `show` or `hide` items.

kenzieschmoll:
Will hold off on landing this until the discussion at https://github.com/flutter/flutter/issues/143170#issuecomment-2108839023 is resolved.

kenzieschmoll:
Closing this since this PR since this issue was resolved with a different solution than using the test import prefix: https://github.com/dart-lang/test/issues/2246

PR #2232, opened 2024-05-31
Fix some lints

kevmoo:


kevmoo:
Playing with enabling some lints

natebosch (approved):

PR #2239, opened 2024-06-04
Remove an unnecessary `dynamic`

natebosch:
The `dynamic` was used to satisfy the purpose that is more directly
expressed with `FutureOr`.

Replace the private `_Command` class with a typedef record with named fields.

### DIFF
--- a/pkgs/test_core/lib/src/runner/console.dart
+++ b/pkgs/test_core/lib/src/runner/console.dart
@@ -47,9 +47,9 @@ class Console {
   ///
   /// The [description] should be a one-line description of the command to print
   /// in the help output. The [body] callback will be called when the user types
-  /// the command, and may return a [Future].
+  /// the command.
   void registerCommand(
-      String name, String description, dynamic Function() body) {
+      String name, String description, FutureOr<void> Function() body) {
     if (_commands.containsKey(name)) {
       throw ArgumentError('The console already has a command named "$name".');
     }
@@ -111,7 +111,7 @@ class _Command {
   final String description;
 
   /// The callback to run when the command is invoked.
-  final Function body;
+  final FutureOr<void> Function() body;
 
   _Command(this.name, this.description, this.body);
 }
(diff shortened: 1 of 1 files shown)


kevmoo (approved):
needs a rebase, but yes!

natebosch:
I expanded the scope a bit from the original review. @jakemac53 WDYT?

pkgs/test_core/lib/src/runner/console.dart
```diff
@@ -101,17 +102,3 @@ class Console {
     }
   }
 }
-
-/// An individual console command.
-class _Command {
```
kevmoo, line 106: Mmm...records!

kevmoo (approved):

pkgs/test_core/lib/src/runner/console.dart
```diff
@@ -13,7 +13,8 @@ import '../util/io.dart';
 /// An interactive console for taking user commands.
 class Console {
   /// The registered commands.
-  final _commands = <String, _Command>{};
+  final _commands = <String,
```
jakemac53, line 16: Fwiw, I don't like having the record type jammed in here - I would prefer a typedef for the record type?

jakemac53 (approved):

pkgs/test_core/lib/src/runner/console.dart
natebosch: Added a typedef. I tried also with a trailing comma at the inline usage site and while I think it's a bit better than this version I don't think I like it as much as the typedef.

pkgs/test_core/lib/src/runner/console.dart
```diff
@@ -102,16 +102,8 @@ class Console {
...
-
-  /// The single-line description of the command.
-  final String description;
-
-  /// The callback to run when the command is invoked.
-  final dynamic Function() body;
-
-  _Command(this.name, this.description, this.body);
-}
+typedef _Command = ({
```
kevmoo, line 105: That works, too!

PR #2242, opened 2024-06-10
Bump min Dart version to 3.4.0 for WASM

Rexios80:
Before 3.4.0:
```console
  Could not find an option with short name "-O".
  
  Usage: dart compile wasm [arguments] <dart entry point>
```

This option was added in commit [92d53c426f9b4764e5c34d33ef7bbcf5fd3b12a6](https://github.com/dart-lang/sdk/commit/92d53c426f9b4764e5c34d33ef7bbcf5fd3b12a6)

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -308,181 +308,6 @@ jobs:
       - job_003
       - job_004
   job_007:
-    name: "unit_test; linux; Dart 3.2.0; PKG: pkgs/test; `xvfb-run -s \"-screen 0 1024x768x24\" dart test --preset travis --total-shards 5 --shard-index 0`"
-    runs-on: ubuntu-latest
-    steps:
-      - name: Cache Pub hosted dependencies
-        uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9
-        with:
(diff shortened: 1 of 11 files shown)


jakemac53:
Thanks, unfortunately we can't fix this for the old published versions, and are past the retraction window for 1.25.6. Because of that I won't bother publishing this change right now, but we should land it nonetheless.

jakemac53:
adding @natebosch for additional review since I pushed a bunch of other related changes here

Rexios80:
Should we update CI to run at the lowest resolvable Dart SDK version to avoid this in the future?

jakemac53:
> Should we update CI to run at the lowest resolvable Dart SDK version to avoid this in the future?

It does already - the problem is wasm is tested through the `integration_tests/wasm/` package, which _was_ updated to require 3.4.0. This was just missed in code review.

We should also add at least a basic wasm test in the actual `test` package probably though.

natebosch (approved):

jakemac53:
It looks like these failures are a result of the change to running on `3.4.0`, so something has changed such a that only certain tests now fail on windows.

We will separately address this problem, landing this for now, on red.

Issue #2243, opened 2024-06-10
Failing windows tests on 3.4.0


jakemac53:
See failures here https://github.com/dart-lang/test/actions/runs/9452121663/job/26034666073?pr=2242.

```
Cannot copy file to 'C:\Users\RUNNER~1\AppData\Local\Temp\dart_test.kernel.4322a797\output.dill', path = 'C:\Users\runneradmin\AppData\Local\Temp\dart_test_fac3c362\.dart_tool\test\incremental_kernel.
```

There is some issue when trying to copy over the incremental dill file. Notably there is a missing hash after the `incremantal_kernel.`. This could happen if the package config fails to find the root package (see  https://github.com/dart-lang/test/blob/master/pkgs/test_core/lib/src/runner/package_version.dart#L21).

I am not sure if this is the core issue, but it could be related.

PR #2244, opened 2024-06-10
try to fix windows failures

jakemac53:
Fixes https://github.com/dart-lang/test/issues/2243

I can't explain why this would only break on 3.4.0 and later, still looking into that.

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.6.5-wip
 
 * Increase SDK constraint to ^3.4.0.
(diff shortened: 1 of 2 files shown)


jakemac53:
Ok, https://gist.github.com/jakemac53/845b60847dbda53f93aeed8056d5fbd2 illustrates what I think is the core of the issue.

I would expect this to print `false` and also hit the catch block for every case. However, on 3.4.0, in the very first case the copy actually succeeds even though it claims the file doesn't exist. I filed https://github.com/dart-lang/sdk/issues/55972 about this.

natebosch (approved):

PR #2251, opened 2024-07-09
Expand error message for async test declaration

natebosch:
Add suggestions for fixing the declared tests. Differentiate the message
based on whether the test `main` is running within the test runner, or
as a standalone executable.


jakemac53 (approved):

natebosch:
I opened https://github.com/dart-lang/test/issues/2253 with some initial thoughts from our discussion so far about how we might want to handle this in the future.

For now I think I'll land this PR and not make further changes to the runner. We can re-evaluate if we hit errors anywhere we need to run standalone.

PR #2258, opened 2024-07-22
Migrate off deprecated Analyzer API

natebosch:
ContextLocator is deprecated in favor of AnalysisContextCollection.

### DIFF
--- a/pkgs/test_api/test/import_restrictions_test.dart
+++ b/pkgs/test_api/test/import_restrictions_test.dart
@@ -68,8 +68,7 @@ class _ImportCheck {
     final contexts =
         AnalysisContextCollection(includedPaths: [packagePath]).contexts;
     if (contexts.length != 1) {
-      throw StateError(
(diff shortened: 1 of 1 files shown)


scheglov (approved):

kevmoo:
Should we change the lower bound on the analyzer pkg? Did we double check?

natebosch:
Good call, we almost certainly do need to change the lower bound (I wouldn't be surprised if it was already too low - it's a `dev_dependency` and we pretty much always end up with latest compatible on CI anyway)

natebosch:
Looks like the API we migrated to has been around for a while.

Introduced in https://github.com/dart-lang/sdk/commit/2e7b0a13f10a

At that time the version was `0.32.4`. https://github.com/dart-lang/sdk/blob/2e7b0a13f10a03c347cb936354e33564d146b915/pkg/analyzer/pubspec.yaml#L2

We should be OK for this PR.

PR #2259, opened 2024-07-26
Support dart2wasm in node.js tests

simolus3:
Support `dart2wasm` as a compiler for tests running in Node.js. `dart2wasm` emits a `.mjs` file exporting definitions to load generated wasm modules, the only additional thing we have to do is wrap that in a simple entrypoint file compiling the wasm module and invoking the startup wrapper.

There's no support for stack trace maps yet. We also don't support precompiled node wasm tests yet (`dart2wasm` is also not currently supported by `build_web_compilers`, so we're blocked on that either way).

In the test runtime, I had to migrate off `require` as that function is not in the global context for `.mjs` files. It looks like we can use `await import` instead though. If we need to support ancient Node versions that lack `import` support, I can adapt that to still use `require` when compiled with `dart2js` for compatibility.

### DIFF
--- a/pkgs/test/test/runner/node/runner_test.dart
+++ b/pkgs/test/test/runner/node/runner_test.dart
@@ -63,6 +63,16 @@ String? skipBelowMajorNodeVersion(int minimumMajorVersion) {
   return null;
 }
 
+String? skipAboveMajorNodeVersion(int maximumMajorVersion) {
+  final (:major, :full) = _nodeVersion ??= _readNodeVersion();
+  if (major > maximumMajorVersion) {
+    return 'This test requires Node $maximumMajorVersion.x or older, '
+        'but is running on $full';
+  }
+
+  return null;
+}
+
 void main() {
   setUpAll(precompileTestExecutable);
 
@@ -226,6 +236,37 @@ void main() {
     expect(test.stdout, emitsThrough(contains('+1 -2: Some tests failed.')));
     await test.shouldExit(1);
   }, skip: skipBelowMajorNodeVersion(22));
+
+  test(
+    'gracefully handles wasm errors on old node versions',
+    () async {
+      // Old Node.JS versions can't read the WebAssembly modules emitted by
+      // dart2wasm. The node process exits before connecting to the server
+      // opened by the test runner, leading to timeouts. So, this is a
+      // regression test for https://github.com/dart-lang/test/pull/2259#issuecomment-2307868442
+      await d.file('test.dart', '''
+        import 'package:test/test.dart';
+
+        void main() {
+          test("test", () {
+            // Should pass on newer node versions
+          });
+        }
+      ''').create();
+
+      var test = await runTest(['-p', 'node', '-c', 'dart2wasm', 'test.dart']);
+      expect(
+        test.stdout,
+        emitsInOrder([
+          emitsThrough(
+              contains('Node exited before connecting to the test channel.')),
+          emitsThrough(contains('-1: Some tests failed.')),
+        ]),
+      );
+      await test.shouldExit(1);
+    },
+    skip: skipAboveMajorNodeVersion(21),
+  );
 
   test('forwards prints from the Node test', () async {
     await d.file('test.dart', '''
--- a/pkgs/test/test/runner/node/runner_test.dart
+++ b/pkgs/test/test/runner/node/runner_test.dart
@@ -116,6 +116,15 @@ void main() {
       expect(test.stdout, emitsThrough(contains('+1: All tests passed!')));
       await test.shouldExit(0);
     });
+
+    test('compiled with dart2wasm', () async {
+      await d.file('test.dart', _success).create();
+      var test =
+          await runTest(['-p', 'node', '--compiler', 'dart2wasm', 'test.dart']);
+
+      expect(test.stdout, emitsThrough(contains('+1: All tests passed!')));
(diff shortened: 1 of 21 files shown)


pkgs/test/test/runner/node/runner_test.dart
```diff
@@ -148,8 +157,18 @@ void main() {
...
+      'node',
+      '-p',
+      'vm',
+      '-c',
+      'dart2js',
+      '-c',
+      'dart2wasm',
+      'test.dart'
+    ]);
+    expect(test.stdout, emitsThrough(contains('+1 -2: Some tests failed.')));
```
jakemac53, line 171: Two of these should pass right?

jakemac53: I think maybe the environment variable isn't being set for dart2wasm

jakemac53:
I am generally comfortable with this, yes.

pkgs/test/test/runner/node/runner_test.dart
simolus3: I think it's correct? The test is

```
if (const bool.fromEnvironment("node")) {
  throw TestFailure("oh no");
}
```

so it should _fail_ on node. Previously it was +1/-1 because the configurations were VM and dart2js+node. Now it's +1/-2 because dart2wasm+node as added.

jakemac53: Oh, geeze I misread it. Then yes this seems correct :)

simolus3:
Speaking of old node versions, the `ubuntu-latest` image on GitHub actions ships Node.js version 18 which is too old to support the WebAssembly modules generated by `dart2wasm`.

I guess we can use `actions/setup-node`, but I'm not sure if `mono_repo` allows adding custom actions steps to stages? And that's probably the kind of maintenance burden you were hoping to avoid...

jakemac53:
> I guess we can use `actions/setup-node`, but I'm not sure if `mono_repo` allows adding custom actions steps to stages? And that's probably the kind of maintenance burden you were hoping to avoid...

Hmm, yeah I don't think it supports that.

simolus3:
I have changed the tests to skip `dart2wasm` if we're running on older node versions. That's not great because it means that this feature is not tested in the CI, but it also means that we won't have to change the CI setup (and at some point `ubuntu-latest` will ship a recent node version).

natebosch (approved):

natebosch:
I fixed up some of the failing CI jobs. It looks like there are still some legitimate failures.

```
[Node.js, Dart2Wasm] loading test/hello_world_test.dart (failed)
  Generated wasm module '/tmp/dart_test_WWPNVK/test_KVONWA/hello_world_test.dart.node_test.dart.wasm.wasm', and JS init file '/tmp/dart_test_WWPNVK/test_KVONWA/hello_world_test.dart.node_test.dart.wasm.mjs'.
  node:internal/process/promises:288
              triggerUncaughtException(err, true /* fromPromise */);
              ^
  [CompileError: WebAssembly.compileStreaming(): expected signature definition 0x60, got 0x4e @+13]
  Node.js v[18](https://github.com/dart-lang/test/actions/runs/10523203385/job/29176912922?pr=2259#step:6:21).20.4
  TimeoutException after 0:12:00.000000: Test timed out after 12 minutes.
  package:test_api/src/backend/invoker.dart 338:28  Invoker._handleError.<fn>
  dart:async/zone.dart 1391:47                      _rootRun
  dart:async/zone.dart 1301:[19](https://github.com/dart-lang/test/actions/runs/10523203385/job/29176912922?pr=2259#step:6:22)                      _CustomZone.run
  package:test_api/src/backend/invoker.dart 336:10  Invoker._handleError
  package:test_api/src/backend/invoker.dart 291:9   Invoker.heartbeat.<fn>.<fn>
  dart:async/zone.dart 1399:13                      _rootRun
  dart:async/zone.dart 1301:19                      _CustomZone.run
  package:test_api/src/backend/invoker.dart 290:38  Invoker.heartbeat.<fn>
  dart:async-patch/timer_patch.dart 18:15           Timer._createTimer.<fn>
  dart:isolate-patch/timer_impl.dart 398:19         _Timer._runTimers
  dart:isolate-patch/timer_impl.dart 4[29](https://github.com/dart-lang/test/actions/runs/10523203385/job/29176912922?pr=2259#step:6:32):5          _Timer._handleMessage
  dart:isolate-patch/isolate_patch.dart 184:12      _RawReceivePort._handleMessage
```

natebosch:
It also looks like some of the tests are timing out with a really poor UX - the test runner isn't noticing it and the tests are spinning. I'm not sure if thats a new issue, or if it's surfacing an existing issue with some types of timeouts

simolus3:
> It also looks like some of the tests are timing out with a really poor UX - the test runner isn't noticing it and the tests are spinning. I'm not sure if thats a new issue, or if it's surfacing an existing issue with some types of timeouts

This is caused by old Node versions not being able to load the `dart2wasm`-generated WebAssembly modules. This causes the node process to exit before connecting to the server opened by the test runner. The runner unfortunately waits for a connection, so it didn't detect this condition.
I've changed the implementation to throw an error in that case, which then considers the test failed. I've added a test exercising this behavior to ensure we're not hitting these timeouts on unsupported Node versions, but it also means that we can't run the integration tests in Node on GitHub actions.

PR #2266, opened 2024-08-12
add health check job

jakemac53:
This should cut down on the number of comments we need to leave regarding changelog updates etc.

### DIFF
--- a/.github/workflows/health.yaml
+++ b/.github/workflows/health.yaml
@@ -1,0 +1,8 @@
+name: Health
+on:
+  pull_request:
+    branches: [ main, master ]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+jobs:
+  health:
+    uses: dart-lang/ecosystem/.github/workflows/health.yaml@main
--- a/.github/workflows/health.yaml
+++ b/.github/workflows/health.yaml
@@ -6,3 +6,5 @@ on:
 jobs:
   health:
     uses: dart-lang/ecosystem/.github/workflows/health.yaml@main
+    permissions:
+      pull-requests: write
--- a/.github/workflows/post_summaries.yaml
+++ b/.github/workflows/post_summaries.yaml
@@ -1,0 +1,17 @@
+name: Comment on the pull request
+
+on:
+  # Trigger this workflow after the Health workflow completes. This workflow will have permissions to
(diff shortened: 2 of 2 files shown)


mosuem:
To work on PRs from forks this also needs another workflow file (`comment_on_pr.yaml` or similar)

```
name: Comment on the pull request

on:
  # Trigger this workflow after the Health workflow completes. This workflow will have permissions to
  # do things like create comments on the PR, even if the original workflow couldn't.
  workflow_run:
    workflows: 
      - Health
      - Publish
    types:
      - completed

jobs:
  upload:
    uses: dart-lang/ecosystem/.github/workflows/post_summaries.yaml@main
    permissions:
      pull-requests: write
```

.github/workflows/health.yaml
```diff
@@ -0,0 +1,8 @@
+name: Health
+on:
+  pull_request:
+    branches: [ main, master ]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+jobs:
+  health:
+    uses: dart-lang/ecosystem/.github/workflows/health.yaml@main
```
mosuem, line 8: This might need

```
    permissions:
      pull-requests: write
```

See https://github.com/dart-lang/i18n/blob/main/.github/workflows/health.yaml

jakemac53:
> To work on PRs from forks this also needs another workflow file (`comment_on_pr.yaml` or similar)

We should add this to the docs as well? I am a bit confused how exactly this solves the issue though

jakemac53:
> We circumvent this by uploading the comment as an artifact, and running a second workflow which downloads the artifact and posts it as a comment

What I am confused about is how that second workflow is able to post the comment? Why is it allowed to when the first one is not?

jakemac53:
Ah, reading the linked doc in the PR you sent to update the usage explains how this works around the issue 👍

natebosch (approved):

natebosch:
@jakemac53 - it looks like the "coverage" check re-runs a bunch of tests that we run in other actions. We might want to skip that one always for the test repo since it's so slow.

Do we have a good way to configure that?

natebosch:
> Do we have a good way to configure that?

Found it. https://github.com/dart-lang/test/pull/2270

PR #2262, opened 2024-08-11
fix: use `toFilePath` in package config Uri

Zekfad:
fixes #2231

Thanks to @d-markey https://github.com/dart-lang/test/issues/2231#issuecomment-2282283629 for figuring it out.

This fix uses the same technique as in precompiled tests:

https://github.com/dart-lang/test/blob/9fbbfdbee18a686e3f84a386a01960ea0543ba01/pkgs/test/test/runner/precompiled_test.dart#L54

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>

PR #2267, opened 2024-08-12
update changelogs per https://github.com/dart-lang/test/pull/2262

jakemac53:


natebosch:
The changes aren't user facing (not even to ourselves as users of the packages). I'd be inclined to skip the changelog check if it had already been enabled on that PR.

I don't know if we've ever decided on concrete guidelines for when to include or omit changelog changes. Maybe we should...

jakemac53:
This is user facing - dart2wasm was broken on windows entirely afaik, and now it will work. Maybe the wording is confusing though?

In other words, what changed isn't that our internal dart2wasm tests now pass, its that dart2wasm tests in general can now work on windows.

natebosch (approved):

natebosch:
I misread the linked PR.

PR #2268, opened 2024-08-14
require approval from core-package-admins for anything under /pkgs

jakemac53:
Trying out the CODEOWNERS workflows, we can remove this if it is annoying.

Any dart-lang admin can edit that group so we can always get something landed if needed that way, but it will require some explicit steps from an admin.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+/pkgs/ @dart-lang/core-package-admins


natebosch:
The github diff view is showing an "Unknown Owner" warning that the group needs to be public. I'm not sure if that will prevent it from working.

natebosch:
Maybe @kevmoo has a suggestion.

For context, we'd like to limit reviewers for PRs in the test repo since the tight coupling (both within the repo, and with flutter and google3) can make it tricker to land things.

devoncarew:
I changed the team to public; I think that'll address the error.

natebosch (approved):

PR #2273, opened 2024-08-29
[pkg:test] Use Set instead of List where appropriate

kevmoo:
Use more specific String type where possible

### DIFF
--- a/pkgs/test_api/lib/src/backend/compiler.dart
+++ b/pkgs/test_api/lib/src/backend/compiler.dart
@@ -41,8 +41,8 @@ enum Compiler {
   /// [Compiler].
   ///
   /// Note that custom [Compiler] implementations are not supported.
-  factory Compiler.deserialize(Object serialized) => builtIn
-      .firstWhere((compiler) => compiler.identifier == serialized as String);
+  factory Compiler.deserialize(String serialized) =>
+      builtIn.firstWhere((compiler) => compiler.identifier == serialized);
 
   /// Converts [this] into a JSON-safe object that can be converted back to a
   /// [Compiler] using [Compiler.deserialize].
--- a/pkgs/test_api/lib/src/backend/runtime.dart
+++ b/pkgs/test_api/lib/src/backend/runtime.dart
@@ -11,38 +11,38 @@ final class Runtime {
 
   /// The command-line Dart VM.
   static const Runtime vm = Runtime('VM', 'vm', Compiler.kernel,
-      [Compiler.kernel, Compiler.source, Compiler.exe],
+      {Compiler.kernel, Compiler.source, Compiler.exe},
       isDartVM: true);
 
   /// Google Chrome.
   static const Runtime chrome = Runtime('Chrome', 'chrome', Compiler.dart2js,
-      [Compiler.dart2js, Compiler.dart2wasm],
+      {Compiler.dart2js, Compiler.dart2wasm},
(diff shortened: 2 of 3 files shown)


kevmoo:
No user-visible changes

pkgs/test_api/lib/src/backend/compiler.dart
```diff
@@ -41,8 +41,8 @@ enum Compiler {
   /// [Compiler].
   ///
   /// Note that custom [Compiler] implementations are not supported.
-  factory Compiler.deserialize(Object serialized) => builtIn
-      .firstWhere((compiler) => compiler.identifier == serialized as String);
+  factory Compiler.deserialize(String serialized) =>
```
jakemac53, line 44: This is technically breaking, this is a public API (part of `backend.dart`).

pkgs/test_api/lib/src/backend/runtime.dart
```diff
@@ -89,7 +89,7 @@ final class Runtime {
   final Compiler defaultCompiler;
 
   /// All the supported compilers for this runtime.
-  final List<Compiler> supportedCompilers;
+  final Set<Compiler> supportedCompilers;
```
jakemac53, line 92: This is also breaking (this class is also exported by backend.dart)

pkgs/test_api/lib/src/backend/compiler.dart
kevmoo: But there is NO world where this would run without failing, right?

pkgs/test_api/lib/src/backend/runtime.dart
kevmoo: Ugh. k.

pkgs/test_api/lib/src/backend/compiler.dart
jakemac53: I am not totally sure... but probably this change is OK in practice?

kevmoo: since we're casting `as String`?

pkgs/test_api/lib/src/backend/runtime.dart
jakemac53: This one would break some things

pkgs/test_api/lib/src/backend/compiler.dart
natebosch: I think this is OK. This constructor is not used outside of the package and I don't think it's worth a major version bump since we only support our known uses.

pkgs/test_api/lib/src/backend/runtime.dart
natebosch: I don't think this will break anything. Only `Iterable` interface methods are used on this field, and only from `test_core`

jakemac53: It is breaking because this field is used in initializing formal parameters, and the constructor is used internally at least and possibly also in flutter but not sure on that.

jakemac53: We could possibly change the constructor to take an iterable and call `toSet` but that is less than ideal probably

kevmoo:
We've already spent more time on this then we should have. Unless we're in agreement this is fine, I'll just close out without landing.

PR #2195, opened 2024-03-02
Fix path to launch firefox on macOS

koji-1009:
Fix firefox startup path, as `dart test -p firefox` can no longer be executed on macOS.

close #2194 

I checked the workflow and both macos and firefox are removed from the execution environment. In relation to PR, do I need to add macos and firefox execution patterns to the workflow?

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>

Issue #2194, opened 2024-03-02
Firefox fails to start when called via `/Applications/Firefox.app/Contents/MacOS/firefox-bin` on macOS


koji-1009:
When I try to run the test code for the browser API, I cannot run the test using firefox.
Note that if `-p chrome` is specified, it works fine.

```console
$ dart test -p firefox
00:03 +0: loading test/crypto_subtle_test.dart
Compiled 11,501,289 input bytes (6,524,318 characters source) to 1,333,060 characters JavaScript in 3.08 seconds using 323.453 MB of memory

00:03 +0 -1: loading test/crypto_subtle_test.dart [E]
  Failed to load "test/crypto_subtle_test.dart":
  Failed to run Firefox: No such file or directory
    Command: /Applications/Firefox.app/Contents/MacOS/firefox-bin --profile /private/var/folders/hr/y37hfwp553b_1bc41k7s8nqm0000gn/T/dart_test_JsusyQ http://localhost:57615/rxxgE01iInxRSACwP2MzZZtXA6TVPGX2/packages/test/src/runner/browser/static/index.html?managerUrl=ws%3A%2F%2Flocalhost%3A57615%2FrxxgE01iInxRSACwP2MzZZtXA6TVPGX2%2F0&debug=false --no-remote.
  dart:io-patch/process_patch.dart 402:33                      _ProcessImpl._start
  dart:io-patch/process_patch.dart 38:20                       Process.start
  package:test/src/runner/browser/firefox.dart 44:33           Firefox._startBrowser
  package:test/src/runner/browser/firefox.dart 35:13           new Firefox.<fn>
  package:test/src/runner/browser/browser.dart 61:39           new Browser.<fn>
  dart:async/zone.dart 1399:13                                 _rootRun
  dart:async/zone.dart 1301:19                                 _CustomZone.run
  dart:async/zone.dart 1804:10                                 _runZoned
  dart:async/zone.dart 1792:12                                 runZonedGuarded
  package:test/src/runner/browser/browser.dart 60:5            new Browser
  package:test/src/runner/browser/firefox.dart 34:9            new Firefox
  package:test/src/runner/browser/browser_manager.dart 161:28  BrowserManager._newBrowser
  package:test/src/runner/browser/browser_manager.dart 117:19  BrowserManager._start
  package:test/src/runner/browser/browser_manager.dart 107:7   BrowserManager.start
  package:test/src/runner/browser/platform.dart 438:33         BrowserPlatform._browserManagerFor
[...]

PR #2274, opened 2024-08-29
[pkg:test] Fix binary location of Firefox on Mac

kevmoo:


kevmoo:
@jakemac53 – thoughts on this?

jakemac53:
> @jakemac53 – thoughts on this?

I have no context so I have no thoughts :). Did it move? Does this location work on older versions? Maybe we need a fallback to the old location etc...

kevmoo:
it DID move @jakemac53

FF tests were failing locally until I made this fix.
I guess I can open an issue that we actually test FF + Mac in CI, right?

jakemac53:
> I guess I can open an issue that we actually test FF + Mac in CI, right?

I don't know if anybody has the bandwidth to actually look into this, there might be an open issue already.

jakemac53:
See also https://github.com/dart-lang/test/issues/2194 and https://github.com/dart-lang/test/pull/2195 ... the tldr; here is that it seems like adding support for multiple paths was never agreed upon and no matter what we do it doesn't work for some people.

But, maybe this change has existed for long enough that it is time to make the switch, and highlight the workaround for anybody on older versions (using a custom path for Firefox via dart_test.yaml).

cc @natebosch thoughts?

natebosch:
https://github.com/dart-lang/test/pull/2276 gives a try at making this non-breaking and keeping the scope of the change as small as possible.

I haven't tested it yet, but I can if we think we want to land it.
I'm 50/50 on whether I want to try to land something minimal like that, or to just make the switch assuming more users will benefit than suffer, and document the workaround.

Do we have any details on when it moved, or how many users are likely to have either path?

kevmoo:
Happy to run with #2276 if that's good.

Want me to close this?

PR #2276, opened 2024-09-03
Add support for multiple full paths on macos

natebosch:
We cannot look up multiple basename commands in the system path and the
current `macOsExecutable` configuration may have existing uses in
`dart_test.yaml` files so it isn't safe to require full paths. Add a
separate `macOsAbsolutePaths` configuration to enable internal
definitions that check multiple full paths and execute the first one
that exists.

Add the basename `firefox` as a fallback. Adding that command to the
path is one workaround for users with firefox installed in an unexpected
location. This is also the only approach that is compatible with mac on
GitHub actions using the `setup-firefox` action.

There is no support here for `dart_test.yaml`. Users that are
configuring an executable for macOS will continue to have support for
only a single value, even if they are specifying an absolute path.

Tests remain skipped because our mono repo setup does not have a way to
include the `setup-firefox` action.

### DIFF
--- a/pkgs/test/lib/src/runner/executable_settings.dart
+++ b/pkgs/test/lib/src/runner/executable_settings.dart
@@ -20,7 +20,7 @@ class ExecutableSettings {
   /// looked up on the system path. It may not be relative.
   final String? _linuxExecutable;
 
-  /// Potential commands to execute on Mac OS.
+  /// Potential commands to execute on Mac OS to launch this executable.
   ///
   /// The values may be an absolute path, or a basename.
   /// The chosen command will be the first value from this list which either is
--- a/pkgs/test/lib/src/runner/executable_settings.dart
+++ b/pkgs/test/lib/src/runner/executable_settings.dart
@@ -20,18 +20,14 @@ class ExecutableSettings {
   /// looked up on the system path. It may not be relative.
   final String? _linuxExecutable;
 
-  /// The command to execute on Mac OS.
+  /// Potential commands to execute on Mac OS.
   ///
-  /// This may be a basename or an absolute path.
-  /// If this is a basename it is resolved by the OS on the system path.
-  /// Otherwise, the path must be absolute, not relative.
-  final String? _macOSExecutable;
-
-  /// Potential absolute paths to the executable on Mac OS.
-  ///
-  /// When multiple paths are provided the first to exist on the filesystem is
-  /// chosen. Paths must be absolute.
-  final Set<String>? _macOSAbsolutePaths;
+  /// The values may be an absolute path, or a basename.
+  /// The chosen command will be the first value from this list which either is
+  /// a full path that exists, or is a basename. When a basename is included it
+  /// should always be at the end of the list because it will always terminate
+  /// the search through the list. Relative paths are not allowed.
+  final List<String>? _macOSExectuables;
 
   /// The path to the executable on Windows.
   ///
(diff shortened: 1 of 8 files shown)


natebosch:
This is the least invasive way I could see to make this change. I'm not sure if it's worth landing, but opening the PR for discussion.

natebosch:
I think if we wanted we could support this in `dart_test.yaml`. We could accept either a single string, or a list of strings, and we could validate when parsing that only full paths are used if it is a list. We already have _some_ inconsistency between the meaning of the arguments across OSes, so I don't think it's a particular problem to add this option/caveat exclusive to macOS.

pkgs/test/lib/src/runner/executable_settings.dart
```diff
@@ -193,5 +209,6 @@ class ExecutableSettings {
       headless: other._headless ?? _headless,
       linuxExecutable: other._linuxExecutable ?? _linuxExecutable,
       macOSExecutable: other._macOSExecutable ?? _macOSExecutable,
+      macOsAbsolutePaths: other._macOSAbsolutePaths ?? _macOSAbsolutePaths,
```
natebosch, line 212: Hmm, this will cause a problem with overriding the path...

pkgs/test/lib/src/runner/executable_settings.dart
```diff
@@ -20,11 +20,14 @@ class ExecutableSettings {
   /// looked up on the system path. It may not be relative.
   final String? _linuxExecutable;
 
-  /// The path to the executable on Mac OS.
+  /// Potential commands to execute on Mac OS.
```
jakemac53, line 23:
```suggestion
  /// Potential commands to execute on Mac OS to launch this executable.
```

pkgs/test/lib/src/runner/executable_settings.dart
```diff
@@ -45,7 +48,15 @@ class ExecutableSettings {
...
 
-    if (Platform.isMacOS) return _macOSExecutable!;
+    if (Platform.isMacOS) {
+      if (_macOSExectuables case final paths?) {
+        for (final path in paths) {
+          if (p.basename(path) == path) return path;
+          if (p.isAbsolute(path) && File(path).existsSync()) return path;
+        }
+      }
+      throw ArgumentError('Missing macOsExecutable');
```
jakemac53, line 58: Can we just rely on private field promotion? It is much easier to read imo if so:

```suggestion
      if (_macOSExectuables != null) {
        for (final path in _macOSExectuables) {
          if (p.basename(path) == path) return path;
          if (p.isAbsolute(path) && File(path).existsSync()) return path;
        }
      }
      throw ArgumentError('Missing macOsExecutable');
```

jakemac53 (approved):

pkgs/test/lib/src/runner/executable_settings.dart
```diff
@@ -192,6 +205,13 @@ class ExecutableSettings {
       arguments: arguments.toList()..addAll(other.arguments),
       headless: other._headless ?? _headless,
       linuxExecutable: other._linuxExecutable ?? _linuxExecutable,
-      macOSExecutable: other._macOSExecutable ?? _macOSExecutable,
+      macOSExecutables: other._macOSExectuables ?? _macOSExectuables,
       windowsExecutable: other._windowsExecutable ?? _windowsExecutable);
 }
+
+List<String>? _normalizeMacExecutable(
```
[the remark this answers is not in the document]
natebosch: We don't currently have a guard, we use `!`. I don't think the error is any more confusing without a specific check here.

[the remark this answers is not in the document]
natebosch: Oh I misread that. I think it's ok to let the list take precedence in that case. I'd rather not add handling for situations that aren't likely to come up.

natebosch:
This fixes the problem for me on a local mac. It looks like maybe firefox doesn't come by default on github actions macos runs? I don't think mono package supports adding a firefox setup step. Any suggestions on a workaround @kevmoo ?

natebosch:
> It looks like maybe firefox doesn't come by default on github actions macos runs?

No, it still fails even if I manually add the setup-firefox steps.

natebosch:
I cannot figure out the mac problem. Landing this is still worth it because it does fix things for me locally. @jakemac53 any concerns about landing this without unskipping the tests on CI, or with using the basename as a fallback?

jakemac53:
> I cannot figure out the mac problem. Landing this is still worth it because it does fix things for me locally. @jakemac53 any concerns about landing this without unskipping the tests on CI, or with using the basename as a fallback?

I am fine with it, we don't have a lot of people asking about this, so it isn't worth dumping a ton of time into. If people do get broken again they will file issues (plus, there is a workaround to set your own path).

jakemac53 (approved):

Issue #1253, opened 2020-06-06
Testing on Safari is not working in Catalina


lucacasonato:
In Catalina opening a file in Safari requires the user to confirm that they want to open the file in Safari. Because of this testing on safari does not work anymore without manual user interaction. According to my research this has been broken since at least Mojave.

grouma:
You are correct as I verified this is an issue on Mojave. A quick Google search shows other testing frameworks also have this issue:
https://github.com/testem/testem/issues/1387

There is a possible workaround:

> I have ran into this as well with Github Actions VMs and locally as well. There's something for Karma via AppleScript to workaround this. https://github.com/brandonocasey/karma-safari-applescript-launcher.

That being said I don't think this is something we will make progress on in the near term.

PR #2280, opened 2024-09-10
Run package:test tests on mac

natebosch:
This will exercise things like launching browsers (where they are
currently working on CI).

Add `osx` to the sharded `package:test` tests alongside `windows`. These
exercise the majority of OS specific behavior.

Skip existing safari tests due to a bug running tests in non-interactive
environments. (#1253)

Skip existing firefox tests due to a changed executable path. (Planned
fix in #2276)

Refactor a ternary chain to an if/else chain to add the extra
conditions for firefox and safari.

### DIFF
--- a/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
+++ b/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
@@ -34,6 +34,8 @@ void main() {
       } else if ([Runtime.firefox, Runtime.nodeJS].contains(runtime) &&
           Platform.isWindows) {
         skipReason = 'https://github.com/dart-lang/test/issues/1942';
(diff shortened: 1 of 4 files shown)


pkgs/test/test/runner/compiler_runtime_matrix_test.dart
```diff
@@ -26,8 +26,19 @@ void main() {
           (runtime == Runtime.safari && !Platform.isMacOS)) {
         continue;
       }
+      String? skipReason;
+      if (runtime == Runtime.safari) {
```
kevmoo, line 30: I really want to nerd snipe you to make this a switch, but don't fall for it!

jakemac53 (approved):

Issue #110, opened 2015-05-13
Add the ability to get the current script or project path when using the new test runner


alextekartik:
http://stackoverflow.com/questions/30214563/get-current-script-path-or-current-project-path-using-new-test-runner
I am porting old vm unittest files using the new test package. Some relies on input files in sub directories of my test folder. Before I was using Platform.script to find the location of such files. This works fine when using

    $ dart test/my_test.dart

However using

    $ pub run test

this is now pointing to a temp folder (tmp/dart_test_xxxx/runInIsolate.dart). I am unable to locate my test input files anymore. I cannot rely on the current path as I might run the test from a different working directory.

Is there a way to find the location of my_test.dart (or event the project root path), from which I could derive the locations of my files?

nex3:
This is the sort of thing that should be exposed through the API (#48).

jmesserly:
/sub ... I'm having the same problem (also covered in #123). Our existing tests use Platform.script to load data, making hard to migrate to test runner.

Anecdotally: most packages I've worked on with VM tests end up using data files.

Anyway, I'm guessing I can work around this with dart:mirrors ... instead of Platform.script, find the script's library URI via mirrors.

enyo:
Is there a simple hack/workaround for this, that we can use until it is implemented? (Apart from setting an environment variable)

alextekartik:
I use the following ugly workaround that get me the directory name of the current test script if I'm running it directly or with pub run test. It will definitely break if anything in the implementation changes but I needed this desperately...

    library test_utils.test_script_dir;
    
    import 'dart:io';
    import 'package:path/path.dart';
    
    // temp workaround using test package
    String get testScriptDir {
      String scriptFilePath = Platform.script.toFilePath();
      print(scriptFilePath);
      if (scriptFilePath.endsWith("runInIsolate.dart")) {
    
        // Let's look for this line:
        // import "file:///path_to_my_test/test_test.dart" as test;
    
        String importLineBegin = 'import "file://';
        String importLineEnd = '" as test;';
        int importLineBeginLength = importLineBegin.length;
    
        String scriptContent = new File.fromUri(Platform.script).readAsStringSync();
    
        int beginIndex = scriptContent.indexOf(importLineBegin);
        if (beginIndex > -1) {
          int endIndex = scriptContent.indexOf(importLineEnd, beginIndex + importLineBeginLength);
          if (endIndex > -1) {
[...]

PR #2281, opened 2024-09-11
[WIP] Add suitePath getter

jakemac53:
Possible resolution to https://github.com/dart-lang/test/issues/110.

Adds a `suitePath` getter which gives the package relative path to the current suite if it is available.

- Only works in the test runner itself (we could maybe support some other scenarios if we wanted, using Platform.script).
- On the web, `Uri.base` already actually contains the full path including the suite, so you don't want to use this there which is a bit weird. On the VM you want to do something like `Uri.base.resolve(suitePath).resolve(<some-suite-relative-path>)`.
  - We could consider instead providing a function for resolving suite relative paths, or make this API give you the absolute path to the suite (which would be an http URI on the web).
  - We could also add a <base> tag to the html page, setting it to be the package root.


natebosch:
@matanlurey would you be able to put together a prototype of your intended usage so we can confirm whether this solves the problem?

jakemac53:
This is only meant to solve the original problem stated in https://github.com/dart-lang/test/issues/110, which is getting the path to the test _script_. It does not help with getting access to the package root, but `Uri.base` does already give you that (whenever using the test runner).

natebosch:
> but `Uri.base` does already give you that

Does this get impacted by changes to the working directory?

jakemac53:
> Does this get impacted by changes to the working directory?

Probably, although I don't know for sure. But, that is partially why I like just exposing the relative suite path. We can be out of the business of the rest of it, which isn't really in our control. We can provide suggestions with caveats (like using Uri.base), which will work for most, but we don't guarantee them or support them directly.

Issue #2185, opened 2024-02-13
Replace broken Observatory URL with DevTools URL


kenzieschmoll:
When running a test with `dart test foo_test.dart --pause-after-load`, the following is printed to the CLI:
```shell
 dart test test/dart_test/b_example_test.dart --pause-after-load
...
00:01 +0: loading test/dart_test/b_example_test.dart                                                  
Observatory URL: http://127.0.0.1:51715/1m3AdwIltJ4=/#/inspect?isolateId=isolates%2F3973115144356875&objectId=libraries%2F%4019317648
The test runner is paused. Open the Observatory and set breakpoints. Once you're finished, return to
this terminal and press Enter.
```

This URL is broken. DevTools is also unavailable from this state because there is not a DDS instance.

We should fix this so that when running a test with `dart test` for debugging, DevTools is served properly and the user can access if from a URL printed to CLI (Flutter tools does this properly for `flutter test`).

CC @jakemac53 @bkonyi

kenzieschmoll:
This is important for the work to support DevTools extensions for dart tests: https://github.com/flutter/devtools/issues/7183.

bkonyi:
This will require two changes:

1) We need to be able to serve DDS and DevTools from `dart test` (this currently isn't configured)
2) We need to update something in `package:test*` to print the right message.

For 1), @a-siva is currently working on a CL to launch DDS (and also serve DevTools) from the VM service isolate instead of from DartDev, so we'll get this for free once that work is done.

For 2), changing the messaging should be easy enough, but I'm not sure we'll be able to mimic the prior behavior since the DevTools debugger doesn't support navigating to a library based on a query parameter like Observatory does.

kenzieschmoll:
> For 2), changing the messaging should be easy enough, but I'm not sure we'll be able to mimic the prior behavior since the DevTools debugger doesn't support navigating to a library based on a query parameter like Observatory does.

I think this is okay. The user can decide what type of debugging they want to do once they get into DevTools.

bkonyi:
> > For 2), changing the messaging should be easy enough, but I'm not sure we'll be able to mimic the prior behavior since the DevTools debugger doesn't support navigating to a library based on a query parameter like Observatory does.
> 
> I think this is okay. The user can decide what type of debugging they want to do once they get into DevTools.

[...]

PR #2283, opened 2024-09-24
Use dev tools URL when debugging

natebosch:
Towards #2185

The Observatory UI is no longer served and the current URL does not
work. Dev Tools does not support deep linking to a particular isolate or
library, so link to the overall devtools app for now.


jakemac53 (approved):

natebosch:
There are still open questions in #2185 but this is worth landing in the meantime because the currently printed URL is not useful.
Given that we haven't had complaints this is probably not a highly used feature...

kevmoo (approved):

PR #2257, opened 2024-07-22
Special case maps with only value keys

natebosch:
Closes #2256

When checking deep equality on Map instances (including maps nested in
other collection types) we allow keys which are instances of other
collection (and don't have a useful `operator ==`) and keys which are
`Condition` callbacks that could potentially match multiple keys of the
actual value. This is maximally flexible, but it loses the nested path
information we are otherwise able to keep for iterables compared by
index.

Check whether expectation maps have any keys which need this special
treatment, and if not, special case to an algorithm close to the
`Iterable` algorithm which can enqueue nested checks and maintain the
direct known path into the collection.

Update the doc comment to describe the new behavior. This does
potentially introduce confusion since a change which adds a `Condition`
key has to potential to make _other_ keys in the same expectation
behaved differently. The benefit is a more actionable failure output for
the much more common case of checking expectations with hardcoded keys.

PR #2284, opened 2024-09-27
Add containsMatchingInOrder  containsEqualInOrder

natebosch:
The joined behavior in `containsInOrder` has some usability issues:
- It mimics the arguments for `deepEquals`, but it doesn't have the same
  behavior for collection typed elements. Checking that a nested
  collection is contained in order requires a `Condition` callback that
  uses `.deepEquals` explicitly.
- The `Object?` signature throws away inference on the `Condition`
  callback arguments. With a method that supports only conditions the
  argument type can be tightened and allow inference.

Deprecate the old `containsInOrder` and plan to remove it before stable.
This is a bit more restrictive, but it's not too noisy to fit a few
`(it) => it.equals(foo)` in a collection that needs mixed behavior and
the collection of two methods is less confusing to document than the
joined behavior.

Lean on the "Matches" verb for cases that check a `Condition` callback
and rename `pairwiseComparesTo` as `pairwiseMatches`.

Fix a type check when pretty printing `Condition` callbacks. Match
more than `Condition<dynamic>` by checking `Condition<Never>`.

### DIFF
--- a/pkgs/checks/CHANGELOG.md
+++ b/pkgs/checks/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 -   Update min SDK constraint to 3.4.0.
 -   Always wrap Condition descriptions in angle brackets.
+-   Add `containsMatchingInOrder` and `containsEqualInOrder` to replace the
+    combined functionality in `containsInOrder`.
 
 ## 0.3.0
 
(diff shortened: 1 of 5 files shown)


natebosch:
#2257 highlights another place where mixing the matches/equals behavior has a negative impact on the UX. There we are mitigating it by special casing arguments that don't use `Condition` in some places. I don't think that it makes sense to try to split between `deepEquals` and `deepMatches` in that case. The distinction comes down to the deep vs non-deep behavior. With the shallow "contains" check we get a bigger benefit from type inference on the condition callbacks (can allow typed `Subject` arguments, instead of `Subject<dynamic>`).

PR #2285, opened 2024-09-27
Skip some wasm tests

natebosch:
Skip some wasm integration tests in chrome and on windows.
Add a `dart2js` test to run in firefox on linux which can start, but
currently fails because it runs no test cases.


integration_tests/wasm/dart_test.yaml
```diff
@@ -1,5 +1,9 @@
-platforms: [chrome, firefox]
-# Node doesn't work because the version available in the current Ubuntu GitHub runners is too
-# old to support WASM+GC, which would be required to run Dart tests.
-#platforms: [chrome, firefox, node]
-compilers: [dart2wasm]
+platforms:
+# TODO(https://github.com/dart-lang/test/issues/1790) - unskip
+# - chrome
```
osa1, line 3: The linked issue is two years old and the link to the CI run in the issue is dead. Do we know if it's still relevant?

There shouldn't be any failures on Chrome. There only a few bugs in `async*` but other than that it work, and if it doesn't we should fix it instead.

osa1:
What is this fixing exactly? The referenced issue in the comments is 2 years old and referring to a time when dart2wasm could run hello world plus a few other programs. It's feature complete now with only a few known bugs. The CI runs in the main branch are also all green.

natebosch:
This was just trying to unblock CI for the failures with the old SDK. I reused the stale issue instead of filing a new one just so we had something to track unskipping them.

The failures are resolved by #2286 so I'll close this one.

We might still want to look in to the weird behavior on firefox.

PR #2286, opened 2024-09-30
Update dart2wasm module runner

osa1:
Update dart2wasm module instantiation, compilation, and invoking `main` to support all the ways dart2wasm implemented.

### DIFF
--- a/pkgs/test/lib/src/runner/browser/static/run_wasm_chrome.js
+++ b/pkgs/test/lib/src/runner/browser/static/run_wasm_chrome.js
@@ -7,23 +7,39 @@
 // affect chrome.
 (async () => {
   // Fetch and compile Wasm binary.
-  let data = document.getElementById('WasmBootstrapInfo').dataset;
+  let data = document.getElementById("WasmBootstrapInfo").dataset;
 
   // Instantiate the Dart module, importing from the global scope.
-  let dart2wasmJsRuntime = await import('./' + data.jsruntimeurl);
+  let dart2wasmJsRuntime = await import("./" + data.jsruntimeurl);
 
-  // Check whether the JS shim has the new instantiation API.
-  if (dart2wasmJsRuntime.CompiledApp !== undefined) {
-    // New API.
-    let compiledModule = await dart2wasmJsRuntime.compileStreaming(fetch(data.wasmurl));
-    let instantiatedModule = await compiledModule.instantiate()
+  // Support three versions of dart2wasm:
+  //
+  // (1) Versions before 3.6.0-167.0.dev require the user to compile using the
+  // browser's `WebAssembly` API, the compiled module needs to be instantiated
+  // using the JS runtime.
+  //
+  // (2) Versions starting with 3.6.0-167.0.dev added helpers for compiling and
+  // instantiating.
+  //
+  // (3) Versions starting with 3.6.0-212.0.dev made compilation functions
+  // return a new type that comes with instantiation and invoke methods.
 
-    // Call `main`. If tasks are placed into the event loop (by scheduling tasks
-    // explicitly or awaiting Futures), these will automatically keep the script
-    // alive even after `main` returns.
-    await instantiatedModule.invokeMain();
+  if (dart2wasmJsRuntime.compileStreaming !== undefined) {
+    // Version (2) or (3).
+    let compiledModule = await dart2wasmJsRuntime.compileStreaming(
+      fetch(data.wasmurl),
+    );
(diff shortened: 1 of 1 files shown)


osa1:
Do we need to update some min. SDK version in this repo to the version with the new API?

I think the test runner is shipped with the SDK (instead of fetched from pub.dev), so maybe we just need to roll this change into the SDK?

osa1:
Note: we are bringing back the deprecated API in https://dart-review.googlesource.com/c/sdk/+/387162.

We should still merge this change in prep. for the removal of the deprecated API.

jakemac53:
It looks like we need to update the min SDK constraints to `3.6.0-271.0.dev`. @osa1 did you want me to push a change to this PR to do that across the repo?

jakemac53:
Given that the new API isn't available in a stable SDK yet we should also discuss whether we should wait to land and/or publish this until 3.6 is out.

When do we plan to remove the deprecated API again?

natebosch:
> I think the test runner is shipped with the SDK (instead of fetched from pub.dev), so maybe we just need to roll this change into the SDK?

The test runner is fetched from pub to allow changes to the user code without breaking SDK changes.

pkgs/test/lib/src/runner/browser/static/run_wasm_chrome.js
```diff
@@ -8,14 +8,14 @@
 (async () => {
   // Fetch and compile Wasm binary.
   let data = document.getElementById('WasmBootstrapInfo').dataset;
-  let modulePromise = WebAssembly.compileStreaming(fetch(data.wasmurl));
 
   // Instantiate the Dart module, importing from the global scope.
-  let dart2wasm = await import('./' + data.jsruntimeurl);
-  let dartInstance = await dart2wasm.instantiate(modulePromise, {});
+  let dart2wasmJsRuntime = await import('./' + data.jsruntimeurl);
+  let compiledModule = await dart2wasmJsRuntime.compileStreaming(fetch(data.wasmurl));
```
natebosch, line 14: Can we check whether an `instantiate` method exists and choose at runtime which approach to use?

jakemac53: That would definitely be ideal, so we can still support 3.5 SDKs.

jakemac53:
Note https://github.com/dart-lang/test/pull/2285/files is skipping some of these for now, we should make sure to unskip them when landing this.

osa1:
> It looks like we need to update the min SDK constraints to 3.6.0-271.0.dev. @osa1 did you want me to push a change to this PR to do that across the repo?

Please feel free to always make any changes you think are necessary to my branches, this one and in the future as well.

> When do we plan to remove the deprecated API again?

Whenever the users are ready. The new API allows deferred loading, but it cannot be used in `package:test` tests anyway as you have to call JS `instantiate` method yourself to provide a callback to fetch the deferred modules.

> Note https://github.com/dart-lang/test/pull/2285/files is skipping some of these for now, we should make sure to unskip them when landing this.

I'm not sure what that PR is fixing exactly. It refers to a 2 year old issue that should be invalid now as dart2wasm is quite stable today compared to two years ago.

osa1:
I added the skip-changelog-check label, but the CI check is still not happy that I don't have a changelog entry..

jakemac53:
> I'm not sure what that PR is fixing exactly. It refers to a 2 year old issue that should be invalid now as dart2wasm is quite stable today compared to two years ago.

It was just adding additional skips to get the bots green and linking re-using an old issue for tracking that.

pkgs/test/lib/src/runner/browser/static/run_wasm_chrome.js
```diff
@@ -7,15 +7,41 @@
...
-  // explicitly or awaiting Futures), these will automatically keep the script
-  // alive even after `main` returns.
-  await dart2wasm.invoke(dartInstance);
+  // Support three versions of dart2wasm:
+  //
+  // (1) Versions before 3.6.0-167.0.dev require the user to compile using the
+  // browser's `WebAssembly` API, the compiled module needs to be instantiated
+  // using the JS runtime.
+  //
+  // (2) Versions starting with 3.6.0-167.0.dev added helpers for compiling and
```
natebosch, line 21: It would be OK to drop support for a strategy that was only in a series of dev releases, but supporting it sounds good too.

After we have any other reason to bump the min SDK we will get this cleaned up to only the version 3 approach.

natebosch (approved):

PR #2296, opened 2024-10-17
update the fake_async version to -wip following this repos standar processes

jakemac53:
As we merge these repos in, we should update the versions as well to follow our standard procedure (otherwise, firehose is going to keep saying we have ready to publish versions).

It is probably best done in a follow-up PR like this though, or in the source repo prior to merging.


natebosch (approved):

mosuem:
I upgrade the version as part of the moving process, and release it after the merge PR as landed to update the repository location in pub.dev. After that, the `-wip` should be restored.

PR #2410, opened 2024-10-21
Enable avoid_dynamic_calls lint

natebosch:
Add argument types to `typedMatches` arguments which were unnecessarily
widening it back to `dynamic`. 

Add some missing `as dynamic` and `as Function` which is the pattern the
lint uses to allow intentional dynamic calls.

Ignore some dynamic calls in a test for a legacy API.

Ignore a couple false positives for calling methods (as opposed to
getters) on a cast expression. The false positive was fixed in
https://dart-review.googlesource.com/c/sdk/+/390640
but not yet published.

Remove the deprecated lint `package_api_docs`.

Replaces https://github.com/dart-lang/matcher/pull/205

### DIFF
--- a/pkgs/matcher/lib/src/core_matchers.dart
+++ b/pkgs/matcher/lib/src/core_matchers.dart
@@ -148,7 +148,7 @@ class _ReturnsNormally extends FeatureMatcher<Function> {
   const _ReturnsNormally();
 
   @override
-  bool typedMatches(Function f, Map matchState) {
+  bool typedMatches(void Function() f, Map matchState) {
     try {
       f();
       return true;
--- a/pkgs/matcher/lib/src/core_matchers.dart
+++ b/pkgs/matcher/lib/src/core_matchers.dart
@@ -148,8 +148,9 @@ class _ReturnsNormally extends FeatureMatcher<Function> {
   const _ReturnsNormally();
 
   @override
-  bool typedMatches(Object f, Map matchState) {
+  bool typedMatches(Function f, Map matchState) {
     try {
+      // ignore: unnecessary_cast
       (f as Function)();
       return true;
     } catch (e, s) {
--- a/pkgs/matcher/lib/src/core_matchers.dart
+++ b/pkgs/matcher/lib/src/core_matchers.dart
@@ -144,7 +144,7 @@ class isInstanceOf<T> extends TypeMatcher<T> {
 /// a wrapper will have to be created.
 const Matcher returnsNormally = _ReturnsNormally();
 
-class _ReturnsNormally extends FeatureMatcher<Function> {
+class _ReturnsNormally extends FeatureMatcher<void Function()> {
   const _ReturnsNormally();
 
   @override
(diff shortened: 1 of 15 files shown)


pkgs/matcher/lib/src/core_matchers.dart
```diff
@@ -148,9 +148,9 @@ class _ReturnsNormally extends FeatureMatcher<Function> {
   const _ReturnsNormally();
 
   @override
-  bool typedMatches(Function f, Map matchState) {
+  bool typedMatches(Object f, Map matchState) {
```
kevmoo, line 151: Is this better? Better to ignore?

natebosch: Either pattern is ugly.

We'd either need to ignore an `unnecessary_cast` or the `avoid_dynamic_calls` if we use an ignore.

Do we think it would be better behavior in the analyzer to suppress `unnecessary_cast` in these cases?
If we think the analyzer would make that change adding an `// ignore: unnecessary_cast` would be a good choice for now.

natebosch: https://github.com/dart-lang/sdk/issues/56939

I'll ignore the `unnecessary_cast` for now and if the discussion on that issue changes we can rethink which pattern to use here.

pkgs/matcher/lib/src/core_matchers.dart
```diff
@@ -150,7 +150,8 @@ class _ReturnsNormally extends FeatureMatcher<Function> {
   @override
   bool typedMatches(Function f, Map matchState) {
     try {
-      f();
+      // ignore: unnecessary_cast
+      (f as Function)();
```
jakemac53, line 154: can we do this?

```suggestion
      (f as void Function())();
```

pkgs/matcher/lib/src/map_matchers.dart
```diff
@@ -15,6 +15,7 @@ class _ContainsValue extends Matcher {
 
   @override
   bool matches(Object? item, Map matchState) =>
+      // ignore: avoid_dynamic_calls
       (item as dynamic).containsValue(_value);
```
jakemac53, line 19: I am curious why we can't cast this to a Map?

jakemac53 (approved):

pkgs/matcher/lib/src/map_matchers.dart
natebosch: I may have been worried about non-map types with `operator []` and `containsKey` or `containsValue`, but thinking about it now that seems like an unlikely use case. I'll change this and `containsValue` to be `FeatureMatcher<Map>`

pkgs/matcher/lib/src/core_matchers.dart
natebosch: Unfortunately not, because it works today to do `expect(<T>(){}, returnsNormally);`, but a `<T>() -> void` is not a subtype of `() -> void`

PR #2415, opened 2024-11-01
Use dynamic methods for map matchers

natebosch:
Allow using `containsKey` and `containsValue` on non-map types like
built collection maps which do not implement the core `Map` but has the
required methods and operators.


natebosch:
I thought I had a passing global test run that included these changes, but it must have been a different snapshot because this change does cause failures.

jakemac53 (approved):

PR #2421, opened 2024-12-03
widen the dep in package:shelf_web_socket

devoncarew:
- add a 2nd argument to the closure passed into package:shelf_web_socket's `webSocketHandler` method
- widen the dep on package:shelf_web_socket

This will allow us to add more type info to the closure that `webSocketHandler` expects; it's currently an untyped Function. See also https://github.com/dart-lang/shelf/issues/457 and https://github.com/dart-lang/shelf/pull/463.

This forward declares compatibility with `3.0` of `package:shelf_web_socket`; I _think_ this is necessary - as `dart test` uses both package:test and package:shelf_web_socket - but happy to hear otherwise.

---

<details>

</details>

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.25.11
+
+* Update the `package:shelf_web_socket` constraint to allow version `3.x`.
+
 ## 1.25.10
 
 * Update the `package:vm_service` constraint to allow version `15.x`.
--- a/pkgs/test/lib/src/runner/browser/compilers/dart2js.dart
+++ b/pkgs/test/lib/src/runner/browser/compilers/dart2js.dart
@@ -174,7 +174,10 @@ class Dart2JsSupport extends CompilerSupport with JsHtmlWrapper {
   @override
   (Uri, Future<WebSocketChannel>) get webSocket {
     var completer = Completer<WebSocketChannel>.sync();
-    var path = _webSocketHandler.create(webSocketHandler(completer.complete));
+    var path =
+        _webSocketHandler.create(webSocketHandler((WebSocketChannel ws, _) {
+      completer.complete(ws);
+    }));
     var webSocketUrl = serverUrl.replace(scheme: 'ws').resolve(path);
     return (webSocketUrl, completer.future);
   }
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test
-version: 1.25.10
+version: 1.25.11
 description: >-
   A full featured library for writing and running Dart tests across platforms.
 repository: https://github.com/dart-lang/test/tree/master/pkgs/test
@@ -29,7 +29,7 @@ dependencies:
   shelf: ^1.0.0
   shelf_packages_handler: ^3.0.0
   shelf_static: ^1.0.0
-  shelf_web_socket: '>=1.0.0 <3.0.0'
+  shelf_web_socket: '>=1.0.0 <4.0.0'
   source_span: ^1.8.0
   stack_trace: ^1.10.0
   stream_channel: ^2.1.0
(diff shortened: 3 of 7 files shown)


jakemac53:
We should be able to land this change without widening the constraint pre-emptively, and then in shelf_web_socket we can add a dependency override on the latest version of package:test (with the forwards compatible change).

Then, when shelf_web_socket is published, we can update the constraint here and publish again.

That avoids publishing any regular dependency constraints in any packages with versions that aren't actually there yet (it will only be an override on the dev dependency in shelf_web_socket, which is fine).

WDYT?

devoncarew:
> That avoids publishing any regular dependency constraints in any packages with versions that aren't actually there yet (it will only be an override on the dev dependency in shelf_web_socket, which is fine).

Sounds great! Forward declaring the dep - even for something we control - does feel like crossing your fingers and hoping.

pkgs/test/pubspec.yaml
```diff
@@ -29,7 +29,7 @@ dependencies:
   shelf: ^1.0.0
   shelf_packages_handler: ^3.0.0
   shelf_static: ^1.0.0
-  shelf_web_socket: '>=1.0.0 <3.0.0'
+  shelf_web_socket: '>=1.0.0 <4.0.0'
```
jakemac53, line 32: All we need to do for this one is just revert this back to 3.x, and update the changelog. Then we can land/publish and unblock the shelf_web_socket change.
 
```suggestion
  shelf_web_socket: '>=1.0.0 <3.0.0'
```

pkgs/test/lib/src/runner/browser/compilers/dart2js.dart
```diff
@@ -174,7 +174,10 @@ class Dart2JsSupport extends CompilerSupport with JsHtmlWrapper {
   @override
   (Uri, Future<WebSocketChannel>) get webSocket {
     var completer = Completer<WebSocketChannel>.sync();
-    var path = _webSocketHandler.create(webSocketHandler(completer.complete));
+    var path =
+        _webSocketHandler.create(webSocketHandler((WebSocketChannel ws, _) {
```
jakemac53, line 178: Without the type here, do we get a lint about the implicit cast from dynamic? Or is it an actual error?

In version 3.0 of shelf_web_socket the type will not need to be here, so we probably want to remove it, but I want to make sure we don't break users on 2.0. If that will be the case, we would want a comment here saying not to remove the type. Or, we would want to set the min constraints on shelf_web_socket to >3.0.

pkgs/test/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.25.11
+
+* Update the `package:shelf_web_socket` constraint to allow version `3.x`.
```
jakemac53, line 3:
```suggestion
* Update to be forward compatible with `package:shelf_web_socket` version `3.x`.
```

pkgs/test/lib/src/runner/browser/compilers/dart2js.dart
devoncarew: We get a lint on the webSocketHandler call (`The type of ws can't be inferred; a type must be explicitly provided.`) and an error on the completer line (`The argument type 'dynamic' can't be assigned to the parameter type 'FutureOr<WebSocketChannel>?'. `).

I'll add comments that the type is needed for compatibility with package:shelf_web_socket v2.

jakemac53 (approved):

devoncarew:
Is it expected here that the publish check fails? We just ignore for this package, but still use publishing automation after this lands?

```
 Package validation found the following potential issues:
  * Your dependency on "test_api" should allow more than one version. For example:
...
```

jakemac53:
> Is it expected here that the publish check fails? We just ignore for this package, but still use publishing automation after this lands?

Yes, we have to add the label to ignore warnings because we pin test_core and test_api.

PR #2426, opened 2024-12-13
Use dart_flutter_team_lints for matcher

natebosch:
Change the dependency and analysis options include to the stricter team
configuration.

Fix newly surfaced diagnostics. Most code changes are related to the
strict modes that are enabled by the team config.
Fix most implicit casts with explicit casts, but fix some arguments in
`typedMatches` implementation so they don't unnecessarily widen the type
back to `dynamic` before using the value.


natebosch:
Replaces https://github.com/dart-archive/matcher/pull/251

kevmoo (approved):

Issue #2441, opened 2024-12-29
`check(...).deepEquals(...)` fails to describe a large-ish Map


matanlurey:
I am not 100% sure of the _cause_ of the failure (yet), but it can be reproduced with the following test case:

```dart
test('can be described failing compared to another large map', () {
  const expected = {
    1: -5,
    2: -4,
    3: -4,
    4: -3,
    5: -3,
    6: -2,
    7: -2,
    8: -1,
    9: -1,
    10: 0,
    11: 0,
    12: 1,
    13: 1,
    14: 2,
    15: 2,
    16: 3,
    17: 3,
    18: 4,
    19: 4,
    20: 5,
    21: 5,
    22: 6,
    23: 6,
    24: 7,
    25: 7,
    26: 8,
    27: 8,
    28: 9,
    29: 9,
    30: 10,
  };
  final actual = {
    1: -4,
    2: -4,
    3: -3,
    4: -3,
    5: -2,
    6: -2,
    7: -1,
    8: -1,
    9: 0,
    10: 0,
    11: 0,
    12: 1,
    13: 1,
    14: 2,
    15: 2,
    16: 3,
    17: 3,
    18: 4,
    19: 4,
    20: 5,
    21: 5,
    22: 6,
    23: 6,
    24: 7,
    25: 7,
    26: 8,
    27: 8,
    28: 9,
    29: 9,
    30: 10
  };
  check(actual).not((a) => a.deepEquals(expected));
});
```

The failure will look like this:

```txt
type 'List<Iterable<String>>' is not a subtype of type 'Iterable<List<String>>' of 'newContents'
dart:collection                                ListBase.replaceRange
package:checks/src/describe.dart 75:14         _prettyPrintCollection
package:checks/src/describe.dart 54:12         _prettyPrint
package:checks/src/describe.dart 17:45         literal
package:checks/src/checks.dart 1006:27         Rejection._fillActual
package:checks/src/checks.dart 706:53          _TestContext.expect.<fn>
package:checks/src/checks.dart 594:65          _Present.apply
package:checks/src/checks.dart 706:16          _TestContext.expect
package:checks/src/extensions/map.dart 103:12  MapChecks.deepEquals
test/extensions/map_test.dart 97:32            main.<fn>.<fn>
package:checks/src/checks.dart 117:12          softCheck
package:checks/src/extensions/core.dart 53:13  CoreChecks.not.<fn>
package:checks/src/checks.dart 706:43          _TestContext.expect.<fn>
[...]

PR #2442, opened 2024-12-29
Fix a type error that occurs comparing two large maps with `deepEquals`.

matanlurey:
Closes https://github.com/dart-lang/test/issues/2441.

I am not 100% sure _why_ this occurs, and why the repro case needs to be of a certain design/size, but here ya go.


pkgs/checks/lib/src/describe.dart
```diff
@@ -72,7 +72,11 @@ Iterable<String> _prettyPrint(
 Iterable<String> _prettyPrintCollection(
     String open, String close, List<Iterable<String>> elements, int maxLength) {
   if (elements.length > _maxItems) {
-    elements.replaceRange(_maxItems - 1, elements.length, [
+    // Force inference as List<String>, otherwise the type is
+    // List<Iterable<String>> which ends up as a type error in dart:collection
+    // when the underlying list is actually a List<List<String>>. See
+    // https://github.com/dart-lang/test/issues/2441 for more details.
+    elements.replaceRange(_maxItems - 1, elements.length, <List<String>>[
```
jakemac53, line 79: Another option could be to make this a top level variable? Then we wouldn't have to force inference and it would be shared across invocations.

jakemac53:
Fwiw this is the actual culprit line that creates a `List<List<String>>` https://github.com/dart-lang/test/blob/master/pkgs/checks/lib/src/describe.dart#L48.

A different solution would be to convert this to return an iterable, which may have (good or bad) performance characteristics, something like this:

```dart
      return key.take(key.length - 1).followedBy(
          ['${key.last}: ${value.first}']).followedBy(value.skip(1));
```

pkgs/checks/test/extensions/map_test.dart
```diff
@@ -29,6 +29,65 @@ void main() {
   test('values', () {
     check(_testMap).values.contains(1);
   });
+  test('can be described failing compared to another large map', () {
```
jakemac53, line 32: Looks like the same issue exists for lists so it would be good to also test that (although I think your fix will work for both).

matanlurey:
It might be a few days until I return to this PR (side-project) so it will need some support if you want it merged sooner :)

jakemac53:
No worries, I don't think there is a rush on this. If you do need to hand it off though let us know and one of us could probably take it over.

matanlurey:
Thanks for taking it over!

jakemac53 (approved):

PR #2445, opened 2025-01-14
Unblock CI: Ignore deprecation and skip a test

natebosch:
Unblock CI until these can be migrated and fixed.


kevmoo:
Or, you know, we could fix these! 😄

kevmoo:
At least a TODO and an open issue to migrate?

jakemac53 (approved):

natebosch:
> At least a TODO and an open issue to migrate?

https://github.com/dart-lang/test/issues/2447

I don't see value in a TODO when a grep for `dart:js_util` already has 100% fidelity.

natebosch:
Holding back the recompile since that is triggering some other CI bug

natebosch:
hmm, I didn't know that we don't support making configuration specific to an OS when it's running on the browser... I thought that was possible but when we evaluate the `OnPlatform` selectors the browser tests always treat the OS as empty.

https://github.com/dart-lang/test/blob/f364fc8291d668d85c702a5b9f9a4f2e5c1ade0e/pkgs/test_api/lib/src/backend/suite_platform.dart#L41-L43

natebosch:
linux firefox tests are also not working, we can skip these entirely on firefox for now to get unblocked.

PR #2448, opened 2025-01-17
Migrate host.dart to new JS interop

natebosch:
Recompile with the latest dart2js.


pkgs/test/lib/src/runner/browser/static/host.dart.js
```diff
@@ -18188,11 +16322,14 @@
...
+    _lazyFinal($, "_v8JsFrame", "$get$_v8JsFrame", () => A.RegExp_RegExp("^\\s*at (?:(\\S.*?)(?: \\[as [^\\]]+\\])? \\((.*)\\)|(.*))$", false));
+    _lazyFinal($, "_v8JsUrlLocation", "$get$_v8JsUrlLocation", () => A.RegExp_RegExp("^(.*?):(\\d+)(?::(\\d+))?$|native$", false));
+    _lazyFinal($, "_v8WasmFrame", "$get$_v8WasmFrame", () => A.RegExp_RegExp("^\\s*at (?:(?<member>.+) )?(?:\\(?(?:(?<uri>\\S+):wasm-function\\[(?<index>\\d+)\\]\\:0x(?<offset>[0-9a-fA-F]+))\\)?)$", false));
     _lazyFinal($, "_v8EvalLocation", "$get$_v8EvalLocation", () => A.RegExp_RegExp("^eval at (?:\\S.*?) \\((.*)\\)(?:, .*?:\\d+:\\d+)?$", false));
     _lazyFinal($, "_firefoxEvalLocation", "$get$_firefoxEvalLocation", () => A.RegExp_RegExp("(\\S+)@(\\S+) line (\\d+) >.* (Function|eval):\\d+:\\d+", false));
-    _lazyFinal($, "_firefoxSafariFrame", "$get$_firefoxSafariFrame", () => A.RegExp_RegExp("^(?:([^@(/]*)(?:\\(.*\\))?((?:/[^/]*)*)(?:\\(.*\\))?@)?(.*?):(\\d*)(?::(\\d*))?$", false));
+    _lazyFinal($, "_firefoxSafariJSFrame", "$get$_firefoxSafariJSFrame", () => A.RegExp_RegExp("^(?:([^@(/]*)(?:\\(.*\\))?((?:/[^/]*)*)(?:\\(.*\\))?@)?(.*?):(\\d*)(?::(\\d*))?$", false));
```
natebosch, line 16330: This is the blocking line and regex.

natebosch:
https://github.com/dart-lang/test/security/code-scanning/149

> Inefficient regular expression
>
> This part of the regular expression may cause exponential backtracking on strings starting with '(' and containing many repetitions of ')('.
>
> First detected in commit on Aug 9, 2023

As far as I can tell this regex is not changing, but other content on the same line as the regex is changing.

Confusingly the error is reported with the _old_ line number, but the regex doesn't change between the two copies.

@devoncarew @kevmoo - do you know what caused this check to start happening when we previously had this regex checked in? Is it something we can turn off temporarily?

@sigmundch - is this potentially inefficient regex a concern for the dart2js team?

sigmundch:
I'm not particularly concerned. This seems to come from https://github.com/dart-lang/tools/blob/ce3003493b86333e7652ee5bb62bc7913b2d11b7/pkgs/stack_trace/lib/src/frame.dart#L71 - this logic is used only to parse frame lines from stack traces in FF/Safari, which usually don't start with `(` and usually only contain one pair of parens.

jakemac53 (approved):

kevmoo:
that's a HUGE drop in code size. Did we drop dart:html here?

natebosch:
> that's a HUGE drop in code size. Did we drop dart:html here?

it does look like there is a big chunk of removed stuff that look like `dart:html` wrappers `A.AreaElement.prototype = {...`

This does not yet migrate to the new interop though. Maybe something made treeshaking work better here? Nothing that is dropped strikes me as something we would be using.

PR #2451, opened 2025-01-24
Make zone handlers call zone.run to run callbacks.

lrhn:
The platform library `Zone` implementation will change to not use `bindCallbackGuarded` in the `Timer` constructors and top-level `scheduleMicrotask` function.
Instead it only calls `registerCallback`, and then the `Zone.create{,Periodic}Timer` and `Zone.scheduleMicrotask` are in charge of using `Zone.runGuarded` on the callback when timer/microtask event happens.

This ensures that a surrounding zone's `registerCallback` can't make the callback throw in a way that is outside of the `runGuarded` call.

Also makes periodic timers that are delayed past multiple tick points, update their `tick` count by more than one instead of running multiple times, like timers should.

Adds testing of using the `FakeAsync`'s `Zone` functions directly, and embedding it in another zone to check that it does run using `Zone.run`.
Adds testing of periodic timers incrementing `tick` by more than one.

### DIFF
--- a/pkgs/fake_async/CHANGELOG.md
+++ b/pkgs/fake_async/CHANGELOG.md
@@ -1,5 +1,14 @@
 ## 1.3.3-wip
 
+* Make the zone `create*Timer` and `scheduleMicrotask`
+  be responsible for running callbacks in the zone they're
+  scheduled in, matching (new) standard zone behavior.
+  (The `Timer` constructors and top-level `scheduleMicrotask`
+  used to bind their callback, but now only registers it,
+  leaving the zone to run in the correct zone and handle errors.)
+* Make periodic timers increment their `tick` by more than one
+  if `elapseBlocking` advanced time past multiple ticks.
+
 ## 1.3.2
 
 * Require Dart 3.3
--- a/pkgs/fake_async/CHANGELOG.md
+++ b/pkgs/fake_async/CHANGELOG.md
@@ -16,7 +16,6 @@
   the callback of a periodic timer would immediately invoke
   the same timer.
 * Move to `dart-lang/test` monorepo.
-* Require Dart 3.5.
 
 ## 1.3.1
 
--- a/pkgs/fake_async/lib/fake_async.dart
+++ b/pkgs/fake_async/lib/fake_async.dart
@@ -152,7 +152,7 @@ class FakeAsync {
   /// Throws an [ArgumentError] if [duration] is negative.
   void elapseBlocking(Duration duration) {
     if (duration.inMicroseconds < 0) {
-      throw ArgumentError('Cannot call elapse with negative duration');
+      throw ArgumentError.value(duration, 'duration', 'Must not be negative');
     }
 
     _elapsed += duration;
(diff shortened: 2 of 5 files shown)


pkgs/fake_async/CHANGELOG.md
```diff
@@ -1,13 +1,21 @@
...
+  if `elapseBlocking` advanced time past multiple ticks.
+
 ## 1.3.2
 
 * Require Dart 3.3
 * Fix bug where a `flushTimers` or `elapse` call from within
   the callback of a periodic timer would immediately invoke
   the same timer.
 * Move to `dart-lang/test` monorepo.
-* Require Dart 3.5.
```
lrhn, line 10: Didn't actually change `pubspec.yaml`.
(Only one of the first and last line here is correct and needed.)

pkgs/fake_async/lib/fake_async.dart
```diff
@@ -207,23 +210,25 @@ class FakeAsync {
...
+    bool flushPeriodicTimers = true,
+  }) {
     final absoluteTimeout = _elapsed + timeout;
     _fireTimersWhile((timer) {
       if (timer._nextCall > absoluteTimeout) {
         // TODO(nweiz): Make this a [TimeoutException].
         throw StateError('Exceeded timeout $timeout while flushing timers');
       }
 
-      if (flushPeriodicTimers) return _timers.isNotEmpty;
```
lrhn, line 220: The list is never empty here, it always contains at least `timer`, which is never removed until it is `_fire`d.

pkgs/fake_async/lib/fake_async.dart
```diff
@@ -207,23 +210,25 @@ class FakeAsync {
...
-      // Continue firing timers until the only ones left are periodic *and*
-      // every periodic timer has had a change to run against the final
-      // value of [_elapsed].
-      return _timers
-          .any((timer) => !timer.isPeriodic || timer._nextCall <= _elapsed);
+      // If no timers are due, continue running timers
+      // (and advancing time to their next due time)
+      // if flushing periodic timers,
+      // or if there is any non-periodic timer left.
+      return flushPeriodicTimers || _timers.any((timer) => !timer.isPeriodic);
```
lrhn, line 231: This function should just check whether to run the *current* timer, which it should unless it's not due, we're not flushing periodic timers, and all remaining timers (including this one) are periodic timers.

pkgs/fake_async/test/fake_async_test.dart
```diff
@@ -27,9 +27,18 @@ void main() {
...
   group('elapseBlocking', () {
-    test('should elapse time without calling timers', () {
-      Timer(elapseBy ~/ 2, neverCalled);
-      FakeAsync().elapseBlocking(elapseBy);
+    test('should elapse time without calling timers or microtasks', () {
+      FakeAsync()
+        ..run((_) {
+          // Do not use [neverCalled] from package:test.
+          // It schedules timers to "pump the event loop",
+          // which stalls the test if you don't call `elapse` or `flushTimers`.
```
lrhn, line 35: I tried, it did not go well.

lrhn:
Should have been update to other CL.

natebosch (approved):

PR #2461, opened 2025-02-25
Introduce an analyzer plugin for the test package.

srawlins:
Note: I'm not strongly suggesting this PR be accepted today as is. But I'm offering it up.

As part of the new analyzer plugin system, I'd like to offer some first-party-maintained analyzer plugins. I don't intend for us to maintain complex plugins necessarily. But good bang-for-our-buck plugins have some great benefits:

* We provide examples of analyzer plugins for others to crib. This plugin clocks in at 160 lines of code, offering a static warning rule and a quick fix.
* We provide value to customers of our first-party packages, like the test package.

If there are other static analyses that the team has dreamed of, I'd love to evaluate them for complexity and possibly add more static analyses to this plugin.

Again, we don't need to land this any time soon, if the team is hesitant to check this in, maintain it, etc. The new analyzer system is not officially released, and the API is probably not super stable. I'm opening this PR for the discussion, and to have URL to point to as a real world analyzer plugin example.

### DIFF
--- a/pkgs/test_analyzer_plugin/README.md
+++ b/pkgs/test_analyzer_plugin/README.md
@@ -1,0 +1,13 @@
+# test_analyzer_plugin
+
+This package is an analyzer plugin that provides additional static analysis for
+usage of the test package.
+
+This analyzer plugin provides the following additional analysis:
+
+* Report a warning when a `test` or a `group` is declared inside a `test`
+  declaration. This can _sometimes_ be detected at runtime. This warning is
+  reported statically.
+
+* Offer a quick fix in the IDE for the above warning, which moves the violating
+  `test` or `group` declaration below the containing `test` declaration.
--- a/pkgs/test_analyzer_plugin/lib/src/fixes.dart
+++ b/pkgs/test_analyzer_plugin/lib/src/fixes.dart
@@ -1,0 +1,58 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analysis_server_plugin/edit/dart/correction_producer.dart';
+import 'package:analysis_server_plugin/edit/dart/dart_fix_kind_priority.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer_plugin/utilities/change_builder/change_builder_core.dart';
+import 'package:analyzer_plugin/utilities/fixes/fixes.dart';
+import 'package:analyzer_plugin/utilities/range_factory.dart';
+
+import 'utilities.dart';
+
+class MoveBelowEnclosingTestCall extends ResolvedCorrectionProducer {
+  static const _wrapInQuotesKind = FixKind(
+      'dart.fix.moveBelowEnclosingTestCall',
+      DartFixKindPriority.standard,
+      "Move below the enclosing 'test' call");
+
+  MoveBelowEnclosingTestCall({required super.context});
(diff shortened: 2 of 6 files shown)


pkgs/test_analyzer_plugin/lib/src/fixes.dart
```diff
@@ -0,0 +1,58 @@
...
+import 'package:analysis_server_plugin/edit/dart/dart_fix_kind_priority.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer_plugin/utilities/change_builder/change_builder_core.dart';
+import 'package:analyzer_plugin/utilities/fixes/fixes.dart';
+import 'package:analyzer_plugin/utilities/range_factory.dart';
+
+import 'utilities.dart';
+
+class MoveBelowEnclosingTestCall extends ResolvedCorrectionProducer {
+  static const _wrapInQuotesKind = FixKind(
```
jakemac53, line 15: nit: rename this :)

pkgs/test_analyzer_plugin/lib/src/fixes.dart
```diff
@@ -0,0 +1,58 @@
...
+      // This fix may break code by moving references to variables away from the
+      // scope in which they are declared.
+      CorrectionApplicability.singleLocation;
+
+  @override
+  FixKind get fixKind => _wrapInQuotesKind;
+
+  @override
+  Future<void> compute(ChangeBuilder builder) async {
+    var methodCall = node;
```
jakemac53, line 33: Where does this `node` come from?

pkgs/test_analyzer_plugin/README.md
```diff
@@ -0,0 +1,13 @@
+# test_analyzer_plugin
```
jakemac53, line 1: We should probably add something to the package:test readme as well, informing users of this plugin and how to enable it for their project.

pkgs/test_analyzer_plugin/lib/src/fixes.dart
srawlins: `node` is a field on the superclass, always set to the error node on which the warning has been reported.

pkgs/test_analyzer_plugin/README.md
srawlins: Since we are very "soft launching" analyzer plugins ATM, I think we don't want to advertise it there yet. WDYT?

pkgs/test_analyzer_plugin/lib/src/fixes.dart
srawlins: Heh, good catch, done.

pkgs/test_analyzer_plugin/README.md
jakemac53: Up to you I guess, but we won't get usage if we don't advertise it

srawlins:
> but what about an annotation on Matchers to specify what types we are expecting to handle and a diagnostic if they aren't being tested against them?

That's a fun idea, you could open an issue against this repo.

For first-party matchers in the matcher package, I don't think it would be worth the extra complexity. So the feature would be for custom matchers implemented outside of the matcher and test packages.

srawlins:
OK I'd love to land this actually. Esp with Dart 3.10 out today. Anyone know what the failing Publish check is about? I can't make heads or tails of it.

kevmoo:
@srawlins 

image

srawlins:
How did you find that, Kevin? I grep the failed Publish CI task for 'error' and nothing :( Just 300 lines of 'Found published package at Directory'.

PR #2467, opened 2025-03-05
Suggest using `dart pub add dev:checks`

jonasfj:
Then we don't have to update the version number in the `README.md` every time we make publish a new release.


natebosch (approved):
Thanks!

PR #2478, opened 2025-04-08
Migrate from deprecated `dart:js`, `dart:js_util`, `package:js_util` to `dart:js_interop`

### LINKED ISSUE #2447: Migrate off dart:js_util
https://github.com/dart-lang/test/pull/2445

osa1:
Closes #2447.

### DIFF
--- a/pkgs/test/lib/src/runner/browser/dom.dart
+++ b/pkgs/test/lib/src/runner/browser/dom.dart
@@ -10,7 +10,6 @@ import 'dart:js_util' as js_util;
 
 extension type Window(EventTarget _) implements EventTarget {
   @pragma('dart2js:as:trust')
-
   Window get parent => _.getProperty('parent'.toJS) as Window;
 
   external Location get location;
--- a/pkgs/test/lib/src/runner/browser/dom.dart
+++ b/pkgs/test/lib/src/runner/browser/dom.dart
@@ -115,7 +115,7 @@ extension type MessageEvent(Event _) implements Event {
   external String get origin;
 
   List<MessagePort> get ports =>
-      (getProperty('ports'.toJS).dartify() as List).cast<MessagePort>();
+      getProperty<JSArray>('ports'.toJS).toDart.cast<MessagePort>();
 
   /// The source may be a `WindowProxy`, a `MessagePort`, or a `ServiceWorker`.
   ///
--- a/pkgs/test/lib/src/runner/browser/dom.dart
+++ b/pkgs/test/lib/src/runner/browser/dom.dart
@@ -2,64 +2,55 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+
 // ignore: deprecated_member_use
 import 'dart:js_util' as js_util;
 
-// ignore: deprecated_member_use
-import 'package:js/js.dart';
-
-@JS()
-@staticInterop
-class Window extends EventTarget {}
-
-extension WindowExtension on Window {
+extension type Window(EventTarget _) implements EventTarget {
   @pragma('dart2js:as:trust')
-  Window get parent => js_util.getProperty<dynamic>(this, 'parent') as Window;
+
+  Window get parent => _.getProperty('parent'.toJS) as Window;
+
   external Location get location;
-  Console get console => js_util.getProperty(this, 'console') as Console;
+
+  Console get console => _.getProperty('console'.toJS) as Console;
+
   CSSStyleDeclaration? getComputedStyle(Element elt, [String? pseudoElt]) =>
-      js_util.callMethod(this, 'getComputedStyle', <Object>[
-        elt,
-        if (pseudoElt != null) pseudoElt
+      _.callMethodVarArgs('getComputedStyle'.toJS, <JSAny?>[
+        elt._._,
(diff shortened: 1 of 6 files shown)


pkgs/test/lib/src/runner/browser/dom.dart
```diff
@@ -2,100 +2,74 @@
...
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
 
 // ignore: deprecated_member_use
-import 'package:js/js.dart';
-
-@JS()
-@staticInterop
-class Window extends EventTarget {}
+import 'dart:js_util' as js_util;
```
osa1, line 9: This remains because there's a `callConstructor` call below and I'm not sure what the `js_interop` equivalent of it is. @srujzs could you advise please?

osa1:
@natebosch This is an internal change, does it need a changelog entry?

osa1:
Thanks -- in that case this should be ready for reviewing and merging.

pkgs/test/tool/host.dart
```diff
@@ -104,7 +103,7 @@ final _currentUrl = Uri.parse(dom.window.location.href);
 /// does mean that the server needs to be sure to nest its [MultiChannel]s at
 /// the same place the client does.
 void main() {
-  dom.window.console.log('Dart test runner browser host running');
+  dom.window.console.log('Dart test runner browser host running'.jsify());
```
srujzs, line 106: `.toJS`.

srujzs (approved):

pkgs/test/lib/src/runner/browser/dom.dart
```diff
@@ -2,100 +2,71 @@
...
 
-// ignore: deprecated_member_use
-import 'package:js/js.dart';
-
-@JS()
-@staticInterop
-class Window extends EventTarget {}
-
-extension WindowExtension on Window {
+extension type Window(EventTarget _) implements EventTarget {
```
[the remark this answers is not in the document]
osa1: I think that would make sense, but let's do it separately. I mainly wanted to learn more about `js_interop` with this PR while also doing something useful.

natebosch:
@osa1 - it looks like the current stable SDK has broken our JS compile. Do you have a link showing the current pattern for exposing a Dart interface to JS?

https://github.com/dart-lang/test/blob/8643fbf375330cd1099cdc7434306532b4bcaaf1/pkgs/test/tool/host.dart#L20-L26

srujzs:
Ah, yeah this would only give you errors if you tried to compile it due to the `void Function()`. The call to the constructor used to have `allowInterop`, so it's just a matter of updating them to use `toJS` and using `JSFunction`.

I expect the correct changes to look like this, but I'm not sure if I need to rebuild anything:

```
diff --git a/pkgs/test/tool/host.dart b/pkgs/test/tool/host.dart
index fe8c6144..bdae52e7 100644
--- a/pkgs/test/tool/host.dart
+++ b/pkgs/test/tool/host.dart
@@ -17,24 +17,23 @@ import 'package:test/src/runner/browser/dom.dart' as dom;
 ///
 /// These are exposed so that tools like IDEs can interact with them via remote
 /// debugging.
-@JS()
-@anonymous
-@staticInterop
-class _JSApi {
-  external factory _JSApi(
-      {void Function() resume, void Function() restartCurrent});
+extension type _JSApi._(JSObject _) implements JSObject {
+  external factory _JSApi({
+    JSFunction resume,
+    JSFunction restartCurrent,
+  });
 }

 extension _JSApiExtension on _JSApi {
   /// Causes the test runner to resume running, as though the user had clicked
   /// the "play" button.
   // ignore: unused_element
-  external Function get resume;
+  external JSFunction get resume;

   /// Causes the test runner to restart the current test once it finishes
   /// running.
   // ignore: unused_element
-  external Function get restartCurrent;
+  external JSFunction get restartCurrent;
 }

 /// Sets the top-level `dartTest` object so that it's visible to JS.
@@ -151,9 +150,9 @@ void main() {
       if (!dom.document.body!.classList.contains('paused')) return;
       dom.document.body!.classList.remove('paused');
       serverChannel.sink.add({'command': 'resume'});
-    }, restartCurrent: () {
+    }.toJS, restartCurrent: () {
       serverChannel.sink.add({'command': 'restart'});
-    });
+    }.toJS);
   }, (error, stackTrace) {
     dom.window.console.warn('$error\n${Trace.from(stackTrace).terse}'.toJS);
   });
```

osa1:
@natebosch how do I reproduce the issue?

I'm not aware of a documentation that shows a full example with Dart functions exposed to JS and the other way around. I mainly followed the links from https://dart.dev/interop/js-interop, and looked at SDK tests that use `dart:js_interop`, like the ones [here][1]. I'm also familiar with how `dart:js_interop` is implemented in dart2wasm, so that also helped.

[1]: https://github.com/dart-lang/sdk/tree/main/tests/lib/js/static_interop_test

natebosch:
> how do I reproduce the issue?

By recompiling `host.dart`.

```
dart compile js tool/host.dart -o lib/src/runner/browser/static/host.dart.js
```

https://github.com/dart-lang/test/blob/0f1e3792fec32ce9e0d5578efaf42aa4d5044bd5/pkgs/test/tool/README.md?plain=1#L6

Thanks both for the pointers.

We have some CI issues with updating the compiled version at the moment but I'll work on getting this unblocked and landed in https://github.com/dart-lang/test/pull/2448

Issue #2029, opened 2023-06-02
Test line/col/filenames are incorrect when using package:test_reflective_loader


DanTup:
Usually when we run tests with the JSON reporter, we get the file/line/col of the test back. We can now also run tests by line number.

However, the file/line/col information is incorrect for tests using `package:test_reflective_loader` so if we try to run by line then things don't work right. Or if you try to "Go to Test" that goes to the wrong place. I believe the issue is that when the `test()` function is run, the actual location of the test method isn't on the stack at all, so what we end up with is the location of the call to `defineReflectiveSuite`:

```
danny@Dannys-MacBook-Pro analysis_server % dart test -r json test/lsp/call_hierarchy_test.dart --plain-name "PrepareCallHierarchyTest | test_method"

// snip ...

"line": 175,
"column": 28,
"url": "package:test_reflective_loader/test_reflective_loader.dart",
// This is the call to defineReflectiveSuite
"root_line": 12,
"root_column": 3,
"root_url": "file:///Users/danny/Dev/Google/dart-sdk/sdk/pkg/analysis_server/test/lsp/call_hierarchy_test.dart"

// snip ...
```

So, the request is to support this style of test in a way that the correct location can be reported. The code that's calling `test()` lives here:

https://github.com/dart-lang/test_reflective_loader/blob/40d61b16647cd61b02d806fea46362ef07e7c502/lib/test_reflective_loader.dart#L175

Looking at the code, it has a `MethodMirror` at the point where `test` is called, and that has a `.location` that looks like it's probably the values we want. However I can't see a way it could pass them to `package:test`. Would it be feasible to allow passing an overridden location to `test()` (or having some similar API that supported this)?

jakemac53:
hmm, yeah the only way I could see this working is through some sort of special information passed through the `test` function. We would likely have a similar issue with a codegen/macro based package that was trying to do the same thing as this package, not sure if we would ever want to expose line/col info to macros though (it would hurt invalidation).

jakemac53:
cc @natebosch wdyt about a way to override the line/col info?

natebosch:
Would we also need to allow overriding the URI for where the caller wants to pretend the `test` call was made?

I think this would be OK to expose. We should consider whether to do something like require importing the definitions from `test_core` and having a copy in `test` that doesn't expose those arguments.

DanTup:
> Would we also need to allow overriding the URI

[...]

PR #2481, opened 2025-04-09
Allow specifying an explicit location for test/groups

DanTup:
Usually test locations (which are used both to report the location of tests in the json reporter, and also for filtering tests by line/col) are inferred from the call stack when test() is called.

This changes adds a new `location` parameter to group/test to allow this information to be provided explicitly. This can be useful where the call to `test()` doesn't contain the actual location of the declaration in the call stack (for example when using `pkg:test_reflective_loader`).

See https://github.com/dart-lang/test/issues/2029
See https://github.com/Dart-Code/Dart-Code/issues/5480


pkgs/test_api/lib/scaffolding.dart
```diff
@@ -16,6 +16,7 @@ export 'src/backend/configuration/skip.dart' show Skip;
 export 'src/backend/configuration/tags.dart' show Tags;
 export 'src/backend/configuration/test_on.dart' show TestOn;
 export 'src/backend/configuration/timeout.dart' show Timeout;
+export 'src/backend/test_location.dart' show TestLocation;
```
DanTup, line 19: I don't know if the changes I made to public API are reasonable (or if I've exporting things in the right ways), so please check this carefully :-)

pkgs/test_api/lib/src/backend/test_location.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// The location of a test or group.
+class TestLocation {
```
DanTup, line 6: I originally just added a line/col to the `GroupEntry` class, but it turns out we need uri too, so it felt better to wrap this up in a class (so you either provide the whole lot or none at all).

We need a URI because filtering by the stack trace, we can say "if any frame of the call stack matches the requested file+line+col", but when using the location override, we wouldn't have a uri/path to match on.

This does mean we need to ensure this class is exported and available to any callers of `test()`.

pkgs/test_core/lib/src/runner.dart
```diff
@@ -324,6 +326,17 @@ class Runner {
             // The absolute path as it will appear in stack traces.
             var absoluteSuitePath = File(path).absolute.uri.toFilePath();
 
+            // First check if we're a match for the overridden location.
+            if (location != null) {
+              if ((line == null || location.line == line) &&
+                  (col == null || location.column == col) &&
+                  location.uri.isScheme('file') &&
+                  location.uri.toFilePath() == absoluteSuitePath) {
```
DanTup, line 334: I've assumed we're ok to only handle file URIs here. The code below for stack traces does handle another case (and mentions `--filesystem-scheme`) but it's not clear to me under what circumstances that would be used. If we need to support some equivalent here, I'll need to better understand what that is.

pkgs/test/test/runner/line_and_col_test.dart
```diff
@@ -80,6 +110,52 @@ void main() {
...
 
+    test('additionally selects groups with a matching custom location',
+        () async {
+      // TODO(dantup): This fails because we don't ever look at a groups
+      //  location.. instead, we only look at tests. The reason we can
+      //  normally run groups by line/col is because they just happen to be
+      //  in the stack trace for the test() call too.
+      //
+      //  So maybe we need to look up the tree to match location (in
+      //  Runner._loadSuites() filter)?
```
DanTup, line 121: Currently this test fails, because we only put the `TestLocation` on the `group`. It turns out that all filtering is just done based on tests, and the reason we can run a test by the groups line+col is because that location also happens to be in the call stack of the `test()` call (and therefore we match on the test, not the group).

So here, we fail to match anything, because the tests `location`s are `null` and don't match, and we never look at the group.

My feeling is that we need to change the filter to also look up the chain at parents when checking the `location`, however at a glance we don't seem to have easy access to that in the place where we're filtering (`Runner._loadSuites`).

DanTup (commented):
@jakemac53 this doesn't quite work yet - see comments on the test about filtering groups.

I'm also not sure if the API changes I've made are reasonable, so interested in feedback on that too.

pkgs/test/test/runner/json_reporter_test.dart
```diff
@@ -587,6 +587,70 @@ void customTest(String name, dynamic Function() testFn) => test(name, testFn);
...
     });
+
+    test('groups and tests with custom locations', () {
+      return _expectReport('''
+      group('group 1 inferred', () {
+        setUpAll(() {});
+        test('test 1 inferred', () {});
+        tearDownAll(() {});
+      });
+      group('group 2 custom', location: TestLocation('file:///foo/group', 123, 234), () {
```
jakemac53, line 598: Maybe these URIs should be package root relative? Or at least support that?

pkgs/test/test/runner/line_and_col_test.dart
jakemac53: Ah yeah that makes sense, I do think we will end up needing to pass through the parent groups then.

pkgs/test_api/lib/src/backend/test_location.dart
jakemac53: I do think wrapping this all in a class makes sense, fewer new arguments to group/test as well.

pkgs/test_api/lib/src/backend/test_location.dart
```diff
@@ -0,0 +1,27 @@
...
+  final Uri uri;
+  final int line;
+  final int column;
+
+  TestLocation(String uri, this.line, this.column) : uri = Uri.parse(uri);
```
jakemac53, line 11: If we want a Uri, I would be inclined to just take a Uri here.

Otherwise you end up in situations where people have a Uri, and they have to call `toString()` before passing it to you, just for you to parse it again. And parsing Uris is actually pretty expensive.

pkgs/test_core/lib/src/runner.dart
jakemac53: We do have other URIs when tests are pre-compiled.

Specifically `org-dartlang-app:///` will be used among others.

jakemac53: Specifically these are used to obfuscate the file system - so the paths are relative to some other root which isn't machine specific (its some path under the repository, usually the root of the package or repository)

pkgs/test_core/lib/src/runner/reporter/json.dart
```diff
@@ -297,8 +303,21 @@ class JsonReporter implements Reporter {
...
-      SuitePlatform platform, String suitePath) {
+  Map<String, dynamic> _locationInfo(
+      SuiteConfiguration suiteConfig,
+      Trace? trace,
+      TestLocation? location,
+      SuitePlatform platform,
+      String suitePath) {
+    // If this test has a location override, always use that.
+    if (location != null) {
+      return {
```
jakemac53, line 314: can this just be `location.serialize`?

pkgs/test_core/lib/src/runner.dart
DanTup: I've extracted the URI check into a new helper `matchesUri` that this now calls (as does the original `matchLineAndCol`).

DanTup: Although I also see that this code won't work right on Windows, since `absoluteSuitePath` is a real path (`.toFilePath()`) and is compared to a URI path (`absoluteSuitePath.endsWith(uri.path)`) that would have the wrong slashes on Windows. I'll tweak this to always use "URI paths".

pkgs/test/test/runner/json_reporter_test.dart
DanTup: I've added test cases for `org-dartlang-app:///` for the filtering, although it's not clear to me what supporting `org-dartlang-app` here would mean for the JSON... I think the clients of the JSON reporters would always expect file paths? If we supported `org-dartlang-app` here, would they need to be translated to real file paths somewhere?

pkgs/test_api/lib/src/backend/test_location.dart
DanTup: Done! When I was writing the tests, it felt weird to have to put `Uri.parse()`/`Uri.file()` repeatedly, but ofc it's likely the caller is some test framework and the user isn't writing these explicitly, so that was a bad optimisation.

pkgs/test/test/runner/line_and_col_test.dart
DanTup: hmm, I'm struggling to see how best to do this - it seems we don't have a reference from a `Group`/`GroupEntry` to the parent. Any pointers?

pkgs/test_core/lib/src/runner/reporter/json.dart
DanTup: We could, but it would be easy to accidentally change the JSON reporter output when changing the internal classes - I don't know if that's reasonable or not. If you think it is, I'll change this and also include a note in the `TestLocation` `serialize`/`deserialize` methods about that.

pkgs/test/test/runner/json_reporter_test.dart
jakemac53: Yeah I do think it will mess up some of these use cases - one thing we can try externally is running tests with build_web_compilers (build_runner test -- -p chrome), and see what all it takes to get things working.

Ultimately none of this works today anyways, so its probably ok if we just leave it in a state where it still doesn't work, as long as internally package:test does something sensible and we can leave any extra bits to later.

pkgs/test/test/runner/line_and_col_test.dart
jakemac53: In the Declarer, the `Group` entries are built bottom-up.

You could add a `Group? parent;` field to `GroupEntry`, and then assign it inside `Declarer.build` - setting the "parent" of each `GroupEntry` in `_entries` to the group you just created.

You could similarly assign the parent groups when deserializing tests over the wire this way (after creating the group inside `_Deserializer.deserializeGroup`).

I really don't know all that much about this part of the codebase though, so this approach could run into additional snags that I am not expecting.

pkgs/test_core/lib/src/runner/reporter/json.dart
jakemac53: I think I would just add the comment to the TestLocation methods and call `serialize` here - otherwise if we add fields or something to TestLocation, we might forget to update this, so there is some risk either way.

pkgs/test/test/runner/line_and_col_test.dart
DanTup: Thanks! I did this and it seems to have work. I wired up the parents in both the declarer and deserialization, although I'm not sure if both are necessary (it felt better than it always being null on one side though?).

I'm not sure having a mutable `parent` is ideal, but I'm not sure if that's avoidable (we can't pass parent to child and child to parent in constructors immutably?).

pkgs/test_core/lib/src/runner/reporter/json.dart
DanTup: Done!

DanTup:
@jakemac53 I think this all works as required now, but I'm unsure about:

- the public API changes (lots of them are in `src`, but I think many of these classes are exported from `lib`?)
- the mutable `parent` on `GroupEntry`
- whether this will affect how Flutter uses this (however once you're happy with it here, I can try updating in Flutter locally to verify before we commit)

pkgs/test_core/lib/src/runner.dart
```diff
@@ -351,13 +350,19 @@ class Runner {
...
-              if ((line == null || location.line == line) &&
-                  (col == null || location.column == col) &&
-                  matchesUri(location.uri)) {
-                return true;
+            // First check if we're a match for the overridden location for this
+            // item or any parents.
+            var current = test as GroupEntry?;
+            while (current != null) {
+              var location = current.location;
+              if (location != null) {
```
jakemac53, line 358: nit: this is a good place to use if/case

```suggestion
              if (current.location case var location?) {
```

pkgs/test_core/lib/src/runner/plugin/platform_helpers.dart
```diff
@@ -138,18 +138,28 @@ class _Deserializer {
...
+    }).toList();
+
+    var result = Group(group['name'] as String, entries,
         metadata: metadata,
         trace: trace,
         location: location,
         setUpAll: _deserializeTest(group['setUpAll'] as Map?),
         tearDownAll: _deserializeTest(group['tearDownAll'] as Map?));
+
+    // Now we have created the group, we can set it as the parent for all
```
jakemac53, line 154: nit: consider extracting this into a method on `group`, called `reparentChildren` or similar.

pkgs/test_api/lib/scaffolding.dart
jakemac53: the API changes LGTM

pkgs/test_core/lib/src/runner/plugin/platform_helpers.dart
jakemac53: Also consider asserting the parent is null

jakemac53 (commented):
This needs a pubspec/changelog update as well

pkgs/test_core/lib/src/runner.dart
DanTup: Oops, I usually do that and somehow forgot here - thanks!

pkgs/test_core/lib/src/runner/plugin/platform_helpers.dart
DanTup: Moving to the constructor is perfect - and avoids doing it in two places - thanks!

DanTup:
@jakemac53 

> This needs a pubspec/changelog update as well

I've added this to all three changelogs that this affected. I used the same description but let me know if these should be more specific (the changes only really work together, and I presume these are always shipped/updated together).

I've also tested with Flutter locally and this doesn't seem to affect anything there (at least, no analysis errors, and no failures from the tests I ran).

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
jakemac53, line 1: Lets add this file to gitignore :)

pkgs/test/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 1.25.16-wip
```
jakemac53, line 1: This is a new feature (new apis) so lets release it as such.

```suggestion
## 1.26.0-wip
```

pkgs/test_core/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## 0.6.9-wip
 
 - Add support for native assets for `dart test` in pub workspaces.
```
jakemac53, line 3:
```suggestion
* Add support for native assets for `dart test` in pub workspaces.
```

.vscode/settings.json
DanTup: You mean you don't keep your SDK at this location? 😄

I carefully skipped this on every change, until I stashed to merge and then brought it back 🤦‍♂️ Fixed!

pkgs/test/CHANGELOG.md
DanTup: hmm, I'm having a hard time making this work (although I'm bumping the minor part for all three packages - is that what you meant?)

By bumping these versions, the `matcher` constraint doesn't match. So I tried bumping that, but `test` depends back on `matcher`, so I end up with:

```
[test] dart pub get --no-example
Resolving dependencies...
Because matcher >=0.12.16+1 depends on test_api >=0.5.0 <0.8.0 and matcher >=0.12.16 <0.12.16+1 depends on test_api >=0.5.0 <0.7.0, matcher >=0.12.16 requires test_api >=0.5.0 <0.8.0.
So, because test depends on matcher >=0.12.16 <0.12.18 and test_workspace depends on test_api, version solving failed.
exit code 1
```

Do I need to also bump the major versions on `matcher`/`checks` and increase their constraints to match the new versions of everything else?

jakemac53: For test_api and test_core, since they are on the `0` major version - all the semantic versions are shifted to the right so they were actually fine (already feature releases). Only package:test needed to be updated

DanTup: Oh yeah!

I've done that, though I know see this:

```
[test] dart pub get --no-example
Resolving dependencies...
Because test_core depends on test_api 0.7.4 and test_workspace depends on test_api, version solving failed.
exit code 1
```

However, I'm also seeing that when I revert all of my changes - so I've pushed here so you can kick the bots while I figure out if this is just a me problem.

jakemac53: That change lgtm

DanTup:
@jakemac53 looks like the asserts I added caused a lot of failures. This is because in `forPlatform()` we create new groups, but we reuse the same child entries, so we try to reassign their parents to the new group:

https://github.com/dart-lang/test/blob/3ac991fa5ea406474e4d8d6c823fc078a6c6c6b4/pkgs/test_api/lib/src/backend/group.dart#L61-L67

I'm not sure how to handle this.. should we keep the parent pointing at the original group (not the new platform-scoped one), or do we need to clone the child entries?

jakemac53:
Looksl ike some legit test failures in the non-null asserts @DanTup

DanTup:
@jakemac53 could you check whether 990f13faf2e54307f92d9d9aed571ea77c52780a seems reasonable (and if so, kick the bots off again)?

jakemac53:
Hmmm, I do think the asserts are generally helpful, bashing over a parent is never intended, so ideally the tests would be updated to copy tests before reparenting them.

We could possibly file a TODO for this, but it makes me a bit uneasy to remove the asserts.

DanTup:
> it makes me a bit uneasy to remove the asserts

Yep, I agree - but is there something nicer we could do than changing all of these tests from something like:

```dart
runnerSuite(Group.root(tests.take(2))), 
```

to:

```dart
runnerSuite(Group.root(tests.take(2).map(_clone)), 
```

Maybe a version of `runnerSuite` that takes a list of tests, and creates the root group itself using clones of the tests?

```dart
runnerSuiteTests(tests.take(2)), 
```

jakemac53:
Either way sounds fine to me

DanTup:
@jakemac53 assuming I'm running them all correctly, I believe the tests are all passing now. I made two further changes:

1. Ensure `Test.filter()` always returns a clone (because when `Group.filter()` is called it returns a new group, and therefore must return new tests to have the parent wired up correctly)
2. Update some tests to clone the tests when they're taking them from the declared group and producing a new group subset for testing (I did this with an private `.asRootGroup()` for that test file)

If these look reasonable, please can you trigger the bots again? Thanks!

DanTup:
Looks like the bots are all green with those changes :-)

jakemac53:
> 1. Ensure `Test.filter()` always returns a clone (because when `Group.filter()` is called it returns a new group, and therefore must return new tests to have the parent wired up correctly)

This is the only part that feels a bit off - does `Group.filter()` need to create a copy? It might be cleaner if instead that doesn't return a copy so we don't have to copy the tests.

If it appears to be doing a copy for a legitimate reason, then LGTM.

DanTup:
@jakemac53 the `filter()` functions are documented as returning new suite/groups:

image

Internally, `Group` has an unmodifiable list for entries, so I assume it's to preserve immutability? We could possibly do the copy inside `group.filter()` but it felt more consistent to me to do it in each class itself (`Suite.filter()` and `Group.filter()` clones themselves, so it seems natural for `Test.filter()` to handle its own).

Happy to change this if you have other ideas, but this is the best I could think of (I'm much less familiar with this code though!).

jakemac53 (approved):

jakemac53:
@DanTup ready for me to merge? Also, how soon do we need this published?

DanTup:
@jakemac53 yep, I'm happy if you are :)

I guess we need to publish before I can update `pkg:test_reflective_loader` to fix the original issue, but it's definitely not high priority. It will improve using the test runner a little (where failures are shown, and where "Go to Test" goes to) for projects using test_reflective_loader, but I don't know how widespread that is outside of the few SDK projects I contribute to that use it.

jakemac53:
SGTM, I will let this bake a bit then before publishing

pkgs/test_api/lib/src/backend/group.dart
```diff
@@ -50,9 +57,22 @@ class Group implements GroupEntry {
...
+      this.setUpAll,
+      this.tearDownAll})
       : entries = List<GroupEntry>.unmodifiable(entries),
-        metadata = metadata ?? Metadata();
+        metadata = metadata ?? Metadata() {
+    for (var entry in entries) {
+      assert(entry.parent == null);
+      entry.parent = this;
+    }
+    assert(setUpAll?.parent == null);
```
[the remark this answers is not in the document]
sigurdm: https://pub.dev/documentation/test/latest/test/setUpAll.html says that setupAll can be nested inside groups. Not sure if that is what we are asserting against here.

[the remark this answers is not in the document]
sigurdm: No worries! Thanks for fixing

Issue #2482, opened 2025-04-09
Tests run with Safari and dart2wasm stall on load


eyebrowsoffire:
We're using this package in the web engine unit tests for Flutter. I am attempting to add some unit tests that run with Safari and dart2wasm (Safari now has WasmGC support) but I am running into a bug where the bootstrapping handshake between the host page and the iframe stalls out and fails. The reason for this is this [check](https://github.com/dart-lang/test/blob/8643fbf375330cd1099cdc7434306532b4bcaaf1/pkgs/test/tool/host.dart#L217)

For whatever reason, `message.source.location?.href` returns the URL of the host page rather than the URL loaded in the iframe. This only happens in Safari, and only happens when running a test compiled via dart2wasm. This seems like a Safari bug, but we might need a workaround for this if we want Safari+dart2wasm to work with this package.

FWIW, just commenting out that check does seem to allow my tests to run, but the web engine unit tests don't need to disambiguate between different test suites because we only run one at a time on the page (only one iframe).

PR #2486, opened 2025-04-15
Add support for dart2wasm+Safari

eyebrowsoffire:
This PR attempts to enable dart2wasm testing with Safari. For whatever reason, in Safari, the `message.source.location?.href` check on the message that sends the port from the iframe to the host window actually returns the parent frame rather than the iframe's href, and only when dart2wasm is used. This changes the logic to explicitly add this value to the message being sent to avoid this problem.

This fixes https://github.com/dart-lang/test/issues/2482


pkgs/test/tool/host.dart
```diff
@@ -201,15 +201,17 @@ StreamChannel<dynamic> _connectToIframe(String url, int id) {
     // running, but it's good practice to check the origin anyway.
     var message = event as dom.MessageEvent;
     if (message.origin != dom.window.location.origin) return;
+
     // Disambiguate between frames for different test suites.
     // Depending on the source type, the `location.href` may be missing.
-    if (message.source.location?.href != iframe.src) return;
+    var data = message.data as Map;
+    if (data['href'] != iframe.src) return;
```
natebosch, line 208: Changing this interface requires careful coordination across test backends. See https://github.com/dart-lang/test/issues/2065

eyebrowsoffire: Can you give some more detail on this? I see that you have a lot of steps listed in the linked issue, but I'm not sure I understand the fundamental issue that requires careful coordination. Are downstream clients not consuming the test package atomically as one unit?

In this case, the code in the iframe (i.e. `postMessageChannel`) is backwards compatible, it just adds a field and if the host page doesn't use it the existing behavior doesn't change. The host page is not exactly backwards compatible though, as it requires that `href` field to be set, and it won't work properly if the code in the iframe doesn't set it. However, if there are other implementations of the iframe-side of the message channel somewhere that need to be updated, we can update them first to include the field before we land this change. Can you clarify the compatibility issues here?

natebosch: > Are downstream clients not consuming the test package atomically as one unit?

Downstream clients define their own platforms. Flutter copy/pasted both the host and iframe side but in different places IIRC and their package update strategy requires cross compatibility in both directions. See https://github.com/flutter/engine/pull/46592 and https://github.com/flutter/flutter/pull/131881 for example

Google3 also has a copy of the host, but I think it uses the external iframe code so the backwards compatibility might already be set there.

Trying the change in google3 is a good place to start checking for potential issues.

eyebrowsoffire: At least for flutter, that's all code I'm familiar with and am happy to update. But since the test package is pinned to an absolute version in flutter, it won't break anything downstream until we actually update the pinned version. I can make the necessary updates and change the pinned version in a single PR.

As for Google3, I'll look into that and see if it breaks anything.

natebosch: > since the test package is pinned to an absolute version in flutter, it won't break anything downstream until we actually update the pinned version

IIRC the flutter copies are used in google3 too so we cannot rely on package pinning to avoid a period of cross compatibility.

natebosch: We'll probably need something like

```dart
    var data = message.data;
    if (message.source.location?.href != iframe.src &&
        (data is! Map || data['href'] != iframe.src)) {
      return;
    }

    message.stopPropagation();
    windowSubscription.cancel();

    switch (message.data) {
      case 'port' || {'messageType': 'port'}:
```

pkgs/test/lib/src/runner/browser/post_message_channel.dart
```diff
@@ -16,8 +16,10 @@ StreamChannel<Object?> postMessageChannel() {
   dom.window.console.log('Suite starting, sending channel to host'.toJS);
   var controller = StreamChannelController<Object?>(sync: true);
   var channel = dom.createMessageChannel();
-  dom.window.parent
-      .postMessage('port', dom.window.location.origin, [channel.port2]);
+  dom.window.parent.postMessage(
+      {'messageType': 'port', 'href': dom.window.location.href},
+      dom.window.location.origin,
+      [channel.port2]);
```
natebosch, line 22: This change is not backwards compatible. A compatible `host.dart` needs to be rolled out everywhere before we can land this change.

natebosch (approved):

pkgs/test/lib/src/runner/browser/post_message_channel.dart
eyebrowsoffire: Hmm, that won't work because we're transferring the port.

PR #2494, opened 2025-05-07
Set a debug name for test isolates

liamappelbe:
The Dart VM treats the main isolate specially. When the main isolate shuts down, the entire VM shuts down, even if there are other isolates still running.

package:coverage needs to collect coverage for all isolates. To do this, the VM is run with `--pause-isolates-on-exit`, and package:coverage waits for each isolate to pause, collects the coverage, then resumes the isolate and allows it to exit. But since the main isolate exiting would cause all isolates to exit, some of which may not have been collected yet, package:coverage needs to treat the main isolate specially, and avoid resuming it until every other isolate is collected.

Unfortunately there's no reliable way of determining which isolate is the main isolate using the VM service API. So package:coverage uses a heuristic: the first isolate is sees with a debug name of "main" is the main isolate. 99% of the time this works. But package:test spawns a new isolate for each test file, and doesn't set a debug name, so the name defaults to "main". If one of those test isolates is seen by package:coverage before the true main isolate, the system breaks. This causes about a 1% flakiness for test coverage in g3.

We're [working](https://dart-review.googlesource.com/c/sdk/+/427062) on extending the VM service API to tell us which isolate is the main one, but it's turning out to be non-trivial. In the meantime, a quick fix for the g3 flakiness is just to set some other debug name on the isolates that package:test spawns.

I've manually tested that this works, but let me know if I missed any isolate spawns that could be hit during `dart test`. Also, do you think this needs a test?

See also: https://github.com/dart-lang/sdk/issues/56732

### DIFF
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -111,7 +111,7 @@ class VMPlatform extends PlatformPlugin {
       if (platform.compiler == Compiler.exe) {
         throw UnsupportedError(
             'Unable to debug tests compiled to `exe` (tried to debug $path with '
-            'the `exe` compiler).');
+          'the `exe` compiler).');
       }
       var info =
           await Service.controlWebServer(enable: true, silenceOutput: true);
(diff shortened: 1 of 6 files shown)


pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -246,8 +246,14 @@ stderr: ${processResult.stderr}''');
...
   Future<Isolate> _spawnIsolateWithUri(Uri uri, SendPort message) async {
-    return await Isolate.spawnUri(uri, [], message,
-        packageConfig: await packageConfigUri, checked: true);
+    return await Isolate.spawnUri(
+      uri,
+      [],
+      message,
+      packageConfig: await packageConfigUri,
+      checked: true,
+      debugName: 'test_main',
```
natebosch, line 255: Where are these surfaced?

Is there any reason to _not_ put more detail in here? WDYT About

```suggestion
      debugName: 'test_suite:$uri',
```

liamappelbe: They're surfaced in the VM service API. They're designed to be human readable names and are handy for debugging. So yeah, it definitely makes sense to include the test URI.

natebosch (approved):

Issue #2498, opened 2025-05-15
Failed assertions when


bkonyi:
When attempting to do a pub package roll into the Flutter framework, we've started encountering failures due to assertions added in https://github.com/dart-lang/test/commit/935b8b04f72ed902b66fd869447cd05bdbd94e6a.

In particular, we're seeing this assertion fail in all three tests referencing `flutter/dev/automated_tests/flutter_test/trivial_widget_test.dart` in `flutter/packages/flutter_tools/test/integration.shard/test_test.dart`:

```
              package:test_api/src/backend/group.dart': Failed assertion: line 73 pos 12: '<optimized out>': is not true.
              dart:core-patch/errors_patch.dart 67:4                _AssertionError._doThrowNew
              dart:core-patch/errors_patch.dart 49:5                _AssertionError._throwNew
              package:test_api/src/backend/group.dart 73:12         new Group
              package:test_api/src/backend/group.dart 95:12         Group.filter
              package:test_core/src/runner/runner_suite.dart 66:26  RunnerSuite.filter
              package:test_core/src/runner.dart 283:34              Runner._loadSuites.<fn>.<fn>
              package:test_core/src/runner/load_suite.dart 185:26   LoadSuite.changeSuite.<fn>.<fn>
              dart:async/zone.dart 1525:13                          _rootRun
              dart:async/zone.dart 1422:19                          _CustomZone.run
              dart:async/zone.dart 1321:7                           _CustomZone.runGuarded
              package:test_core/src/runner/load_suite.dart 184:12   LoadSuite.changeSuite.<fn>
              dart:async/zone.dart 1538:47                          _rootRunUnary
              dart:async/zone.dart 1429:19                          _CustomZone.runUnary
              dart:async/future_impl.dart 951:45                    Future._propagateToListeners.handleValueCallback
              dart:async/future_impl.dart 980:13                    Future._propagateToListeners
              dart:async/future_impl.dart 714:7                     Future._complete
              dart:async/future_impl.dart 116:12                    _SyncCompleter.complete
              package:test_core/src/runner/load_suite.dart 108:19   new LoadSuite.<fn>.<fn>
```

Full logs are [here](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8714790960587909953/+/u/run_test.dart_for_tool_integration_tests_shard_and_subshard_8_9/stdout).

I'm going to pin the dependency on `package:test_api` to `0.7.4` prevent the latest change from being pulled in.

[...]

PR #2499, opened 2025-05-15
Fix an issue with failed assertions using setUpAll/tearDownAll

DanTup:
When groups are filtered using `filter()` and not `onPlatform`, we would clone the group and tests without also cloning `setUpAll`/`tearDownAll`, which led to an assertion failure when trying to set the parents.

This change ensures we clone tests in `filter()` like we do in `forPlatform()`.

Fixes https://github.com/dart-lang/test/issues/2498

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 1.26.1-wip
 
 * Set a debug name for test isolates.
+* Use version `0.7.6` of `packge:test_api` to fix an assertion failure when
+  using `setUpAll` or `tearDownAll` and running with asserts enabled.
 
 ## 1.26.0
 
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -1,8 +1,8 @@
-## 1.26.1-wip
+## 1.26.1
 
 * Set a debug name for test isolates.
-* Use version `0.7.6` of `packge:test_api` to fix an assertion failure when
-  using `setUpAll` or `tearDownAll` and running with asserts enabled.
+* Fix an assertion failure when using `setUpAll` or `tearDownAll` and running
+  with asserts enabled.
 
 ## 1.26.0
 
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test
-version: 1.26.1-wip
+version: 1.26.1
 description: >-
   A full featured library for writing and running Dart tests across platforms.
 repository: https://github.com/dart-lang/test/tree/master/pkgs/test
@@ -37,7 +37,7 @@ dependencies:
 
   # Use an exact version until the test_api and test_core package are stable.
   test_api: 0.7.6
-  test_core: 0.6.10-wip
+  test_core: 0.6.10
 
   typed_data: ^1.3.0
   web_socket_channel: '>=2.0.0 <4.0.0'
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -36,7 +36,7 @@ dependencies:
   stream_channel: ^2.1.0
 
   # Use an exact version until the test_api and test_core package are stable.
-  test_api: 0.7.5
+  test_api: 0.7.6
   test_core: 0.6.10-wip
 
   typed_data: ^1.3.0
--- a/pkgs/test/test/runner/set_up_all_test.dart
+++ b/pkgs/test/test/runner/set_up_all_test.dart
@@ -95,4 +95,19 @@ void main() {
     expect(test.stdout, neverEmits(contains('(setUpAll)')));
     await test.shouldExit(0);
   });
+
+  test('does not fail with asserts enabled - issue #2498', () async {
+    await d.file('test.dart', r'''
+        import 'package:test/test.dart';
+
+        void main() {
+          setUpAll(() {});
+
+          test("test", () {});
+        }
+        ''').create();
+
+    var test = await runTest(['test.dart'], vmArgs: ['--enable-asserts']);
+    await test.shouldExit(0);
+  });
 }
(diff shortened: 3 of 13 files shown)


pkgs/test/pubspec.yaml
```diff
@@ -36,7 +36,7 @@ dependencies:
   stream_channel: ^2.1.0
 
   # Use an exact version until the test_api and test_core package are stable.
-  test_api: 0.7.5
+  test_api: 0.7.6
```
DanTup, line 39: @jakemac53 not sure what you want to do about versions here - these changes could be a `+1`, but it might be better to prepare them on a branch from the last release if you don't want to release any of the other changes pushed to `main` since the release.

pkgs/test/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## 1.26.1-wip
 
 * Set a debug name for test isolates.
+* Use version `0.7.6` of `packge:test_api` to fix an assertion failure when
```
jakemac53, line 4: nit: just copy the changelog from test_api directly - we pretend as if these are all just one package.

pkgs/test_core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## 0.6.10-wip
 
 * Set a debug name for test isolates.
+* Use version `0.7.6` of `packge:test_api` to fix an assertion failure when
```
jakemac53, line 4: same

pkgs/test/test/runner/set_up_all_test.dart
```diff
@@ -95,4 +95,19 @@ void main() {
     expect(test.stdout, neverEmits(contains('(setUpAll)')));
     await test.shouldExit(0);
   });
+
+  test('does not fail with asserts enabled - issue #2498', () async {
```
DanTup, line 99: I'm not sure why no existing tests triggered this, but did trigger when we didn't have the fix in `onPlatform`, but I added specific tests for these referencing the issue.

jakemac53 (commented):
Go ahead and update the other package versions for a release as well.

pkgs/test_api/lib/src/backend/group.dart
```diff
@@ -96,8 +96,9 @@ class Group implements GroupEntry {
         metadata: metadata,
         trace: trace,
         location: location,
-        setUpAll: setUpAll,
-        tearDownAll: tearDownAll);
+        // Always clone these because they are being re-parented.
+        setUpAll: setUpAll?.filter((_) => true),
+        tearDownAll: tearDownAll?.filter((_) => true));
```
DanTup, line 101: This makes me wonder if we should just have a `clone()` as well as `.filter()`... I guess the code above for `onPlatform` could also just pass `(_) => true` to be consistent with this, but since the `setUpAll`/`tearDownAll` functions always have the same `platform` in the metadata as the enclosing group, it should always be the same.

pkgs/test_api/lib/src/backend/group.dart
```diff
@@ -96,8 +96,9 @@ class Group implements GroupEntry {
         metadata: metadata,
         trace: trace,
         location: location,
-        setUpAll: setUpAll,
-        tearDownAll: tearDownAll);
+        // Always clone these because they are being re-parented.
+        setUpAll: setUpAll?.filter((_) => true),
```
jakemac53, line 100: Should we expose more directly a copy function?

DanTup (commented):
@jakemac53

pkgs/test/pubspec.yaml
jakemac53: This is fine, we can release with the other changes as well.

pkgs/test_api/lib/src/backend/group.dart
jakemac53: lol I left the same comment, it does feel wrong to do this just to clone

DanTup: Are there any implications of adding a new `clone()` to the `Test` class? (might anyone outside of here be `implement`ing it for example?)

DanTup:
@jakemac53 

> Go ahead and update the other package versions for a release as well.

Did you mean remove the `-wip`s? If so, done - if not, let me know what :)

pkgs/test_api/lib/src/backend/group.dart
DanTup: @jakemac53 I added a `clone()` in aef5437bd2754ee3eb2aed9af588ef7424d14b59 and updated the existing test `filter()`s to call it that were only there to clone anyway.

It does mean there are two almost identical implementations for `Group.filter()` and `Group.clone()` though (I did consider making `clone()` internally just call `filter((_) => true)` but I don't know how you'd feel about that!).

DanTup:
@jakemac53 I don't understand the Publish bot failure - the only messages I can see are about widening the version constraints, but all I did was remove `-wip` so I'm not sure what the problem is.

jakemac53:
> @jakemac53 I don't understand the Publish bot failure - the only messages I can see are about widening the version constraints, but all I did was remove `-wip` so I'm not sure what the problem is.

This is expected, I will add the label to skip the check. Pub always warns about pinning deps - actually in this case though it should warn if we _weren't_ pinning deps since we export those packages. But, it doesn't understand that.

pkgs/test_api/lib/src/backend/group.dart
jakemac53: > Are there any implications of adding a new clone() to the Test class? (might anyone outside of here be implementing it for example?)

It isn't part of the public API. If flutter is doing this, we can update it when the package rolls in (although it's possible this will make the google3 roll complicated).

DanTup:
Got it, thanks!

GitHub says there are workflows needing approval, although I', not sure what since everything seems to have a green tick. I think everything is done here, but let me know if you think there's anything outstanding I need to do.

jakemac53 (approved):

PR #2501, opened 2025-05-21
[native assets] Graduate to preview

dcharkes:
The experiment flag will be on by default so it is not passed as argument anymore.

Accompanying Dart CL: https://dart-review.googlesource.com/c/sdk/+/429920


dcharkes:
FYI @mosuem It looks like this publish lint violation is intentional.

goderbauer (approved):
LGTM

Issue #2463, opened 2025-03-01
Timeout for `LoadSuite` is not configurable


nilsreichardt:
https://github.com/dart-lang/test/blob/db8cf091506a67eba1e523215e0e49a0db7cd1fd/pkgs/test_core/lib/src/runner/load_suite.dart#L23-L29

The timeout for loading the test suit is fixed. I assume this the root cause for the issue that we have with running `flutter test integration_test` with heavy apps: https://github.com/flutter/flutter/issues/105913:

```
12:00 +0 -1: loading /Users/runner/work/sharezone-app/sharezone-app/app/integration_test/app_test.dart [E]                                                                                             
  TimeoutException after 0:12:00.000000: Test timed out after 12 minutes.
  package:test_api/src/backend/invoker.dart 333:28  Invoker._handleError.<fn>
```

As the documentation of the variables says, determines this variable how long the test command tries to compile the application. Therefore, I assume that this is the issue.

People pass the `--timeout` flag (like `--timeout none`) to the `flutter test` command but are confused why their integration test timeout. To proper fix this issue, I think we need to make this timeout configure.

Another workaround would be to add the `--ignoreTimeout` flag of the Dart test command also the Flutter test command and pass it.

bkonyi:
@natebosch it would be nice if there was a way to bypass or change this timeout so users wouldn't need to disable all timeouts to work around it.

PR #2505, opened 2025-05-24
[Feedback needed] Add `--suite-load-timeout` option

nilsreichardt:
This PR adds a `--suite-load-timeout` argument to remove the hard-coded timeout (see #2463). [More description is added later]

---

<details>

</details>

### DIFF
--- a/pkgs/test_core/lib/src/runner/suite.dart
+++ b/pkgs/test_core/lib/src/runner/suite.dart
@@ -57,7 +57,8 @@ final class SuiteConfiguration {
       tags: null,
       onPlatform: null,
       metadata: null,
-      ignoreTimeouts: null);
+      ignoreTimeouts: null,
+      suiteLoadTimeout: null);
 
   /// Whether or not duplicate test (or group) names are allowed within the same
   /// test suite.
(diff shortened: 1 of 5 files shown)


nilsreichardt:
@natebosch I need a bit of feedback here. Would be a `--suite-load-timeout` option a good solution for this? Should we use a different name?

If you are not the right person for this, could you tag someone else?

After this, I will add tests for this PR 👍

pkgs/test_core/lib/src/runner/suite.dart
```diff
@@ -149,6 +150,11 @@ final class SuiteConfiguration {
   final bool? _ignoreTimeouts;
   bool get ignoreTimeouts => _ignoreTimeouts ?? false;
 
+  /// The timeout for loading a test suite.
+  final Timeout? _suiteLoadTimeout;
+  Timeout get suiteLoadTimeout =>
+      _suiteLoadTimeout ?? const Timeout(Duration(minutes: 12));
```
nilsreichardt, line 156: I implemented `--suite-load-timeout` the same way as `--ignore-timeouts` was implemented. However, it's not optimal that we have two locations for the default value. Here and in `pkgs/test_core/lib/src/runner/configuration/args.dart`. Should we improve this? Or not? Or just writing a comment to update the default value also in the other file?

nilsreichardt:
I'm marking this as ready for review in the hope of getting some feedback.

natebosch:
I think it would be fine to add a `--suite-load-timeout` option. I'm not sure if it needs to be in `SuiteConfiguration` or if it can live in `Configuration` alone. I'll also take a look at adding support in the yaml file.

natebosch (approved):

PR #2508, opened 2025-06-17
Add --json-help argument

natebosch:
Add a hidden `--json-help` argument which writes the usage information
in JSON Schema format.


natebosch:
This is using an unpublished feature in `package:args`.

https://github.com/dart-lang/core/pull/897

natebosch:
Opening this PR to have the code in a shared place. Not sure if this is going to be worth landing (or when) but this branch can be used to generate the schema.

natebosch:
The ecosystem has moved away from patterns where this was useful.

PR #2509, opened 2025-06-18
Add the platform level channel to zone values

natebosch:
We make a test channel available in the zones running tests to enable
features like spawning VM isolates but it's not available in the zone
that runs `main`. Add a new zone variable that holds the test channel.


natebosch:
I don't think we ended up needing this.

Issue #2511, opened 2025-06-26
Align `dart test --coverage` with `package:coverage`'s `test_with_coverage` and `flutter test --coverage`


liamappelbe:
I'd like to make `dart test --coverage` behave more like [`test_with_coverage`](https://github.com/dart-lang/tools/blob/main/pkgs/coverage/bin/test_with_coverage.dart) and `flutter test --coverage`. Context is https://github.com/dart-lang/sdk/issues/60958

### `test_with_coverage` behavior:
- starts `dart test` with `--pause-isolates-on-exit` (and optionally `--branch-coverage`)
- calls [`collect`](https://pub.dev/documentation/coverage/latest/coverage/collect.html)
  - `resume` and `waitPaused` enabled
  - `scopedOutput` set to the package's name (by default)
  - Optionally with `functionCoverage` and `branchCoverage` enabled
- calls [`formatLcov`](https://pub.dev/documentation/coverage/latest/coverage/FileHitMapsFormatter/formatLcov.html) to create an lcov file
  - also checks for ignored lines here
- outputs to `<package-dir>/coverage/` by default. The `lcov.info` file is placed in this directory

### `flutter test --coverage` behavior:
- broadly the same, except that `resume` and `waitPaused` are disabled in the [`collect` call](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/test/coverage_collector.dart#L312), and isolate lifecycles are managed by the test runner. Does multiple collect calls in a single run, so uses `coverableLineCache` to speed up collection
- No function coverage option, due to performance issues
- output is also in `<package-dir>/coverage/lcov.info` by default, but the flag (`--coverage-path`) specifies the output file, not directory

### `dart test --coverage` behavior:
- flow is most similar to `flutter test --coverage`
  - `resume` and `waitPaused` are disabled in the [`collect` call](https://github.com/dart-lang/test/blob/master/pkgs/test_core/lib/src/runner/vm/platform.dart#L349), so I assume that the test runner is managing the isolate lifecycles manually
  - passing `isolateIds`, which could miss coverage if the test is spawning isolates that aren't in the same group (eg using `spawnUri`)
    - Is this used to avoid double counting, because tests are run in separate isolate groups within the same process?
  - not setting `scopedOutput`, leaving filtering to a later step
  - not using `coverableLineCache` which could be slightly slower if it's doing multiple collect calls
- No function coverage or branch coverage options
[...]

PR #2517, opened 2025-07-09
WIP: Add `--coverage-lcov` and `--branch-cov` options

liamappelbe:
Fixes https://github.com/dart-lang/test/issues/2511

### DIFF
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -14,7 +14,7 @@ dependencies:
   async: ^2.5.0
   boolean_selector: ^2.1.0
   collection: ^1.15.0
-  coverage: ^1.0.1
+  coverage: ^1.15.0
   http_multi_server: ^3.0.0
   io: ^1.0.0
   js: '>=0.6.4 <0.8.0'
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -37,7 +37,7 @@ dependencies:
 
   # Use an exact version until the test_api and test_core package are stable.
   test_api: 0.7.6
-  test_core: 0.6.11
+  test_core: 0.6.12
 
   typed_data: ^1.3.0
   web_socket_channel: '>=2.0.0 <4.0.0'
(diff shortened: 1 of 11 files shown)


liamappelbe:
[One of the test failures](https://github.com/dart-lang/test/blob/master/pkgs/test/test/runner/coverage_test.dart#L56) is because I'm setting `collect`'s `scopedOutput` param to the name of the package under test. In this case that's the name in the [workspace's pubspec](https://github.com/dart-lang/test/blob/master/pubspec.yaml): `test_workspace`. So the scope is filtering out the fake `test.dart` file that's created [here](https://github.com/dart-lang/test/blob/master/pkgs/test/test/runner/coverage_test.dart#L35).

The way to fix this would be to construct a small fake package directory, instead of the single file. I'll probably need to add a source file and a pubspec. The `runTest` function runs the test using `d.sandbox` as the working dir, so I just have to create these files in there.

But if the test is being run in `d.sandbox`, why is the package name being set to `test_workspace`? I tried printing the working dir in there, and it's still set to the top level dir of the monorepo.

liamappelbe:
The remaining test failures don't seem related to this PR

liamappelbe:
This is ready for review. @jakemac53 @natebosch PTAL.

liamappelbe:
@jakemac53 @natebosch Friendly ping

natebosch (approved):
I think if we rename `--coverage-lcov` to `--coverage-path` as mentioned on the issue we'll be pretty well set up to normalize things in the future. It's a bit less obvious in the mean time, but I think it'll be a better place to end up long term.

natebosch:
This may need some attention before we can publish it. The dependency on `package:coverage` from `lib/src/runner/coverage.dart` causes a cycle in the targets split out for blaze.

liamappelbe:
> This may need some attention before we can publish it. The dependency on `package:coverage` from `lib/src/runner/coverage.dart` causes a cycle in the targets split out for blaze.

Can you give some more details? package:coverage doesn't depend on package:test afaik (aside from a dev dep)

DanTup:
Oh, never mind - this is because it depends on the version of `pkg:test` my project is using, not what was in the SDK. I'll test it using a git dependency, and wait for a new `pkg:test` release before shipping anything. Thanks!

PR #2518, opened 2025-07-18
Expand Analyzer constraints to allow 8.x

natebosch:
Prepare to publish with no other package changes.

### DIFF
--- a/pkgs/test_core/lib/src/runner/parse_metadata.dart
+++ b/pkgs/test_core/lib/src/runner/parse_metadata.dart
@@ -214,6 +214,8 @@ class _Parser {
       Object? skip;
       for (var expression in expressions) {
         if (expression is InstanceCreationExpression) {
+          // Replace name2 with name when bumping min analyzer dependency
+          // ignore: deprecated_member_use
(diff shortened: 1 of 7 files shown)


pkgs/test_core/lib/src/runner/parse_metadata.dart
```diff
@@ -348,6 +350,8 @@ class _Parser {
 
   String? _findConstructorNameFromInstantiation(
       InstanceCreationExpression constructor, String className) {
+    // Replace name2 with name when bumping min analyzer dependency
+    // ignore: deprecated_member_use
```
scheglov, line 354: Why not doing it now?
When we will start removing old APIs from the analyzer, this will cause issues.
And than we will have to do one more iteration to clean this up.

kevmoo (approved):

pkgs/test_core/lib/src/runner/parse_metadata.dart
natebosch: I think we should have at least 1 release which allows both analyzer versions so I'll publish first with the ignore and then fix it in a followup.

PR #2532, opened 2025-08-26
Serve dart2wasm source map files

osa1:
dart2wasm generates a source map file at the same location with the .wasm file. Serve it to help with debugging in the browser with `--pause-after-load`.

Note: somehow the source map file is not requested by the browser when testing with just `test -p chrome -c dart2wasm` (without `--pause-after-load` and manually adding breakpoints). I'm investigating why the source map isn't requested by default. When I run the same failing test with `-c dart2js` the source map is used.


sigurdm (approved):

osa1:
CI failure seems unrelated, #2529 also has the same failure.

Should I merge this?

sigurdm:
Yeah, seems to be failing at main as well

Probably fine to merge

Issue #2535, opened 2025-09-03
"dart test --version" doesn't work in workspace folders


DanTup:
Running `dart test --version` inside a project that is part of a Pub Workspace fails with "Couldn't find version number". It works fine in the root of the workspace:

```
PS D:\Dev\Temp Projects\workspace_bug_repro> dart test --version
Building package executable... (2.1s)
Built test:test.
1.26.2


PS D:\Dev\Temp Projects\workspace_bug_repro> cd .\packages\application\
PS D:\Dev\Temp Projects\workspace_bug_repro\packages\application> dart test --version       
Couldn't find version number.
```

Dart-Code uses this command to determine the `pkg:test` capabilities (to know if we can use `--ignore-timeouts`) and while we could locate the workspace root, this feels like a bug.

PR #2537, opened 2025-09-04
Get version from a hard coded value instead of the pubspec.lock

jakemac53:
Fixes https://github.com/dart-lang/test/issues/2535

- Adds a hard coded version value to package:test, and a test to ensure it stays in line with the pubspec.
- Add optional arguments to the test_core apis to accept a version, and use that for the `--version` option.

Open question: Does this work out of the box for flutter or would they separately have to pass a version?

### DIFF
--- a/pkgs/test/test/version_test.dart
+++ b/pkgs/test/test/version_test.dart
@@ -13,10 +13,8 @@ import 'package:test/test.dart';
 
 void main() {
   test('testVersion is up to date', () {
-    final pubspec = Pubspec.parse(
-      File('pkgs/test/pubspec.yaml').readAsStringSync(),
-      sourceUrl: Uri.file('pkgs/test/pubspec.yaml'),
-    );
+    final file = File('pubspec.yaml');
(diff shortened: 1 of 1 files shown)


jakemac53:
Ugh, this actually probably won't work - https://github.com/flutter/flutter/blob/8a5c6df06fe2773c6c25c5c2ac13111823b04189/packages/flutter_tools/lib/src/test/test_wrapper.dart#L31C11-L31C15.

package:test is not even a dependency here

natebosch:
> @natebosch idk if you have a better solution...

Reading it from the new files that pub writes makes the most sense to me, but I'm not sure if the pub team has any concerns about us taking a dependency on the new file formats.

https://github.com/dart-lang/test/pull/2538

natebosch:
> which would work because of the pinning but seems wrong to do still.

Yeah I don't think I want to add a new and very concrete blocker for moving to proper semantic versioning between the test runner packages. If the pub team doesn't want external dependencies on these files hardcoding is probably the best set of tradeoffs.

PR #2538, opened 2025-09-04
Support --version in pub workspace

### LINKED ISSUE #2535: "dart test --version" doesn't work in workspace folders

natebosch:
Closes #2535

Add a check for the `workspace_ref.json` file in place of the
`pubspec.lock` file that exists for single package pub solutions.

When the workspace ref exists, follow it and read the
`package_graph.json` file from the workspace root.

### DIFF
--- a/pkgs/test_core/lib/src/runner/version.dart
+++ b/pkgs/test_core/lib/src/runner/version.dart
@@ -2,60 +2,99 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert';
 import 'dart:io';
 
+import 'package:collection/collection.dart';
+import 'package:path/path.dart' as p;
 import 'package:yaml/yaml.dart';
 
 /// The version number of the test runner, or `null` if it couldn't be loaded.
 ///
 /// This is a semantic version, optionally followed by a space and additional
 /// data about its source.
-final String? testVersion =
-    (() {
-      dynamic lockfile;
-      try {
-        lockfile = loadYaml(File('pubspec.lock').readAsStringSync());
-      } on FormatException catch (_) {
-        return null;
-      } on IOException catch (_) {
-        return null;
-      }
+final String? testVersion = _readWorkspaceRef() ?? _readPubspecLock();
 
-      if (lockfile is! Map) return null;
-      var packages = lockfile['packages'];
-      if (packages is! Map) return null;
-      var package = packages['test'];
-      if (package is! Map) return null;
+String? _readWorkspaceRef() {
+  try {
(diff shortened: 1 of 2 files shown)


natebosch:
@jonasfj - do you have any concerns about the test running reading `workspace_ref.json` and `package_graph.json` files?

pkgs/test_core/lib/src/runner/version.dart
```diff
@@ -2,60 +2,99 @@
...
+    final workspaceRefFile = File(workspaceRefPath);
+    if (!workspaceRefFile.existsSync()) return null;
+    final ref = jsonDecode(workspaceRefFile.readAsStringSync());
+    if (ref is! Map) return null;
+    final relativeRoot = ref['workspaceRoot'];
+    if (relativeRoot is! String) return null;
+    final packageGraphPath = p.normalize(
+      p.join(
+        '.dart_tool',
+        'pub',
```
jakemac53, line 30: consider just `p.dirname(workspaceRefPath)` or saving this part as a separate variable above

```suggestion
        p.dirname(workspaceRefPath),
```

natebosch:
> If removing it doesn't break much on CI we could also drop that behavior entirely.

Checking the feasibility in #2539

natebosch:
@sigurdm do you have any concerns about this usage? Are `workspace_ref.json` and `package_graph.json` stable enough to rely on this way?

sigurdm:
> @jonasfj - do you have any concerns about the test running reading `workspace_ref.json` and `package_graph.json` files?

I don't think reading workspace_ref is any big issue. Flutter tools already relies on the file being there, so it is probably not going anywhere. But I think it would be easier to use https://api.dart.dev/dart-isolate/Isolate/packageConfigSync.html to find the package config, and then find package_graph.json next to that.

You can rely on the package_graph always being present, so if you are ok with just the version number (no information about the source), you should be able to leave out the code reading pubspec.lock

jakemac53 (approved):

DanTup:
Is there anything preventing this from landing?

FWIW I don't know how involved it is or what the process/schedule is, but it would be great to have a `pkg:test` release if that's possible... there have been changes to support coverage in Dart in the same format as Flutter but no release since that landed.

PR #2539, opened 2025-09-06
Remove extra version info

natebosch:
Follow up to https://github.com/dart-lang/test/pull/2538

This is only relevant when using the test runner as a git or path
dependency. In these cases the version will already end in `-WIP` with
our current development practices and in the places where it may come up
the users don't need help understanding where the code is coming from.


kevmoo (approved):
Nice cleanup!

PR #2543, opened 2025-09-18
Allow exceptions from operator ==

natebosch:
It's confusing to treat an object that throws from `operator ==`
identically to one that returns `false` normally. Intentional exceptions
in `operator ==` are very unlikely, so this change is not likely to
cause spurious failures.


natebosch:
Merging on red because it looks like there is a problem with a new version of chrome.

https://github.com/dart-lang/test/issues/2615

Issue #2542, opened 2025-09-17
Non-solo (skipped) tests spam on compact mode


FMorschel:
> I did just confirm that skipped tests always spam even in compact mode, we should just fix that imo.

_Originally posted by @jakemac53 in https://github.com/dart-lang/tools/issues/2107#issuecomment-3005741735_

So we don't lose this.

Small repro:

```dart
void main() {
  test('2 * 2 equals 4', () {
    expect(2 * 2, equals(4));
  });
  test('3 - 1 equals 2', () {
    expect(3 - 1, equals(2));
  }, solo: true);
}
```

Current output for compact:

```
00:00 +0: test\a_test.dart: 2 * 2 equals 4
  Skip: does not have "solo"
00:00 +1 ~1: All tests passed!
```

If we replace the `solo` above with `skip`:

```
00:00 +1 ~1: All tests passed!
```

PR #2546, opened 2025-09-26
Solo skipped tests shouldn't be reported by failures-only or compact reporters

FMorschel:
The issue I'm trying to fix is https://github.com/dart-lang/test/issues/2542.

I've added a new `soloSkip` getter to `Metadata`, which is set (obviously) only when the test is being skipped because it doesn't have `solo`.

I've then used that information to skip all `Message`s writes when inside the relevant reporters.

I've already copied and added some tests here.

I'd like some guidance on whether what I did was the correct implementation (first contribution here). And once I've resolved all requests from the team, I'll add the new entry to the changelog and bump the version as needed.

---

<details>

</details>

### DIFF
--- a/pkgs/test/test/runner/compact_reporter_test.dart
+++ b/pkgs/test/test/runner/compact_reporter_test.dart
@@ -467,6 +467,88 @@ void main() {
     });
   });
 
+  group('solo:', () {
+    test('displays non-solo skipped tests separately', () {
+      return _expectReport(
+        '''
+          test('skip 1', () {});
+          test('skip 2', () {});
+          test('solo 3', () {}, solo: true);''',
+        '''
+          +0: loading test.dart
+          +0: skip 1
+          +0 ~1: skip 1
+          +0 ~1: skip 2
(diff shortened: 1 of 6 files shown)


natebosch:
I'll start a conversation with a few folks to make a decision on what behavior we want to switch to and revisit this.

FMorschel:
> I don't think I agree with treating solo differently than other skipped tests. I think I can get on board with suppressing all skips in the compact and failures only reporters.

Well, I guess it makes no difference too. Now that you said it, I realized that skipped tests wouldn't have any message to send so this would work fine too.

I'll wait for your answer (after the discussions), to continue working on this. Thank you for taking a look at it already!

FMorschel:
I've updated this to remove the `soloSkip` since it made little sense. Any news on what the team wants to do @natebosch? 

Small note: in the related issue, I mentioned this idea came from @jakemac53 at https://github.com/dart-lang/tools/issues/2107#issuecomment-3005741735 to help out in debugging analyzer (and related) tests.

PR #2548, opened 2025-10-02
Add missing Compiler Selectors heading

elliotwutingfeng:
Add missing `Compiler Selectors` heading and update the `@TestOn` example to include a compiler identifier.

### DIFF
--- a/pkgs/test/README.md
+++ b/pkgs/test/README.md
@@ -371,6 +371,16 @@ only supports boolean operations. The following identifiers are defined:
 * `posix`: Whether the test is running on a POSIX operating system. This is
   equivalent to `!windows`.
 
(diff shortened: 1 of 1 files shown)


pkgs/test/README.md
```diff
@@ -371,6 +371,16 @@ only supports boolean operations. The following identifiers are defined:
 * `posix`: Whether the test is running on a POSIX operating system. This is
   equivalent to `!windows`.
 
+### Compiler Selectors
```
natebosch: Hmm, I'm not sure it makes sense to add a separate header for this section, but I think we can do some rephrasing above to make it more clear how general a "platform" is. I'll take a try at a rewrite.

PR #2555, opened 2025-10-30
Bump requirements to allow/require analyzer 9.0.0

srawlins:
The various changelogs already mention "the latest version of the analyzer package." I guess that remains as true now as it did when it was written.

### DIFF
--- a/pkgs/test_core/CHANGELOG.md
+++ b/pkgs/test_core/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Restrict to latest version of analyzer package.
 * Require Dart 3.7
 * Add `--coverage-path` and `--branch-coverage` options to `dart test`.
+* Allow `analyzer` major version 9.
 
 ## 0.6.12
 
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Serve dart2wasm source map files.
 * Fix Node.js dart2wasm tests on windows.
 * Doc edit: clarify in README that a platform selector includes the compiler.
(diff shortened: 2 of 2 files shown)


pkgs/test_core/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ environment:
   sdk: ^3.7.0
 
 dependencies:
-  analyzer: ^8.0.0
+  analyzer: ^9.0.0
```
natebosch:
```suggestion
  analyzer: '>=8.0.0 <10.0.0'
```

natebosch:
> Is it easy to bump the GitHub Actions to exclude Dart 3.7 and 3.8? It looks like a lot of references in `.github/workflows/dart.yml`.

Yeah we update that file with `mono_repo`. I can take over the PR.

pkgs/test_api/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
   channels to the test runner.
 * Restrict to latest version of analyzer package.
 * Require Dart 3.7
+* Allow `analyzer` major version 9.
```
srawlins: A bit awkward with the text on line 5.

pkgs/test_core/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Restrict to latest version of analyzer package.
 * Require Dart 3.7
 * Add `--coverage-path` and `--branch-coverage` options to `dart test`.
+* Allow `analyzer` major version 9.
```
srawlins: A bit awkward with the text on line 3.

PR #2561, opened 2025-11-19
Add .fromEnvironment flags to the global declarer

johnniwinther:
The global declarer is used by package:test_reflective_loader which has no means of configuring the test output. It is for instance used to run all tests in package:analyzer, and the current output prints a line for each test (of which there is currently >32000) and uses colors in the output. This makes it very hard to output to find failing tests, since the failures are hidden within all the succeeding tests and piping the output to a file for easier search still leaves the color in the output.

This change adds .fromEnvironment variables to the creation of the global declarer, such that it can be configured at the command line. This allow for enabling compact output which only displays the failing tests, and whether to use colors in the output.

### DIFF
--- a/pkgs/test_core/pubspec.yaml
+++ b/pkgs/test_core/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test_core
-version: 0.6.13
+version: 0.6.14
 description: A basic library for writing tests and running them on the VM.
 repository: https://github.com/dart-lang/test/tree/master/pkgs/test_core
 issue_tracker: https://github.com/dart-lang/test/issues?q=is%3Aissue+is%3Aopen+label%3Apackage%3Atest


pkgs/test_core/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.6.14
+* Support environment flags for configuring the global declarer.
```
natebosch: This reads like we support reading from `Platform.environment`.

I wonder, _should_ we be trying to do that instead? For the VM case it's probably little consequence either way, for AOT native we'd probably prefer `Platform.environment` support, and for web we'd prefer what's implemented.

Would there be value in supporting a `Platform.environment` fallback in places where we can import it? I'd be happy to write up the conditional import version if we think it's useful.

PR #2566, opened 2025-11-25
Add `--coverage-package` option

liamappelbe:
At the moment `dart test` only ever gathers coverage data for the current package. But sometimes part of out test coverage comes from other packages, especially in the case of monorepos.

To handle this case, `flutter test` has a `--coverage-package` option, which is a multi-option of `RegExp`s. When this option is present, we list all the transitive packages the current package depends on, and include them in the coverage report if they match *any* of the `RegExp`s.

This PR adds the exact same flag to `dart test`.

Fixes https://github.com/dart-lang/sdk/issues/60958

### DIFF
--- a/pkgs/test_core/lib/src/runner/configuration.dart
+++ b/pkgs/test_core/lib/src/runner/configuration.dart
@@ -76,8 +76,10 @@ class Configuration {
   final bool? _branchCoverage;
 
   /// A list of regular expressions to match against the package names during
-  /// coverage collection. Only packages that match at least one of the regexps
-  /// will be included in the coverage report.
+  /// coverage collection.
+  ///
+  /// Only packages that match at least one of the regexps will be included in
+  /// the coverage report.
   final List<RegExp>? coveragePackages;
 
   /// The path to the file from which to load more configuration information.
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -427,7 +427,7 @@ Future<Map<String, dynamic>> _gatherCoverage(
     false,
     false,
     false,
-    await _getCoveragePackages(config.coveragePackages),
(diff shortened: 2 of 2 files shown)


pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -427,7 +427,7 @@ Future<Map<String, dynamic>> _gatherCoverage(
     false,
     false,
     false,
-    {(await currentPackage).name},
+    await _getCoveragePackages(config.coveragePackages),
```
natebosch: How about `_coveragePackages(config.coveragePackages),`, or pick a more descriptive verb.

https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get

pkgs/test_core/lib/src/runner/configuration.dart
```diff
@@ -75,6 +75,11 @@ class Configuration {
   bool get branchCoverage => _branchCoverage ?? false;
   final bool? _branchCoverage;
 
+  /// A list of regular expressions to match against the package names during
+  /// coverage collection. Only packages that match at least one of the regexps
```
natebosch: [nit] add a blank line between header sentence and the details.

https://dart.dev/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

natebosch (approved):

natebosch:
I have a few more changes I'd like to get out that have been slow to land internally. I expect I should have a release ready sometime next week.

DanTup:
sgtm, thanks!

DanTup:
@natebosch no worries, thanks for the update! It's definitely not a priority so don't push to do a release for me - mostly I was just trying to decide whether to include the related Dart-Code change in the upcoming release or bump to next month (I don't want to ship it without having tested it end-to-end with an actual published version). I'll just bump it until Jan instead. Thanks!

PR #2569, opened 2025-12-02
Require a definition named `main` in test suites

### LINKED ISSUE #2567: Missing main under `_test.dart` file
I ran the tests in a project where I wrote the `_test.dart` file, but since it was a plugin, I should use `test_reflective_loader`. I wrote the class, but I forgot to add `@reflectiveTest` and the `main` function and tried running the tests from the terminal:

```
00:11 +31 -1: loading essential_lints/test\fixes\replace_with_squared_box_test.dart [E]
  Failed to load "essential_lints/test\fixes\replace_with_squared_box_test.dart":
  /C:/Users/FELIP_~1/AppData/Local/Temp/dart_test.kernel.2beb11dc/test.dart:14:42: Error: Undefined name 'main'.
        internalBootstrapVmTest(() => test.main, sendPort);
                                           ^^^^

To run this test again: C:\Users\felip_0vh5fa6\dart\bin\dart.exe test essential_lints/test\fixes\replace_with_squared_box_test.dart -p vm --plain-name "loading essential_lints/test\fixes\replace_with_squared_box_test.dart"
```

From inside VS Code:

```
Connecting to VM Service at ws://127.0.0.1:15018/jbQlmHQSBV4=/ws
Connected to the VM Service.
Failed to load "D:/dev/dart/essential_lints/essential_lints/test/fixes/replace_with_squared_box_test.dart":
/C:/Users/FELIP_~1/AppData/Local/Temp/dart_test.kernel.37e10478/test.dart:14:42: Error: Undefined name 'main'.
test.dart:14
      internalBootstrapVmTest(() => test.main, sendPort);
                                         ^^^^
package:test_core/src/runner/vm/platform.dart 290:7   VMPlatform._compileToKernel
platform.dart:290
package:test_core/src/runner/vm/platform.dart 260:11  VMPlatform._spawnIsolate
platform.dart:260
package:test_core/src/runner/vm/platform.dart 83:19   VMPlatform.load
platform.dart:83
package:test_core/src/runner/loader.dart 244:27       Loader.loadFile.<fn>
loader.dart:244
package:test_core/src/runner/load_suite.dart 104:21   new LoadSuite.<fn>.<fn>
load_suite.dart:104
✖ loading D:/dev/dart/essential_lints/essential_lints/test/fixes/replace_with_squared_box_test.dart

Exited (1).
```

Basically, this is _trying_ to say I have no `main`. But I feel like this should be a very specific message.

natebosch:
Closes #2567

The error when there is a compilation failure due to a missing `main` is
less direct than it could be and exposes details of wrapping the test
suite in another library. Add an explicit check for a function declared
with that name when parsing metadata before attempting any compiles.

Add a check to `parseMetadata` that throws a `FormatException` if there
is no function declared directly in the library named `main`. This will
prevent some previously working designs such as exposing a main in a
part or with `export`, but we do not expect those designs are used in
practice.

### DIFF
--- a/pkgs/test/test/runner/runner_test.dart
+++ b/pkgs/test/test/runner/runner_test.dart
@@ -193,8 +193,7 @@ $_usage''');
         test.stdout,
         containsInOrder([
           'Failed to load "test.dart":',
-          "test.dart:1:9: Error: Expected ';' after this.",
-          'invalid Dart file',
+          "Error: Expected ';' after this.",
         ]),
       );
 


natebosch:
We currently allow a `_test.dart` file to be independent of the precompiled output for that test. We parse the file for metadata as if it was the same content though. Precompiling and then changing the content isn't a supported workflow so I think it's OK to require that the `_test.dart` file stay well-formed with a `main` defined even after it's been precompiled.

PR #2574, opened 2025-12-05
[coverage] Support workspaces in coverage filters

liamappelbe:
Fixes https://github.com/dart-lang/tools/issues/2269

The bug is that when `dart test --coverage...` is run in a package that's part of a workspace, the default coverage filter is set to the name of the workspace, rather than the package under test. In `package:coverage` I solved this by searching for all the nested packages, and using those as the default filter. Another approach would be to restrict the default filter to only include the package under test, but I think users are more likely to want all the workspace's packages included in the report.

This change introduces a dependency on `package:pubspec_parse`. Hopefully that's ok.

### DIFF
--- a/pkgs/test_core/lib/src/util/package_config.dart
+++ b/pkgs/test_core/lib/src/util/package_config.dart
@@ -59,7 +59,7 @@ Set<String> _allWorkspaceNames(Uri packageRoot, Set<String> results) {
     final pubspec = Pubspec.parse(yaml, sourceUrl: pubspecUri);
     results.add(pubspec.name);
     for (final package in pubspec.workspace ?? const <String>[]) {
-      _getAllWorkspaceNames(packageRoot.resolve('$workspace/'), results);
+      _allWorkspaceNames(packageRoot.resolve('$package/'), results);
     }
   }
   return results;


pkgs/test_core/lib/src/util/package_config.dart
```diff
@@ -42,3 +43,24 @@ Future<Uri> absoluteUri(String path) async {
...
+    _getAllWorkspaceNames(package.root, <String>{});
+
+Set<String> _getAllWorkspaceNames(Uri packageRoot, Set<String> results) {
+  final pubspecUri = packageRoot.resolve('pubspec.yaml');
+  final pubspecFile = File(pubspecUri.toFilePath());
+  if (pubspecFile.existsSync()) {
+    final yaml = pubspecFile.readAsStringSync();
+    final pubspec = Pubspec.parse(yaml, sourceUrl: pubspecUri);
+    results.add(pubspec.name);
+    for (final workspace in pubspec.workspace ?? <String>[]) {
```
natebosch: [nit] According to the docs the `workspace` collection holds  "subpackages", not "workspaces"


```suggestion
    for (final package in pubspec.workspace ?? const <String>[]) {
```

natebosch (approved):

PR #2584, opened 2026-01-07
Allow analyzer major version 10.

scheglov:


scheglov:
I don't understand why "Publish / publish / validate (pull_request)" is not happy.

PR #2586, opened 2026-01-20
test: add library doc comments

kevmoo:


pkgs/test/lib/expect.dart
```diff
@@ -2,4 +2,10 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// The `expect` API for test assertions.
+///
+/// This is a re-export of `package:matcher/expect.dart`.
+/// Generally you should import `package:test/test.dart` instead.
```
natebosch: I think it's OK to say this for now. The main intention for this import had been to separate the usage of these APIs through `test/test` into `test/scaffolding` and `test/expect` where the latter is an easy-to-grep indicator of un-migrated tests. Multiple things have changed in the `dart fix` tool since then and we might not end up ever pushing towards import of this library.

pkgs/test/lib/fake.dart
```diff
@@ -6,5 +6,11 @@
 // so until Mockito stops implementing its own version of `Fake`, because there
 // is code in the wild that imports both test_api.dart and Mockito.
 
+/// The `Fake` class for faking dependencies.
+///
+/// This is a re-export of `package:test_api/fake.dart`. It is kept separate
+/// to avoid a breaking change for users who import `package:mockito` directly.
```
natebosch: I don't think this is worth documenting. The `test_api` library is not user facing so it's not worth mentioning in user facing docs. The history of it's landing in the test package isn't really interesting for users either. I'm not convinced we'd want to necessarily export `Fake` from` test/teat` outside of `Mockito` considerations, but I agree it's the case that would fit in more with the existing library design.

Issue #2581, opened 2025-12-30
Test Coverage is Empty when Files are not in lib/ directory


stefanhk31:
After upgrading a Dart Frog project to version 1.28.0 of `test`, I noticed that coverage was not getting outputted for files in my `routes` directory--running coverage would produce an empty `lcov.info` file. When I rolled back to version 1.25.8, the `lcov.info` once again had data.

Dart SDK version: 3.9.0

Here is [a repository with a minimal reproduction](https://github.com/stefanhk31/test_cov/tree/main). 

Steps:

1. Download repository
2. Run `dart test -j 4 --coverage=coverage --platform=vm && dart pub global run coverage:format_coverage --lcov --in=coverage --out=coverage/lcov.info --report-on=foo` and verify that `coverage/lcov.info` has data
3. In `pubspec.yaml`, change test from 1.25.8 to 1.28.0
4. Run `dart pub get` 
5. Run `dart test -j 4 --coverage=coverage --platform=vm && dart pub global run coverage:format_coverage --lcov --in=coverage --out=coverage/lcov.info --report-on=foo`, and see that `coverage/lcov.info` is now empty

mrverdant13:
I think this is no longer an issue in test 1.29.0

camsim99:
Thanks for looking into it @felangel. I'll see if I can escalate the triage of this issue!

felangel:
@camsim99 I looked into it a bit more and it looks like the line that causes the regression is https://github.com/dart-lang/test/blob/fe41d5408bfd7c73cc4309da6ab6f0ba46cbf417/pkgs/test_core/lib/src/runner/vm/platform.dart#L430

It seems like it was originally introduced in https://github.com/dart-lang/test/pull/2517/changes#r2806436414 by @liamappelbe.

Previously `collect` was called without scopedOutput and the newly introduced filter now excludes files that are outside of the `lib` directory which were previously being included.

The following one line change restores coverage reporting in the reproduction sample above:

```diff
Future<Map<String, dynamic>> _gatherCoverage(
  Environment environment,
  Configuration config,
) async {
  final isolateId =
      Uri.parse(
        environment.observatoryUrl!.fragment,
      ).queryParameters['isolateId'];
  return await collect(
    environment.observatoryUrl!,
    false,
    false,
    false,
-   await _filterCoveragePackages(config.coveragePackages),
+  {},
    isolateIds: {isolateId!},
    branchCoverage: config.branchCoverage,
  );
}
```

[...]

PR #2594, opened 2026-02-14
fix: report all coverage when no filters are specified

felangel:
- When no coverage filters are specified, coverage is reported for all packages to maintain backward compatibility.
- Fixes https://github.com/dart-lang/test/issues/2581
- Tagging @liamappelbe since this change was originally introduced in https://github.com/dart-lang/test/pull/2517

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -459,10 +459,10 @@ Future<Set<String>> _filterCoveragePackages(
   String? coverageLcov,
 ) async {
   if (coverageLcov == null && coveragePackages == null) {
-    // If no filters were provided and we're using json workflow, report
-    // coverage for all packages. This is required to maintain backward
-    // compatibility particularly in cases where coverage is required for files
-    // outside of the lib directory.
+    // If no filters were provided and the JSON workflow is used, report all
+    // coverage. This is required to maintain backward compatibility
+    // particularly in cases where coverage is required for files outside of the
+    // lib directory.
     // See https://github.com/dart-lang/test/issues/2581.
     return {};
   }


kevmoo (commented):
No tests?

felangel:
> No tests?

I didn’t see any tests in pkg:test_core but happy to write some. Let me know how you want me to proceed.

kevmoo:
it's a workspace / monorepo – guessing the tests would be in `test_api` or just `test`

felangel:
> it's a workspace / monorepo – guessing the tests would be in `test_api` or just `test`

Okay I’ll take a look tomorrow. Just heads up the PR that introduced this bug didn’t appear to include any tests.

https://github.com/dart-lang/test/pull/2517/changes#diff-72178683dabb293d9a0ae3f372f9b182381c95a0f3d5d91db4efa7b52ff067c4

felangel:
> If the goal is to change the default behaviour to include coverage for _all_ packages, including transitive deps, then this approach works. But for most users that's going to mean that the default behaviour includes a lot of coverage data that they don't care about. It also means package:coverage is going to waste time gathering all that coverage.
> 
> We do need to support the use case in #2581 somehow, but it's an obscure use case. So I think it would be reasonable to ask those users to change the flags they pass, if this makes the common use case simpler.
> 
> Before we make a UX decision like that though, I have a few questions:
> 
> 1. @felangel Am I right that this changes the default behaviour to include coverage data for all the transitive deps?
> 2. @stefanhk31, if you pass `--coverage-package=.*` to package:test, does that fix your bug?
> 3. What about `--coverage-package=` with no value?
> 
> I think if either of those flags work, then we don't need to change anything in package:test. I suspect they won't work though. I think `--coverage-package=.*` won't work because the files missing in #2581 are not part of a package path at all, and I'm guessing `--coverage-package=` won't work because package:args expects a value.
> 
> So I think the best option would be to make `--coverage-package=` or `--coverage-package=.*` have this behaviour of zero filters. I suppose another (more hacky) option would be to default to no filter if outputting JSON, and keep the current default filter if outputting LCOV.
> 
> Ofc, if @natebosch says maintaining backwards compatibility is more important, then this PR LGTM.

@liamappelbe the specific goal was to restore coverage reporting for Dart Frog which includes code in a top level routes directory (see the newly added test case). This used to be reported by default and I don’t think either of those command line options work unfortunately so coverage data that we were previously relying is no longer accessible :(

felangel:
> > maybe we can go with the other option I mentioned, where the default behaviour in LCOV mode remains the same (only reporting coverage on the package under test), but we revert the behaviour for JSON mode (back to not filtering anything).
> 
> This sounds like a good choice to me. We can figure out a way to migrate away from the old JSON behavior later and for now use the nicer behavior only in LCOV mode.

Sounds good I’ll update this PR shortly 👍

natebosch (approved):

kevmoo:
@felangel – good question for Monday morning 😉

Issue #2599, opened 2026-02-26
Test locations are missing for skipped tests


DanTup:
Given a simple test file where one test has `solo: true`:

```dart
import 'package:test/test.dart';

void main() {
  test('calculate1', () {
    expect(1, 1);
  });
  test('calculate2', () {
    expect(1, 1);
  }, solo: true);
}
```

Running `dart test -r json` results in the following:

```
{"test":{"id":3,"name":"calculate1","suiteID":0,"groupIDs":[2],"metadata":{"skip":true,"skipReason":"does not have \"solo\""},"line":null,"column":null,"url":null},"type":"testStart","time":537}
{"testID":3,"messageType":"skip","message":"Skip: does not have \"solo\"","type":"print","time":537}
{"testID":3,"result":"success","skipped":true,"hidden":false,"type":"testDone","time":538}
{"test":{"id":4,"name":"calculate2","suiteID":0,"groupIDs":[2],"metadata":{"skip":false,"skipReason":null},"line":8,"column":3,"url":"file:///D:/Dev/Test%20Projects/dartcode_5931/test/dartcode_5931_test.dart"},"type":"testStart","time":539}
{"testID":4,"result":"success","skipped":false,"hidden":false,"type":"testDone","time":546}
```

The test that was executed has a location, but the test that was skipped does not. I don't think there's a reason we shouldn't have this, and think this is just a bug.

Edit: Interestingly, seems like this only affects `solo` and not `skip`:

```
{"group":{"id":2,"suiteID":0,"parentID":null,"name":"","metadata":{"skip":false,"skipReason":null},"testCount":2,"line":null,"column":null,"url":null},"type":"group","time":261}
{"test":{"id":3,"name":"calculate1","suiteID":0,"groupIDs":[2],"metadata":{"skip":false,"skipReason":null},"line":4,"column":3,"url":"file:///D:/Dev/Test%20Projects/dartcode_5931/test/dartcode_5931_test.dart"},"type":"testStart","time":261}
{"testID":3,"result":"success","skipped":false,"hidden":false,"type":"testDone","time":270}
{"test":{"id":4,"name":"calculate2","suiteID":0,"groupIDs":[2],"metadata":{"skip":true,"skipReason":null},"line":7,"column":3,"url":"file:///D:/Dev/Test%20Projects/dartcode_5931/test/dartcode_5931_test.dart"},"type":"testStart","time":271}
[...]

PR #2600, opened 2026-02-26
Fix missing test locations on solo-skipped tests by passing location/trace through

DanTup:
Fixes https://github.com/dart-lang/test/issues/2599 by ensuring stack traces/locations are carried across when tests are replaced with skips because of `solo`.

Since `test_api` 0.7.9 has already been published, I bumped this to `0.7.10`, which also meant updating in `test` and `test_core` for things to resolve correctly. I'm not sure if I did this correctly, so if I need to change anything, please let me know!

### DIFF
--- a/pkgs/test_api/CHANGELOG.md
+++ b/pkgs/test_api/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.10
+## 0.7.10-wip
 
 * Fix missing locations on tests skipped due to the use of `solo`.
 


pkgs/test_api/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: test_api
-version: 0.7.9
+version: 0.7.10
```
jakemac53: These all get published at the same time, so they should all be stable release numbers or `-wip` numbers. Lets start with `0.7.10-wip` for test_api, so we can confirm this works internally. Then we will publish everything.

PR #2613, opened 2026-03-23
Respect common environment variables for color output detection

kevmoo:
* Respect NO_COLOR, CLICOLOR, FORCE_COLOR, CLICOLOR_FORCE, and TERM=dumb.

### DIFF
--- a/pkgs/test_core/lib/src/util/io.dart
+++ b/pkgs/test_core/lib/src/util/io.dart
@@ -105,18 +105,19 @@ final _tempDir =
 ///
 /// Otherwise only printable ASCII characters should be used.
 bool get canUseSpecialChars {
-  // Respect common environment variables for disabling color output.
-  // See https://no-color.org/
-  if (Platform.environment.containsKey('NO_COLOR')) return false;
-  if (Platform.environment['TERM'] == 'dumb') return false;
+  switch (Platform.environment) {
+    // Respect common environment variables for disabling color output.
+    // See https://no-color.org/
+    case {'NO_COLOR': _}:
+    case {'TERM': 'dumb'}:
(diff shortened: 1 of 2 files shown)


pkgs/test_core/lib/src/util/io.dart
natebosch: [optional] I like how this reads when it's fully a switch expression, but the `_ when` pattern for a few cases that don't depend on `Platform.environment` is not an established idiom so I may be the only one who likes it.

```suggestion
bool get canUseSpecialChars => switch (Platform.environment) {
  // Respect common environment variables for disabling color output.
  // See https://no-color.org/
  {'NO_COLOR': _} ||
  {'TERM': 'dumb'} ||
  // Respect CLICOLOR=0. See https://bixense.com/clicolors/
  {'CLICOLOR': '0'} => false,

  // Respect FORCE_COLOR and CLICOLOR_FORCE.
  {'FORCE_COLOR': _} || {'CLICOLOR_FORCE': _} => true,
  _ when inTestTests => false,
  // On Linux and Mac, `supportsAnsiEscapes` always returns `false` on most
  // modern terminals, see https://github.com/dart-lang/sdk/issues/31606 for
  // details.
  //
  // Instead of relying on `supportsAnsiEscapes`, we assume that all modern
  // terminals support colors and check that `stdout` is a tty.
  _ when Platform.isWindows => stdout.supportsAnsiEscapes,
  _ => stdioType(stdout) == StdioType.terminal ,
};
```

natebosch (approved):

pkgs/test_core/lib/src/util/io.dart
kevmoo: OOO! Then we get a lambda? Hrm... I kinda like it.

Issue #2615, opened 2026-03-24
wasm tests are failing in chrome windows


natebosch:
```
[9944:6116:0324/230749.650:INFO:CONSOLE:47] "Uncaught (in promise) CompileError: WebAssembly.compileStreaming(): Imported builtin function "wasm:js-string" "fromCharCodeArray" has incorrect signature", source: http://localhost:63725/YuhhO6QRNi4=/test/hello_world_test.dart.browser_test.dart.js (47)
```

The relevant bit is `Imported builtin function "wasm:js-string" "fromCharCodeArray" has incorrect signature"` - @osa1 does this look like a problem in the SDK?

PR #2618, opened 2026-03-30
Fix hang on windows with multiple exe tests

### LINKED ISSUE #2617: `-c exe` tests hang on exit on windows
On windows, when running with the `exe` compiler, tests fail to exit properly.

After the last test, you will not see the "X tests passed" message.

If you hit Ctrl+c, then you see the message right away, and even saying "N" to the "terminate batch job" prompt will allow it to exit. Not sure what exactly that indicates.

jakemac53:
Fixes https://github.com/dart-lang/test/issues/2617

Not sure exactly what is happening here, but re-ordering the `socket.destroy` and `serverSocket.close` calls appears to resolve the issue.

Also enabled `-c exe` tests for test_api to help get better coverage of these kinds of tests.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -715,7 +715,7 @@ jobs:
       - job_005
       - job_006
   job_018:
-    name: "unit_test; linux; Dart 3.7.0; PKG: pkgs/test_api; `dart test --preset travis -x browser`"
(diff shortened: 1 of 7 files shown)


jakemac53:
The test failures here look to be pre-existing fwiw

natebosch:
> The test failures here look to be pre-existing fwiw

https://github.com/dart-lang/test/issues/2615

natebosch (approved):

PR #2622, opened 2026-04-09
Allow analyzer ^13.0.0, with breaking changes.

scheglov:
This change refers to not yet published `analyzer 13.0.0`, because I cannot land corresponding [analyzer CL](https://dart-review.googlesource.com/c/sdk/+/488624) (and then publish) without pulling a compatible version of `package:test` into the Dart SDK. So, I want to land this PR with a version of `test` that is compatible with `analyzer 13.0.0` (to be), and the pull this git commit into [the CL](https://dart-review.googlesource.com/c/sdk/+/488624).

I tested it locally with
```yaml
dependency_overrides:
  _fe_analyzer_shared:
    path: /Users/scheglov/Source/Dart/sdk.git/sdk/pkg/_fe_analyzer_shared
  analyzer:
    path: /Users/scheglov/Source/Dart/sdk.git/sdk/pkg/analyzer
```
Obviously, GitHub bots will be red until `analyzer 13.0.0` is published.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -3,7 +3,7 @@
 * Ignore an error locating the SDK directory on platforms where the
   `resolvedExecutable` is unexpectedly `null`.
 * Fix a bug where `-c exe` tests would hang on exit on windows.
-* Require `analyzer: ^13.0.0`
+* Allow `analyzer` version  `13.x.x`.
 
 ## 1.31.0
 
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Ignore an error locating the SDK directory on platforms where the
   `resolvedExecutable` is unexpectedly `null`.
 * Fix a bug where `-c exe` tests would hang on exit on windows.
+* Require `analyzer: ^13.0.0`
 
 ## 1.31.0
 
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -35,7 +35,7 @@ dependencies:
   stream_channel: ^2.1.0
 
   # Use an exact version until the test_api and test_core package are stable.
-  test_api: 0.7.11
+  test_api: 0.7.12-wip
   test_core: 0.6.18-wip
 
   typed_data: ^1.3.0
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -10,7 +10,7 @@ environment:
   sdk: ^3.7.0
 
 dependencies:
-  analyzer: ^13.0.0
+  analyzer: '>=8.0.0 <14.0.0'
   async: ^2.5.0
   boolean_selector: ^2.1.0
   collection: ^1.15.0
--- a/pkgs/test/pubspec.yaml
+++ b/pkgs/test/pubspec.yaml
@@ -10,7 +10,7 @@ environment:
   sdk: ^3.7.0
 
 dependencies:
-  analyzer: '>=8.0.0 <13.0.0'
+  analyzer: ^13.0.0
   async: ^2.5.0
   boolean_selector: ^2.1.0
   collection: ^1.15.0
--- a/pkgs/test_api/CHANGELOG.md
+++ b/pkgs/test_api/CHANGELOG.md
@@ -1,10 +1,13 @@
+## 0.7.12-wip
+
+* Allow `analyzer` major version 13.
+
 ## 0.7.11
 
 * Add `vmAsan`, `vmMsan` and `vmTsan` runtimes.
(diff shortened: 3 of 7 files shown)


pkgs/test/pubspec.yaml
```diff
@@ -10,7 +10,7 @@ environment:
   sdk: ^3.7.0
 
 dependencies:
-  analyzer: '>=8.0.0 <13.0.0'
+  analyzer: ^13.0.0
```
jakemac53: Given we are past the release cutoff, this effectively would mean that no flutter users could get new versions of package:test and friends until the next stable release after this one, which is over 3 months away.

If we have to make this as a breaking change that might be OK... but we should consider it as a data point.

jakemac53:
> Obviously, GitHub bots will be red until analyzer 13.0.0 is published.

Can we instead use a git dependency for the time being? This will slow down pub get and subsequently CI jobs, but should give better confidence. Otherwise we can't validate any future changes until analyzer is published which puts this package in a bad state.

scheglov:
> > Obviously, GitHub bots will be red until analyzer 13.0.0 is published.
> 
> Can we instead use a git dependency for the time being? This will slow down pub get and subsequently CI jobs, but should give better confidence. Otherwise we can't validate any future changes until analyzer is published which puts this package in a bad state.

Sounds interesting.
Can we specify a PR in DEPS file?
Or we still need to land the PR? And then I don't understand the difference.

pkgs/test_core/lib/src/runner/parse_metadata.dart
```diff
@@ -144,7 +144,9 @@ class _Parser {
   ///
   /// [annotation] is the annotation.
   PlatformSelector _parseTestOn(Annotation annotation) =>
-      _parsePlatformSelector(annotation.arguments!.arguments.first);
+      _parsePlatformSelector(
+        annotation.arguments!.arguments.first.argumentExpression,
```
jakemac53: Could we potentially add `argumentExpression` as an extension getter instead (it would be an extension on the old type of `arguments.first`), and it would just return `this`?

That may be able to allow us to support both versions at least for a short time.

scheglov: If we say `Argument` in `NodeList<Argument> _parseArguments(Expression expression)`, this probably will not be enough?

scheglov:
> until analyzer is published

My intention is to get approvals for both PRs for `dart_style` and `test`, then on the CL, then land PRs and CLs in close succession. There might be a gap in `dart_style` until `test` is published, I have no control here. But might be OK, `dart_style` is not moving quickly.

pkgs/test_core/lib/src/runner/parse_metadata.dart
scheglov: Actually, you gave me an idea. We can encapsulate argument into an extension type, and use dynamic to provide information for both before and after breaking change.

scheglov:
OK, I updated the PR, it is compatible with both analyzer 12, and with analyzer 13 (when using dependency overrides).

kevmoo:
I'm wondering if Chrome changed so the wasm generated by Dart 3.7 is now invalid. 🤔

jakemac53:
> Sounds interesting.
> Can we specify a PR in DEPS file?
> Or we still need to land the PR? And then I don't understand the difference.

We can depend on a git branch of the SDK if necessary for a period of time (in this repo), then the SDK can pick up the latest from the repo and land your CL there, then you can publish, and then we remove the dependency override.

jakemac53:
> I'm wondering if Chrome changed so the wasm generated by Dart 3.7 is now invalid. 🤔

Yeah we have seen this in several repos

pkgs/test_api/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 * Change return type on the `body` callback argument to `group` to `void` from
   `dynamic`. This may surface cases where the group callback was erroneously
   returning an ignored value.
-* Require `analyzer: '>=8.0.0 <13.0.0'`
+* Allow `analyzer` major version 13.
```
jakemac53: 0.7.11 is published so this should go under a new section `0.7.12-wip` and the pubspec version should also be changed.

Then test_core and test should update to pin 0.7.12 also.

pkgs/test/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Ignore an error locating the SDK directory on platforms where the
   `resolvedExecutable` is unexpectedly `null`.
 * Fix a bug where `-c exe` tests would hang on exit on windows.
+* Require `analyzer: ^13.0.0`
```
jakemac53:
```suggestion
* Allow `analyzer` version  `13.x.x`.
```

jakemac53 (approved):
LGTM with some small changelog/version comments

PR #2624, opened 2026-04-10
Bump to language version 3.10

### LINKED ISSUE #2623: Proposal/Discussion: Bump min SDK version to something >3.7
Wasm tests have issues on 3.7 now, and a hotfix for that is probably unlikely.

We have two potential resolutions to get the tests fixed:

- Skip wasm tests on 3.7
- Bump the min SDK to something higher that works

I propose the latter, 3.7 is quite old now, we could bump probably even as high as 3.10 which would also give access to dot shorthands, but even 3.8 would give us access to null aware elements which would be nice.

natebosch:
Closes #2623

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -530,6 +530,191 @@ jobs:
       - job_005
       - job_006
   job_013:
+    name: "unit_test; linux; Dart 3.10.0; PKG: pkgs/test; `xvfb-run -s \"-screen 0 1024x768x24\" dart test --preset travis --total-shards 5 --shard-index 0`"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Cache Pub hosted dependencies
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306
+        with:
+          path: "~/.pub-cache/hosted"
+          key: "os:ubuntu-latest;pub-cache-hosted;sdk:3.10.0;packages:pkgs/test;commands:command_01"
+          restore-keys: |
+            os:ubuntu-latest;pub-cache-hosted;sdk:3.10.0;packages:pkgs/test
+            os:ubuntu-latest;pub-cache-hosted;sdk:3.10.0
+            os:ubuntu-latest;pub-cache-hosted
+            os:ubuntu-latest
+      - name: Setup Dart SDK
+        uses: dart-lang/setup-dart@e51d8e571e22473a2ddebf0ef8a2123f0ab2c02c
+        with:
+          sdk: "3.10.0"
+      - id: checkout
+        name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      - id: pkgs_test_pub_upgrade
+        name: pkgs/test; dart pub upgrade
+        run: dart pub upgrade
+        if: "always() && steps.checkout.conclusion == 'success'"
+        working-directory: pkgs/test
+      - name: "pkgs/test; xvfb-run -s \"-screen 0 1024x768x24\" dart test --preset travis --total-shards 5 --shard-index 0"
+        run: "xvfb-run -s \"-screen 0 1024x768x24\" dart test --preset travis --total-shards 5 --shard-index 0"
+        if: "always() && steps.pkgs_test_pub_upgrade.conclusion == 'success'"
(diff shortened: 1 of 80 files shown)


jakemac53 (approved):
LGTM although usually we do changelog entries for these, and test_api should also go to `-wip`.

natebosch:
> usually we do changelog entries for these

Some folks always do - I usually do when I'm asked, but I've never really understood the motivation. The vast majority of CHANGELOG readers don't care when we change SDK bounds. The folks who do want to know which versions are compatible with their SDK can glean that much easer from https://pub.dev/packages/test/versions than from skimming a CHANGELOG.md

I think I pointed this out to @kevmoo recently and IIRC we're OK with skipping CHANGELOG entries for changes that only impact `pubspec.yaml`. If we agree I can update https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md with a formal style choice.

natebosch:
```
  Expected: {
              'type': 'start',
              'protocolVersion': '0.1.1',
              'runnerVersion': null,
              'pid': 5992
            }
    Actual: {
              'protocolVersion': '0.1.1',
              'runnerVersion': null,
              'pid': 6988,
              'type': 'start'
            }
     Which: at location ['pid'] is <6988> instead of <5992>
```

I don't think I've seen this before. I'm not sure if it's flakey.

jakemac53:
> I don't think I've seen this before. I'm not sure if it's flakey.

I have never seen that either...

jakemac53:
> I think I pointed this out to @kevmoo recently and IIRC we're OK with skipping CHANGELOG entries for changes that only impact `pubspec.yaml`. If we agree I can update https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md with a formal style choice.

I don't feel super strongly, but I  usually do include them. My general rule is if there is a user visible impact then include it, and changing constraints (of the SDK or packages) does have user visible impact.

kevmoo:
*SEE* @natebosch ! (I always liked you, @jakemac53 )

natebosch:
Who is reading those? Have _you_ ever read a changelog with the intent to see it's language version or dependencies? What question can you answer by reading that information in the changelog that isn't better answered somewhere else?

kevmoo:
@natebosch 🖐️

kevmoo:
@natebosch – do we have any ideas of what's failing here?

jakemac53:
> Who is reading those? Have _you_ ever read a changelog with the intent to see it's language version or dependencies? What question can you answer by reading that information in the changelog that isn't better answered somewhere else?

It is useful to know what version to roll back to if you need support for an older SDK or dependency. You can look at a single file and find which version enforced that newer version.

kevmoo:
```diff
diff --git a/pkgs/test/test/io.dart b/pkgs/test/test/io.dart
index ef7eb967..8507f1ab 100644
--- a/pkgs/test/test/io.dart
+++ b/pkgs/test/test/io.dart
@@ -27,7 +27,7 @@ final String sdkDir = p.dirname(p.dirname(Platform.resolvedExecutable));
 
 /// The platform-specific message emitted when a nonexistent file is loaded.
 final String noSuchFileMessage = Platform.isWindows
-    ? 'The system cannot find the file specified.'
+    ? 'The system cannot find the file specified'
     : 'No such file or directory';
 
 /// An operating system name that's different than the current operating system.
diff --git a/pkgs/test/test/runner/json_reporter_utils.dart b/pkgs/test/test/runner/json_reporter_utils.dart
index e064c027..82959cde 100644
--- a/pkgs/test/test/runner/json_reporter_utils.dart
+++ b/pkgs/test/test/runner/json_reporter_utils.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:convert';
+import 'dart:io';
 
 import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
@@ -42,7 +43,7 @@ Future<void> expectJsonReport(
     'protocolVersion': '0.1.1',
     // Runner version cannot be read in the synthetic package
     'runnerVersion': null,
-    'pid': testPid,
+    'pid': Platform.isWindows ? isA<int>() : testPid,
   };
   expect(decoded.first, equals(start));
 
```

So the pid on windows can be weird – maybe due to a wrapper
And the text output from the SDK dropped the period in the error message in 3.10 it seems.

PR #2626, opened 2026-04-13
Fix dart2wasm tests with custom HTML files

### LINKED ISSUE #2625: `dart test` with `dart2wasm` fails with `TypeError: Cannot read properties of null (reading 'dataset')` when using a custom HTML file
When running web tests with `dart test -p chrome -c dart2wasm` and providing a custom HTML file (e.g., `test/web_app_test.html` alongside `test/web_app_test.dart`), the test execution fails in the browser with a JavaScript error early in the load process. This causes the test runner to time out waiting for the test suite to load.

The same tests pass successfully when compiled with `dart2js` or when no custom HTML file is used (relying on the test runner's generated default HTML).

### Reproduction Steps
1. Create a Dart web test `test/example_test.dart`.
2. Create a custom HTML file `test/example_test.html` in the same directory with the required tags:
   ```html
   <!DOCTYPE html>
   <html>
   <head>
     <link rel="x-dart-test" href="example_test.dart">
   </head>
   <body>
     <script src="packages/test/dart.js"></script>
   </body>
   </html>
   ```
3. Run the test using the WebAssembly compiler:
   ```bash
   dart test test/example_test.dart -p chrome -c dart2wasm
   ```

### Expected Behavior
The test should compile to Wasm, load in Chrome, and execute successfully.

### Actual Behavior
The test runner hangs and eventually times out. The browser output (visible in verbose mode or when tests time out) shows:

```
[INFO:CONSOLE:10] "Uncaught (in promise) TypeError: Cannot read properties of null (reading 'dataset')", source: http://localhost:55822/.../example_test.dart.browser_test.dart.js (10)
Trying to load the allocator multiple times. This is *not* supported.
```

### Diagnostic Context
It appears that the JS bootstrap file generated for `dart2wasm` tries to read configuration or paths from the `dataset` property of a DOM element (likely the `<script>` tag that loads it). In `dart2js` mode, or when using the default generated HTML, this lookup succeeds. However, when using a custom HTML file, this element is either not found or is missing the expected `data-*` attributes.

***

Now that you have the issue description, how should we proceed to fix it? Since you have access to the source, should we try to locate where this `dataset` read happens in the `package:test` source or the SDK's Wasm web bootstrap templates?

kevmoo:
When running tests with `dart2wasm` and a custom HTML file, the JS bootstrap file `run_wasm_chrome.js` failed because it tried to read `dataset` from a DOM element with ID `WasmBootstrapInfo`, which is not present in custom HTML files.

This commit updates `run_wasm_chrome.js` to fallback to inferring the `.wasm` and `.mjs` URLs from `document.currentScript.src` if the `WasmBootstrapInfo` element is not found.

Added a regression test to verify the fix.

Fixes #2625


kevmoo:
Waiting on https://github.com/dart-lang/test/pull/2624 RE https://github.com/dart-lang/test/issues/2623

kevmoo:
@natebosch – this should be good to go now, too!

integration_tests/wasm/test/custom_html_wasm_test.dart
natebosch, line 1: [nit] move this to `integration_tests/wasm`

kevmoo:
@natebosch – PTAL

natebosch:
Looks like it's failing on firefox? Since this doesn't work at all today I'd be fine landing a fix for chrome initially and filing a bug for firefox if it's easier to skip the tests for now.

kevmoo:
this is just config delta since you had me move directories, I'm guessing

natebosch:
> this is just config delta since you had me move directories, I'm guessing

Right, it probably wasn't testing on firefox earlier and now that it is we either need to skip it or fix it.

kevmoo:
@natebosch – LGTY?

natebosch (approved):

PR #2630, opened 2026-04-13
Ignore test pid on windows

natebosch:
The tests are flaky or always failing due to inconsistencies with PID
behavior on windows. Change the PID expectation argument to take a
matcher and use `anything` on windows.


kevmoo (approved):
Nice!

Issue #2616, opened 2026-03-25
CoreChecks.has() usage could perhaps be improved


time4tea:
because 

```dart
Subject<R> has<R>(R Function(T) extract, String name) {}
```
has two arguments, you cannot easily maintain a library of functions to access things across a codebase:

how about (sorry for naming, i didn't have enough time to think of a better one):

```dart
class Extractor<T, R> {
  final R Function(T) extract;
  final String name;

  Extractor(this.extract, this.name);
}

 Subject<R> get<R>(Extractor<T, R> ex) {
    return context.nest(() => ['has $ex.name'], (value) {
      try {
        return Extracted.value(ex.extract(value));
      } catch (e, st) {
        return Extracted.rejection(which: [
          ...prefixFirst('threw while trying to read ${ex.name}: ', literal(e)),
          ...const LineSplitter().convert(st.toString())
        ]);
      }
    });
  }
```

then you can say:

```dart
Extractor<Uri, String> scheme() => Extractor((it) => it.scheme, "scheme");
```

```dart
    test("scheme", () {
      check(Uri.parse("ftp://example.com"))
          ..get(scheme()).equals("ftp");
    });
```

the functions to extract the things you need can be shared across a codebase.
i'm sure this could be improved further, just an idea.

time4tea:
or another example:
```dart
Extractor<Request, Header?> header(String name) => Extractor((it) => it.header(name), "header ${name}");
```

time4tea:
Actually, scratch this - however i didn't realise immediately from the docs that the idiomatic way to write reusable checks was to make extensions e.g.

```dart
extension UriChecks on Subject<Uri> {
  Subject<String?> param(String name) => has((u) => u.queryParameters[name], 'param ${name}');
  Subject<String> get host => has((u) => u.host, 'host');
  Subject<String> get path => has((u) => u.path, 'path');
  Subject<String> get scheme => has((u) => u.scheme, 'scheme');
  Subject<String> get value => has((u) => u.toString(), 'toString value');
}
```

the docs only mention extensions as ways to create composite or custom checks, maybe this would be a useful addition? thx!

jonasfj:
I really think that `package:checks` is awesome, but I'll grant that perhaps a few more illustrative examples could be useful.

Especially, in the API references where sometimes you forget how to use something.

Proposed: https://github.com/dart-lang/test/pull/2631

natebosch:
[...]

PR #2631, opened 2026-04-17
Example of how to use `has`

jonasfj:
Maybe better docs is useful, context: https://github.com/dart-lang/test/issues/2616


kevmoo:
I like!

natebosch (approved):

PR #2632, opened 2026-04-17
Add package:checks_codegen

natebosch:
Initial open sourcing of the code generation companion to
`package:checks`. This generates extension using the `.has()` utility
for each field in the classes specified for generation.

Add an annotation to mark imports of the generated libraries as a
trigger for the codegen as configuration for which types to generate
for. Add the builder using `package:source_gen` to target the
annotation. Add tests using `package:build_test`. Add an example with a
checked in generated file.


pkgs/checks_codegen/lib/src/annotation.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Annotation specifying types to generate Subject extensions with `has`
+/// getters for fields.
+///
+/// Annotate an import to the `.checks.dart` library.
+class CheckExtensions {
```
jonasfj:
```suggestion
final class CheckExtensions {
```

pkgs/checks_codegen/lib/src/builder.dart
```diff
@@ -0,0 +1,211 @@
...
+        );
+      }
+    } on UnresolvableAssetException {
+      // Asset may be in a summary.
+      return exportingLibrary.uri.toString();
+    }
+  }
+}
+
+class InvalidCheckExtensions extends Error {
```
jonasfj:
```suggestion
final class InvalidCheckExtensions extends Error {
```

pkgs/checks_codegen/README.md
```diff
@@ -0,0 +1,46 @@
+[![pub package](https://img.shields.io/pub/v/checks_codegen.svg)](https://pub.dev/packages/checks_codegen)
+[![package publisher](https://img.shields.io/pub/publisher/checks_codegen.svg)](https://pub.dev/packages/checks_codegen/publisher)
+
+`package:checks_code` is a companion to [`package:checks`] for generating
```
jonasfj:
```suggestion
`package:checks_codegen` is a companion to [`package:checks`] for generating
```

pkgs/checks_codegen/README.md
```diff
@@ -0,0 +1,46 @@
...
+1. Use `checks` that read fields in your test code:
+
+```dart
+@CheckExtensions([TypedData])
+import 'some_test.checks.dart';
+
+// The `elementSizeInBytes` and `lengthInBytes` extensions are generated.
+void main() {
+  test('sample test', () {
+    check(typedData)
```
jonasfj, line 44: **crazy idea:** We could detect at code-gen time that there are no extensions for `Subject<TypedData>` and on the basis of this generate such an extension.

Probably there are some potential issues around what to do if there are extensions for a super type... and maybe it's a bit fragile.

But it would be cool, to do in a future where users commonly have `dart run build_runner watch` sitting in the background, then you'd just write `checks(foo)` and it'd magically work on demand.

Do ignore this comment, because, yes I can see lots of reasons not to be this magical 🤣

pkgs/checks_codegen/lib/src/annotation.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Annotation specifying types to generate Subject extensions with `has`
+/// getters for fields.
+///
+/// Annotate an import to the `.checks.dart` library.
```
jonasfj:
```suggestion
/// Annotate an import to the `.checks.dart` library.
///
/// {@example /example/test/example_test.dart}
```

pkgs/checks_codegen/lib/src/builder.dart
```diff
@@ -0,0 +1,211 @@
...
+import 'package:source_gen/source_gen.dart' hide LibraryBuilder;
+
+import 'annotation.dart';
+
+Builder checksBuilder(BuilderOptions? _) => source_gen.LibraryBuilder(
+  const ChecksGenerator(),
+  generatedExtension: '.checks.dart',
+);
+
+class ChecksGenerator extends GeneratorForAnnotation<CheckExtensions> {
```
jonasfj: Is there any good reasons for keeping this class public? Just asking?
```suggestion
final class ChecksGenerator extends GeneratorForAnnotation<CheckExtensions> {
```

pkgs/checks_codegen/test/builder_test.dart
```diff
@@ -0,0 +1,119 @@
...
+    int get intField;
+}
+''',
+          'a|test/bar.dart': '''
+abstract class Bar {}
+''',
+        },
+        readerWriter: readerWriter,
+        outputs: {
+          'a|test/some_test.checks.dart': decodedMatches(
```
jonasfj: Doesn't `testBuilder` return something you can feed into `checks()`, or are we missing some `Subject` extensions 🤣 

I don't mind, but really shouldn't we be using `checks()`.

pkgs/checks_codegen/lib/src/builder.dart
```diff
@@ -0,0 +1,211 @@
...
+    final basename = p.url.basenameWithoutExtension(buildStep.inputId.path);
+    final expectedImport = '$basename.checks.dart';
+
+    if (directive case LibraryImport(
+      :final DirectiveUriWithRelativeUri uri,
+    ) when uri.relativeUriString == expectedImport) {
+      // Annotation is on the correct import
+    } else {
+      throw InvalidCheckExtensions(
+        'must annotate an import of $expectedImport',
```
jonasfj: This limitation should probably be documented on `CheckExtensions`.

--------

**Orthogonal question:** is this a good idea? Is it not a bit excessive to have a `<name>_test.checks.dart` file for each `<name>_test.dart` file?

I'll grant that I'm not 100% sure how to otherwise do it.

In my projects so far I've often done a `test/checks_ext.dart` file and then I write all the checks extensions I need/want into that file.

Often times I also end up making a `test/harness.dart` which exports _checks extensions_ and one or more wrappers for `test()`, like `testWithContext(String name, FutureOr<void> Function(Context ctx) fn)` for some `Context`.

-------------

We could also have `test/my_checks_stuff.dart` like:
```dart
@CheckExtensions()
library;

import 'dart:io';
import 'dart:typed_data';

// Then just generate Subject<T> extensions for all classes T that are
// imported in this library, into the part file below:
part 'my_checks_stuff.g.dart';
```

This would:
 * Give users a single library that contains all the _checks extensions_ (granted with two part files)
 * Avoid a library with _checks extensions_ for each `_test.dart` file.
 * Let users control if they want to limit how much is generated using `show/hide` modifiers.
 * Let users write `import '...'` and `export '...'`, so that their tests just imports `test/my_checks_stuff.dart` and nothing else.
 * Give users a place to put their manually written _checks extensions_, and a natural place to write test helpers.
 * Stick to the common `<name>.g.dart` pattern for generated files.

Granted there is a serious risk that people will import too much and generate HUGE part files. So it might make sense to keep the `@CheckExtensions([TypeData, ...])` annotation. On the other hand, generating a lot of extensions is hopefully fairly harmless :rofl:

pkgs/checks_codegen/lib/src/builder.dart
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }
+}
+
+class InvalidCheckExtensions extends Error {
+  final String message;
+  InvalidCheckExtensions(this.message);
+  @override
+  String toString() => 'Invalid `CheckExtensions` annotation: $message';
+}
```
jonasfj: Why not use [`InvalidGenerationSource`](https://pub.dev/documentation/source_gen/latest/source_gen/InvalidGenerationSource-class.html) from `package:source_gen` ?

pkgs/checks_codegen/lib/src/annotation.dart
jonasfj: nit:

Would be nice to use:
```
@Target(.importDirective)
class CheckExtensions {
```

But it doesn't exist :/

We could `.directive` but it's deprecated:
https://pub.dev/documentation/meta/latest/meta_meta/TargetKind/directive-constant.html

pkgs/checks_codegen/lib/src/builder.dart
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  @override
+  dynamic generateForAnnotatedElement(
+    Element element,
+    ConstantReader annotation,
+    BuildStep buildStep,
+  ) {
+    final basename = p.url.basenameWithoutExtension(buildStep.inputId.path);
+    throw InvalidCheckExtensions(
```
jonasfj: Consider using source span from `element`, it's not perfect, but better than nothing.

Makes more sense if using `InvalidGenerationSource`.

For `package:typed_sql` I ended up abandoning `GeneratorForAnnotation` inorder to make sure I could access the annotation source spans:

https://github.com/google/dart-neats/blob/7cb984e7d85b0201b5b110629afa45a9ed580adf/typed_sql/lib/src/codegen/analyzer_utils.dart

Granted, this is probably something we should put into `package:source_gen` at some point.

pkgs/checks_codegen/lib/src/builder.dart
```diff
@@ -0,0 +1,211 @@
...
+
+  Future<Extension> _createExtension(
+    List<LibraryElement> imports,
+    DartObject dartObject,
+    Resolver resolver,
+    String entryAssetPath,
+  ) async {
+    final type = dartObject.toTypeValue();
+    if (type is! InterfaceType) {
+      throw StateError('Got a non interface type: $type');
```
jonasfj: Probably, this is a user error? If I put a function type into the annotation:
```dart
typedef Foo = int Function(int);

@CheckExtensions([Foo])
```

Granted it's a corner case.

jonasfj (commented):
I LOVE more generation support for checks.

I'm a little unsure if a  `foo_test.checks.dart` file for each test file is desirable.

natebosch:
> I'm a little unsure if a `foo_test.checks.dart` file for each test file is desirable.

I'm unsure about generating for entire imported surface area, but I think we can also add that down the line if we make the argument to the annotation optional.  I do think the shared test utility file is a pattern worth supporting, and it's pretty easy to do if we expand the targets for the annotation to exports as well. WDYT about the new example?

pkgs/checks_codegen/lib/src/annotation.dart
natebosch: I don't see `@example` listed at https://github.com/dart-lang/dartdoc/blob/main/doc/directives.md

Is there support for custom doc directives with `@example` defined somewhere?

pkgs/checks_codegen/README.md
natebosch: I think I like the idea of generating for all definitions of imported libraries a bit more than I like parsing test bodies to detect which types are used.

pkgs/checks_codegen/lib/src/annotation.dart
jonasfj: I think it might not have been documented, I'll ask..

pkgs/checks_codegen/lib/src/builder.dart
```diff
@@ -0,0 +1,212 @@
...
+
+    if (directive
+        case LibraryImport(:final DirectiveUriWithRelativeUri uri) ||
+            LibraryExport(:final DirectiveUriWithRelativeUri uri)
+        when uri.relativeUriString == expectedImport) {
+      // Annotation is on the correct import or export
+    } else {
+      throw InvalidGenerationSourceError(
+        'must annotate an import or export of $expectedImport',
+      );
```
jonasfj, line 44: nit:

```suggestion
      throw InvalidGenerationSourceError(
        'must annotate an import or export of $expectedImport',
        element: directive,
      );
```

I think that might work, and probably better than hanging it off nothing, not sure we can easily hang it off the annotation.

pkgs/checks_codegen/lib/src/builder.dart
```diff
@@ -0,0 +1,212 @@
...
+  Future<Extension> _createExtension(
+    List<LibraryElement> imports,
+    DartObject dartObject,
+    Resolver resolver,
+    String entryAssetPath,
+  ) async {
+    final type = dartObject.toTypeValue();
+    if (type is! InterfaceType) {
+      throw InvalidGenerationSourceError(
+        'Only interface types may be used for checks extensions:: $type',
```
jonasfj:
```suggestion
        'Only interface types may be used for checks extensions: $type',
```

pkgs/checks_codegen/lib/src/builder.dart
```diff
@@ -0,0 +1,212 @@
...
+            ..symbol = 'Subject'
+            ..url = 'package:checks/context.dart'
+            ..types.add(refer(element.displayName, import)),
+        )
+        ..methods.addAll(hasGetters),
+    );
+  }
+
+  bool _isCheckableField(FieldElement field) =>
+      field.name != 'hashCode' && !field.isStatic;
```
jonasfj: I was trying to think through heuristics that could work here. But I think they are bad. Because it becomes hard to predict if a property is generated or not.

And for anything nullable `.isNull()` is useful, and using analyzer to globally search for `Subject<ReturnType>` is both complex and hard for the user to understand.

Maybe, properties that return non-nullable function types should not have a getter. But then again, those could be useful some day.

jonasfj (approved):
> WDYT about the new example?

This is a great example!

I can see that there is really no need for a part-file, just generating a `<basename>.checks.dart` is perfectly fine.

And you're probably right to keep the `<basename>.checks.dart` constraint on the filename, because that means there is one canonical place where the `@CheckExtensions` annotation is made.

Even if people might import `<basename>.checks.dart` from other libraries (granted they probably shouldn't).

pkgs/checks_codegen/test/builder_test.dart
natebosch: Took a bit of learning after some changes to` build_test`, but yes it has what we need to write this as a check. Done.

pkgs/checks_codegen/lib/src/builder.dart
natebosch: I think it makes sense to omit function typed fields to start. We can always add them later and it shouldn't be breaking. Removed function typed fields and added a test.

pkgs/checks_codegen/lib/src/builder.dart
natebosch: The problem here is an `ElementDirective` isn't an `Element`. I didn't see a way to get a relevant `Element` instance in this code path.

PR #2638, opened 2026-05-01
Override test packages in coverage tests

natebosch:
These tests use the `runTest` test utility which uses a precompiled test
runner built from the repo sources. They also use `runPub`  to run fetch
published versions of the test packages using a pubspec it creates. This
causes a version mismatch between the runner and the version of the test
packages the test file is importing when it is compiled.

Add a `dependency_overrides` section to each generated pubspec in the
coverage tests. Override all 3 test runner implementation packages to
use the repo sources.

### DIFF
--- a/pkgs/test/test/runner/coverage_test.dart
+++ b/pkgs/test/test/runner/coverage_test.dart
@@ -22,6 +22,28 @@ void main() {
   group('with the --coverage flag,', () {
     late Directory coverageDirectory;
     late String pkgDir;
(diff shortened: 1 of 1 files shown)


natebosch:
Discovered this was a problem in #2449 but I think it's worth landing separately.

natebosch:
In the long term, these might do better as integration tests

kevmoo (approved):
Solid

Issue #2577, opened 2025-12-09
Test harness crashes/hangs on test crash


rmacnak-google:
```dart
import "package:test/test.dart";
import "dart:ffi";

main() {
  test("fault", () {
    Pointer.fromAddress(0).cast<Long>()[0] = 0;
  });
}
```

With `dart test example.dart`, the test harness itself crashes.

With `dart test -c exe example.dart`, the test harness hangs.

jakemac53:
> With `dart test example.dart`, the test harness itself crashes.

In this mode we run the test in an isolate spawned by the main test runner (it is given a kernel file). Is the exception thrown in this case something we could catch? I am not sure how exactly that would work.

> With dart test -c exe example.dart, the test harness hangs.

This case we should be able to handle better I think.

rmacnak-google:
> Is the exception thrown in this case something we could catch?

There is no exception. The process terminates with SIGSEGV or SIGABRT. The test harness and the tests need to be in separate processes to be robust.

jakemac53:
> There is no exception. The process terminates with SIGSEGV or SIGABRT. The test harness and the tests need to be in separate processes to be robust.

Yeah ok then this is basically just working as expected today.

We could run all tests in a separate process instead of in an isolate, I am not sure if there would be a performance impact to that. This would however have an extra benefit of `exit()` calls not exiting the entire test runner (although it would still exit the suite). But, we could detect early exits in the runner and make them failures.

natebosch:
I think moving vm tests out of process by default is unlikely to get prioritized. Crashes are not a common scenario and the negative side effects are limited to a slightly confusing UX and not seeing the results of some tests that might have run anyway. It doesn't turn a failure into a incorrect success, nor cause spurious failures. The `exit` loophole is a bit riskier since it can make failures look like successes with a 0 exit code.

The hang we should address.

PR #2640, opened 2026-05-15
Disallow exiting from VM tests

natebosch:
As suggested in #2577

Tests exiting early with a hard exit is an infrequent but persistent
source of confusion. Use `IOOverrides` to treat any call to `exit` as a
test failure instead of a VM exit.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -10,6 +10,7 @@
   independent from any OS specific configuration. For instance a wide skip of
   all tests with OS `'windows'` would previously still run browser tests on
   windows, but will now skip all tests including browser tests.
+* Treat calls to `exit` as test failures in VM tests.
 
 ## 1.31.1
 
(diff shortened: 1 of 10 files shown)


natebosch:
Diff will look nicer once we land https://github.com/dart-lang/test/pull/2641

In case a reviewer is looking before it lands, easier to read just the relevant commits: https://github.com/dart-lang/test/pull/2640/changes/BASE..2c6de259c62c6880b3da8d81051a46a9050342f7

jakemac53:
Should we possibly pair this with a matcher that expects `exit` to be called?

Of course users could do their own IOOverrides to check this, but it might we worth having something simple for them to use as a workaround here.

jakemac53 (approved):

natebosch:
> Should we possibly pair this with a matcher that expects `exit` to be called?

Sounds interesting. I'd be inclined to implement something for this in `package:checks` but it's not a very urgent need since we haven't generally gotten requests for that capability as much as complaints about the confusing behavior.

PR #2644, opened 2026-05-18
More consistent unexpected error formatting

natebosch:
- Include stack traces in the rejection output for more extensions which
  have catch blocks for unexpected exceptions.
- Use more consistent phrasing and formatting with a line ending in
  `at:` before the stack trace.
- Add indentation to all outputs that include stack traces.
- Use the `LineSplitter.split` utility method everywhere lines are split.

### DIFF
--- a/pkgs/checks/lib/src/extensions/async.dart
+++ b/pkgs/checks/lib/src/extensions/async.dart
@@ -27,7 +27,7 @@ extension FutureChecks<T> on Subject<Future<T>> {
           actual: ['a future that completes as an error'],
           which: [
             ...prefixFirst('threw ', postfixLast(' at:', literal(e))),
-            ...indent(const LineSplitter().convert(st.toString())),
+            ...indent(LineSplitter.split(st.toString())),
           ],
         );
       }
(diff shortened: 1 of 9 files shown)


pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -27,7 +27,7 @@ extension FutureChecks<T> on Subject<Future<T>> {
           actual: ['a future that completes as an error'],
           which: [
             ...prefixFirst('threw ', postfixLast(' at:', literal(e))),
-            ...const LineSplitter().convert(st.toString()),
+            ...indent(const LineSplitter().convert(st.toString())),
```
kevmoo: Maybe https://api.dart.dev/dart-convert/LineSplitter/split.html instead?

kevmoo (approved):
One tiny nit...

pkgs/checks/lib/src/extensions/async.dart
natebosch: Updated this usage across the package.

PR #2646, opened 2026-05-19
Fix output for multiline clauses in failures

natebosch:
In a failure message the parts of the expectation which were checked and
satisfied are represented in both the "Expected" and "Actual" portions.
This is handled by tracking how many lines of the text overlaps.
Previously this tracked the depth of nesting and assumed each level
contributed only a single line, but this assumption is violated by some
expectations which print with multiple lines. Correct the overlap
accounting to include the actual number of lines added by a label.


kevmoo (approved):
Nice! Excited to see this getting tighter!

PR #2647, opened 2026-05-19
Handle async conditions in mayEmit

natebosch:
Fix a bug where the allowed `AsyncCondition` is used with `softCheck`
instead of `softCheckAsync`. Make it feasible to catch the exception for
using an asynchronous condition in `describe`, and use that exception to
detect when to fallback on a more general clause description for
`mayEmit` and `mayEmitMultiple`.

Note that callers of `nestAsync` and `expectAsync` should synchronously
surface a potential synchronous exception and update all existing
callers to do so.

### DIFF
--- a/pkgs/checks/CHANGELOG.md
+++ b/pkgs/checks/CHANGELOG.md
@@ -3,6 +3,9 @@
 - Require Dart 3.7
 - Improve speed of pretty printing for large collections.
 - Improve formatting for failures involving unexpected exceptions.
+- Fix a bug when using asynchronous conditions with `mayEmit` or
+  `mayEmitMultiple`. Note that extensions using `nestAsync` should synchronously
+  forward exceptions from that call.
 
 ## 0.3.1
 
--- a/pkgs/checks/lib/context.dart
+++ b/pkgs/checks/lib/context.dart
@@ -5,6 +5,7 @@
 export 'src/checks.dart'
     show
         AsyncCondition,
+        AsyncConditionDisallowed,
         CheckFailure,
         Condition,
         Context,
--- a/pkgs/checks/lib/src/checks.dart
+++ b/pkgs/checks/lib/src/checks.dart
@@ -1072,5 +1072,5 @@ class AsyncConditionDisallowed implements Exception {
   final String flavor;
   AsyncConditionDisallowed._(this.flavor);
   @override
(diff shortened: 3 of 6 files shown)


kevmoo (approved):
It all seemed reasonable to me. I asked Gemini, too...

This pull request addresses a bug in the `checks` package where `mayEmit` and `mayEmitMultiple` were incorrectly using synchronous checking (`softCheck`) for potentially asynchronous conditions. It also improves how asynchronous conditions are handled during description generation.

### Key Changes

1.  **Bug Fix in `mayEmit` and `mayEmitMultiple`**:
    *   Switched from `softCheck` to `await softCheckAsync` to correctly handle asynchronous conditions when peeking at stream events.
2.  **Introduction of `AsyncConditionDisallowed`**:
    *   A new exception type specifically for cases where an asynchronous expectation is attempted on a subject that doesn't support it (e.g., during a synchronous `describe` call).
3.  **Synchronous Exception Surfacing**:
    *   Modified `expectAsync` and `nestAsync` to perform the `_allowAsync` check synchronously before returning a `Future`.
    *   Updated built-in extensions (like `completes`, `throws`, `emits`) to return the `Future` from `nestAsync`/`expectAsync` directly rather than being `async` and `await`ing. This ensures that `AsyncConditionDisallowed` is thrown immediately when the extension is called, allowing `describe` to catch it.
4.  **Graceful Fallback for Descriptions**:
    *   `mayEmit` and `mayEmitMultiple` now catch `AsyncConditionDisallowed` during description generation and fall back to a generic description: `"may emit a value satisfying an asynchronous condition"`.

### Review Findings

The PR is technically sound and follows best practices for the `checks` package:
*   **Correctness**: The fix for `mayEmit` is essential for supporting async matchers in stream checks.
*   **Robustness**: The new exception and synchronous surfacing logic make the API more predictable and easier to debug when async matchers are misused.
*   **Maintainability**: Updating the built-in extensions sets a clear pattern for how third-party extensions should be implemented to support synchronous description.
*   **Testing**: The added tests in `async_test.dart` adequately cover both the fix and the new description fallback logic.

### Recommendation

The PR is ready for approval. The changes are focused, well-tested, and correctly solve the reported issue while improving the overall robustness of the async checking API.


- you impressed the robot, @natebosch !

PR #2648, opened 2026-05-19
Indent string diffs lines

natebosch:
I noticed that the failure messages are a little easier to read if the
lines showing the point in the string that deviates are indented
relative to the "Which" header.

Before

    Expected: a String that:
      equals 'some other String'
    Actual: 'some String'
    Which: differs at offset 5:
    some other Stri ...
    some String
         ^

After

    Expected: a String that:
      equals 'some other String'
    Actual: 'some String'
    Which: differs at offset 5:
      some other Stri ...
      some String
           ^


kevmoo (approved):
Agreed! Looks better.

PR #2651, opened 2026-05-20
Add support for collapsed error messages

natebosch:
Checks against a subject are commonly nested against some other subject,
for example the check for an exact list length is in practice an
`equals` expectation checked against an integer which is the "has
length" subject, nested under the subject holding the list instance.
Failure messages have a rigid structure that represents this nesting
pattern and always indent the "clauses" (the descriptions of what was
checked in expectations) following the "label" which describes the
subject.

For very simple conditions the noise of the structure is greater than
the signal of the description output by the specific failing
expectation. For instance a trivial check for a future completing to an
expected value looks like:

```
  Expected: a Future<List<int>> that:
    completes to a value that:
      is not empty
  Actual: a Future<List<int>> that:
    completes to a value that:
    Actual: []
    Which: is empty
```

Add support for collapsing this to:

```
  Expected: a Future<List<int>> that completes to a non-empty iterable
  Actual: a Future<List<int>> that completes to []
  Which: is empty
```

Add a `predicateNoun` callback argument to `expect`, `expectAsync`, and
`expectUnawaited`. Extensions must pass a callback which provides the
"clause" representation like "is empty". Now they may also pass a
callback which provides a predicate noun formed by combining the clause
with a noun for the subject like "an empty iterable".

Add an `addPredicate` callback argument to `nest`, and `nestAsync`.
Extensions must pass a callback which provides a label for the value
they are deriving from the original subject like "completes to a value".
Now they may also pass a callback which takes a collapsed form of the
noun as described by nested conditions, and attaches the predicate it
contributes.

Check for possibilities to collapse failure messages down to predicate
nouns when generating failure messages. When full single-line collapsing
is not possible, child clauses and `Actual:` lines still collapse where
supported within the multi-line format.

Add predicate noun callbacks to existing extensions where sensible.
Tweak some wording where I noticed inconsistencies.

Use of these new arguments is completely optional for extension
implementations and impacts only the failure message formatting which is
not considered a breaking change.

The user impact is that many simple expectation failures will be more
compact and readable, but the position of the failing expectation in a
series of cascaded expectations may cause a significantly different
output for the same checks usage. It has always been the case that
expectations which follow a failure have no impact on the output,
because they are never reached to run. Now when the first expectation in
a series has a collapsible representation, its failure triggers an
overall collapsed format which will read differently from the expanded
format if any condition other than the first fails.


pkgs/checks/lib/src/checks.dart
```diff
@@ -243,15 +254,29 @@ extension ContextExtension<T> on Subject<T> {
...
+///   ends with 'z'
 /// ```
-/// {@endtemplate}
 ///
+/// An expectation can optionally provide a [predicateNoun] callback.
+/// [predicateNoun] should return a single-line noun phrase including this
+/// expectation modifying the noun (often just a representation of the expected
+/// value) to be used when collapsing simple expectations.
+///
+/// Examples of predicate nouns:
```
brianquinlan, line 271: Very nice! Love the examples.

brianquinlan (approved):
This looks goods if you are convinced that the complexity is worth it.

pkgs/checks/lib/src/extensions/core.dart
```diff
@@ -124,27 +126,34 @@ extension CoreChecks<T> on Subject<T> {
 
   /// Expects that the value is equal to [other] according to [operator ==].
   void equals(T other) {
-    context.expect(() => prefixFirst('equals ', literal(other)), (actual) {
-      if (actual == other) return null;
-      return Rejection(which: ['are not equal']);
-    });
+    context.expect(
+      () => prefixFirst('equals ', literal(other)),
+      predicateNoun: () => literal(other).singleOrNull,
```
lrhn, line 131: Probably not worth it, but:

Doing `literal(other)` twice, which calls the potentially expensive `prettyPrint`.
(It returns an `Iterable` which is _always_ a `List`. Just change the return type to `List`, instead of risking code that depends on it being cheap to iterate and safe to iterate twice, without actually being promised that.)

Could do `List<String>? otherLiteral;` before the  `context.expect` call and `() => prefixFirst('equals ', otherLiteral ??= literal(other))`, and 
`predicateNoun: () => (otherLiteral ??= literal(other)).singleOrNull`.

_I know this only gets called if a test fails, and it's not like any of the other list-of-string operations are optimized, so it's _probably_ not performance critical. I just don't like seing duplicate work._

It happens a lot that there is an `other` and both callbacks call `literal` on it, but it's not _every_ `context.expect` that has an `other` value.
(Could have a helper extension function `context.expectWithOther(other, (otherLiteral) => prefixFirst('equals ', literal(other)), predicateName: (otherLiteral) => otherLiteral.singleOrNull, ...`, or a `LazyLiteral(other)` with a caching `lazyOtherLiteral.literal`.)

pkgs/checks/lib/src/extensions/core.dart
```diff
@@ -124,27 +126,34 @@ extension CoreChecks<T> on Subject<T> {
...
+      (actual) {
+        if (actual == other) return null;
+        return Rejection(which: ['is not equal']);
```
lrhn, line 134: I'd keep the `are`. 

Being equal is a binary relation. It needs either two subjects (so s plural "are") or a `to <other>` subject complement (a term I just looked up, so I hope it's correct).

Unless the "to" gets added by the `predicateNoun` handling, but I don't really nuderstand the flow flow, and _there is no documentation!_

pkgs/checks/lib/src/extensions/core.dart
```diff
@@ -124,27 +126,34 @@ extension CoreChecks<T> on Subject<T> {
...
-    ) {
-      if (identical(actual, other)) return null;
-      return Rejection(which: ['is not identical']);
-    });
+    context.expect(
+      () => prefixFirst('is identical to ', literal(other)),
+      predicateNoun: () => literal(other).singleOrNull,
+      (actual) {
+        if (identical(actual, other)) return null;
+        return Rejection(which: ['is not identical']);
```
lrhn, line 146: "are", unless `to $other` gets added somewhere.
(And I doubt that, because won't always be "to" that's the right word, so unless it's a default, this function should still mention the "to".)

pkgs/checks/lib/src/extensions/core.dart
```diff
@@ -170,7 +180,9 @@ extension NullableChecks<T> on Subject<T?> {
   }
 
   void isNull() {
-    context.expect(() => const ['is null'], (actual) {
+    context.expect(() => const ['is null'], predicateNoun: () => 'null', (
+      actual,
+    ) {
       if (actual != null) return Rejection();
```
lrhn: (Total nit: In all the other functions, the stucture is `if (successTest) return null; return Rejection(...);`. this one is the other way around.)

lrhn: (So no `which` because it's obvious that `Expected: null` and `Actual: <something else>` is not the same. Makes sense.
Do we check anywhere if `actual.toString()` equals `expected.toString()`, and print something more to distinguish it, like the runtime type or print strings with quotes?)

pkgs/checks/lib/src/extensions/iterable.dart
```diff
@@ -10,58 +10,80 @@ import 'core.dart';
...
-    while (iterator.moveNext()) {
-      current = iterator.current;
-    }
-    return Extracted.value(current);
-  });
+  Subject<T> get last => context.nest(
+    () => ['has last element'],
+    addPredicate: (predicateNoun) => 'has last element: $predicateNoun',
+    (actual) {
+      final iterator = actual.iterator;
```
lrhn, line 29: (Unrelated: Could this be using the `actual.last` instead. It might be more efficient.
It'd have to catch a `StateError` and recognize that as meaning "no element", which isn't great, or iterate twice, which is also not great.
_I tried testing an iterable, and was surprised when `check(myIteralble).last.equals(x)` didn't call my `last` code. I guess you have to do `check(() => myIterable.last)...`, putting the actual code you want to check inside the `check`, to be sure. Might need to be documented somewhere.)

pkgs/checks/lib/src/extensions/iterable.dart
lrhn, line 89: Nit: Iterates iterable twice.
Consider at least only doing the `actual.isEmpty` check after the `contains` check, when you know it'll be a rejection anyway, so the success path only iterates once.

pkgs/checks/lib/src/extensions/function.dart
```diff
@@ -17,8 +17,10 @@ extension FunctionChecks<T> on Subject<T Function()> {
   /// If this function is async and returns a [Future], this expectation will
   /// fail. Instead invoke the function and check the expectation on the
   /// returned [Future].
-  Subject<E> throws<E>() {
-    return context.nest<E>(() => ['throws an error of type $E'], (actual) {
+  Subject<E> throws<E>() => context.nest<E>(
+    () => ['throws an error of type $E'],
```
lrhn: (Could check if `const Object() is E`, meaning there is no type check and any thrown is accepted, and if so drop the `of type $E`.)

pkgs/checks/lib/src/extensions/string.dart
```diff
@@ -11,46 +11,75 @@ import 'core.dart';
...
-    context.expect(() => const ['is not empty'], (actual) {
-      if (actual.isNotEmpty) return null;
-      return Rejection(which: ['is empty']);
-    });
+    context.expect(
+      () => const ['is not empty'],
+      predicateNoun: () => 'a non-empty string',
+      (actual) {
+        if (actual.isNotEmpty) return null;
+        return Rejection(which: ['is empty']);
```
lrhn, line 48: Is this neeed. If the line above is `Actual: <''>`, the `which` line should be no more needed than it is for `isNull`/`isNotNull`.

pkgs/checks/lib/src/extensions/string.dart
```diff
@@ -11,46 +11,75 @@ import 'core.dart';
...
+    context.expect(
+      () => prefixFirst('starts with ', literal(other)),
+      predicateNoun: () {
+        final l = literal(other).singleOrNull;
+        return l != null ? 'a string starting with $l' : null;
+      },
+      (actual) {
+        if (actual.startsWith(other)) return null;
+        return Rejection(
+          which: prefixFirst('does not start with ', literal(other)),
```
lrhn, line 63: Also possibly not needed.
A more relevant error would be `which: 'differs from the expectation at index $i'` to say where it differs.
(Or consider:
```
Expected: '$other...'
Actual  : '$actual...'
Which   :        ^ differs here
```
as a strethch goal 😁 )

pkgs/checks/lib/src/extensions/string.dart
```diff
@@ -11,46 +11,75 @@ import 'core.dart';
...
+    context.expect(
+      () => prefixFirst('ends with ', literal(other)),
+      predicateNoun: () {
+        final l = literal(other).singleOrNull;
+        return l != null ? 'a string ending with $l' : null;
+      },
+      (actual) {
+        if (actual.endsWith(other)) return null;
+        return Rejection(
+          which: prefixFirst('does not end with ', literal(other)),
```
lrhn, line 79: Same, but with `'differs at index $i from the end'.

pkgs/checks/lib/src/checks.dart
```diff
@@ -83,16 +83,27 @@ Subject<T> check<T>(T value, {String? because}) => Subject._(
...
+          ? (0, ['Expected: ${collapsed.$1}'], ['Actual: ${collapsed.$2}'])
+          : (
+              f.detail.depth,
+              prefixFirst('Expected: ', f.detail.expected),
+              [
+                ...prefixFirst('Actual: ', f.detail.actual),
+                if (!f.detail.hasCustomActual)
+                  ...indent(
+                    prefixFirst('Actual: ', f.rejection.actual),
+                    f.detail.depth,
```
lrhn, line 97: So output is:
```
Expected: ...
Actual: ....
  Actual: ....
```
Is it expected that `f.detail.actual` ends with `  Expected: ...`?

pkgs/checks/lib/src/checks.dart
```diff
@@ -296,42 +335,57 @@ extension ContextExtension<T> on Subject<T> {
...
+///    collapse into compact, single-line "Expected:" and "Actual:" lines:
 ///
-/// But if the failure happens on a nested subject, the actual starts with a
-/// description of the nesting or non-nesting expectations that succeeded, up
-/// to nesting point of the failure, then the "actual" and "which" from the
-/// rejection are indented to that level of nesting.
+///    ```
+///    Expected: a User that has name: 'Alice'
+///    Actual: a User that has name: 'Bob'
+///    Which: is not equal
```
lrhn, line 359: "are" here too.

pkgs/checks/lib/src/checks.dart
```diff
@@ -296,42 +335,57 @@ extension ContextExtension<T> on Subject<T> {
...
+///    use the "actual" and "which" arguments from the rejection at that level:
+///
+///    ```
+///    Expected: a Future<String> that:
+///      completes to a value that:
+///        starts with 'a'
+///        ends with 'z'
+///    Actual: a Future<String> that:
+///      completes to a value that:
+///      Actual: 'abcd'
```
lrhn, line 422: (Ah, tha's where the indented `Actual:` is used.
Could it use `Value: ` instead of `Actual: `?  And end the prior line with `completes to a value:`?

The "... that: Actual: .... Which:...` doesn't work as a sentence. The "Actual" feels like it should be in relation to an expectation, but there is no expectation mentioned.

pkgs/checks/lib/src/checks.dart
```diff
@@ -852,37 +941,174 @@ final class _TestContext<T> implements Context<T>, _ClauseDescription {
     return current;
   }
 
+  /// Returns `true` if this context nests under [clause] (in other words,
```
lrhn: "Returns `true` if" -> "Whether".
(Always! This doc doesn't say what it returns if this context doesn't nest under the clause, you have to deduce that it's probably `false`.)

pkgs/checks/lib/src/checks.dart
```diff
@@ -852,37 +941,174 @@ final class _TestContext<T> implements Context<T>, _ClauseDescription {
...
+  /// whether [clause] is this context itself or one of its ancestors).
+  ///
+  /// This is used to identify which clauses are on the path to the failing
```
lrhn: (Nit: Avoid unqualified "this" or similar words. I'd say "The result is used ...".
I know this is a private function, but if it's worth spending time to write a doc, and for someone to read it, it's worth doing awesomely!)

pkgs/checks/lib/src/checks.dart
```diff
@@ -852,37 +941,174 @@ final class _TestContext<T> implements Context<T>, _ClauseDescription {
...
+  /// expectation, and should therefore be formatted with the rejection details.
+  bool _nestsUnder(_ClauseDescription clause) {
+    if (identical(clause, this)) return true;
+    if (clause.isLeaf) return false;
+    var current = _parent;
```
lrhn, line 991: Could combine the first `identical` check into the loop.
```dart
  if (clause.isLeaf) return false;
  for (var current = this; !identical(current, clause);) {
     var parent = current._parent;
     if (parent == null) return false;
     current = parent;
  }
  return true;
}
```
(A little more complicated than I first expected, to avoid having to declare `current` as nullable.)

pkgs/checks/lib/src/checks.dart
```diff
@@ -852,37 +941,174 @@ final class _TestContext<T> implements Context<T>, _ClauseDescription {
...
+          thisContextFailed: thisContextFailed,
+        ) ??
+        _fullDetails(
+          failingContext,
+          rejection,
+          thisContextFailed: thisContextFailed,
+        );
+  }
+
+  /// Returns a collapsed single-line [FailureDetail], or `null` if this context
```
lrhn: Drop "Returns ". (Possibly also "a".)
Function name is noun phrase, docs should follow suit.

(Never start a DartDoc with the word "Returns". Reserve that for the paragraph that describes the returned value.)

pkgs/checks/lib/src/checks.dart
```diff
@@ -852,37 +941,174 @@ final class _TestContext<T> implements Context<T>, _ClauseDescription {
...
+  /// Returns `null` if there is not exactly one clause, if the child clause did
+  /// not collapse, or if expanding the child representation at this level fails
+  /// (e.g., if `_addPredicate` returns `null` or the resulting root string
+  /// exceeds length limits).
+  FailureDetail? _collapsedDetails(
+    _TestContext failingContext,
+    Rejection? rejection, {
+    required bool thisContextFailed,
+  }) {
+    final clause = _clauses.singleOrNull;
```
lrhn: (Wouldn't it be awesome if `return` was an expression:
```dart
   final clause = _clauses.singleOrNull ?? (return null);
```
😁 )

pkgs/checks/lib/src/checks.dart
```diff
@@ -852,37 +941,174 @@ final class _TestContext<T> implements Context<T>, _ClauseDescription {
...
+    (String, String)? expandedCollapsed;
+    if (_addPredicate != null) {
+      final exp = _addPredicate(childExpected);
+      final act = _addPredicate(childActual);
+      if (exp != null && act != null) expandedCollapsed = (exp, act);
+    } else if (_parent == null) {
+      if (clause.isLeaf) {
+        expandedCollapsed = (childExpected, childActual);
+      } else {
+        final rootLabel = _label().single;
```
lrhn: (Confusing to read, could `_label` be named `_labelFactory` or `_labelFunction`?)

pkgs/checks/lib/src/checks.dart
```diff
@@ -852,37 +941,174 @@ final class _TestContext<T> implements Context<T>, _ClauseDescription {
...
+        : (thisContextFailed ? 0 : -1);
+    return FailureDetail(
+      [expandedCollapsed.$1],
+      -1,
+      depth,
+      collapsedDetails: expandedCollapsed,
+    );
+  }
+
+  /// Computes the uncollapsed, multi-line [FailureDetail] for this context and
```
lrhn: Noun-phrased function, drop the "Computes". (Which is better than "Returns", but not more necessary.)

pkgs/checks/lib/src/checks.dart
```diff
@@ -925,25 +1155,44 @@ final class _SkippedContext<T> implements Context<T> {
...
   final Iterable<String> Function() _expected;
-  _ExpectationClause(this._expected);
+  final String? Function()? _predicateNoun;
+  _ExpectationClause(this._expected, [this._predicateNoun]);
   @override
-  FailureDetail detail(_TestContext failingContext) =>
-      FailureDetail(_expected(), -1, -1);
+  FailureDetail detail(_TestContext failingContext, Rejection? rejection) {
+    final collapsedExpected = _predicateNoun?.call();
+    final collapsedActualString = collapsedExpected != null
```
lrhn: (Remove `String` from name to match the others. And to not repeat type in name.)

pkgs/checks/lib/src/checks.dart
```diff
@@ -925,25 +1155,44 @@ final class _SkippedContext<T> implements Context<T> {
...
+        ? rejection?.actual.singleOrNull
+        : null;
+    final collapsedDetails = collapsedActualString != null
+        ? (collapsedExpected!, collapsedActualString)
+        : null;
```
lrhn: Hard to follow the logic. Maybe nesting would be better?
```dart
(String, String)? collapsedDetails;
if (_predicateNoun?.call() case final collapsedExpected?) {
  if (rejection?.actual.singleOrNull case final collapsedActual?) {
    collapsedDetails = (collapsedExpected, collapsedActual);
  }
}
```
I think it's the same computation, but I find it much easier to read.
(And it doesn't need `!`, which is a sign that the code has the same logic as the programmer.)

lrhn (approved):
(Can't claim to actually be able to follow the computations, but it _looks_ plausible and it's tested!)

pkgs/checks/lib/src/extensions/core.dart
natebosch: This failure looks like:

```
  Expected: <2>
  Actual: <1>
  Which: is not equal
```

I think initially I had this longer and shortened it because it felt redundant:

```
  Expected: <2>
  Actual: <1>
  Which: is not equal to <2>
```

I think the "is" reads better, but we can adjust this in a followup. We don't consider is breaking to change the particular wording of failure messages.

pkgs/checks/lib/src/extensions/core.dart
natebosch: I think we can address this later and it will be non-breaking. For now I'll keep them separate since the callbacks are called (hopefully) rarely, and a memoization approach add complexity.

pkgs/checks/lib/src/extensions/core.dart
natebosch: Done.

pkgs/checks/lib/src/checks.dart
natebosch: It should be the case that `f.detail.actual` ends with a line like "\<some label\> that:". For instance if a failure happens on an expectation checked on a nested subject you can end up with (non-collapsed) failure messages like:

```
  Expected: a List<int> that:
    has length that:
      is less than <10>
      is greater than <0>
  Actual: a List<int> that:
    has length that:
    Actual: <0>
    Which: is not greater than <0>
```

So the `f.detail.actual` ends with `has length that:`.

pkgs/checks/lib/src/checks.dart
natebosch: I'll do some more thinking on this. It's not a pattern introduced by this PR (which makes this pattern less likely to appear) so I won't wait to land it.

The way we nest subjects leads to it feeling natural at the implementation level to reuse "actual" to describe both the overall actual value (as corresponds to the `check` argument) and the derived value which failed a specific expectation (as corresponds to whatever \<label\> that: nested subject).

I think using "Value" would be a difference without a distinction but there might be some other ways we can make this a little more clear.

pkgs/checks/lib/src/checks.dart

pkgs/checks/lib/src/checks.dart

pkgs/checks/lib/src/checks.dart
natebosch: > to avoid having to declare `current` as nullable

We also have to declare it a `_TestContext<dynamic>` to avoid it inferring `_TestContext<T>` and after that I think it reads nicer as-is.

pkgs/checks/lib/src/checks.dart
natebosch: Dropped "Returns", retained "a".

pkgs/checks/lib/src/checks.dart
natebosch: It's not the first time I've wanted that, but it's close 😄

pkgs/checks/lib/src/checks.dart
natebosch: How about `_labelCallback`?

Done.

pkgs/checks/lib/src/checks.dart

pkgs/checks/lib/src/checks.dart

pkgs/checks/lib/src/checks.dart

pkgs/checks/lib/src/extensions/iterable.dart
natebosch: https://github.com/dart-lang/test/issues/2746

pkgs/checks/lib/src/extensions/iterable.dart
natebosch: Done.

pkgs/checks/lib/src/extensions/string.dart
natebosch: For some of these I went back an forth on how verbose to be in the failures.

I went ahead and dropped this and the same in the iterable extensions.

Do you think the `'is not empty'` version should be removed too?

pkgs/checks/lib/src/extensions/string.dart
natebosch: https://github.com/dart-lang/test/issues/2747

pkgs/checks/lib/src/extensions/function.dart
natebosch: Done.

pkgs/checks/lib/src/extensions/string.dart
natebosch: Actually, I restored these for now, we can reconsider later.

In the collapsed format I think this may still have some value

```
  Expected: a Future<List<int>> that completes to a non-empty iterable
  Actual: a Future<List<int>> that completes to []
  Which: is empty
  ```

PR #2654, opened 2026-05-22
Fix unawaited_return_in_try_block

dcharkes:


dcharkes:
@natebosch are these unawaited futures intentional? E.g. should we add a lint suppression comment instead?

jakemac53 (approved):

jakemac53:
> @natebosch are these unawaited futures intentional? E.g. should we add a lint suppression comment instead?

This is just a new lint, but its not intentional, the try/catch won't work properly without the await and maybe some other things.

natebosch (approved):

PR #2655, opened 2026-05-26
New flag --shard-by-suite

### LINKED ISSUE #2614: Enable dart test to shard on files, not individual tests
We have a rather large test suite and most tests have a relatively time consuming setup and teardown. We have implemented sharding with `dart test`, which helps, but not at all to the degree we expected. 

It appears that the sharding cuts across the files, i.e. tests from the same test file occur in more than one shard. When splitting the work like this, the same setUpAll and tearDownAll chains need to be run multiple times. This is very inefficient.

We would like a new flag for `dart test` to make the shard splitter split on files instead of tests.

This has an additional benefit. Often a bug will cause a cluster of test cases to fail, and these are often in the same file, which would mean that they would be reported in the same shard's output and thus easier to get an overview of.

sigurdm:
Running all setupAll hooks is expensive when what we are trying to do is to minimize the time spent for each shard.

This allows choosing a less fine-grained strategy of sharding by individual test files.

Fixes https://github.com/dart-lang/test/issues/2614


pkgs/test/test/runner/runner_test.dart
```diff
@@ -84,6 +84,8 @@ $_runtimeCompilers
                                       (defaults to "$_defaultConcurrency")
     --total-shards                    The total number of invocations of the test runner being run.
     --shard-index                     The index of this test runner invocation (of --total-shards).
+    --shard-by                        How to distribute tests across shards.
+                                      [test (default), file]
```
lrhn: Any expectations of other granularities?
(What would that even be? "Directory", but most packages have all the tests in one directory anyway.)


Is there any reason to not always use `--shard=by=file`?
(If you have only one big file with many tests, I guess?)

pkgs/test/README.md
```diff
@@ -204,12 +204,35 @@ dart test --total-shards 3 --shard-index 0 path/to/test.dart
...
+### Interaction with Test Filters
+
+The sharding modes interact differently with filters like `--name` or `--tags`:
+
+* When sharding by `test`, the sharding partition is calculated *after*
+  applying filters. This guarantees that the matching tests are distributed
+  as evenly as possible across all shards.
+* When sharding by `file`, the files are partitioned *before* they are loaded
+  and filtered. If a filter only matches tests in a few files, some shards
+  might run no tests because those files were allocated to other shards.
```
lrhn: _Why_ not shard after filtering?

I'm assuming there is a reason, but it would be nice to know what it is.
I  guess each shard will individually do the same "find all tests" computation and sharding, but if  it can do that normally, it should be able to do it sharding by file as well. As long as all the shards agree on the test-set, they should agree after filtering too.

pkgs/test_core/lib/src/runner.dart
```diff
@@ -269,25 +270,76 @@ class Runner {
...
+      allFiles.sort();
+
+      final shardSize = allFiles.length / _config.totalShards!;
+      final shardStart = (shardSize * _config.shardIndex!).round();
+      final shardEnd = (shardSize * (_config.shardIndex! + 1)).round();
+      final shardFiles = allFiles.sublist(shardStart, shardEnd);
+
+      source = StreamGroup.merge(
+        shardFiles.map((path) {
+          final key = _config.testSelections.keys.firstWhere(
```
lrhn: (Can use `_config.testSelections.entries.firstWhere((e) => path == e.key || p.isWithin(e.key, path), ...`, then you don't need to do a `...[key]!` afterwards)

pkgs/test_core/lib/src/runner.dart
```diff
@@ -269,25 +270,76 @@ class Runner {
   /// Only tests that match [_config.patterns] will be included in the
   /// suites once they're loaded.
   Stream<LoadSuite> _loadSuites() {
-    return StreamGroup.merge(
-      _config.testSelections.entries.map((pathEntry) {
+    final Stream<LoadSuite> source;
+    if (_config.totalShards != null && _config.shardBy == 'file') {
+      final allFiles = <String>[];
+      for (var pathEntry in _config.testSelections.entries) {
```
lrhn: (Can use patter assignment:
```dart
for (var MapEntry(key: testPath, value: ...)) ...
```
but I'm not sure the value is actually used, so it could just iterate keys?)

pkgs/test/README.md
jakemac53: Some filters are runtime only, but we should be able to respect the top level `@TestOn` annotation filters without loading the test.

jakemac53:
Meta comment, but we call test "files" test "suites". So it might be best to make this `--shard-by=suite`. Although it may not be as intuitive for users, but it would match our general terminology.

I am curious what @natebosch thinks though.

pkgs/test/test/runner/shard_test.dart
```diff
@@ -172,6 +172,147 @@ void main() {
     await test.shouldExit(79);
```
jakemac53, line 172: Should add some edge cases, 0 test suites, less test suites than shards, more test suites than shards, etc.

jakemac53, line 172: I am also curious how this interacts with the `@TestOn` annotation as well as tags.

pkgs/test/test/utils.dart
```diff
@@ -232,6 +232,7 @@ Configuration configuration({
   int? concurrency,
   int? shardIndex,
   int? totalShards,
+  String? shardBy,
```
jakemac53: Lets make this an enum

pkgs/test_core/lib/src/runner.dart
```diff
@@ -269,25 +270,76 @@ class Runner {
   /// Only tests that match [_config.patterns] will be included in the
   /// suites once they're loaded.
   Stream<LoadSuite> _loadSuites() {
-    return StreamGroup.merge(
-      _config.testSelections.entries.map((pathEntry) {
+    final Stream<LoadSuite> source;
```
jakemac53: I would break this out into separate functions instead of inlining it all.

Or instead of making `shardBy` an enum make it a sealed class where the subclasses contain implementations so this is just `_config.shardBy.loadSuites(...)` or something.

pkgs/test_core/lib/src/runner.dart
```diff
@@ -269,25 +270,76 @@ class Runner {
...
         if (Directory(testPath).existsSync()) {
-          return _loader.loadDir(testPath, suiteConfig);
+          allFiles.addAll(
+            Directory(testPath)
+                .listSync(recursive: true)
+                .whereType<File>()
+                .map((f) => f.path)
+                .where((path) => _config.filename.matches(p.basename(path))),
+          );
         } else if (File(testPath).existsSync()) {
```
lrhn: This test does nothig, you add the `testPath` to `allFiles` whether it's true or false.

pkgs/test/test/runner/runner_test.dart
sigurdm: changed to --shard-by-file. I cannot come up with other granularities.

Yeah, I think if you have many slow tests in a single file, then you don't want this. But for most use cases I expect that --shard-by-file would be preferable.

pkgs/test/test/runner/shard_test.dart
sigurdm: Done

sigurdm: The file sorting happens before @testOn/tags are read from the files.

Added tests.

pkgs/test/test/utils.dart
sigurdm: Done

pkgs/test/README.md
sigurdm: For me I guess I think it is to have the simplest mental model of what happens. And the least preprocessing before starting running actual tests.

If we start looking into files before sharding it will be a big job just specifying which features to take into account and which not. (If we do top-level @TestOn, tags, it seems to me asymetrical that we don't look at the single-test annotations) We would also need to spin up the analyzer to parse these (I guess) subtracting from the efficiency of raw file splitting.

pkgs/test_core/lib/src/runner.dart
sigurdm: Done.

pkgs/test_core/lib/src/runner.dart
sigurdm: Split it out. I guess we will only ever have two choices here.

pkgs/test_core/lib/src/runner.dart
sigurdm: Fixed

pkgs/test_core/lib/src/runner.dart
sigurdm: Done

sigurdm:
> Meta comment, but we call test "files" test "suites". So it might be best to make this --shard-by=suite. Although it may not be as intuitive for users, but it would match our general terminology.

I think as this is raw sharding by file, I like to keep the terminology there. Also I don't really see a reason for package:test to invent its own name for a file. I added a parenthetical explanation in the help-string and readme for people who know what a suite is, but don't know what a file is.

jakemac53:
> I think as this is raw sharding by file, I like to keep the terminology there. Also I don't really see a reason for package:test to invent its own name for a file. I added a parenthetical explanation in the help-string and readme for people who know what a suite is, but don't know what a file is.

I don't disagree with you but the decision was made a long time ago to call test files suites. I couldn't tell you why. This does leak out to users via `--suite-load-timeout`, but I don't see a lot of other places. cc @natebosch for thoughts on this

natebosch:
Sorry for the slow response. I spent some time thinking through alternative sharding designs that can take into account at least some of the differences between test suites but I came up short. I think I lean towards landing something very close to this design. Some thoughts:

RE `--shard-by-file` vs `--shard-by-suite`

> I think as this is raw sharding by file

With the "file" naming, as a user I _might_ expect that a pattern like `all_tests_test.dart` which imports and runs the `main` of a bunch files with test cases would still shard based on which file the `test()` is actually written in. "Suite" might cause a reader to need to look for a definition of the term, but is unlikely to result in an incorrect interpretation like "file" might.

RE shard before or after reading top level `@TestOn` annotations:

> If we do top-level `@TestOn`, tags, it seems to me asymetrical that we don't look at the single-test annotations

It feels quite natural to me that we would honor suite-level configuration when we are doing suite-level sharding. We should expect that test suites which are entirely configured for one platform would do that at the suite level instead of the test or group level

The fact that we do sorting by file path could make a situation where a big run of tests for a single platform end up in a single shard that ends up doing no testing because they were all filtered out. We should maybe consider doing a stable shuffle then sublists, or sort and round-robin distribute suites to shards so that similarly named tests are more likely to end up on different shards.

pkgs/test/test/runner/runner_test.dart
```diff
@@ -163,10 +164,15 @@ $_usage''');
   group('fails gracefully if', () {
     test('an invalid option is passed', () async {
       var test = await runTest(['--asdf']);
-      expectStderrEquals(test, '''
-Could not find an option named "--asdf".
+      var expected1 = 'Could not find an option named "asdf".\n\n$_usage';
+      var expected2 = 'Could not find an option named "--asdf".\n\n$_usage';
```
natebosch: Is this related to the change or was it an existing problem?

pkgs/test/README.md
```diff
@@ -204,12 +204,37 @@ dart test --total-shards 3 --shard-index 0 path/to/test.dart
...
+By default, sharding is done by individual tests within each suite.
+You can shard by entire test files (suites) using the `--shard-by-file` flag.
+
+* When sharding by test (default): Distribute individual test cases across
+  shards. This balances the test load at the test case level. Test cases from
+  each suite are sliced continuously, which minimizes how often a suite is
+  split across shards and helps maximize the re-use of `setUpAll` and
+  `tearDownAll` setups.
+* When sharding by file (using `--shard-by-file`): Distribute entire test files (suites)
+  across shards. This can be faster for suites with many small files as it
```
natebosch: [nit] This usage of "suites" is wrong - replace with maybe "packages" or "test invocations"?

pkgs/test/README.md
```diff
@@ -204,12 +204,37 @@ dart test --total-shards 3 --shard-index 0 path/to/test.dart
...
+  avoids loading every file in every shard. Because test files are not split,
+  any `setUpAll` and `tearDownAll` setups in a file are guaranteed to run only
```
natebosch: Here is where a potential confusion can come in - it's not the most common situation, but it's entirely plausible that a `setUpAll` is defined in a shared place used from multiple `_test.dart` files. I think we need to be careful not to say something about "a `setUpAll` in a file" because that conveys very strongly a textual/source code relationship to me.

pkgs/test/README.md
```diff
@@ -204,12 +204,37 @@ dart test --total-shards 3 --shard-index 0 path/to/test.dart
...
+
+The sharding modes interact differently with filters like `--name` or `--tags`:
+
+* When sharding by test (default), the sharding partition is calculated *after*
+  applying filters. This guarantees that the matching tests are distributed
+  as evenly as possible across all shards.
+* When sharding by file (using `--shard-by-file`), the files are partitioned
+  *before* they are loaded and filtered. If a filter only matches tests in a
+  few files, some shards might run no tests because those files were allocated
+  to other shards.
```
natebosch: I think we should make top level test annotations work for this strategy.

I think the way we should talk about the test-case level filtering is a little bit more general. Something maybe like

> Test filters which apply to individual test cases or groups (as opposed to suite level annotations at the top of a file) will be reflected in the distribution when sharding by test case, but not when sharding by test suite. Test invocations with uneven number of tests cases to run between test suites can cause uneven workloads across the shards.

natebosch:
I had gemini take a try at handling the filtering before sharding. At a skim the solution it found looks fine to me - add a cache of the `MetaData` to the `Loader` so that we can read metadata earlier than we used to without doubling up. The `Runner` where the sharding is implemented already has a reference to the `Loader` so it's feasible to introduce suite filtering eagerly when using suite sharding.

sigurdm:
Fair, I can try to give it a go next week to have top-level filtering before splitting.

pkgs/test/test/runner/runner_test.dart
sigurdm: Unrelated. Removed

pkgs/test/test/runner/shard_test.dart
sigurdm: No longer true. Now suite-level annotations are processed before sharding.

sigurdm (commented):
Now suites are filtered by top-level annotations before being sharded.
The flag is now --shard-by-suite, and the phrasing has been updated everywhere.

pkgs/test/README.md
sigurdm: Fixed.

pkgs/test/README.md
sigurdm: Everything is phrased in terms of suites now.

jakemac53 (approved):
LGTM from me

pkgs/test_core/lib/src/runner/loader.dart
```diff
@@ -29,6 +31,84 @@ import 'vm/platform.dart';
...
+          File(path).readAsStringSync(),
+          _runtimeVariables.toSet(),
+        );
+      } catch (e) {
+        return e;
+      }
+    });
+    if (cached is Metadata) return cached;
+    throw cached;
+  }
```
natebosch, line 53: [nit] I prefer being a bit more explicit than `Metadata`/ not `Metadata`. WDYT about using `Result` from `package:async` (even though we're only using it in a synchronous way)

```suggestion
  /// Cache of metadata parsed from test files.
  final _metadataCache = <String, Result<Metadata>>{};

  /// Parses and returns the suite metadata for [path], caching the result.
  Metadata parseSuiteMetadata(String path) {
    final cached = _metadataCache.putIfAbsent(path, () {
      if (!File(path).existsSync()) return .value(Metadata.empty);
      try {
        return .value(
          parseMetadata(
            path,
            File(path).readAsStringSync(),
            _runtimeVariables.toSet(),
          ),
        );
      } catch (e, stackTrace) {
        return .error(e, stackTrace);
      }
    });
    return switch (cached) {
      ValueResult(:var value) => value,
      ErrorResult(:var error, :var stackTrace) => Error.throwWithStackTrace(
        error,
        stackTrace,
      ),
      _ => throw StateError('Unexpected result type: $cached'),
    };
  }
```

pkgs/test_core/lib/src/runner/loader.dart
```diff
@@ -29,6 +31,84 @@ import 'vm/platform.dart';
...
+
+  /// Returns whether the test suite at [path] matches target platform and tag filters
+  /// specified in [suiteConfig].
+  bool matchesSuite(String path, SuiteConfiguration suiteConfig) {
+    if (!File(path).existsSync()) return true;
```
natebosch, line 58: I would have expected `false`. Is this a code path we'll never hit in practice? If this branch does get executed I think we should describe it in the doc.

pkgs/test_core/lib/src/runner/loader.dart
```diff
@@ -29,6 +31,84 @@ import 'vm/platform.dart';
...
+
+    if (_config.includeTags != BooleanSelector.all &&
+        mergedConfig.metadata.tags.isNotEmpty) {
+      if (!_config.includeTags.evaluate(mergedConfig.metadata.tags.contains)) {
+        return false;
+      }
+    }
+
+    var hasMatchingPlatform = false;
+    for (var runtimeName in mergedConfig.runtimes) {
```
natebosch, line 82: I think this is somewhat duplicative. I think we could pull out a `List<SuitePlatform> _matchingPlatforms(SuiteConfiguration suiteConfig)` method if the `throw` behavior is OK for both.

Would it cause any problem if we use the exception behavior here instead of silently ignoring unknown platforms? I think we wouldn't expect them here any more than we would when loading would we?

natebosch (approved):

pkgs/test_core/lib/src/runner/loader.dart
sigurdm: If a file doesn't exist (e.g. a typo in a CLI path), `Runner` includes it in `allFiles` so that downstream loading can report a `LoadException('Does not exist.')`. Similarly, if a file has malformed metadata, returning `false` here would silently omit the file from sharding and hide the syntax error. By returning `true`, we ensure the file is assigned to a shard so `loadFile` can surface the failure properly.

I've updated the doc comment on `matchesSuite` to make this explicit.

pkgs/test_core/lib/src/runner/loader.dart
sigurdm: Good call. Runtimes come from CLI/runner configuration rather than suite metadata, so throwing early on an unknown runtime during sharding is completely fine and matches `loadFile`. Extracted `_matchingPlatforms`.

pkgs/test_core/lib/src/runner/loader.dart
sigurdm: Updated to use `Result<Metadata>` from `package:async`.

pkgs/test/README.md
sigurdm: Adopted in ac341e56.

PR #2660, opened 2026-06-01
[test] Add compiler `dart test -c cli` to run hooks

dcharkes:
Bug:

* https://github.com/dart-lang/sdk/issues/63372

This PR uses `dart build cli` under the `-c cli` flag.

Implementation:

* The wrapper Dart script around the test that package test creates is the new entry-point and **must** be within the package root to run the correct hooks.
  * So, this PR updates the temp dir to be in `.dart_tool/test/temp/<...>` rather than in the system temp.
* Gates the implementation on the next dev release after https://dart-review.googlesource.com/c/sdk/+/506242.

Testing:

* The integration tests run with the `cli` compiler
* The integration tests must run on a valid package with a valid pubspec and package_config.json (otherwise the hooks-runner cannot determine which hooks to run).
* The integration tests do _not_ actually run a hook and include c code. `dart build cli` bundles are already properly tested in the dartdev tests for `dart build cli` in the SDK. If you feel we should add an integration test running native code here, I'm happy to add one.
* Skips the tests before the next dev release.
  * **TODO**: Rerun tests on Tuesday after new dev release has come out.

### DIFF
--- a/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
+++ b/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
@@ -26,34 +26,8 @@ final bool supportsCliCompiler = () {
 }();
 
 void main() {
-  late String testPath;
-  late String testCorePath;
-  late String testApiPath;
-
   setUpAll(() async {
     await precompileTestExecutable();
-
-    testPath = await packageDir;
-    final testPathParts = p.split(testPath);
-    testCorePath = p.joinAll(
-      testPathParts.take(testPathParts.length - 1).followedBy(['test_core']),
-    );
-    testApiPath = p.joinAll(
-      testPathParts.take(testPathParts.length - 1).followedBy(['test_api']),
-    );
-  });
-
-  // The test sandbox must be set up as a valid Dart package (with its own
-  // pubspec.yaml, running `pub get` to generate a valid package_config.json,
-  // and putting the entrypoint under the correct package root). This is required
-  // for the `cli` compiler (`dart build cli`), which requires the entrypoint
-  // target to reside inside a package defined in the package config.
-  setUp(() async {
-    await d
-        .file('pubspec.yaml', _pubspec(testPath, testCorePath, testApiPath))
-        .create();
-
-    await (await runPub(['get'], workingDirectory: d.sandbox)).shouldExit(0);
   });
 
   for (var runtime in Runtime.builtIn) {
@@ -93,7 +67,7 @@ void main() {
         skip: skipReason,
         () {
           final testArgs = [
-            'test/test.dart',
+            'test.dart',
             '-p',
             runtime.identifier,
             '-c',
@@ -101,15 +75,8 @@ void main() {
           ];
 
           test('can run passing tests', () async {
-            await d.dir('test', [d.file('test.dart', _goodTest)]).create();
-            var test = await runTest(
-              testArgs,
-              packageConfig: p.join(
-                d.sandbox,
-                '.dart_tool/package_config.json',
-              ),
-              workingDirectory: d.sandbox,
-            );
+            await d.file('test.dart', _goodTest).create();
+            var test = await runTest(testArgs);
 
             expect(
               test.stdout,
@@ -119,17 +86,8 @@ void main() {
           });
 
           test('fails gracefully for invalid code', () async {
-            await d.dir('test', [
-              d.file('test.dart', _compileErrorTest),
-            ]).create();
-            var test = await runTest(
-              testArgs,
-              packageConfig: p.join(
-                d.sandbox,
-                '.dart_tool/package_config.json',
-              ),
-              workingDirectory: d.sandbox,
-            );
+            await d.file('test.dart', _compileErrorTest).create();
+            var test = await runTest(testArgs);
 
             expect(
               test.stdout,
@@ -143,22 +101,15 @@ void main() {
           });
 
           test('fails gracefully for test failures', () async {
-            await d.dir('test', [d.file('test.dart', _failingTest)]).create();
-            var test = await runTest(
-              testArgs,
-              packageConfig: p.join(
-                d.sandbox,
-                '.dart_tool/package_config.json',
-              ),
-              workingDirectory: d.sandbox,
-            );
+            await d.file('test.dart', _failingTest).create();
+            var test = await runTest(testArgs);
 
             expect(
               test.stdout,
               containsInOrder([
                 'Expected: <2>',
                 'Actual: <1>',
-                '${p.join('test', 'test.dart')} 5',
+                'test.dart 5',
                 '+0 -1: Some tests failed.',
               ]),
             );
@@ -167,37 +118,21 @@ void main() {
           });
 
           test('fails gracefully if a test file throws in main', () async {
-            await d.dir('test', [d.file('test.dart', _throwingTest)]).create();
-            var test = await runTest(
-              testArgs,
-              packageConfig: p.join(
-                d.sandbox,
-                '.dart_tool/package_config.json',
-              ),
-              workingDirectory: d.sandbox,
-            );
+            await d.file('test.dart', _throwingTest).create();
+            var test = await runTest(testArgs);
             expect(
               test.stdout,
               containsInOrder([
-                '-1: [${runtime.name}, ${compiler.name}] loading test/test.dart [E]',
-                'Failed to load "test/test.dart": oh no',
+                '-1: [${runtime.name}, ${compiler.name}] loading test.dart [E]',
+                'Failed to load "test.dart": oh no',
               ]),
             );
             await test.shouldExit(1);
           });
 
           test('captures prints', () async {
-            await d.dir('test', [
-              d.file('test.dart', _testWithPrints),
-            ]).create();
-            var test = await runTest(
-              [...testArgs, '-r', 'json'],
-              packageConfig: p.join(
-                d.sandbox,
-                '.dart_tool/package_config.json',
-              ),
-              workingDirectory: d.sandbox,
-            );
+            await d.file('test.dart', _testWithPrints).create();
+            var test = await runTest([...testArgs, '-r', 'json']);
 
             expect(
               test.stdout,
@@ -213,18 +148,8 @@ void main() {
             test(
               'forwards stdout/stderr',
               () async {
-                await d.dir('test', [
-                  d.file('test.dart', _testWithStdOutAndErr),
-                ]).create();
-                var test = await runTest(
-                  testArgs,
-                  reporter: 'silent',
-                  packageConfig: p.join(
-                    d.sandbox,
-                    '.dart_tool/package_config.json',
-                  ),
-                  workingDirectory: d.sandbox,
-                );
+                await d.file('test.dart', _testWithStdOutAndErr).create();
+                var test = await runTest(testArgs, reporter: 'silent');
 
                 expect(test.stdout, emitsThrough('hello'));
                 expect(test.stderr, emits('world'));
@@ -285,20 +210,3 @@ void main() async {
   stderr.writeln('world');
   test('success', () {});
 }''';
-
-String _pubspec(String testPath, String testCorePath, String testApiPath) =>
-    '''
-name: mypackage
-version: 1.0.0
-environment:
-  sdk: ^3.5.0
-dev_dependencies:
-  test: any
-dependency_overrides:
-  test:
-    path: $testPath
-  test_core:
-    path: $testCorePath
-  test_api:
-    path: $testApiPath
-''';
--- a/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
+++ b/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
@@ -8,15 +8,50 @@ library;
 
 import 'dart:io';
 
+import 'package:path/path.dart' as p;
+import 'package:pub_semver/pub_semver.dart';
 import 'package:test/test.dart';
 import 'package:test_api/backend.dart';
 import 'package:test_descriptor/test_descriptor.dart' as d;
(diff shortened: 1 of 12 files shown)


pkgs/test/test/runner/compiler_runtime_matrix_test.dart
```diff
@@ -8,15 +8,52 @@ library;
...
 import 'package:test_descriptor/test_descriptor.dart' as d;
 
 import '../io.dart';
 
+final bool supportsCliCompiler = () {
+  try {
+    var current = Version.parse(Platform.version.split(' ').first);
+    return current > Version.parse('3.13.0-139.0.dev');
+  } catch (_) {
+    return false;
```
jakemac53: Do we need this try/catch? We shouldn't really (given this is only a test, and it would be good to know if this started failing)?

pkgs/test/test/runner/compiler_runtime_matrix_test.dart
```diff
@@ -8,15 +8,52 @@ library;
...
     await precompileTestExecutable();
+
+    testPath = await packageDir;
+    final testPathParts = p.split(testPath);
+    testCorePath = p.joinAll(
+      testPathParts.take(testPathParts.length - 1).followedBy(['test_core']),
+    );
+    testApiPath = p.joinAll(
+      testPathParts.take(testPathParts.length - 1).followedBy(['test_api']),
+    );
```
jakemac53:
```suggestion
    testCorePath = p.join(p.dirname(testPath), 'test_core');
    testApiPath = p.join(p.dirname(testPath), 'test_api');
```

pkgs/test/test/runner/compiler_runtime_matrix_test.dart
```diff
@@ -8,15 +8,52 @@ library;
...
+  });
+
+  // The test sandbox must be set up as a valid Dart package (with its own
+  // pubspec.yaml, running `pub get` to generate a valid package_config.json,
+  // and putting the entrypoint under the correct package root). This is required
+  // for the `cli` compiler (`dart build cli`), which requires the entrypoint
+  // target to reside inside a package defined in the package config.
+  setUp(() async {
```
jakemac53: Lets do this stuff lower down and only for the `cli` compiler? It's a pretty expensive operation requiring the network because of the pub get.

pkgs/test/test/runner/subprocess_crash_test.dart
```diff
@@ -5,35 +5,106 @@
...
 
+import 'dart:io';
+import 'package:path/path.dart' as p;
+import 'package:pub_semver/pub_semver.dart';
 import 'package:test/test.dart';
 import 'package:test_descriptor/test_descriptor.dart' as d;
 
 import '../io.dart';
 
+final bool supportsCliCompiler = () {
```
jakemac53: Lets move this to a util

pkgs/test/test/runner/subprocess_crash_test.dart
```diff
@@ -5,35 +5,106 @@
...
+    await precompileTestExecutable();
+
+    testPath = await packageDir;
+    final testPathParts = p.split(testPath);
+    testCorePath = p.joinAll(
+      testPathParts.take(testPathParts.length - 1).followedBy(['test_core']),
+    );
+    testApiPath = p.joinAll(
+      testPathParts.take(testPathParts.length - 1).followedBy(['test_api']),
     );
```
jakemac53:
```suggestion
    testCorePath = p.join(p.dirname(testPath), 'test_core');
    testApiPath = p.join(p.dirname(testPath), 'test_api');
```

pkgs/test/test/runner/subprocess_crash_test.dart
```diff
@@ -5,35 +5,106 @@
...
-    await test.shouldExit(1);
   });
+
+  // The test sandbox must be set up as a valid Dart package (with its own
+  // pubspec.yaml, running `pub get` to generate a valid package_config.json,
+  // and putting the entrypoint under the correct package root). This is required
+  // for the `cli` compiler (`dart build cli`), which requires the entrypoint
+  // target to reside inside a package defined in the package config.
+  setUp(() async {
```
jakemac53: Same comment as above

pkgs/test/test/runner/subprocess_crash_test.dart
```diff
@@ -5,35 +5,106 @@
...
+        await test.shouldExit(1);
+      },
+      skip: compiler == 'cli' && !supportsCliCompiler
+          ? 'Dart version does not support build cli'
+          : null,
+    );
+  }
 }
+
+String _pubspec(String testPath, String testCorePath, String testApiPath) =>
```
jakemac53: This should also move out to a util, and possibly encapsulate the logic for computing all the paths in that so it just takes zero args.

pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -41,7 +41,20 @@ class VMPlatform extends PlatformPlugin {
     p.join(p.current, '.dart_tool', 'test', 'incremental_kernel'),
   );
   final _closeMemo = AsyncMemoizer<void>();
-  final _tempDir = Directory.systemTemp.createTempSync('dart_test.vm.');
+  // The temporary directory must be located within the package workspace
```
jakemac53: This seems to be causing problems for both package:test tests and also the actual runtime here - I am generally quite uncomfortable with this change actually because this directory will not get automatically cleaned up by the OS and people do force quite the test runner often.

Can we fix `dart build cli` so that it can pass a package config explicitly when invoking the entry point?

jakemac53 (commented):
See the note about not using a temp dir any more, I would like to see an upstream change in `dart build cli` to make using real temp directories possible to remove this logic from both the tests and real code.

If that is not possible, then I think we need some reliable form of persistent cleanup here - as in the test runner doing its own periodic cleanup of old directories. But I much prefer `dart build cli` being able to build scripts anywhere and just pass the package config path, which is consistent with all the other compilers.

dcharkes:
> If that is not possible, then I think we need some reliable form of persistent cleanup here - as in the test runner doing its own periodic cleanup of old directories. But I much prefer `dart build cli` being able to build scripts anywhere and just pass the package config path, which is consistent with all the other compilers.

The problem with a script outside a package is that we cannot infer for what packages a hook should be run.

We could theoretically add "pretend this file was part of package 'foo'" as cli parameter, but that feels hacky. E.g. you pass in a `package_config.json` but then pass in a target entry point that isn't governed by the package config. I guess it's kinda fine, because the package config is only used for resolving imports? cc @bkonyi WDYT?

(A workaround to say make a _new_ package in temp that then imports the real package doesn't work, because we need the dev dependencies of the real package, and we don't include dev dependencies of the non-root packages.)

jakemac53:
> The problem with a script outside a package is that we cannot infer for what packages a hook should be run.

Can we just use the current working dir for this?

jakemac53:
Or possibly, infer the package based on the package config location?

dcharkes:
> > The problem with a script outside a package is that we cannot infer for what packages a hook should be run.
> 
> Can we just use the current working dir for this?

You can do `dart run foo:bar` which does not include package `baz` as the current directory. You need to run the hooks from package `foo` and its dependencies for that.

> Or possibly, infer the package based on the package config location?

With pub workspaces that will be the pub workspace directory, not a package root.

We'll need to pass in the "current package" or infer it from the entry point.

@bkonyi how do you feel about `dart build cli --target=$TMP/bar/baz.dart  --packages=$TMP/foo/package_config.json --run-package-name=bla`? (we use [runPackageName](https://pub.dev/documentation/hooks_runner/latest/hooks_runner/PackageLayout/runPackageName.html) in the hooks_runner public API)

jakemac53:
> You can do `dart run foo:bar` which does not include package `baz` as the current directory. You need to run the hooks from package `foo` and its dependencies for that.

This seems like more of a special case where the package name is explicit and so it shouldn't fall back on the current working dir.

But that probably would require a flag like `--root-package` or something to `dart build cli` anyways, which we could also use to resolve this issue. So maybe just adding that flag is the best option?

dcharkes:
> > You can do `dart run foo:bar` which does not include package `baz` as the current directory. You need to run the hooks from package `foo` and its dependencies for that.
> 
> This seems like more of a special case where the package name is explicit and so it shouldn't fall back on the current working dir.

Yeah I guess you're right, that if we do `dart run foo:bar`, it's already using `foo` explicitly. So we could fall back to a cwd approach if the target is not in a package, but it's a very implicit behavior. So adding an extra param is probably cleaner. Yeah `--root-package` or `--hooks-root-package` or `--hooks-root` is a better name than `--run-package-name` (that name refers to the package named after "dart run").

pkgs/test_core/lib/src/runner/vm/platform.dart
dcharkes: https://dart-review.googlesource.com/c/sdk/+/508122

jakemac53 (approved):

dcharkes:
> * **TODO**: Rerun tests on Tuesday after new dev release has come out.

3.13.0-155.0.dev does not yet contain https://dart-review.googlesource.com/c/sdk/+/508122. So waiting till Thursday to get the correct dart dev version for the PR and doing a CI run with a dev version that doesn't skip the test.

dcharkes:
```
test\runner\configuration\top_level_test.dart: uses the specified suite load timeout (failed)
  
      00:00 +0: loading test.dart
      00:00 +0 -1: loading test.dart [E]
        TimeoutException after 0:00:01.000000: Test timed out after 1 seconds.
        package:test_api/src/backend/invoker.dart 398:28  Invoker._handleError.<fn>
        dart:async/zone.dart 1517:47                      _rootRun
        dart:async/zone.dart 1422:19                      _CustomZone.run
        package:test_api/src/backend/invoker.dart 396:10  Invoker._handleError
        package:test_api/src/backend/invoker.dart 351:9   Invoker.heartbeat.<fn>.<fn>
        dart:async/zone.dart 1525:13                      _rootRun
        dart:async/zone.dart 1422:19                      _CustomZone.run
        package:test_api/src/backend/invoker.dart 350:38  Invoker.heartbeat.<fn>
        dart:async-patch/timer_patch.dart 18:15           Timer._createTimer.<fn>
        dart:isolate-patch/timer_impl.dart 423:19         _Timer._runTimers
        dart:isolate-patch/timer_impl.dart 454:5          _Timer._handleMessage
        dart:isolate-patch/isolate_patch.dart 193:12      _RawReceivePort._handleMessage
        
      00:00 +0 -1: Some tests failed.
      
      Failing tests:
        test.dart: loading test.dart
  package:test_api/src/backend/invoker.dart 398:28  Invoker._handleError.<fn>
  dart:async                                        _CustomZone.run
  package:test_api/src/backend/invoker.dart 396:10  Invoker._handleError
  package:test_api/src/backend/invoker.dart 351:9   Invoker.heartbeat.<fn>.<fn>
  dart:async                                        _CustomZone.run
  package:test_api/src/backend/invoker.dart 350:38  Invoker.heartbeat.<fn>
  PathAccessException: Deletion failed, path = 'C:\Users\runneradmin\AppData\Local\Temp\dart_test_a4aa779a' (OS Error: The process cannot access the file because it is being used by another process.
  , errno = 32)
  package:test_api/src/backend/invoker.dart 299:9  Invoker.runTearDowns.<fn>
  package:test_api/src/backend/invoker.dart 318:9  Invoker._waitForOutstandingCallbacks.<fn>
❌ test\runner\configuration\top_level_test.dart: uses the specified suite load timeout (failed after test completion)
  Expected: <1>
    Actual: <-1>
  Process `dart bin/test.dart` had an unexpected exit code.
  
  package:matcher/src/expect/expect.dart 187:31        fail
  package:matcher/src/expect/expect.dart 182:3         _expect
  package:matcher/src/expect/expect.dart 65:3          expect
  test\runner\configuration\top_level_test.dart 476:5  main.<fn>
  package:test_api/src/backend/declarer.dart 253:15    Declarer.test.<fn>.<fn>
  package:test_api/src/backend/declarer.dart 250:11    Declarer.test.<fn>
  package:test_api/src/backend/invoker.dart 318:9      Invoker._waitForOutstandingCallbacks.<fn>
```

This failure seems to consistently repro on this Dart 3.10 bot, but not on the dev release branch.

Also, I can't seem to the test locally on a cloudtop:

```
                 *   Failed to load "test.dart": Chrome exited before connecting.
```

@jakemac53 On previous iterations of this PR I just re-hit the bot and it turned green. Are the tests flaky?

jakemac53:
> @jakemac53 On previous iterations of this PR I just re-hit the bot and it turned green. Are the tests flaky?

I have definitely seen windows flakes in the past but you should be able to get a green run. I am retrying it now.

dcharkes:
> > @jakemac53 On previous iterations of this PR I just re-hit the bot and it turned green. Are the tests flaky?
> 
> I have definitely seen windows flakes in the past but you should be able to get a green run. I am retrying it now.

I also tried 3 times. It keeps being red. The weird thing is that the other bots are green. So it must be something old-Dart + Windows specific?

Edit: We're not running the `dart build cli` by default anywhere. So we should be regressing any existing tests right?

jakemac53:
While I don't see what your PR could have had to do with the failure, I also don't see any recent flakes for this test so it does feel somehow related....

jakemac53:
cc @natebosch

dcharkes:
> I have definitely seen windows flakes in the past but you should be able to get a green run. I am retrying it now.

Maybe you have the magic touch, it went green. 😄

jakemac53:
> Maybe you have the magic touch, it went green. 😄

Haha ok, weird that it took so many tries. I think Nate has been trying to work on some of these flakes. The windows ones are often very hard to track down though.

PR #2662, opened 2026-06-01
Handle objects with empty toString

natebosch:
The approach to building text descriptions involves appending and
prepending text to string iterables. If the "base" of the string
description is an empty iterable it propagates through as empty, which
can cause missing lines from the failure output.

Add a fallback of `'empty toString()'` in the case any object returns a
completely empty string representation.


natebosch:
I noticed this when trying the [migration of `package:package_config` to checks](https://github.com/dart-lang/tools/pull/2415), it's possible to construct a `LanguageVersion` instance where to `.toString()` is empty.

kevmoo (approved):
Although CI is sad. 🤷

PR #2663, opened 2026-06-01
Throw when an async matcher is used with isNot

### LINKED ISSUE #1637: isNot matcher is not negating the result of a wraped matcher
This issue was originally reported on flutter/flutter as https://github.com/flutter/flutter/issues/94170

## Steps to Reproduce

1. Use this test 
<details>

pubspec.yaml as 
```yaml
dev_dependencies:
  flutter_test:
    sdk: flutter
```
``` dart
main(){
  test("given a async closure that dosnt throw, then it shouldnt throw", () async {
    await expectLater(() => Future.value("this dosnt throw and just returns"), isNot(throwsA(anything)));
  });

  test("samezys just with expect", () {
    expect(() => Future.value("same"), isNot(throwsA(anything)));
  });

  test("given sync funcs, the test should detect throwing", () {
    expect(() => "test", isNot(throwsA(anything)));
    expect(() => throw Error(), throwsA(anything));
  });

  test("given a throwing future, then the closure should throw", (){
    expect(() => Future.error(Error()), throwsA(anything));
  });
}
```

</details>

2. Execute `flutter test` on the code appended. 
3. I would expect that both tests should satisfy this criterion.

#### What actually happens

**Expected results:** 
I expect that all test cases finish without error.

**Actual results:** 
Both async isNot tests fail with equivalent error messages.

```
  Expected: not throws anything
    Actual: <Closure: () => Future<String>>

  package:test_api                                    expectLater
  package:flutter_test/src/widget_tester.dart 498:10  expectLater
  test\example.dart 60:15                     main.<fn>.<fn>.<fn>
  test\example.dart 58:41                     main.<fn>.<fn>.<fn>
```

<details>
<summary>flutter doctor -v</summary>
```
PS C:\Users\user\Documents\Developer\medlog> flutter doctor -v
[√] Flutter (Channel stable, 2.5.3, on Microsoft Windows [Version 10.0.22000.282], locale en-US)
    • Flutter version 2.5.3 at C:\Users\user\Documents\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 18116933e7 (6 weeks ago), 2021-10-15 10:46:35 -0700
    • Engine revision d3ea636dc5
    • Dart version 2.14.4

[√] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Android SDK at C:\Users\user\AppData\Local\Android\Sdk
    • Platform android-31, build-tools 31.0.0
    • ANDROID_SDK_ROOT = C:\Users\user\AppData\Local\Android\Sdk
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)
    • All Android licenses accepted.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[√] Android Studio (version 2020.3)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
       https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
       https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)

[√] IntelliJ IDEA Ultimate Edition (version 2020.3)
    • IntelliJ at C:\Program Files\JetBrains\IntelliJ IDEA 2020.3
    • Flutter plugin can be installed from:
       https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
       https://plugins.jetbrains.com/plugin/6351-dart

[√] VS Code (version 1.62.2)
    • VS Code at C:\Users\user\AppData\Local\Programs\Microsoft VS Code
    • Flutter extension version 3.28.0

[√] Connected device (1 available)
    • Chrome (web) • chrome • web-javascript • Google Chrome 95.0.4638.69

• No issues found!
```

</details>

natebosch:
Closes #1637

Async matchers are implemented as a hack where they usually appear to
succeed synchronously, and may report an error for the test case later.
This is incompatible with meta matchers or "operator" matchers `isNot`
and `anyOf` where the synchronous answer is trusted and other values are
derived from it. In the case of `isNot` this causes false positives, and
for `anyOf` it causes false negatives.

Historically it was infeasible to add knowledg about `AsyncMatcher` in
the definition of these operators, because async matching was
implemented in the test runner directly and the matchers exposed by the
test runner are a superset of those defined by the matcher package. Now
that `expect` and the async matchers are definied in `package:matcher`
it is possible to detect when an async matcher is used this way.

There is one async matcher when does often serve as a synchronous
matcher in practice - the `throwsA` matcher is used for both synchronous
and asynchronous errors, the same matcher is uesed for `Function()`,
`Future Function()` and `Future` values. I think the combination of
`throwsA` and `isNot` or `anyOf` is likely to be rare, but if errors
surface in practice we may need to add an exception for the `Throws`
matcher.

### DIFF
--- a/pkgs/matcher/CHANGELOG.md
+++ b/pkgs/matcher/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.12.21-wip
+
+* Throw an `ArgumentError` when `isNot` or `anyOf` is used with an `AsyncMatcher`.
+
 ## 0.12.20
 
 * Allow exceptions from `operator ==` to bubble up and fail the test instead of
--- a/pkgs/matcher/lib/src/operator_matchers.dart
+++ b/pkgs/matcher/lib/src/operator_matchers.dart
@@ -2,11 +2,18 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'expect/async_matcher.dart';
 import 'interfaces.dart';
 import 'util.dart';
 
 /// Returns a matcher that inverts [valueOrMatcher] to its logical negation.
-Matcher isNot(Object? valueOrMatcher) => _IsNot(wrapMatcher(valueOrMatcher));
+Matcher isNot(Object? valueOrMatcher) {
+  final matcher = wrapMatcher(valueOrMatcher);
+  if (matcher is AsyncMatcher) {
+    throw ArgumentError('isNot cannot be used with AsyncMatchers');
+  }
+  return _IsNot(matcher);
+}
 
 class _IsNot extends Matcher {
   final Matcher _matcher;
@@ -108,6 +115,9 @@ class _AnyOf extends Matcher {
   @override
   bool matches(dynamic item, Map matchState) {
     for (var matcher in _matchers) {
+      if (matcher is AsyncMatcher) {
+        throw ArgumentError('anyOf cannot be used with AsyncMatchers');
+      }
       if (matcher.matches(item, matchState)) {
         return true;
       }
--- a/pkgs/matcher/pubspec.yaml
+++ b/pkgs/matcher/pubspec.yaml
@@ -1,5 +1,5 @@
 name: matcher
-version: 0.12.20
+version: 0.12.21-wip
 description: >-
   Support for specifying test expectations via an extensible Matcher class.
   Also includes a number of built-in Matcher implementations for common cases.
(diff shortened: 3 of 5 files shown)


natebosch:
This is still a tricky import to add internally because `AsyncMatcher` uses `fail` from `package:test_api/hooks.dart` and internally that library is `testonly`, but there are some non-testonly uses of matcher.

PR #2664, opened 2026-06-03
Resolve process leak in test compiler

natebosch:
Check for early closure after async operations and resolve a deadlock
during close where the frontend server is busy and is holding a pool
resource during shutdown.

Add a mechanism for injecting a fake `FrontendServerClient` to allow
controlling timing during a test without arbitrary delays trying line up
with the external process.


natebosch:
I think this is contributing to the CI flakes, but I don't think it's the only problem.

jakemac53:
I am curious what the situation is here that actually causes `dispose` to get called during an active compile, especially on CI?

jakemac53 (approved):

natebosch:
> I am curious what the situation is here that actually causes `dispose` to get called during an active compile, especially on CI?

It's the new suite load timeout feature - with a very short load timeout it can abort during the compile.

PR #2666, opened 2026-06-03
Disable chrome throttling options

natebosch:
I don't expect this to have an effect in typical tests, but in case a
test opens a new tab or similar it's a good idea to disable throttling
when the browser is controlled by automation.


kevmoo (approved):
Great find!

PR #2670, opened 2026-06-11
Support extension types in checks codegen

natebosch:
The element model normalizes extension types to their representation
type, so we need to use the AST to read field names and types for
extension types.

Initial vibe-coded approach - will need review/refactor before landing.


natebosch:
@scheglov - is there any way to find the interface of an extension type using the element model, or is the AST requirement fundamental for this feature?

scheglov:
Can you describe the problem that you are trying to solve?
Theoretically `InterfaceType.element` already can be `ExtensionTypeElement`.

PR #2677, opened 2026-06-22
[test] More `dart test -c cli` fixes

dcharkes:
More fixes to make `dart --enable-experiment=record-use test -c cli` work:

* Fix forwarding experiment flags to the Dart compiler.
* Fix concurrency issues:
  * Don't override compiled app bundles, dir otherwise only `-j 1` succeeds.
  * https://dart-review.googlesource.com/c/sdk/+/515600/3/pkg/dartdev/lib/src/commands/build.dart also for `-j` more than 1.
* Fix pub workspace behavior: Specify the correct root package when running tests in workspaces.
* Fix windows (the bundle executable ends with `.exe`).

After this PR and https://dart-review.googlesource.com/c/sdk/+/515600, the test project in https://github.com/dart-lang/native/pull/3424 works as expected.

~~A release to pub.dev. Should it be 1.33 instead because we add a new feature? Any in flight changes that block a release @natebosch @jakemac53 ?~~ Edit: First landing the changes before bumping the versions.

### DIFF
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -25,6 +25,7 @@ import '../../runner/plugin/platform_helpers.dart';
 import '../../runner/plugin/shared_platform_helpers.dart';
 import '../../runner/runner_suite.dart';
 import '../../runner/suite.dart';
+import '../../util/dart.dart';
 import '../../util/io.dart';
 import '../../util/package_config.dart';
 import '../package_version.dart';
@@ -281,6 +282,8 @@ class VMPlatform extends PlatformPlugin {
 
     var outputDir = p.join(_tempDir.path, 'cli_build');
     var processResult = await Process.run(Platform.resolvedExecutable, [
+      for (var experiment in enabledExperiments)
+        '--enable-experiment=$experiment',
       'build',
       'cli',
       '--target',
@@ -334,6 +337,8 @@ stderr: ${processResult.stderr}''');
     );
     var output = File(p.setExtension(bootstrapPath, '.so'));
     var processResult = await Process.run(Platform.resolvedExecutable, [
+      for (var experiment in enabledExperiments)
+        '--enable-experiment=$experiment',
       'compile',
       'aot-snapshot',
       bootstrapPath,
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -280,7 +280,12 @@ class VMPlatform extends PlatformPlugin {
           await rootPackageLanguageVersionComment,
     );
 
-    var outputDir = p.join(_tempDir.path, 'cli_build');
+    // Unique output per path to avoid overriding in concurrent tests.
+    var outputDir = p.join(
+      _tempDir.path,
+      'cli_build',
+      p.withoutExtension(path),
+    );
     var processResult = await Process.run(Platform.resolvedExecutable, [
       for (var experiment in enabledExperiments)
         '--enable-experiment=$experiment',
--- a/pkgs/test_core/lib/src/runner/vm/platform.dart
+++ b/pkgs/test_core/lib/src/runner/vm/platform.dart
@@ -286,6 +286,12 @@ class VMPlatform extends PlatformPlugin {
       'cli_build',
       p.withoutExtension(path),
     );
+    // Find the package the test belongs to. If the test is outside the package
+    // config, fall back to "current package" which might be a workspace. In
+    // this case no hooks are run. Alternatively, this should be a failure.
+    final rootPackage =
+        (await currentPackageConfig).packageOf(await absoluteUri(path))?.name ??
+        (await currentPackage).name;
     var processResult = await Process.run(Platform.resolvedExecutable, [
       for (var experiment in enabledExperiments)
         '--enable-experiment=$experiment',
@@ -298,7 +304,7 @@ class VMPlatform extends PlatformPlugin {
       '--packages',
       (await packageConfigUri).toFilePath(),
       '--root-package',
-      (await currentPackage).name,
+      rootPackage,
       if (platform.runtime == Runtime.vmAsan) '--target-sanitizer=asan',
       if (platform.runtime == Runtime.vmMsan) '--target-sanitizer=msan',
       if (platform.runtime == Runtime.vmTsan) '--target-sanitizer=tsan',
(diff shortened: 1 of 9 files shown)


dcharkes:
The test failures are also happing on main (https://github.com/dart-lang/test/commit/1020c783adbebea6dc64627e2418cc6f7576fde6).

pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -279,8 +280,21 @@ class VMPlatform extends PlatformPlugin {
...
+    // Unique output per path to avoid overriding in concurrent tests.
+    var outputDir = p.join(
+      _tempDir.path,
+      'cli_build',
+      p.withoutExtension(path),
+    );
+    // Find the package the test belongs to. If the test is outside the package
+    // config, fall back to "current package" which might be a workspace. In
+    // this case no hooks are run. Alternatively, this should be a failure.
+    final rootPackage =
```
jakemac53, line 292: I am not quite understanding what the difference is here between `currentPackage` and `(await currentPackageConfig).packageOf(await absoluteUri(path))`.

Is there some broken logic for `currentPackage` where its not returning the actual current package for workspaces?

jakemac53: Ok yeah I took a look and that seems to be the case. We should instead just fix it. Possibly it can just do something like `(await currentPackageConfig).packageOf(Directory.current.absolute.uri)`?

dcharkes: The package config is the directory of the pub workspace when workspaces are used. So it's not returning the package of the test entry point:

https://github.com/dart-lang/test/blob/1020c783adbebea6dc64627e2418cc6f7576fde6/pkgs/test_core/lib/src/util/package_config.dart#L43-L46

The name `currentPackage` is misleading.

(AFAIK you can run tests from multiple packages at the same time via `dart test pkg1/test/ pkg2/test/`, so we can't have a _single_ `currentPackage` that is a global var.)

jakemac53: It looks like we might need a separate `workspaceRootPackage` getter as well

dcharkes: But in this case we're actually interested in the correct root package for each test. If you run test from multiple packages, we should pass for each test the correct "rootPackage". Passing the correct root package is necessary for running the correct build hooks.

jakemac53:
> The test failures are also happing on main ([1020c78](https://github.com/dart-lang/test/commit/1020c783adbebea6dc64627e2418cc6f7576fde6)).

Working on some fixes here https://github.com/dart-lang/test/pull/2678

pkgs/test_core/lib/src/util/package_config.dart
```diff
@@ -40,11 +40,16 @@ Future<Uri> absoluteUri(String path) async {
   }
 }
 
-/// Returns the current package.
-final Future<Package> currentPackage = () async {
+/// Returns the current workspace root package.
+final Future<Package> currentWorkspaceRootPackage = () async {
```
natebosch: If we're renaming anyway I think we can pick a better name. The "current" word wasn't all that useful to start. The trailing "Package" doesn't seem necessary either - it's a `Package` by type, but not a "package" in the way that we tend to think of them in the Dart ecosystem. Generally we don't need to repeat information about the type in the name, so I think the name `workspaceRoot` is nicer.


```suggestion
final Future<Package> workspaceRoot = () async {
```

pkgs/test_core/lib/src/util/package_config.dart
```diff
@@ -40,11 +40,16 @@ Future<Uri> absoluteUri(String path) async {
...
   return (await currentPackageConfig).packageOf(await packageConfigUri)!;
 }();
 
+/// Returns the package of [path].
+Future<Package?> packageOf(String path) async {
```
natebosch: [nit] Avoid redundant docs - expand this if we can add details, otherwise remove it.


```suggestion
Future<Package?> packageOf(String path) async {
```

pkgs/test_core/lib/src/runner/vm/platform.dart
natebosch: What is the "root" part of the name conveying? What is the effect in practice of the `--root-package` flag in the command below?  I'm worried about conflating "package root" (a concept and established naming in `package_config`) and "root package" (a concept which is slightly ambiguous, has meaning in different dev tooling packages that largely predates pub workspaces, and here seems to also get a new meaning in the specific context of "workspaceRootPackage").

If the effect in practice of the `--root-package` CLI argument is "the build hooks configured by this package are run" I'd like to call this variable `final runBuildHooksForPackage` or similar.

dcharkes: It's the root used for following the dependency graph. So all hooks of packages reachable via dependencies are run. So `runBuildHooksForPackage` doesn't work, it's all reachable packages from the "root".

> workspaceRootPackage

It has nothing to do with a workspace root "package". So, another overload of the term. 🙈 

Also, we already decided on `--root-package` after some rounds on the PR for the `dart build cli` args (it used to be "run package"). I think we should be consistent here. If you feel really strong about it I can change it, but we don't want to change the flag for `dart build cli` now, so it would be inconsistent then.

jakemac53 (approved):

pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -279,8 +280,20 @@ class VMPlatform extends PlatformPlugin {
           await rootPackageLanguageVersionComment,
     );
 
-    var outputDir = p.join(_tempDir.path, 'cli_build');
+    // Unique output per path to avoid overriding in concurrent tests.
+    var outputDir = p.join(
+      _tempDir.path,
+      'cli_build',
+      p.withoutExtension(path),
+    );
```
natebosch, line 288: @jakemac53 - do you think we should expect problems with long paths on windows here? I'm trying to recall whether we needed to do something more careful in build runner when we were extending file paths with a preceding temp directory.

natebosch (approved):

pkgs/test_core/lib/src/runner/vm/platform.dart
jakemac53: I don't think we have to worry about it, these paths are going to be shorter overall the the actual path to the real test in most situations (the temp dir will be higher up the tree probably).

If this becomes an issue, we can address it at that time.

PR #2678, opened 2026-06-22
fix package:test CI jobs

jakemac53:
Fix windows edge test expectations.

Skip Node+WASM tests and link issue https://github.com/dart-lang/test/issues/2679

### DIFF
--- a/pkgs/test/test/io.dart
+++ b/pkgs/test/test/io.dart
@@ -33,7 +33,7 @@ final bool supportsCliCompiler = () {
 
 /// The platform-specific message emitted when a nonexistent file is loaded.
 final String noSuchFileMessage = Platform.isWindows
-    ? 'The system cannot find the file specified.'
+    ? 'The system cannot find the file specified'
     : 'No such file or directory';
 
 /// An operating system name that's different than the current operating system.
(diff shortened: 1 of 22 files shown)


jakemac53:
cc @biggs0125 @nshahan  any ideas on the wasm stuff? I don't know much about it.

Previous failure: https://github.com/dart-lang/test/actions/runs/27837205323/job/82388037780#step:6:133
Current failure: https://github.com/dart-lang/test/actions/runs/27963306665/job/82750618168?pr=2678#step:6:143

pkgs/test/test/io.dart
```diff
@@ -33,7 +33,7 @@ final bool supportsCliCompiler = () {
 
 /// The platform-specific message emitted when a nonexistent file is loaded.
 final String noSuchFileMessage = Platform.isWindows
-    ? 'The system cannot find the file specified.'
+    ? 'The system cannot find the file specified'
```
natebosch, line 36: I don't have any problems landing this - but is it related to the failures?

natebosch (commented):
Can you update the first comment with the expected commit message? I don't think it matches the diff currently.

Edit nevermind I see how it relates.

pkgs/test/test/io.dart
natebosch: Oh this is in a different place than I thought, I see how it relates.

natebosch (approved):

PR #2680, opened 2026-06-22
Disable more backgrounding features on chromium

natebosch:
Add more flags to disable backgrounding when launching chromium based
browsers. These came up in an internal CL but should be helpful to
include externally.


kevmoo (approved):
Yes, please!

PR #2683, opened 2026-06-23
Allow similar use between sync and async checks

### LINKED ISSUE #2695: Allow multiple patterns for nested checks to allow higher consistency between synchronous and asynchronous checks
There are currently 2 styles of writing tests which check "nested" conditions - conditions that are checked against a value derived from the original subject. Synchronous checks nest with method chaining like `check(foo).isNotNull().isGreaterThan(1)`, while asynchronous checks nest with `Condition` callbacks `await check(foo).completes((it) => it.isGreaterThan(1));`. The reason for the divergence is the awkward syntax needed to handle the `await` for nesting `await (await check(foo).completes()).isGreaterThan(1);`

Some authors may have a preference to match the styles. We could normalize these and allow either style - synchronous nested checks with `Condition` arguments and asynchronous nested checks that return `Future<Subject>`. The reason not to do this is it violates the preference for having one way of accomplishing goals and forcing a choice onto every author. If we don't do this there will be authors who are disappointed not to be able to use their preferred style.

I have a PR which does this #2683 - but I haven't asked folks to decide their preference because I was pushing on a PR for #2694 too which is worth considering together.

natebosch:
Closes #2695

Originally async expectations which extract a value also returned a
`Subject` like synchronous extensions still do, but they were changed to
accept `AsyncCondition` callback arguments for ergonomics. Add support
for synchronous `Condition` callbacks on a few expectations that return
subjects for authors who prefer to use the async style for synchronous
checks too.

Affected APIs:
- `Subject.isA`
- `Subject.isNotNull`
- `Subject<T Function()>.throws`
- `Subject<T Function()>.returnsNormally`

Restore the `Future<Subject>` return value from async check extensions
that accept an optional `AsyncCondition` argument. This allows usage of
other patterns which some authors may feel are better aligned with the
sync checks of the original pattern.

Affected APIs:
- `Subject<Future>.completes`
- `Subject<Future>.throws`
- `Subject<StreamQueue>.emits`
- `Subject<StreamQueue>.emitsError`

Refactor tests into finer grained cases and add cases for the new
arguments.


natebosch:
The ergonomics of this is also impacted by https://github.com/dart-lang/test/pull/2697 - I'm more inclined to land this one if we'll also be landing the other (or if I rework it to land first). I implemented #2697 on top of this but it could also be landed separately.

pkgs/checks/test/extensions/async_test.dart
```diff
@@ -83,6 +86,11 @@ void main() {
           (it) => it.deepEquals(['  completes to an error of type StateError']),
         );
       });
+      test('returns Future<Subject> and can be awaited', () async {
+        (await check(_futureFail()).throws<UnimplementedError>())
+            .has((p0) => p0.message, 'message')
+            .isNull();
```
jonasfj, line 91: These are just "feeling", feel free to ignore.

I suspect I would rarely use `(await check(...)).whatever()`, because inline `await` is just ugly when it has to be wrapped that way.

I suppose I might find myself occasionally wanting to do:
```dart
final it = await check(...);
it.has(...);
```

I could imagine that being attractive in some complicated scenarios.

pkgs/checks/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
...
   `not((it) => it.isA<R>())`.
 - Require Dart 3.11
+- Updated `Context.nest` to accept an optional named `nestedCondition` argument
+  which is executed against the nested subject.
+- Updated `Subject.isA`, `Subject.isNotNull`, `Subject.throws` (sync), and
+  `Subject.returnsNormally` to accept an optional `Condition` callback to apply
+  to the extracted value.
+- Updated `Subject.completes`, `Subject.throws` (async), `StreamChecks.emits`,
+  and `StreamChecks.emitsError` to return `Future<Subject>` in addition to
+  accepting an optional `AsyncCondition` callback.
```
jonasfj: Consider showing examples like:
```md
 - Updated `Subject.isA` to allow `check(42).isA<int>((it) => it.lessThan(100));`
```

pkgs/checks/test/extensions/function_test.dart
```diff
@@ -53,6 +67,16 @@ void main() {
           which: ['threw <Bad state: oops!> at:', '  fake trace'],
         );
       });
+      test('evaluates condition', () {
+        check(() => 1).isRejectedBy(
+          (it) => it.returnsNormally((it) => it.equals(2)),
```
jonasfj, line 80: This is the new thing?

jonasfj (commented):
Maybe, write some examples with old and new things, it a bit hard to follow the explanations.

From what I understand, and this is just my personal feelings.
I would probably rather avoid `(await check(...)`, but I also see little harm in allowing it.

I would probably also prefer `check(...).isA<T>().something(..)` over `check(...).isA<T>((it) => it.something(..))`.

But I can see no harm in allowing both things. And it kind of makes both sync and async APIs behave similar, even if the async API still requires an `await` inside the expression.

---------

There is obviously no way we are going to do an extension on `Future<Subject<int>>` that will allow people to do:
```dart
await check(Future.value(42)).lessThan(100);
```

That would be cool, but subclassing `Future` is probably a crime. and making a global expando mapping from `Future<Subject>` to SomeKindOfHack, is probably borderline insane 🤣🤣🤣

So even if it could be done, it probably shouldn't be done 🤣
 -- But mentally that's what I would prefer to write, but if I had to reason about how it works, I probably wouldn't prefer it.

pkgs/checks/test/extensions/async_test.dart
natebosch: > I suspect I would rarely use `(await check(...)).whatever()`, because inline `await` is just ugly when it has to be wrapped that way.

Yes, this is why I moved await from this pattern and didn't keep it as an option. We've since had at least a few requests to include it. I suspect it will bear out that it won't get used much in practice, but I also think that the corresponding synchronous API that will get used drastically reduces the cost of having 2 ways to do this.

natebosch:
After discussions we plan on moving forward with this and #2697

pkgs/checks/test/extensions/function_test.dart
natebosch: This is not the new Condition syntax. I'll land that after this one.

pkgs/checks/CHANGELOG.md
natebosch: Done.

brianquinlan (approved):
I mainly looked at the tests, which seem reasonable.

PR #2685, opened 2026-06-25
Use compact reporters by default for direct runs

natebosch:
Since we cannot import `dart:io` for tests running on the web we
historically defaulted to the expanded reporter as the only alternative
to the compact reporter. This doesn't match most users preferences since
it makes a verbose output in all cases.

Default instead to the failures only reporter which is quieter, and add
a platform specific library to automatically switch to the compact
reporter when it's supported by both the Dart platform and the runtime
environment.

### DIFF
--- a/pkgs/test_core/lib/src/runner/reporter/direct.dart
+++ b/pkgs/test_core/lib/src/runner/reporter/direct.dart
@@ -1,0 +1,5 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+export 'direct_stub.dart' if (dart.library.io) 'direct_io.dart';
--- a/pkgs/test_core/lib/src/runner/reporter/direct.dart
+++ b/pkgs/test_core/lib/src/runner/reporter/direct.dart
@@ -2,4 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// A utility to pick a platform safe default reporter for cases where a test
+/// suite is run directly instead of through the test runner.
+library;
+
 export 'direct_stub.dart' if (dart.library.io) 'direct_io.dart';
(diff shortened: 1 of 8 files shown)


pkgs/test_core/lib/src/runner/reporter/direct.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+export 'direct_stub.dart' if (dart.library.io) 'direct_io.dart';
```
jakemac53, line 9: nit: maybe add a comment to this library explaining what it is

jakemac53 (approved):
I think this is probably OK to release as non-breaking but I wouldn't be super surprised if somebody does get broken by it 🤷‍♂️

pkgs/test_core/lib/src/runner/reporter/direct.dart
natebosch: Done.

natebosch:
> I wouldn't be super surprised if somebody does get broken by it

I think we can plan to resolve whatever comes up if it does. Overall I expect this to have little impact because most invocations use the test runner, but for the places where it does have an impact I think it's very likely to do more good than harm..

dcharkes:
> I think this is probably OK to release as non-breaking but I wouldn't be super surprised if somebody does get broken by it 🤷‍♂️

Yep, broken 😆 I'll just update the tests 👍 

* https://github.com/dart-lang/sdk/issues/63697

DanTup:
This really confused me - I have PowerShell aliases to run tests, one of which filters out most of the noise with `Select-String` and another which provided the full output, but now it isn't 😄

But, I'm in favour of the change (because now I don't need the `Select-String` hacks!), but I wonder if there should be an env var or something to allow forcing it back to the verbose one?

natebosch:
> I wonder if there should be an env var or something to allow forcing it back to the verbose one?

I think it's probably a good idea for us to support more environment variables like this in the test runner in general, it makes sense to me to honor it in the direct run case as well.

https://github.com/dart-lang/test/pull/2692

PR #2687, opened 2026-06-28
ci: pin dart-lang/ecosystem reusable workflows to commit SHA to prevent supply-chain attacks

XananasX7:
## Summary

The `publish.yaml`, `health.yaml`, and `post_summaries.yaml` workflows all call reusable workflows from `dart-lang/ecosystem` using the **mutable branch reference `@main`**:

```yaml
uses: dart-lang/ecosystem/.github/workflows/publish.yaml@main
```

Because `@main` is a mutable branch pointer, any commit pushed to `dart-lang/ecosystem`'s `main` branch is immediately picked up by every caller. If the `dart-lang/ecosystem` repository were compromised (supply-chain attack, maintainer account takeover, malicious PR merged), an attacker could inject arbitrary code that runs in the context of this repository with `id-token: write` permissions — enabling unauthorized package publishing to pub.dev.

## Vulnerability class

Supply-chain attack via mutable reusable workflow reference (CWE-829). See [GitHub Security hardening](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#reusing-third-party-workflows).

## Fix

All three workflow files now pin `dart-lang/ecosystem` reusable workflows to the full commit SHA `848b3bf3b757d2e9ae4d60030eeed5756c87783f` (current `main` HEAD at time of writing), with `# main` as a comment for readability.


natebosch:
@kevmoo - do you know what it would take to have dependabot  manage these for us?

kevmoo:
> @kevmoo - do you know what it would take to have dependabot manage these for us?

once they are SHAs dependabot will keep them updated AFAIK, @natebosch

natebosch (approved):

PR #2696, opened 2026-07-16
Make tests run in their correct zone.

lrhn:
(This is an attempt and WIP. Creating PR only to have the tests run properly.)

Add test that zones run in the zones they're created in.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -11,6 +11,7 @@ analyzer:
     avoid_catching_errors: ignore
     avoid_dynamic_calls: ignore
     comment_references: ignore
+    discarded_futures: ignore
     lines_longer_than_80_chars: ignore
     only_throw_errors: ignore
     unawaited_futures: ignore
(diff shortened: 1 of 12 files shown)


natebosch:
This seems like it makes `setUp` and `test` callbacks run in different error zones, which is breaking because it's common for tests to create a `StreamController` or `Completer` in a `setUp` callback, then use that controller to emit errors in the test. The following test changes behaviors and is broken by this, it will start to time out because the `throwsA` expectation is never satisfied since the error cannot reach that zone.

```dart
import 'dart:async';

void main() {
  late Completer<void> completer;

  setUp(() {
    completer = Completer();
  });

  group('nested group', () {
    test('completer completed across zones', () {
      completer.completeError('error');
      expect(completer.future, throwsA('error'));
    });
  });
}
```

lrhn:
That's ... a problem. So for each test, there should be an error zone, and the registered `setUp` and `tearDown`s should run inside that, as well as the test. May be in different actual zones, but not different error zones.

Maybe just rewrite the test framework from scratch? Maybe AI can do it for me :)

PR #2697, opened 2026-07-17
Restore Condition as a class

### LINKED ISSUE #2694: Restore `Condition` as a class instead of a function typedef
https://github.com/dart-lang/test/pull/1956 replaced a `Condition` class with a typedef for a callback for functions that take `Subject` arguments. Discussion in https://github.com/dart-lang/test/issues/2673 highlighted that the ergonomics have changed with dot shorthands, and there may be more enthusiasm for a class based API.

natebosch:
Closes #2694

Reintroduce `Condition<T>` as a class in place of callback typedefs
(`void Function(Subject<T>)` / `FutureOr<void> Function(Subject<T>)`).

Previously, the `Condition` class was dropped in favor of callback
functions, in part to avoid `it()` appearing like a language feature.
However, using callback functions introduced type inference limitations.

- Callbacks required declaring explicit parameters, introducing
  unnecessary boilerplate across expectation calls.
- When explicit types are required it takes a full `Subject<SomeType>`
  as the argument type annotation, mixing the details of the
  implementation with the goal of passing a type argument.
- When type inference failed (such as in collection matching or
  heterogeneous checks), callbacks resulted in cryptic runtime errors
  regarding missing extension methods on `Subject` with no static
  assistance.

Reintroducing `Condition<T>` with a `Condition.it<T>()` static helper.

- Dot shorthands provide a terse, readable syntax
  (`.it()..isGreaterThan(0)`) without polluting the top-level namespace.
- When inference fails `Condition.it<SomeType>()` is more explicit and
  readable than `(Subject<SomeType> it) =>`.
- Consolidating `softCheck`, `softCheckSync`, `describe`, and
  `describeSync` as instance methods on `Condition<T>` (and eliminating
  standalone top-level functions `softCheck`, `softCheckAsync`,
  `describe`, `describeAsync`) improves overall API discoverability and
  ergonomics, despite being a breaking change.

Example Migration:

Before:
```dart
check(numbers).any((it) => it.isGreaterThan(5));

check(map).containsKeyThat((key) => key
  ..startsWith('a')
  ..has((s) => s.length, 'length').isGreaterThan(2));
```

After:
```dart
check(numbers).any(.it()..isGreaterThan(5));

check(map).containsKeyThat(.it()
  ..startsWith('a')
  ..has((s) => s.length, 'length').isGreaterThan(2));
```


natebosch:
You can see the impact on test code in the diff here - somewhat eaiser if you [start at `async_test.dart`](https://github.com/dart-lang/test/pull/2697/changes#diff-27d562e0218f9cfb527b2260de683025a057c1ede87a50f0f4ce9b006744f007), expand it, and read down from there.

Probably even easier to see the impact on tests in `package:dart_mcp`:
- example of migrating tests that can't use dot shorthands yet: https://github.com/dart-lang/ai/commit/eb15b49cb8bc7bb4a75be91ce895c88bf98c9563
- If we were on a later language version: https://github.com/dart-lang/ai/commit/16680e1c4f14bb008cbc20e5a197541e9c2642b9


This PR sits on top of #2683 because it's a more drastic change when there are more extensions using this syntax. We can land this without the other easily enough if we decide to. If we decide _not_ to land this one I think I'd be less inclined to land the other.

lrhn:
Probably _could_ do something monadic like:
```dart
check(numbers).any.isGreaterThan(5);
```
but it's more complexity, and monads may not always nest nicely.

I like this approach. It generalizes abstracting over operations on a `Subject<T>`.
It does so by recording the operations on the context, so it can replay them on a another context.
That's neat, any syntactic sugar on top of the basic operations will just work.


I'd love to take it a bit further ... but that's not realistic.
Take for example:
```dart
check(numbers).any(.it()..isGreaterThan(5));
```
If we could just make `any` return a `ConditionSubject`, so you could write:
```dart
check(numbers).any.isGreaterThan(5);
```
That sadly has serious timing issues, the `any` call has lost control before the condition
has been filled in. It doesn't know when it can start checking.
Would require serious hackery to get around that, and probably remove any benefit.
(That's where having a specific list-monad-subject would work, because it would get the
calls directly, not indirectly from a `Condition`.)

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -78,30 +79,29 @@ extension FutureChecks<T> on Subject<Future<T>> {
...
+              actual: prefixFirst('completed to error ', literal(e)),
+              which: [
+                'threw an exception that is not a $E at:',
+                ...indent(LineSplitter.split(st.toString())),
+              ],
+            );
+          }
+        },
+        errorCondition,
+      );
```
lrhn: (I had an idea, so I'll put it here 😁:
What if there was `Future<Subject<E>> ifThrows<E extends Object>([Condition<E>? condition])` which:
* If the future throws an `E`, it's checked againsts condition, and if successful, returned.
* If it throws a non-`E`, throws an `E` not satisfying condition, or it completes with a value, the returned subject is a "skipped" subject that can't fail. 

So more of a guard than a check.

And similar for `ifCompletes`

Then you could do
```dart
  check(parseFuture)
    ..ifCompletes(.it.equals(42))
    ..ifThrows(.it.isA<FormatException>());
```
Have to make sure all error futures are properly handled on all branches.)

pkgs/checks/lib/src/checks.dart
```diff
@@ -30,16 +30,6 @@ final class Subject<T> {
...
-/// A callback that synchronously checks expectations against a subject.
-///
-/// Asynchronous expectations should not be used within a `Condition` callback.
-typedef Condition<T> = void Function(Subject<T>);
-
-/// A callback that asynchronously checks expectations against a subject.
-///
-/// Any expectations may be used within an `AsyncCondition` callback.
-typedef AsyncCondition<T> = FutureOr<void> Function(Subject<T>);
-
```
lrhn, line 42: (Thanks. I hate typedefs in APIs. I look at a signature _in an IDE_ and it just says `AsyncCondition<T>`, and then I still don't know what type I have to write. I need to go look at the declaration anyway.)

pkgs/checks/lib/src/checks.dart
```diff
@@ -102,104 +93,104 @@ Subject<T> check<T>(T value, {String? because}) => Subject._(
...
-    ),
-  );
-  condition(subject);
-  return failure;
-}
+/// Synchronous variants [applySync], [softCheckSync], and [describeSync] return
+/// non-[Future] values always, but will throw an [AsyncConditionDisallowed]
+/// exception if asynchronous expectations were invoked on the condition.
+abstract final class Condition<T> {
+  static ConditionSubject<T> it<T>() => ._();
```
lrhn, line 118: (It's really annoying that this has to be a method, just because it has to be generic. We should have generic getters!)

pkgs/checks/lib/src/checks.dart
```diff
@@ -102,104 +93,104 @@ Subject<T> check<T>(T value, {String? because}) => Subject._(
...
-        failure ??= f;
-      },
-      allowAsync: true,
-      allowUnawaited: false,
-    ),
-  );
-  await condition(subject);
-  return failure;
-}
+  /// Run these exepctations against [subject].
```
lrhn: Typo -> `expectations`

pkgs/checks/lib/src/checks.dart
```diff
@@ -102,104 +93,104 @@ Subject<T> check<T>(T value, {String? because}) => Subject._(
...
+  FutureOr<CheckFailure?> softCheck(T value);
+
+  /// Checks whether [value] satisfies all expectations invoked in this
+  /// condition, without throwing an exception.
+  ///
+  /// The future will complete to `null` if all expectations are satisfied,
+  /// otherwise it will complete to the [CheckFailure] for the first expectation
+  /// that fails.
+  ///
+  /// Asynchronous expectations are not allowed.
```
lrhn: (Is this a failure represented by a check failure or will _that_ throw an exception? In which case the first line is slighlty misleading)

pkgs/checks/lib/src/checks.dart
```diff
@@ -102,104 +93,104 @@ Subject<T> check<T>(T value, {String? because}) => Subject._(
...
-    },
-    allowAsync: false,
-    allowUnawaited: true,
-  );
-  condition(Subject._(context));
-  return context.detail(context).expected.skip(1);
+  /// Run these expectations against [subject].
+  ///
+  /// Asynchronous expectations are not allowed and will cause a runtime error
+  /// if they were used.
```
lrhn, line 126: (Use this paragraph for the other `*Sync` functions too, it's more explicit than "is not allowed".)

pkgs/checks/lib/src/extensions/async.dart
lrhn, line 86: Is [ErrorCondition] needed.

Is `(await subject.throws<E>()).tests` equivalent to `await subject.throws<E>(it()..tests)`?

If you can do both, which one should we recommend?
(The former shoud work with `expectSync`, the latter ... might?)

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -78,30 +79,29 @@ extension FutureChecks<T> on Subject<Future<T>> {
   ///
   /// The returned future will complete when the subject future has completed,
   /// and [errorCondition] has optionally been checked.
```
lrhn, line 89: This `errorCondition` is check-functions called on the `Subject<E>` that is caught.

I'm not sure it matters whether the condition is used as a real check, or only as a predicate, other than whether the error of 
```dart
await check(Future.error(StateError('Apricot'))
    .throws(.it().message.equals('Bananas'));
```
Is
> Did not throw a StateError that has a message which equals "Bananas".
or
> Threw a StateError which has a message which does not equal "Bananas".

That is, whether the test is "did it throw E + do checks on E" or "did it throw E which satisfies predicate".

I think it should be the latter.
(Also because the former is equivalent to `(await subject.throws<E>()).checks()` anyway, so you can get that if you want it.)

Also, some tests promote, like `isA`.
Someone might question why:
```dart
  check(future).completes<Object?>(.it().isA<int>())
```
does not return `Future<Subject<int>>`.
(There are good reasons, but "it's only used as a predicate" is an easier explanation.)

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -224,16 +224,21 @@ extension StreamChecks<T> on Subject<StreamQueue<T>> {
...
-        'emits any values then emits a value that:',
-        ...describe(condition),
-      ],
+      () {
+        final description = condition.describe();
+        if (description is Iterable<String>) {
+          return ['emits any values then emits a value that:', ...description];
+        }
+        return [
+          'emits any values then emits a value satisfying an asynchronous condition',
```
lrhn, line 259: (Reading this, I didn't read "any values" as allowing no values. The name didn't help me - I have no intuition what "emits through" means.
I guess it's the inclusive version of `emitsUntil`, and it's on a `StreamQueue`.
Consider `emitsEventual` or `emitsAny`, and just document that it accepts any events until such one.)

pkgs/checks/lib/src/extensions/async.dart
lrhn, line 282: (Can maybe say something like "one event per condition, in order" early. I had to read to here to figure out whether one event can satsify more than one condition.
Also "consumed" is not a word used earlier.)

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -340,20 +343,20 @@ extension StreamChecks<T> on Subject<StreamQueue<T>> {
...
-                (longestAccepted == null ||
-                    copy.eventsDispatched >
-                        longestAccepted!.eventsDispatched)) {
-              longestAccepted = copy;
-            }
-            return failure;
-          }),
-        );
+        final descriptionFuture = conditions
+            .map((c) async => await c.describe())
```
lrhn, line 379: (That's just `(c) => c.describe()` with more words, and maybe more delays.)

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -340,20 +343,20 @@ extension StreamChecks<T> on Subject<StreamQueue<T>> {
...
+            .wait;
```
lrhn, line 380: Can `.describe()` throw? 
(If not, is it documented on the function that it must not throw?)

If it can, then this future may complete with an unhandled async error while doing the next statement.

(If it can't, it'd be nice to know.)

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -340,20 +343,20 @@ extension StreamChecks<T> on Subject<StreamQueue<T>> {
...
+        final failures = await conditions.map((condition) async {
+          final copy = transaction.newQueue();
+          final failure = await condition.softCheck(copy);
+          if (failure == null &&
+              (longestAccepted == null ||
+                  copy.eventsDispatched > longestAccepted!.eventsDispatched)) {
+            longestAccepted = copy;
+          }
+          return failure;
+        }).wait;
```
lrhn, line 392: If any of the futures in `failures` can end in an error, which they can if any of the `softCheck`s throw, and those can be synchronous and calls user code so they almost certainly can throw, ... then the `failures` future may complete with an _unhandled asynchronous error_ while `await descriptionFuture` runs below.

This is just a test, and you probably know it won't throw, so document that.

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -340,20 +343,20 @@ extension StreamChecks<T> on Subject<StreamQueue<T>> {
...
+        final descriptionFuture = conditions
+            .map((c) async => await c.describe())
+            .wait;
+        final failures = await conditions.map((condition) async {
+          final copy = transaction.newQueue();
+          final failure = await condition.softCheck(copy);
+          if (failure == null &&
+              (longestAccepted == null ||
+                  copy.eventsDispatched > longestAccepted!.eventsDispatched)) {
+            longestAccepted = copy;
```
lrhn, line 389: Should the shorter of `longestAccepted` and `copy` be closed in some way?

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -340,20 +343,20 @@ extension StreamChecks<T> on Subject<StreamQueue<T>> {
...
+          }
+          return failure;
+        }).wait;
+
         descriptions.addAll(await descriptionFuture);
         if (longestAccepted != null) {
```
lrhn: `case final accepted?` for the promotion.

pkgs/checks/lib/src/collection_equality.dart
```diff
@@ -15,14 +15,12 @@ import '../context.dart';
...
-/// Elements, keys, or values within [expected] which are [Condition] callbacks
-/// are run against the value in the same position within [actual].
-/// Condition callbacks must take a `Subject<Object?>` or `Subject<dynamic>` and
-/// may not use a more specific generic.
-/// Use `(Subject<Object?> s) => s.isA<Type>()` to check expectations for
+/// Elements, keys, or values within [expected] which are [Condition]
+/// instances are run against the value in the same position within [actual].
+/// Condition instances must be a `Condition<Object?>` or
+/// `Condition<dynamic>` and may not use a more specific generic.
+/// Use `Condition.it<Object?>()..isA<Type>()` to check expectations for
```
lrhn, line 22: Should it be `.isA<Type>()` and not `..isA<Type>()` here? (So you can use checks on `Type`s on it.)

pkgs/checks/lib/src/extensions/core.dart
lrhn: Old format.
```dart
which(it()..isLessThan(10)..isGreaterThan(0));
```

pkgs/checks/lib/src/extensions/core.dart
```diff
@@ -66,19 +66,19 @@ extension CoreChecks<T> on Subject<T> {
   ///     ..isLessThan(10)
   ///     ..isGreaterThan(0));
   /// ```
-  void which(Condition<T> condition) => condition(this);
+  void which(Condition<T> condition) => condition.applySync(this);
```
lrhn, line 71: This change may make `which` more important.
It introduces `Condition`, an application of checks, as a first-class value that you may want to store for _reuse_.
You can have whole libraries of conditions. (Probably shouldn't, those should be extensions on subjects of their types instead.)
But still people may want to use `which` more.

Do we need both `which` and `isA`?

What if `isA` was
```dart
/// Checks that the valie is an [R] which satisfies [condition].
R isA<R extends T>([Condition<R>? condition]);
```
(Only allowing downcasts to make `.isA(.it()...)` work better, maybe have `castsTo<T>` for general casts.
And because also having an operation that only allows downcasts is better than only having one that allows anything without warning that it's unrelated.)

pkgs/checks/lib/src/checks.dart
```diff
@@ -102,104 +93,104 @@ Subject<T> check<T>(T value, {String? because}) => Subject._(
...
-      throw UnimplementedError();
-    },
-    allowAsync: true,
-    allowUnawaited: true,
-  );
-  await condition(Subject._(context));
-  return context.detail(context).expected.skip(1);
+/// A [Subject] that records the checks that are performed and expose them as an
+/// [Condition].
+final class ConditionSubject<T> implements Condition<T>, Subject<T> {
```
lrhn, line 178: `ConditionSubject` has public methods, which means that you can't use those names as the first check on a condition.
```dart
check(myFunctionAbstraction).where(.it()..apply().equals(42));
```
where `apply` is a custom extension on `Subject<MyFunctionAbstraction>`, won't work,
but if it was `.it()..isA<MyFunctionAbstraction>().apply().equals(42)`, it would because the static type of `isA` is just a `Subject`.

Can keep the methods public, but change `it()` to return `Subject<T>` instead of `ConditionSubject<T>`.
(Does the class and methods need to be public?)

lrhn (approved):
I like the approach. I tries to sidestep the "monadic subject" (like `AsyncSubject`) by abstracting over the things you'd do those subjects, and then injecting them into the `Subject<Monad<T>>`.
That way there doesn't have to be special subjects for each kind of weird context a value can be in. (And it generalizes.)

It shouldn't be more or less powerful than `FutureOr<void> Function(Subject<T> subject)`, but it's probably easier to work with for users. (And may be easier to type-infer.)

Maybe the `Condition` class, and the `it` method, should state very directly that you should _always_ us `.it()..` to ensure the result is a `Condition`, since the check calls on it return a normal `Subject`.

pkgs/checks/lib/src/extensions/async.dart
jonasfj: see https://github.com/dart-lang/test/pull/2683

pkgs/checks/lib/src/checks.dart
jonasfj: Could there not be a reason to pass a descriptive name as argument?

```dart
Condition.it<String>().contains(' '); // it contains ' '

Condition.it<String>('fullName').contains(' '); // fullName contains ' '
```

If doing things like `.isRejectedBy` do we not want to print a Condition?

jonasfj:
It took me a while to learn that `Condition<T>` was just something like `(it) => it.has(...)`.

And it's still awkward to work with `Condition<T>` as a general thing.

But `.it` always looks weird to me, so I'm not sure how I should feel about it.

----------------

Writing tests I have found it very useful to make a helper function that takes some data and `Condition<T>` objects and then do stuff and check the conditions.

For example [`testCodeGeneration(name, source, error, output)`](https://github.com/google/dart-neats/blob/eb845395ad854add561f792f097852d53e18001e/typed_sql/test/codegen/test_code_generation.dart#L24-L46), where `error` and `output` are `Condition<String>` on the errors thrown and the output produced. Making it easy to write tests like

```dart
  testCodeGeneration(
    name: 'Only one PrimaryKey is allowed',
    source: r'''
      abstract final class TestDatabase extends Schema {
        Table<Item> get items;
      }

      @PrimaryKey(['id'])
      @PrimaryKey(['id'])
      abstract final class Item extends Row {
        int get id;

        String get value;
      }
    ''',
    error: (s) => s.contains(
      'Only one `PrimaryKey` annotation is allowed',
    ),
  );
```

I would probably prefer to write:

```dart
  testCodeGeneration(
    name: 'Only one PrimaryKey is allowed',
    source: r'''
      abstract final class TestDatabase extends Schema {
        Table<Item> get items;
      }

      @PrimaryKey(['id'])
      @PrimaryKey(['id'])
      abstract final class Item extends Row {
        int get id;

        String get value;
      }
    ''',
    error: .it().contains(
      'Only one `PrimaryKey` annotation is allowed',
    ),
  );
```

IMO, this is the use-case where `Condition<T>` objects are something you want to pass around.

It's possible, but not likely that people will build collections of `Condition<T>` objects and use `.which(...)` with that.

But for making a declarative tests utility, where you provide data and some assertions over output, I can see `Condition<T>` being extremely useful, and `.it()` might make it better. I think this is especially the case if you want to combine two `Condition<T>` objects, then callbacks can be a bit too smart.

But I don't have strong feelings either way here.

jonasfj:
> Consolidating softCheck, softCheckSync, describe, and
describeSync as instance methods on Condition<T> (and eliminating
standalone top-level functions softCheck, softCheckAsync,
describe, describeAsync) improves overall API discoverability and
ergonomics, despite being a breaking change.

Could softCheck be an extension method on `R Function(Subject<T>)`?

pkgs/checks/test/extensions/function_test.dart
```diff
@@ -62,14 +62,14 @@ void main() {
...
-          (it) => it.returnsNormally(),
+          .it()..returnsNormally(),
           actual: ['a function that throws'],
           which: ['threw <Bad state: oops!> at:', '  fake trace'],
         );
       });
       test('evaluates condition', () {
         check(() => 1).isRejectedBy(
-          (it) => it.returnsNormally((it) => it.equals(2)),
+          .it()..returnsNormally(.it()..equals(2)),
```
jonasfj, line 80: @natebosch just curious, why not just do:
```suggestion
          .it().returnsNormally(.it().equals(2)),
```

munificent:
A while back, @natebosch asked for my thoughts on this API change.

I'm still ramping on how the checks API looks, but overall, this looks like an improvement to me. Definitely anything that gets type inference working better will lead to a friendlier UX. In particular, Dart users are well trained to not have to put type annotations on lambda parameters. The previous design that required that in some places feels like an annoying source of friction.

I'm a little iffy about `it()` being a constructor on `Condition` versus a top-level generic function. I see the merit because it means that shorthand name is only available in contexts where it makes sense. It avoids polluting the top level namespace with an extremely short name. That's good.

However, `.it()..` is a lot of dots and feels a bit like the API is pulling the language in a strange direction. I don't hate it though. It just looks kind of weird for someone like me who still isn't very used to dot shorthands (and also generally avoids cascades). I suspect I would quickly get used to it.

The overall model of hanging extension methods off `Subject` feels like a really smart design to me.

pkgs/checks/lib/src/extensions/async.dart
natebosch: I vacillated on this design a bunch, but we recently landed on moving forward with allowing both patterns.

We'll recommend  (by way of primarily using this pattern in docs) the `await check(future).throws<E>(it()..tests)` style for async, and `check(callback).throws<E>().tests` for sync, while allowing either style for both sync and async.

pkgs/checks/lib/src/extensions/async.dart
natebosch: > That is, whether the test is "did it throw E + do checks on E" or "did it throw E which satisfies predicate".
> 
> I think it should be the latter. (Also because the former is equivalent to `(await subject.throws<E>()).checks()` anyway, so you can get that if you want it.)

They'll remain equivalent for now after deciding to move forward with #2683

pkgs/checks/lib/src/extensions/async.dart
natebosch: I played with this a bit and I don't think I want to land it. Most tests should have a single hardcoded behavior, and if we want to allow testing either/or style expectations I think we should do it at a higher level than the throws/completes epectations.

pkgs/checks/lib/src/extensions/async.dart
natebosch: https://github.com/dart-lang/test/issues/2750

pkgs/checks/lib/src/extensions/async.dart
natebosch: #2750

pkgs/checks/lib/src/extensions/async.dart
natebosch: Almost.

It's written this way in favor  of `(c) => switch(c.describe()) {final Future result => result, final syncResult => Future.syncValue(syncResult)}` because [someone](https://github.com/dart-lang/sdk/issues/30280#issuecomment-318935820) argued against accepting `FutureOr` in lists of futures 😝

pkgs/checks/lib/src/extensions/async.dart
natebosch: It should not throws exceptions. I'm not sure that it won't throw errors, but I think it's OK if those are surfaced only as unhandled async errors - that's relatively safe when we know we're running in a test zone.

pkgs/checks/lib/src/extensions/async.dart
natebosch: #2751

pkgs/checks/lib/src/extensions/async.dart
natebosch: Added a comment above.

pkgs/checks/lib/src/extensions/async.dart
natebosch: Done.

pkgs/checks/lib/src/extensions/core.dart
natebosch: Done.

pkgs/checks/lib/src/extensions/core.dart
natebosch: I've also added an async `which` in the upstream change to allow both interaction styles across sync and async expectations.

I think I'd rather not overload `isA`, there are good reasons to use `which` that don't related to type checks.

pkgs/checks/lib/src/checks.dart
natebosch: I can't think of a place we'd use the `'fullname'` value in our errors, but I'll keep this in mind.

pkgs/checks/lib/src/checks.dart

pkgs/checks/lib/src/checks.dart
natebosch: Done.

pkgs/checks/lib/src/checks.dart
natebosch: Expanded the first sentence a bit.

```dart
  /// Checks whether [value] satisfies all expectations invoked in this
  /// condition, without throwing an exception for failed expectations.
```

I think Errors from improper use can still surface, but not exceptions from failed expectations.

pkgs/checks/lib/src/checks.dart
natebosch: > Can keep the methods public, but change `it()` to return `Subject<T>` instead of `ConditionSubject<T>`.
> (Does the class and methods need to be public?)

The type does need to be public so that we can disallow `.which(check(foo))` statically. The methods need to be exposed in some way - they were originally top level methods and we _could_ leave them that way while making these class members private.

I think I'm inclined to keep them public here instead of moving them back to top level functions. I think conflicts are not likely to become a problem, because the verbs would be used different for expectations. Like `check(foo).applies(someAppliedThing)` instead of `check(foo).apply(someAppliedThing)`. Feel free to raise a new issue if you still think we should hide these.

pkgs/checks/lib/src/collection_equality.dart
natebosch: `Condition.it()` is likely to _always_ be followed by `..` in practice. It needs to be a `ConditionSubject` so we can statically restrict arguments to safe types, but once you start calling expecations you always end up back to a plain `Subject` statically.

pkgs/checks/test/extensions/function_test.dart
natebosch: `returnsNormally` and `equals` return a `Subject` statically. We need to use cascades to satisfy the type requirements in both directions (towards the method argument, and towards the extension methods)

natebosch:
We had some more conversations about these designs and decided to move forward with this.

PR #2698, opened 2026-07-17
Make workflow files more secure

jakemac53:
This is just a result of running `zizmore --fix=all` on this package


jakemac53:
~We might end up blocked on https://github.com/google/mono_repo.dart/issues/523 but we will see~

This is resolved now in mono_repo version 6.7.1.

natebosch (approved):

PR #2699, opened 2026-07-18
Group messages with passing tests

### LINKED ISSUE #2636: `Passing tests` groups on GitHub actions don't include "loading" tests
You end up with far more lines than you want, because the loading tests won't get included into existing passing tests groups (ideally, you see just one "passing tests" group, if all tests pass):

Image

This should have just been a single group ideally.

cc @kevmoo

natebosch:
Closes #2636

Web tests unconditionally print information about the compile, so
treating messages during loading as separate output from passing tests
causes significantly longer output when web and VM tests are
interleaved.

Avoid closing groups for tests that print messages. This will impact
more than loading tests, but won't impact errors which are the important
output.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -4,6 +4,8 @@
   when tests are run directly on platforms which support `dart:io`. The
   environment variable takes precedence over configuration in `dart_test.yaml`
   but is overridden by the `--reporter` flag when passed to the test runner.
+* Include passing test and suite loading messages within the `Passing tests`
+  group in `GithubReporter`.
 
 ## 1.31.2
 
--- a/pkgs/test/test/runner/github_reporter_test.dart
+++ b/pkgs/test/test/runner/github_reporter_test.dart
@@ -205,13 +205,34 @@ void main() {
           print("four");
         });''',
         '''
-        ::group::✅ test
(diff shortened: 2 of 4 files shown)


natebosch:
@jakemac53 @kevmoo - what is your opinion on whether or not to collapse even the `print` output that users wrote in their tests? I'm in favor of it, but I'm open to arguments against.

natebosch:
After some reflection I'm pretty confident this is a good idea.

jakemac53:
You can always expand the output, so I think its a good idea yes.

> Web tests unconditionally print information about the compile

Fwiw, I don't think this is or has ever been desirable, it is just extra noise the vast majority of the time. It could print in a verbose mode or something but that should be it imo.

kevmoo (approved):
Yes, please! Thank you!

jakemac53 (approved):

natebosch:
> Fwiw, I don't think this is or has ever been desirable, it is just extra noise the vast majority of the time.

I was feeling that way too, but this solution felt like a good idea either way, and so I didn't bother to think too much about the implications of deleting that message entirely.

How do you feel about doing both?

jakemac53:
> How do you feel about doing both?

I am in favor of both (but you can do that separately). Also I think the exe compiler does the same thing

PR #2703, opened 2026-07-21
Launch browsers with redirecting files

natebosch:
The test runner creates a secret key used for communication with the
browsers to prevent third party processes from connecting as if they
were running test suites, but this key is current leaking through the
command arguments when starting the browser process. We are already
using temp directories which should prevent external reads and the
safari browser launcher already used this pattern. Extend the
`redirect.html` pattern to chromium and firefox browsers.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -7,6 +7,8 @@
 * Suppress dart2js compiler output for successful compiles.
 * Include output from passing tests  messages within the `Passing tests` group
   in `GithubReporter`.
+* Use a redirect html file for browser tests to avoid leaking websocket details
+  through process starting arguments.
 
 ## 1.31.2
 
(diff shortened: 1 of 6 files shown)


jakemac53 (commented):
Doesn't this defeat the purpose of the URL secret? You can now get the secret just by going to redirect.html.

As far as leaking the command line argument, this is only available to attackers which already have access to run arbitrary code on the machine, which I don't find concerning (they already have access to the files as well).

This redirect if I understand correctly means a remote attacker could get access to your files and only needs to know the redirect.html path.

natebosch:
> Doesn't this defeat the purpose of the URL secret? You can now get the secret just by going to redirect.html.
> 
> As far as leaking the command line argument, this is only available to attackers which already have access to run arbitrary code on the machine, which I don't find concerning (they already have access to the files as well).

From what I understand in the report this approach is preferred because files in fresh temp directories have stronger per user encapsulation than process arguments on multi-user hosts.

natebosch:
Internal bug: b/536186824

jakemac53:
> From what I understand in the report this approach is preferred because files in fresh temp directories have stronger per user encapsulation than process arguments on multi-user hosts.

Ok, so the URL being passed to Chrome is now a file path, and presumably Chrome will respect any user specific permissions when reading file paths directly? Or maybe the OS already just enforces this?

jakemac53 (approved):
LGTM assuming my understanding is correct

natebosch:
> Or maybe the OS already just enforces this?

It's the OS level file permissions - the file in a fresh temp directory won't be world readable like process arguments are.

PR #2705, opened 2026-07-23
Remove unnecessary remote debugging flag

natebosch:
The comment above the argument is outdated - chromium based browsers
launched with the arguments we are using, including a unique user
profile directory not associated with any running browser instance, will
work without also opening a debugging port.

### DIFF
--- a/pkgs/test/lib/src/runner/browser/chromium.dart
+++ b/pkgs/test/lib/src/runner/browser/chromium.dart
@@ -58,10 +58,6 @@ enum ChromiumBasedBrowser {
         '--headless',
         '--disable-gpu',
       ],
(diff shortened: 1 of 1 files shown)


natebosch:
b/536188175

kevmoo (approved):

natebosch:
In case it comes up again, this is only safe for chrome versions 128 (released 8/20/24) and up.

stuartmorgan-g:
> In case it comes up again, this is only safe for chrome versions 128 (released 8/20/24) and up.

Thanks for leaving this comment, it made figuring out an OOB breakage in flutter/packages (on shards still running an old version of Chrome) much faster than it would have been otherwise 🙂

PR #2708, opened 2026-07-30
[test] Support CLI compiler for sanitizer runtimes

HamdaanAliQuatil:
Part of dart-lang/sdk#63489.

`package:test` already supports the sanitizer VM runtimes with the `exe` compiler. The Dart SDK can now build CLI bundles with native assets and sanitizer instrumentation, so this extends the same runtimes to `-c cli`.

## What changes

- Allow `Compiler.cli` with `vm-asan`, `vm-msan`, and `vm-tsan`, while retaining `Compiler.exe` as the default compiler.
- Pass the existing sanitizer environment defaults when running a CLI bundle, without replacing values supplied by the user.
- Gate CLI sanitizer support on the first Dart dev SDK containing the required `dartcliruntime_*` artifacts.
- Validate the compiler/runtime combinations against the appropriate CLI runtime artifacts and update the command-line help.
- Add matrix coverage for default and user-provided sanitizer environment options.

## Why

This enables commands such as:

```console
dart test -c cli -p vm-asan
```

The CLI compiler path runs native build hooks, unlike `dart test -c exe`, so packages with hook-built native assets can exercise those assets under ASan, MSan, and TSan.

A follow-up Dart SDK change will roll this `package:test` revision and add an end-to-end native-assets test using a hook-built fixture.

## Validation

- 24 focused CLI sanitizer compiler/runtime matrix cases pass.
- The focused CLI help test passes.
- `dart analyze --fatal-infos` passes for `pkgs/test`, `pkgs/test_api`, and `pkgs/test_core`.

### DIFF
--- a/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
+++ b/pkgs/test/test/runner/compiler_runtime_matrix_test.dart
@@ -21,6 +21,18 @@ void main() {
 
   for (var runtime in Runtime.builtIn) {
     for (var compiler in runtime.supportedCompilers) {
+      final sanitizerSuffix = switch (runtime) {
+        Runtime.vmAsan => 'asan',
+        Runtime.vmMsan => 'msan',
+        Runtime.vmTsan => 'tsan',
+        _ => null,
+      };
+      final sanitizerEnvironment = switch (runtime) {
+        Runtime.vmAsan => 'ASAN_OPTIONS',
+        Runtime.vmMsan => 'MSAN_OPTIONS',
+        Runtime.vmTsan => 'TSAN_OPTIONS',
+        _ => null,
+      };
       // Ignore the platforms we can't run on this OS.
       if ((runtime == Runtime.edge && !Platform.isWindows) ||
           (runtime == Runtime.safari && !Platform.isMacOS) ||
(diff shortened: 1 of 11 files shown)


HamdaanAliQuatil:
@dcharkes please take a look. thanks!

pkgs/test/test/runner/compiler_runtime_matrix_test.dart
```diff
@@ -149,6 +163,43 @@ void main() {
...
+          if (sanitizerEnvironment != null && compiler == Compiler.cli) {
+            test(
+              'sets sanitizer environment defaults',
+              () async {
+                await d
+                    .file(
+                      'test.dart',
+                      _sanitizerEnvironmentTest(
+                        sanitizerEnvironment,
+                        'halt_on_error=1:exitcode=6:symbolize=1',
```
dcharkes, line 176: cc @rmacnak-google Does this look right?

jakemac53 (approved):
@natebosch should probably also take a pass at this

pkgs/test/test/runner/compiler_runtime_matrix_test.dart
rmacnak-google: Looks right.

dcharkes (approved):

rmacnak-google (approved):

PR #2712, opened 2026-08-03
Avoid transitive imports to package:analyzer

### LINKED ISSUE #63923: Failures on [deps] rev core, dartdoc, ecosystem, i18n, shelf, test, tools, web
There are new test failures on [[deps] rev core, dartdoc, ecosystem, i18n, shelf, test, tools, web](https://github.com/dart-lang/sdk/commit/2610f35ecf4c2df8bb79ac40710ffe663c18d8c7).

The tests
```
co19/VM/digit_separators_t01 Timeout (expected Pass)
co19/VM/primary_constructors_t01 Timeout (expected Pass)
co19/VM/primary_constructors_t02 Timeout (expected Pass)
co19/VM/primary_constructors_t03 Timeout (expected Pass)
co19/VM/primary_constructors_t04 Timeout (expected Pass)
co19/VM/primary_constructors_t05 Timeout (expected Pass)
co19/VM/primary_constructors_t06 Timeout (expected Pass)
co19/VM/primary_constructors_t07 Timeout (expected Pass)
co19/VM/private_named_parameters_t01 Timeout (expected Pass)
co19/VM/private_named_parameters_t02 Timeout (expected Pass)
co19/VM/static_extensions_t01 Timeout (expected Pass)
co19/VM/static_extensions_t02 Timeout (expected Pass)
pkg/vm_service/test/fetch_all_types_test/0 Timeout (expected Pass)
pkg/vm_service/test/fetch_all_types_test/1 Timeout (expected Pass)
pkg/vm_service/test/object_graph_identity_hash_test Timeout (expected Pass)
samples/ffi/httpIG/test/http_test Timeout (expected Pass)
vm/dart/heapsnapshot_cli_test Timeout (expected Pass)
```
are failing on configurations
```
cfe-strong-mac
vm-dyn-linux-debug-x64
vm-dyn-mac-debug-arm64
vm-linux-release-x64
vm-mac-debug-arm64
vm-mac-release-arm64
```

/cc @natebosch

natebosch:
Closes https://github.com/dart-lang/sdk/issues/63923

The analyzer package sources dominate the compile time for small tests.
Avoid importing the analyzer transitively from `package:test/test.dart`.

Move the tests of the `test_api` import restrictions to the integration
test directory so it's easier to reuse for tests across the packages.
Add a test that `package:test/test.dart` does not transitively import
any library under the analyzer package.

Use a more narrow import of the parts of the `Platform` library which
are needed in the JSON reporter implementation.


alexmarkov (approved):

integration_tests/regression/test/import_restrictions_test.dart
```diff
@@ -0,0 +1,248 @@
...
+
+  static Future<_ImportCheck> create() async {
+    final packageUri = await Isolate.resolvePackageUri(
+      .parse('package:test_api/'),
+    );
+    final packageConfig = await findPackageConfig(.fromUri(packageUri!));
+    if (packageConfig == null) {
+      throw StateError('Could not find package_config.json.');
+    }
+    return ._(packageConfig);
```
jakemac53, line 79: Hmm, I don't know how I feel about dot shorthands in this case.... it isn't something you have to change but I probably wouldn't use them here because the visible context type is so far away, I am just left wondering what the heck this is.

integration_tests/regression/test/import_restrictions_test.dart
```diff
@@ -0,0 +1,248 @@
...
+                u.pathSegments.isNotEmpty &&
+                u.pathSegments.first == package,
+          )
+          .toSet();
+
+      yield _Source(current, samePackageImports);
+
+      for (final import in samePackageImports) {
+        if (!visited.contains(import)) {
+          queue.add(import);
```
jakemac53: This check gives a false sense of security, its only really a slight performance optimization but queue may contain duplicate items, and we are reliant on the additional check at the top of the loop to discard those still.

I would change this to a `seen` set, and guard this add behind `if (seen.add(current)) {` instead. Then you can remove the add to the visited check above, and we will for sure never add duplicates to the queue.

You will also want to initialize `seen` with `entryPoints`.

integration_tests/regression/test/import_restrictions_test.dart
```diff
@@ -0,0 +1,248 @@
...
+        }
+      }
+    }
+  }
+
+  Future<List<Uri>?> findPathToPackage({
+    required Uri startUri,
+    required String targetPackage,
+  }) async {
+    var queue = Queue<List<Uri>>();
```
jakemac53: A queue of all the full paths here is pretty inefficient, maybe it doesn't matter for this test though... but in the import_path package instead of this I just keep a map of closest "parent" import, so it's just a `Map<String, String?>`, no larger than the total number of libraries. And that is sufficient to then recreate the shortest path very cheaply once you find the desired import.

See https://github.com/jakemac53/import_path/blob/master/bin/import_path.dart#L34

integration_tests/regression/test/import_restrictions_test.dart
natebosch: I did hesitate a bit on this one, but I think I'm inclined to try a shorthand-maximalist style to see whether it does come up that we miss the extra type hints in practice.

I did flip it back and forth a few more times before landing on keeping it as-is and I think I do like the shorthand version when I consider them side by side.

integration_tests/regression/test/import_restrictions_test.dart
natebosch: Done.

integration_tests/regression/test/import_restrictions_test.dart
natebosch: Done.

integration_tests/regression/test/import_restrictions_test.dart
jakemac53: I am fine with  you leaving it as is, but I did end up feeling the need to scroll up when doing the review, fwiw.

In an IDE its less of an issue than in github etc.

jakemac53 (approved):

Issue #2225, opened 2024-05-11
Add extension points to the test runner to support user defined behaviour


TheFriendlyCoder:
Coming from a Python background where I have made extensive use of the `pytest` library, I became accustomed to the level of customizability the library offered. One of the most powerful / useful features of it are it's extensible [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html). These allow users to heavily customize the mechanics of the test runner via a set of hooks / extension points, allowing custom behaviour to be "injected" into various parts of the runner sequencing. Some of the most notable features are:

- being able to [override test fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#overriding-fixtures-on-various-levels) on a folder basis
- being able to [extend command line parameters](https://docs.pytest.org/en/6.2.x/fixture.html#overriding-fixtures-on-various-levels) for the test runner to customize fixture behaviour 
- being able to [scope test fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) so you can customize behaviour for specific tests, classes, or modules ... or an entire suite.
- hooking into the [test runner collection phase](https://docs.pytest.org/en/7.1.x/reference/reference.html#collection-hooks) so you can customize the test behaviour depending on which tests are being run in the current invocation

These are just a few examples of the kinds of extension points that I haven't been able to find comparables for in the Dart testing framework. I did notice that there were some plans to add support for [plugins](https://github.com/dart-lang/test/issues?q=is%3Aissue+is%3Aopen+plugin+label%3A%22plugin+API%22) to the test runner, but I'm not sure the tasks I found there offer nearly the same level of customization that other test frameworks like PyTest offer. In particular, the ones that allow for scoping of test fixtures to different levels, and hooking into specific stages with the test orchestration pipeline. Such features greatly enhance the ability of developers to customize the behaviour for unique workflows.

Below are some workflows that I've used personally in other test suites that would benefit greatly from having this flexibility:

- being able to aggregate log output either per-test, per-module or per suite/subset of tests
- being able to orchestrate shared state that is only required within certain scopes (ie: shared credentials for a group of tests, shared resources like Docker environments, etc.)
[...]

PR #2714, opened 2026-08-05
Add globalSetup and addGlobalTearDown APIs for cross-suite setup and teardown

sigurdm:
Adds `globalSetup()` and `addGlobalTearDown()` APIs in `package:test_api` / `package:test` to execute shared global setup scripts once on the host test runner, memoizing results across all test suites and executing registered teardowns when the test runner closes.

Addresses #2225

The main motivation is to support setup shared across multiple test suites (`_test.dart` files). Examples include:
* Precompiling a binary used by multiple end-to-end tests
* Provisioning test databases or starting local test servers

### Example
```dart
// test/setup_server.dart
import 'dart:io';

Future<Map<String, dynamic>> setUp() async {
  var server = await HttpServer.bind('localhost', 0);
  addGlobalTearDown(() async => await server.close(force: true));
  return {'port': server.port};
}

// test/service_test.dart

void main() {
  test('connects to server', () async {
    final config = await globalSetup(
      Uri.parse('/test/setup_server.dart'),
    ) as Map;
    final port = config['port'] as int;
    // ...
  });
}
```

### Key Features
* **`globalSetup(Uri uri)`**:
  * Runs the Dart file at `uri` in an isolate on the host runner and returns its JSON-encodable result.
  * Memoizes the result so all test suites calling `globalSetup` with the same canonical URI share the exact same setup output and the script only executes once.
  * Bootstraps isolates via an in-memory `data:` URI using `Isolate.spawnUri` (sharing architecture with `spawnHybridUri`), avoiding filesystem compilation or runner-specific compiler hooks.
* **`addGlobalTearDown(FutureOr<void> Function() callback)`**:
  * Called from within a setup script to register cleanup tasks.
  * Executed in reverse registration order (LIFO) when the test runner closes.
  * Best-effort cleanup: all registered teardowns run even if an earlier teardown throws, with errors reported to stderr and runner exiting with code 1.
  * 30-second per-isolate timeout to prevent hanging teardowns.
* **URI Resolution**:
  * **Root-relative** (`Uri.parse('/test/setup.dart')`): Resolved relative to package root (`pubspec.yaml`).
  * **`package:` URIs** (`Uri.parse('package:my_pkg/test_helpers.dart')`): Resolved synchronously via package config.
  * **Relative URIs** (`Uri.parse('setup.dart')` or `Uri.parse('../setup.dart')`): Resolved relative to the executing test suite's directory.
  * **`file:` URIs**: Resolved as absolute filesystem paths.
  * Fragments are automatically removed for consistent compilation caching. Relative, root-relative, and file URIs strictly prohibit query parameters.
* **Cross-Platform Support**:
  * Dispatches over `test_channel`, allowing browser, Node, and VM tests alike to trigger host-side setup.
* **Standalone Execution Support**:
  * Tests run directly via `dart <test_file>.dart` execute setup scripts using a standalone scaffolding fallback and run teardowns on process exit.
* **Flutter Compatibility**:
  * Works out of the box with `flutter test` without runner-level code changes.


lrhn:
I just saw this, and had to ask stupid questions :)


Why is this needed? (More motivational bits! :wink: )

Is it because:
- There is one host runner
- Multiple separate runners, not necessarily in the same process, or even on the same machine?
- Which still need to do _something_ only once for setup.

?

> `globalSetup<T>(Uri uri):`

What is return type. Sounds like it should be `Future<Object?>` (is definitely async and returns JSON).

> * Runs the Dart file at `uri` in an isolate on the host runner and returns its JSON-encodable result.

I'm not sold on having a `main` function with a return type that's actually used. The standard is that `main` is a `void` function. Consider if you can use a different name than `main`.

(Does it _have_ to be an isolate. Could it be run in a different process? (If it could, then docs may want to not be too specific. If not, why not?)

> * Memoizes the result so all test suites calling `globalSetup` with the same canonical URI share the exact same setup output and the script only executes once.

I assume this is the goal here. Otherwise you could just have a memoizing local function called using `setUp` that caches its result. 
But it still feels weird to use a _script_ instead of just spawning an isolate to run a function. That could be the `main` function of the script file. _Why a script?_

Where can you call `globalSetup`? Not inside `test` I assume. Inside a group? Inside `setUp`/`setUpAll`? (Also definitely not inside `teardown`.)

> Uses the shared incremental Frontend Server (TestCompiler) to compile setup scripts quickly.

> `addGlobalTearDown(FutureOr<void> Function() callback):`
> * Called from within a setup script to register cleanup tasks.

What is "setup script" here? One run by `addGlobalSetUp`, or just a `setUp` callback?
If the former, how (and when) does it communicate the global tear down back to the host runner?

If the former, I guess the setup script isolate must stay alive until it needs to run tear-down,
and there is some communication between it and the host runner.

> * Executed in reverse registration order (LIFO) when the test runner closes.

If global setup scripts are run in separate isolates, they can run in parallel, and then it's not obvious how to order their registrations.
(I guess if there is any doubt, they're not related and the order doesn't matter.)

* Best-effort cleanup: all registered teardowns run even if an earlier teardown throws, with errors reported to stderr and runner exiting with code 1.
* 30-second per-isolate timeout to prevent hanging teardowns.

(So definitely sounds like isolate stays alive.)

> URI Resolution:

Is this only for the URI passed to `addGlobalSetUp`.
(Can that script use a different package resolution? Be in a different (maybe nested) package?)

pkgs/test/test/runner/global_setup_test.dart
```diff
@@ -0,0 +1,304 @@
...
+    'runs global setup once across multiple suites and caches result',
+    () async {
+      await d.dir('tool', [
+        d.file('setup.dart', '''
+        import 'dart:io';
+
+        Future<String> main() async {
+          final file = File('counter.txt');
+          final count = file.existsSync() ? int.parse(file.readAsStringSync()) : 0;
+          file.writeAsStringSync('\${count + 1}');
```
lrhn: Conside opening the file in non-shared mode to avoid concurrent read/writes.
And/or appending to it instead of overwriting, then check the length afterwards instead of the content.

It's quite possible for multiple threads to run this code in parallel and end with a result that is the original value plus one.

pkgs/test/test/runner/global_setup_test.dart
```diff
@@ -0,0 +1,304 @@
...
+      await test.shouldExit(0);
+
+      expect(File('${d.sandbox}/counter.txt').readAsStringSync(), equals('1'));
+    },
+  );
+
+  test('supports Map and JSON-encodable return values', () async {
+    await d.dir('tool', [
+      d.file('config.dart', '''
+        Map<String, dynamic> main() {
```
lrhn: I would not expect `Main` to have a return value.
The traditional ways to invoke `main` (as a script, with `Uri.spawnUri`) do not look at the returned value. 
If this `main`'s return is not ignored, I expect it's because this `main` is not the real entry point. Also supported by global teardowns being registered and run later on some signal.
(Does the function have to be named `main` then? Could you call it `setUp`, just so it _doesn't_ look like a `main` function. We have restrictions on `main` functions which probably won't get in the way, but maybe you'll want to allow an argument later, say parsed from the URI query, and then having to have a first argument typed `List<String>` can be a problem.)

pkgs/test_api/lib/src/scaffolding/global_setup.dart
```diff
@@ -0,0 +1,94 @@
...
+///   Interpreted relative to the directory containing the test suite file being executed.
+///   Note: When calling [globalSetup] from a shared helper library imported by tests in different directories,
+///   prefer root-relative (`/tool/...`) or `package:` URIs so the path resolves consistently regardless of which
+///   test file imports the helper.
+/// * **`file:` URIs** (e.g. `Uri.file('/abs/path/setup.dart')`):
+///   Interpreted as absolute file paths on the filesystem.
+///
+/// It is an error if [globalSetup] is called outside of `dart test`.
+///
+/// Throws whatever exception was thrown by the setup script if execution fails.
```
lrhn: (Does not sound possible. If it's a different script, it may throw objects of types that don't exist in this program. Checking .... I think it throws something that has the same `toString()` and `stackTrace.toString()`, but not the same object type.)

pkgs/test_api/lib/src/scaffolding/global_setup.dart
```diff
@@ -0,0 +1,94 @@
...
+///   Interpreted relative to the root of the package (the directory containing `pubspec.yaml`).
+/// * **Relative URIs** (paths without a scheme and without a leading `/`, e.g. `Uri.parse('setup.dart')` or `Uri.parse('../tool/setup.dart')`):
+///   Interpreted relative to the directory containing the test suite file being executed.
+///   Note: When calling [globalSetup] from a shared helper library imported by tests in different directories,
+///   prefer root-relative (`/tool/...`) or `package:` URIs so the path resolves consistently regardless of which
+///   test file imports the helper.
+/// * **`file:` URIs** (e.g. `Uri.file('/abs/path/setup.dart')`):
+///   Interpreted as absolute file paths on the filesystem.
+///
+/// It is an error if [globalSetup] is called outside of `dart test`.
```
lrhn: What does that mean?
Can you only run `dart test test/some_test.dart` and not `dart test/some_test.dart`?

If so, I'll be very annoyed when I hit that. :)

Also contradicts the comment:
> * **Standalone Execution Support:**
>    * Tests run directly via `dart <test_file>.dart` execute setup scripts in a standalone isolate and run teardowns on process exit.

pkgs/test_core/lib/src/bootstrap/vm.dart
```diff
@@ -55,3 +55,68 @@ void internalBootstrapNativeTest(
...
+  List<String> args,
+  SendPort sendPort,
+) async {
+  final tearDowns = <FutureOr<void> Function()>[];
+  final commandPort = ReceivePort();
+
+  try {
+    final mainFn = getMain();
+    final result = await runZoned(() async {
+      if (mainFn is FutureOr<Object?> Function(List<String>)) {
```
lrhn: What if it accepts two arguments?

A valid `main` function must be able to be called either:
- with zero arguments,
- with one positional argument which is a `List<String>`,
- with two positional arguments where the first is a `List<String>`

(Don't use `main`. You don't need those restrictions, they may get in your way one day.)

pkgs/test_core/lib/src/bootstrap/vm.dart
```diff
@@ -55,3 +55,68 @@ void internalBootstrapNativeTest(
...
+        return await mainFn(args);
+      } else if (mainFn is FutureOr<Object?> Function()) {
+        return await mainFn();
+      } else {
+        return await (Function.apply(mainFn, args.isEmpty ? [] : [args])
```
lrhn: That's not going to work, you can just throw instead.

A function that is neither `Object? Function(List<String>)` or `Object? Function()` will not succeed being called by `Function.apply` with zero arguments or one argument which is a `List<String>`.

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,203 @@
...
+              'Global setup "$url" failed:\n${list[0]}\n${list[1]}',
+            ),
+          );
+        }
+      });
+
+      final responseSub = responsePort.listen((response) {
+        final map = response as Map<Object?, Object?>;
+        if (map['success'] == true) {
+          if (map['hasTearDown'] == true && map['commandPort'] is SendPort) {
```
lrhn: Patterns are awesome, here too!
```dart
 if (map case {'hasTearDown': == true, 'commandPort': SendPort commandPort}) {
    ....
    commandPort: commandPort
```
with no redundant lookup or cast! 🥳

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,203 @@
...
+    _activeSetups.clear();
+  });
+
+  Future<String> _normalizeUrl(String url, Suite suite) async {
+    final parsedUri = Uri.parse(url);
+
+    String normalized;
+    switch (parsedUri.scheme) {
+      case '':
+        var isRootRelative = parsedUri.path.startsWith('/');
```
lrhn: [`... = parsedUri.hasAbsolutePath;`](https://api.dart.dev/dart-core/Uri/hasAbsolutePath.html)

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,203 @@
...
+          );
+        } else {
+          var suitePath = suite.path!;
+          normalized = p.url.join(
+            p.url.dirname(p.toUri(p.absolute(suitePath)).toString()),
+            parsedUri.toString(),
+          );
+        }
+      case 'package':
+        final resolvedUri = await Isolate.resolvePackageUri(parsedUri);
```
lrhn: There is a sync version of that. Looks like this call is the only reason the function is async.

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,203 @@
...
+        if (resolvedUri == null) {
+          throw ArgumentError.value(
+            url,
+            'uri',
+            'Could not resolve the package URI',
+          );
+        }
+        normalized = resolvedUri.toString();
+      default:
+        normalized = url;
```
lrhn, line 237: Does not remove query or fragment.
The no-scheme-absolute-path extracts the `path` of the URI and only uses that, which drops a query or fragment. The `package` scheme case will throw if there is a query, because package URIs can't have queries. May retain the fragment.
This one retains both.

Do something consistent. I'd throw if there is a fragment, or remove it using `.removeFragment()`. Throw if there is a query and the scheme is, or becomes, `file`.

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,203 @@
...
+    }
+    return Uri.parse(normalized).normalizePath().toString();
```
lrhn: Calling `normalizePath` does nothing. Sorry for the trap API :)

> The default implementation of Uri always normalizes paths, so calling this function has no effect.

pkgs/test_core/lib/src/util/io.dart
```diff
@@ -286,6 +286,18 @@ extension RetryDelete on FileSystemEntity {
...
 }
 
+/// Attempts to recursively delete [directory], ignoring any [IOException] that
+/// occurs (for example, if the directory is locked or does not exist).
+void tryDeleteDirectory(Directory directory) {
+  try {
+    if (directory.existsSync()) {
+      directory.deleteSync(recursive: true);
+    }
+  } on IOException {
```
lrhn: `FileSystemException` should be enough?

pkgs/test_core/lib/src/util/io.dart
```diff
@@ -286,6 +286,18 @@ extension RetryDelete on FileSystemEntity {
   }
 }
 
+/// Attempts to recursively delete [directory], ignoring any [IOException] that
+/// occurs (for example, if the directory is locked or does not exist).
```
lrhn: "or does not exist" is unlikely right after checking "existsSync()` ... I know it's _possible_, but not what I'd mention in the docs. 
For precision
```dart
/// Deletes [directory] and all its contents, if it exists.
///
/// The deletion is best-effort. If the directory, or any content in it, cannot be deleted,
/// the operation aborts silently.
```

natebosch:
This type of feature needs coordination to make sure it works in flutter_test and google3 testing as well before we land it. I'll take an initial pass on the code soon.

pkgs/test/test/runner/global_setup_test.dart
sigurdm: Good point. Renaming to setUp

pkgs/test_api/lib/src/scaffolding/global_setup.dart
sigurdm: Sorry, that was a limitation at an earlier stage of this PR. Didn't get to remove this comment.

sigurdm (commented):
Yes, standalone test execution (`dart test/some_test.dart`) is fully supported. Inside `test_core/lib/src/scaffolding.dart` it falls back to `standaloneGlobalSetup` using `globalSetupStandaloneFallback` if the test runner zone variable is absent. 

I've also removed the dartdoc snippet that erroneously claimed it isn't supported.

pkgs/test_api/lib/src/scaffolding/global_setup.dart
sigurdm: Updated the dartdoc to clarify this. It now states that it throws an exception containing the string representation and stack trace of the setup error, rather than promising to throw the original exception object type.

pkgs/test_core/lib/src/bootstrap/vm.dart
sigurdm: No longer using main

pkgs/test_core/lib/src/bootstrap/vm.dart
sigurdm: It now checks if (mainFn is FutureOr<Object?> Function(List<String>, SendPort)) (and the 1-arg / 0-arg variants). If it matches none of those signatures, we immediately throw a ArgumentError rather than relying.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Switched to sync version

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done

pkgs/test_core/lib/src/util/io.dart
sigurdm: This helper was dead code from some earlier version. Removed

sigurdm:
> (Does it have to be an isolate. Could it be run in a different process?)

Currently package:test leans heavily on isolates for script execution. Could we run it in a different process natively? Yes, but Isolate.spawnUri is currently used because it aligns perfectly with the test runner's existing architecture and shares the TestCompiler. Also, downstream consumers (like flutter tools / google3) are free to hook and execute these core messages differently in their environment if they need process isolation.

> But it still feels weird to use a script instead of just spawning an isolate to run a function... Why a script?

Because test suites in package:test can run in separate processes entirely (Node.js, Chrome, iOS/Flutter targets, etc.).  So we need the test script to communicate the setup with the test_runner.
Dart fundamentally cannot pass function closures/pointers across process boundaries back to the host test runner. Taking a script Uri is the only safe, cross-platform way (I think?) to instruct the unified host runner what to do.

> Where can you call globalSetup? Not inside test I assume. Inside a group? Inside setUp/setUpAll?

You can call it anywhere! It is a standard asynchronous dispatch. It works inside top-level main() (before a test), inside setUpAll, inside setUp, or directly inside a test() itself! As long as the execution enters the async context, it blocks on the isolate setup.

> (Can that script use a different package resolution? Be in a different package?)

Because we rely on Isolate.spawnUri, the setup script strictly shares the same .dart_tool/package_config.json resolution boundaries as the overarching test run. I'd argue that is the right behavior and equivalently to how we run the tests themselves!

It can be in another package (resolved by spawnUri)

natebosch:
> This type of feature needs coordination to make sure it works in flutter_test and google3 testing

I had Gemini try this out internally and google3 looks like it only needs fairly minor changes, but there is one that we need to double check the performance for. We'll still need to test flutter.

sigurdm:
I switched to using Isolate.spawnUri with a data url directly instead of compiling the globalSetup scripts with the frontendserver-based compiler. That should be compatible with current flutter test I believe

sigurdm:
Tested this end-to-end in a Flutter project by overriding `test_core` and `test_api` in `packages/flutter_tools`, and verified that both `globalSetup` and `addGlobalTearDown` work as expected under `flutter test`.

Summary of what is required on the Flutter side:

1. **No runtime code changes in `flutter_tools`**:
   The host runner picks up the `global-setup` IPC messages automatically from `test_core`. Both setup and teardown run cleanly on runner exit.

2. **Dependency bump in `packages/flutter_tools`**:
   - Roll `test_core` to the version containing this change.
   - `flutter_tools` currently pins `analyzer: 10.1.0`, while `test_core` requires `analyzer: >=13.0.0 <15.0.0`. Flutter will need to bump its `analyzer` dependency in `flutter_tools` to `>=13.0.0` as part of the roll (or `test_core` could widen its lower bound, since the code runs fine on 10.1.0 via `CompatibleArgument`).

3. **Surface `globalSetup` in Flutter tests**:
   - Because `package:test_api/scaffolding.dart` exports `test()`, importing it into a Flutter test causes a name collision with `package:flutter_test`.
   - Users will need to use `import 'package:test_api/scaffolding.dart' show globalSetup;`, or `package:flutter_test` could re-export `globalSetup`.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done: we now throw an `ArgumentError` if query parameters are present on file/relative URIs, and strip fragments using `.removeFragment()` on the returned URI.

sigurdm:
As a quick update: I refactored the implementation to share the isolate-spawning architecture with `spawnHybridUri` (bootstrapping via an in-memory `data:` URI with `Isolate.spawnUri`). This avoids the need for filesystem compilation or runner-specific compiler hooks.

Because of this, it works out of the box with `flutter test`, which delegates directly to `package:test_core`'s `test.main()` and awaits completion. I verified against `flutter/flutter:main` that the updated test packages roll cleanly with zero changes and passing checks.

All review comments have been addressed and CI is green—ready for another look when you have a moment.

pkgs/test/README.md
```diff
@@ -833,8 +833,70 @@ void main() {
...
+  test('uses shared server', () async {
+    final config = await globalSetup(Uri.parse('/test/setup_server.dart')) as Map;
+    final port = config['port'] as int;
+    // ...
+  });
+}
+```
+
+The setup script only executes once regardless of how many test suites or tests
+call `globalSetup()`. Global teardown callbacks registered with
```
lrhn, line 880: Is that "only one run per script URI" or also "only one script URI ever"?

Consider: "... tests call `globalSetup()` with that URL.".

pkgs/test_api/lib/src/scaffolding/global_setup.dart
```diff
@@ -0,0 +1,102 @@
...
+  channel.sink.add({
+    'type': 'global-setup',
+    'url': url,
+    'channel': virtualChannel.id,
+  });
+
+  final completer = Completer<Object?>();
+  virtualChannel.stream.listen(
+    (message) {
+      if (message is Map) {
```
lrhn, line 65: How about a switch with a map pattern? (Almost always a good way to parse JSON, except if there are optional fields.)
```dart
switch (message) {
  case {'type': 'data', 'data': final data}:
    completer.complete(data);
  case {'type': 'error', 'error': final Map errorEncoding}:
    final error = RemoteException.deserialize(errorEncoding);
    completer.completeError(error.error, error.stackTrace);
}
```
It's OK to do nothing on an unrecognize message?

(The `type` field isn't really needed. Can just look for either a `type` key or an `error` key.)

pkgs/test_api/lib/src/scaffolding/global_setup.dart
```diff
@@ -0,0 +1,102 @@
...
+
+  var virtualChannel = channel.virtualChannel();
+  channel.sink.add({
+    'type': 'global-setup',
+    'url': url,
+    'channel': virtualChannel.id,
+  });
+
+  final completer = Completer<Object?>();
+  virtualChannel.stream.listen(
```
lrhn, line 63: Looks like a listen that wants to be `cancelOnError: true`.

Both the `onData` and `onError` completes the completer unconditionally (and `onDone` too if not already completed), so you only ever expect one event, but you don't cancel the subscription when you get it.

I'd remember the subscription and have the `onData` cancel when it receives an event too. No ned to wait for an `onDone` which is then ignored anyway.

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+        );
+      } on IsolateSpawnException catch (error) {
+        throw ApplicationException(
+          'Global setup "$url" failed to compile:\n$error',
+        );
+      }
+
+      final completer = Completer<Object?>();
+
+      final errorSub = errorPort.listen((errorAndStack) {
```
lrhn, line 97: (Style guide says "No abbreviations". Didn't need to say that if it wasn't so tempting 😉. Just write `errorSubscription` out.)

pkgs/test/README.md
sigurdm: Clarified that it only runs once per URI.

pkgs/test_api/lib/src/scaffolding/global_setup.dart
sigurdm: Switched to map pattern matching and added a default branch to fail with a `StateError` on unexpected messages.

pkgs/test_core/lib/src/bootstrap/vm.dart
```diff
@@ -55,3 +55,82 @@ void internalBootstrapNativeTest(
...
+
+    if (tearDowns.isEmpty) {
+      commandPort.close();
+      return;
+    }
+
+    await for (var msg in commandPort) {
+      final tuple = msg is List ? msg : [null, msg];
+      final replyPort = tuple[0] as SendPort?;
+      final command = tuple[1];
```
lrhn, line 101: Consider:
```dart
final (replyPort, command) = switch (msg) {
   [final replyPort, final command] => (replyPort, command),
   _ => (null, msg).
});
```
(Doesn't throw if it's a list with length different from 2. Don't know if that's a good thing or a bad thing.)

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+      }
+    }());
+  }
+
+  Future<Object?> _runSetup(String url) async {
+    final responsePort = ReceivePort();
+    final errorPort = ReceivePort();
+    try {
+      final code =
+          '''
```
lrhn, line 71: Indentation doesn't match code lines, maybe indent them a little more?

pkgs/test_core/lib/src/runner/global_setup.dart
lrhn, line 97: If the only thing you do with the port is `listen` for data, and you only use the subscription to cancel it, you could just use a `RawReceivePort`. 
Set the handler instead of calling `listen`, and close the port instead of cancelling the subscription. Less overhead without the stream wrapper.

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+      return result;
+    } finally {
+      responsePort.close();
+      errorPort.close();
+    }
+  }
+
+  Future<void> close() => _closeMemo.runOnce(() async {
+    for (var active in _activeSetups.reversed) {
+      final replyPort = ReceivePort();
```
lrhn, line 156: Could maybe reuse the receive port.
(If we trust the code at the other end to only send one response per command, we can just respond to the each message in turn. Or use a `RawReceivePort`, or a `StreamIterator` on the response port.)

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+          if (parsedUri.hasQuery) {
+            throw ArgumentError.value(
+              url,
+              'uri',
+              'root-relative URIs cannot have query parameters',
+            );
+          }
+          normalized = p.url.join(
+            p.toUri(p.current).toString(),
+            parsedUri.path.substring(1),
```
lrhn, line 201: This is to get rid of the leading `/` because `join` doesn't actually join, it "resolves" and throws away the first part if the second starts with `/`. Right?

We haven't checked if the URI has an authority. If it does, then that's thrown away, but `path` can also start with `//` (or more!). So might need to skip multiple leading `/`s.

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+            throw ArgumentError.value(
+              url,
+              'uri',
+              'relative URIs cannot have query parameters',
+            );
+          }
+          var suitePath = suite.path!;
+          normalized = p.url.join(
+            p.url.dirname(p.toUri(p.absolute(suitePath)).toString()),
+            parsedUri.path,
```
lrhn, line 214: Doesn't check if `parsedUri.path` starts with `..`. Should it?

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+    }
+    _activeSetups.clear();
+  });
+
+  String _normalizeUrl(String url, Suite suite) {
+    final parsedUri = Uri.parse(url);
+
+    String normalized;
+    switch (parsedUri.scheme) {
+      case '':
```
lrhn, line 190: (Total nit!)
Would use `Uri.hasScheme` to check for this case. If you want to do it inside a switch, you can do:
```dart
  switch (parsedUri) {
     case Uri(hasScheme: false, hasAbsolutePath: true, hasQuery: true):   
         throw ...
     case Uri(hasScheme: false): 
         normalized = ...
     case Uri(scheme: 'file'): ... etc
```

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+      default:
+        normalized = url;
+    }
+
+    return Uri.parse(normalized).removeFragment().toString();
+  }
+}
+
+Future<String> _readUri(Uri uri) async => switch (uri.scheme) {
+  '' || 'file' => await File.fromUri(uri).readAsString(),
```
lrhn, line 245: If you do:
```dart
  switch (uri) {
    Uri(hasScheme: false) || Uri(scheme: 'file) => ...
    Uri(: final data?) => data.contentAsString(),
    _ => throw ... 
  };
```
you don't need the `!` on `uri.data!`.
(I really like getting rid of `!`s!!)

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+    'Only data and file uris (as well as relative paths) are supported',
+  ),
+};
+
+Future<String> _languageVersionCommentFor(String url) async {
+  var parsedUri = Uri.parse(url);
+
+  var result = parseString(
+    content: await _readUri(parsedUri),
+    path: parsedUri.scheme == 'data' ? null : p.fromUri(parsedUri),
```
lrhn, line 259: Use `uri.isScheme('data')` instead of `uri.scheme == 'data'` when possible. 
If it's a `SimpleUri`, reading scheme is a string allocation. Also does the check properly case-insensitively.

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+    throwIfDiagnostics: false,
+  );
+  var languageVersionComment = result.unit.languageVersionToken?.value();
+  if (languageVersionComment != null) return languageVersionComment.toString();
```
lrhn, line 263: If-case?
```dart
  if (result.unit.languageVersionToken?.value() case final versionComment?) {
    return versionComment.toString();
  }
```

Does cost an extra line, but limits the scope of the variable.

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+
+  if (parsedUri.scheme == 'file' || parsedUri.scheme == '') {
```
lrhn, line 265: `!parsedUri.hasScheme || parsedUri.isScheme('file')`

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+    var packageConfig = await currentPackageConfig;
+    var package = packageConfig.packageOf(parsedUri);
+    var version = package?.languageVersion;
+    if (version != null) return '// @dart=$version';
```
lrhn, line 269: `if (package?.langaugeVersion case var version?) return '// @dart=$version';`

(Might just fit on a line!)

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  if (parsedUri.scheme == 'data') {
```
lrhn, line 272: `.isScheme('data')`.

pkgs/test_core/lib/src/runner/parse_metadata.dart
```diff
@@ -330,12 +331,12 @@ class _Parser {
...
-      {
-        for (var argument in arguments)
-          if (argument is NamedArgument)
-            argument.name.lexeme: argument.argumentExpression,
-      };
+  Map<String, Expression> _parseNamedArguments(
+    Iterable<CompatibleArgument> arguments,
+  ) => {
+    for (var argument in arguments)
+      if (argument.name case String name) name: argument.argumentExpression,
```
lrhn, line 338: (The `name` is just a `String?`, so you can use a null-aware map element instead of the `if`: `?argument.name: argument.argumentExpression`.)

lrhn (approved):

pkgs/test_api/lib/src/scaffolding/global_setup.dart
sigurdm: Done. Stored the subscription to cancel on data and set cancelOnError: true.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Switched to RawReceivePort, which eliminates the subscriptions and cancellation overhead altogether.

pkgs/test_core/lib/src/bootstrap/vm.dart
sigurdm: Done. Switched to pattern matching destructuring on the list.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Aligned the template code indentation.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Switched to RawReceivePort for both response and error ports.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Switched to RawReceivePort and close it cleanly in finally.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. We now reject URIs with authorities and use pathSegments to cleanly strip leading slashes.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: For relative URIs (e.g. '../setup.dart'), '..' is intentionally allowed to reference shared helpers in parent directories. For root-relative paths, we now reject paths starting with '/..'.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Switched to pattern matching on Uri.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Pattern matching on Uri(:final data?) removes the !.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Switched to .isScheme('data').

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Switched to if-case.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Switched to !parsedUri.hasScheme || parsedUri.isScheme('file').

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Switched to if-case.

pkgs/test_core/lib/src/runner/global_setup.dart
sigurdm: Done. Switched to .isScheme('data').

pkgs/test_core/lib/src/runner/parse_metadata.dart
sigurdm: Done. Switched to null-aware map element ?argument.name: argument.argumentExpression.

sigurdm:
@natebosch are you ok with the PR as it stands?

pkgs/test/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## 1.32.0-wip
 
+* Support `globalSetup` and `addGlobalTearDown` in `package:test` /
+  `package:test_api` to execute global setup scripts once on the host runner
+  and cache results across all tests and test suites.
```
natebosch, line 5: [nit]
```suggestion
* Support `globalSetup` and `addGlobalTearDown` to execute global setup scripts
  once on the host runner and cache results across all tests and test suites.
```

pkgs/test/test/runner/global_setup_test.dart
```diff
@@ -0,0 +1,544 @@
...
+        'test/test2_test.dart',
+      ]);
+      expect(test.stdout, emitsThrough(contains('+2: All tests passed!')));
+      await test.shouldExit(0);
+
+      expect(File('${d.sandbox}/counter.txt').readAsStringSync(), equals('x'));
+    },
+  );
+
+  test('supports Map and JSON-encodable return values', () async {
```
natebosch, line 70: We should have a test for what happens when it is not JSON encodable. We should make sure the error is clearly associated to the `globalSetup` call or the Uri of the setup script that errors.

pkgs/test_api/lib/src/scaffolding/global_setup.dart
```diff
@@ -0,0 +1,110 @@
...
+Future<Object?> Function(Uri uri)? globalSetupStandaloneFallback;
+
+/// Runs a global setup Dart script at [uri] on the host test runner if it
+/// has not yet run, and returns its result.
+///
+/// If multiple test suites or tests invoke [globalSetup] with the same [uri],
+/// the script runs only once and all callers receive the same cached result.
+///
+/// The Dart file at [uri] must define a top-level `setUp()` function that
+/// returns a JSON-encodable value (or `Future` of one).
```
natebosch, line 23: Can we expand the docs around this value? We might mention that `void setUp()` is OK and results in a `null` value. We might also mention the memoization detail that is mentioned in the README, and that each call to `globalSetUp` gets an independent value returned (mutations are not shared across calls).

pkgs/test_api/lib/src/scaffolding/global_setup.dart
```diff
@@ -0,0 +1,110 @@
...
+  return completer.future;
+}
+
+/// Registers a callback to be run during global teardown when the test runner
+/// closes.
+///
+/// This must be called from within a global setup script invoked via [globalSetup].
+///
+/// It is an error if this is called outside of a global setup script.
+void addGlobalTearDown(FutureOr<void> Function() callback) {
```
natebosch, line 101: I wonder if we should move this to a separate top level library. A global setup script can't use most of `test/scaffolding.dart`, and a test can't use `addGlobalTearDown`.

WDYT about adding `package:test/global.dart` for just this API?

pkgs/test_core/lib/src/runner/loader.dart
```diff
@@ -81,6 +108,8 @@ class Loader {
     _config.validateRuntimes(allRuntimes);
 
     _registerRuntimeOverrides();
+
+    GlobalSetupManager.current = globalSetupManager;
```
natebosch, line 112: As much as zones cause confusion - since we're already using zones to communicate stuff like `Configuration.current` using `Configuration.asCurrent`. Can we do similar here?

pkgs/test_core/lib/src/runner/parse_metadata.dart
```diff
@@ -144,7 +144,7 @@ class _Parser {
   /// [annotation] is the annotation.
   PlatformSelector _parseTestOn(Annotation annotation) =>
       _parsePlatformSelector(
-        annotation.arguments!.arguments.first.argumentExpression,
+        annotation.arguments!.compatibleArguments.first.argumentExpression,
```
natebosch, line 147: How are the changes in this file related? Was this a mistake during a merge from master?

pkgs/test_core/lib/src/runner/loader.dart
```diff
@@ -26,12 +28,21 @@ import 'plugin/environment.dart';
...
 
 /// A class for finding test files and loading them into a runnable form.
 class Loader {
   /// The test runner configuration.
   final _config = Configuration.current;
 
+  /// The compiler used for compiling VM tests and hooks incrementally.
+  final _compiler = TestCompiler(
+    p.join(p.current, '.dart_tool', 'test', 'incremental_kernel'),
+  );
```
natebosch, line 41: I don't think we should be leaking VM specific concepts like this into `Loader`. Can this be moved back?

pkgs/test/README.md
```diff
@@ -833,8 +833,70 @@ void main() {
...
 
+## Global Setup and Teardown
+
+If tests across multiple files or test suites require shared setup (such as
+precompiling binaries, provisioning test resources, or starting backend services),
+you can use [`globalSetup()`]. Setup scripts should typically live inside your `test/`
+directory (for example, `test/setup_server.dart` or `test/helpers/setup.dart`).
+
+The `globalSetup()` function executes a Dart script on the host test runner in an
+isolate compiled with the incremental Frontend Server. The script's `setUp()`
```
natebosch, line 844: Is the detail about the incremental Frontend Server still true or was that an earlier iteration?

pkgs/test/README.md
```diff
@@ -833,8 +833,70 @@ void main() {
...
+  });
+}
+```
+
+The setup script for a given URI only executes once regardless of how many test
+suites or tests call `globalSetup()` with that URI. Global teardown callbacks
+registered with [`addGlobalTearDown()`] are executed when the test runner closes.
+
+`globalSetup()` takes a [`Uri`][Uri] resolved with the following rules:
+* **Root-relative URIs** (beginning with `/`, like `Uri.parse('/test/setup.dart')`):
```
natebosch, line 884: [nit] I prefer using `Uri(path:'/test/setup.dart')` style Uri constructors for anything with hardcoded strings, reserving `Uri.parse` for the situations where all I have is an single non-hardcoded string. I think the docs and examples should do the same.

natebosch (changes requested):

pkgs/test/README.md
```diff
@@ -833,8 +833,70 @@ void main() {
...
+import 'package:test/scaffolding.dart' show globalSetup;
+import 'package:test/test.dart';
+
+void main() {
+  test('uses shared server', () async {
+    final config = await globalSetup(Uri.parse('/test/setup_server.dart')) as Map;
+    final port = config['port'] as int;
+    // ...
+  });
+}
```
natebosch, line 876: What happens if they call `globalSetup` outside of a `test`? I suspect that will be a common pattern given how it matches a `setUp` in `main`.
```dart
void main() {
  final config = await globalSetup(Uri.parse('/test/setup_server.dart')) as Map;
  test('uses shared server', () async {
    final port = config['port'] as int;
    // ...
  });
}
```

natebosch: It might make sense to disallow this and require it's called from a `test` or a `setUp`/`setUpAll`. If we want to do that we should document it clearly.

pkgs/test/CHANGELOG.md
sigurdm: Done.

pkgs/test/test/runner/global_setup_test.dart
sigurdm: Done. Added `json.encode(result)` validation in `internalBootstrapVmHook` to fail fast with an error referencing the script URI, and added a test in `global_setup_test.dart`.

pkgs/test_api/lib/src/scaffolding/global_setup.dart
sigurdm: Done, expanded doc comments to cover `void setUp()`, memoization, and caller value independence.

pkgs/test_api/lib/src/scaffolding/global_setup.dart
sigurdm: Added `package:test/global.dart` and `package:test_api/global.dart` exporting `addGlobalTearDown` (and `globalSetup`), and removed `addGlobalTearDown` from `scaffolding.dart`.

pkgs/test_core/lib/src/runner/loader.dart
sigurdm: Done, removed the mutable static assignment and made `GlobalSetupManager.current` purely zone-scoped via `asCurrent`.

pkgs/test_core/lib/src/runner/parse_metadata.dart
sigurdm: That was an older workaround for an earlier analyzer constraint in Flutter that is no longer needed. Reverted to match `master` exactly.

pkgs/test_core/lib/src/runner/loader.dart
sigurdm: Done, moved back to `VMPlatform` and restored `VMPlatform.new`.

pkgs/test/README.md
sigurdm: That was from the earlier iteration. Removed.

pkgs/test/README.md
sigurdm: Done, updated the examples and docs to use `Uri(path: ...)`.

pkgs/test/README.md
sigurdm: Added an `Invoker.current` check to throw a descriptive `StateError` if called directly from `main()`, documented the requirement, and added a test for it.

pkgs/test_api/lib/global.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Utilities for shared global setup and teardown across test suites.
+library;
+
+export 'src/scaffolding/global_setup.dart' show addGlobalTearDown, globalSetup;
```
natebosch:
```suggestion
export 'src/scaffolding/global_setup.dart' show addGlobalTearDown;
```

pkgs/test_api/lib/test_api.dart
```diff
@@ -9,4 +9,4 @@
 library;
 
 export 'hooks.dart' show TestFailure;
-export 'scaffolding.dart';
+export 'scaffolding.dart' hide globalSetup;
```
natebosch: Can you switch this to a `show` with a positive list of included members?

pkgs/test/README.md
```diff
@@ -855,8 +855,74 @@ void main() {
...
+        await globalSetup(Uri(path: '/test/setup_server.dart')) as Map;
+    final port = config['port'] as int;
+    // ...
+  });
+}
+```
+
+The setup script for a given URI only executes once regardless of how many test
+suites or tests call `globalSetup()` with that URI. Each call receives an
+independent copy of the return value, so mutations to collections are not shared
```
natebosch: Is this guaranteed? Are they reused for the direct run use case? I wonder if we should json encode in all cases and cache the encoded result to make this true? It would also avoid differences between platforms if they all return collections resulting from a json decode.

pkgs/test_core/lib/src/runner/vm/test_compiler.dart
```diff
@@ -57,7 +57,11 @@ class TestCompiler {
 
   /// Compiles [mainDart], using a separate compiler per language version of
   /// the tests.
-  Future<CompilationResponse> compile(Uri mainDart, Metadata metadata) async {
+  Future<CompilationResponse> compile(
+    Uri mainDart,
+    Metadata metadata, {
+    VmTestType testType = VmTestType.isolate,
```
natebosch: Is there a use for this argument? Can we go back to hardcoding it below?

pkgs/test_api/lib/src/scaffolding/global_setup.dart
```diff
@@ -0,0 +1,125 @@
...
+/// Relative, root-relative, and file URIs strictly prohibit query parameters.
+///
+/// If execution fails, throws an exception containing the string representation
+/// and stack trace of the error thrown by the setup script.
+Future<Object?> globalSetup(Uri uri) async {
+  if (Invoker.current == null && globalSetupStandaloneFallback == null) {
+    throw StateError(
+      'globalSetup() must be called from within a test or setUp/setUpAll callback.',
+    );
+  }
```
natebosch: I think this condition is too narrow. `globalSetupStandaloneFallback` may be non-null despite not being in a `test` or `setUp` body. I think this design allows

```dart
void main() {
  setUp(() {});
  globalSetUp('');
}
```

because after the `setUp` call (_before_ setup callback is invoked) the `globalSetupStandaloneFallback` will be non-null.


```suggestion
  if (Invoker.current == null) {
    throw StateError(
      'globalSetup() must be called from within a test or setUp/setUpAll callback.',
    );
  }
```

pkgs/test_core/lib/src/runner/standalone_global_setup.dart
```diff
@@ -0,0 +1,225 @@
...
+    normalizedUrl,
+    () => _runStandaloneSetup(normalizedUrl),
+  );
+}
+
+Future<Object?> _runStandaloneSetup(String url) async {
+  final scriptUri = Uri.parse(url);
+  _standaloneTempDir ??= await Directory.systemTemp.createTemp(
+    'global_setup_standalone_',
+  );
```
natebosch: It isn't safe to use an `await` in an expression using a `??=` for a do-once step.


```suggestion
  _standaloneTempDir ??= Directory.systemTemp.createTempSync(
    'global_setup_standalone_',
  );
```

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,291 @@
...
+
+  /// The current global setup manager set using [asCurrent].
+  static GlobalSetupManager? get current =>
+      Zone.current[_currentKey] as GlobalSetupManager?;
+
+  final _setups = <String, Future<Object?>>{};
+  final _activeSetups = <_ActiveSetup>[];
+  final _closeMemo = AsyncMemoizer<void>();
+
+  GlobalSetupManager();
```
natebosch: [nit] Shouldn't need to define the default constructor unless I'm overlooking another one.


```suggestion
```

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,291 @@
...
+
+    try {
+      final code =
+          '''
+${await _languageVersionCommentFor(url)}
+
+import "dart:isolate";
+import "package:test_core/src/bootstrap/vm.dart";
+
+import "${url.replaceAll(r'$', '%24')}" as lib;
```
natebosch: Why do we need this `replaceAll`? Do we also need it on the `import '$scriptUri' as test;` line in the standalone implementation?

Are `$` characters in file names something we've seen in practice, and does this replacement work?

pkgs/test_core/lib/src/runner/global_setup.dart
```diff
@@ -0,0 +1,291 @@
...
+          'relative URIs cannot have an authority',
+        ),
+      Uri(hasScheme: false, hasAbsolutePath: true, hasQuery: true) =>
+        throw ArgumentError.value(
+          url,
+          'uri',
+          'root-relative URIs cannot have query parameters',
+        ),
+      Uri(hasScheme: false, hasAbsolutePath: true) => () {
+        if (url.startsWith('/..')) {
```
natebosch: Do we know this url is already normalized? If so, we should document it, maybe with an `assert`.

If not, we need to normalize this if we want a `startsWith` to reliably identify paths that step up. `/foo/../../escaped`.

pkgs/test_core/lib/src/runner/standalone_global_setup.dart
```diff
@@ -0,0 +1,225 @@
...
+      } else {
+        if (uri.hasQuery) {
+          throw ArgumentError.value(
+            uri,
+            'uri',
+            'relative URIs cannot have query parameters',
+          );
+        }
+        // Relative to current directory/entrypoint
+        normalized = p.url.join(p.toUri(p.current).toString(), uri.path);
```
natebosch: Are we sure that `cwd` is the right way to target URIs? I don't recall if there was already discussion about this...

The PR opening comment says:

> Resolved relative to the executing test suite's directory.

Is that no longer true or is that happening somewhere else?

natebosch (changes requested):

PR #2715, opened 2026-08-06
Fix race conditions in Chrome coverage collection

sigurdm:
Fixes race conditions causing intermittent empty coverage results in Chrome tests (e.g. in `test/runner/coverage_test.dart`):

1. **Await DevTools & profiler initialization before suite execution**: `BrowserManager.load` now awaits `Chrome.whenReady` before sending `'loadSuite'`, ensuring `Profiler.startPreciseCoverage` is active before the test script executes in the browser.
2. **Listen to `onScriptParsed` before enabling debugger**: Registers the `onScriptParsed` listener before calling `tabConnection.debugger.enable()`, preventing initial `Debugger.scriptParsed` events from being missed on the broadcast stream.

### DIFF
--- a/pkgs/test/lib/src/runner/browser/browser_manager.dart
+++ b/pkgs/test/lib/src/runner/browser/browser_manager.dart
@@ -133,9 +133,12 @@ class BrowserManager {
         });
 
     future
-        .then((webSocket) {
+        .then((webSocket) async {
           if (completer.isCompleted) return;
-          completer.complete(BrowserManager._(browser, runtime, webSocket));
+          var manager = BrowserManager._(browser, runtime, webSocket);
+          await manager._environment;
+          if (completer.isCompleted) return;
+          completer.complete(manager);
         })
         .onError((Object error, StackTrace stackTrace) {
           browser.close();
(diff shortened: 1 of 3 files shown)


pkgs/test/lib/src/runner/browser/browser_manager.dart
```diff
@@ -268,6 +268,10 @@ class BrowserManager {
     );
 
     var suite = _pool.withResource<RunnerSuite>(() async {
+      if (_browser case Chrome chrome) {
+        await chrome.whenReady;
+      }
```
natebosch: I'd like to avoid a Chrome specific type check here.

After working out an alternative suggestion it was easier to just open it as a separate PR we can merge into this one if you agree with the direction.

https://github.com/dart-lang/test/pull/2716

natebosch (changes requested):

pkgs/test/lib/src/runner/browser/browser_manager.dart
sigurdm: Yeah, that is much better! Feel free to merge into this one.

natebosch (approved):

PR #2722, opened 2026-08-10
Keep the browser executable env override through a settings merge

### LINKED ISSUE #2609: Document `CHROME_EXECUTABLE` and friends
Searching the code it appears to me that it's possible to override default browser locations using environment variables:
* `CHROME_EXECUTABLE`
* `MS_EDGE_EXECUTABLE`
* `FIREFOX_EXECUTABLE`
* `SAFARI_EXECUTABLE`

But from what I can see `CHROME_EXECUTABLE` the only one documented, and its only documented in the `CHANGELOG.md` 🤣 

I propose that we:
 * Manually attempt to verify that this works.
 * Document it extremely briefly in `README.md`
 * Document it a bit more thoroughly in [doc/configuration.md](https://github.com/dart-lang/test/blob/master/pkgs/test/doc/configuration.md)

I suspect the [Configuring Platforms](https://github.com/dart-lang/test/blob/master/pkgs/test/doc/configuration.md#configuring-platforms) section is a good place to put it, perhaps under the `executable` key explanation.

------------

IMO, these environment variables are extremely useful when trying to run in a custom CI environment. Where you might not want to hardcode the executable path into `dart_test.yaml`.

Yusufihsangorgel:
Fixes #2609

The docs half is the easy part. Neither `README.md` nor
`configuration.md` named the four variables. Both get them now, briefly
in the first and with the precedence spelled out in the second. The two
links to the browser settings section used the wrong anchor and went
nowhere.

Writing them down turned up a bug. `ExecutableSettings.merge` dropped
`_environmentOverride` when it rebuilt the settings, and
`customizePlatform` merges the defaults with whatever you configured.
Any browser setting at all silenced `CHROME_EXECUTABLE`, not just
`executable:`. `arguments:` on its own was enough. The fix is one line
in `merge`.

The new test in `custom_platform_test.dart` configures chrome with
`arguments:` and points `CHROME_EXECUTABLE` at a path that does not
exist. Without the fix the inner run launches real Chrome and exits 0.

Two tests in that file now set the variable themselves. 1.23.0
documented that it outranks the config file, and with the override
reaching the getter again, a machine that has it set would run a real
browser instead of the missing path they assert on. They point it at the
same missing path, which keeps their assertion working. What they no
longer cover is the key itself: the getter returns as soon as it finds
the variable and never reads `executable:`. `test_process` does expose
`includeParentEnvironment`, and `runTest` could pass it through to drop
one variable instead; pinning was the smaller change. Happy to do it
that way if you would rather those two kept testing the key.

### DIFF
--- a/pkgs/test/CHANGELOG.md
+++ b/pkgs/test/CHANGELOG.md
@@ -19,6 +19,10 @@
   restricted temporary authentication file to prevent secret leakage in command
   line process arguments.
 * Enable asserts in the `exe` compiler.
+* Keep the `CHROME_EXECUTABLE`, `MS_EDGE_EXECUTABLE`, `FIREFOX_EXECUTABLE`, and
+  `SAFARI_EXECUTABLE` overrides working when the browser is also configured
+  through `override_platforms` or `define_platforms`. Any setting on those
+  platforms used to drop the variable.
 
 ## 1.31.2
 
--- a/pkgs/test/README.md
+++ b/pkgs/test/README.md
@@ -633,7 +633,10 @@ last, just as they would in nested groups. This means that for configuration
 like duration-based timeouts, the last matching value wins.
 
 You can also set up global platform-specific configuration using the
-[package configuration file][configuring platforms].
+[package configuration file][configuring platforms]. The `CHROME_EXECUTABLE`,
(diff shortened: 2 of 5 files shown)


natebosch (approved):
Thanks!

PR #2723, opened 2026-08-13
Export TestFailure from package:test_api/scaffolding.dart

### LINKED ISSUE #2073: Should `fail` and `TestFailure` be exposed by scaffolding.dart?
I have a test that looks like:

```dart
try {
  functionExpectedToFail();
  fail('Failed to fail.');
} on ExpectedException catch (e) {
  // Significant code that extracts properties of `e` and that verifies the results...
}
```

While it's probably possible to rewrite the test to use a more typical `check(functionExpectedToFail).throws<ExpectedException>()` style, there's a significant amount of code in my `catch` block, so I think it's more readable with the explicit `try`-`catch`.

However, `fail` (and the `TestFailure` exception it wants to throw) are not exposed by `package:test/scaffolding.dart`.  Should they be?  Is there something else I should use instead?

(And yes, I also could just explicitly throw some other exception.)

kevmoo:
Re-export `TestFailure` in `package:test_api/scaffolding.dart` (which in turn
exports it in `package:test/scaffolding.dart`).

This allows users migrating from `package:test/test.dart` to
`package:test/scaffolding.dart` (such as when using `package:checks`) to
throw or handle `TestFailure` exceptions without needing an explicit import
of `package:test_api/hooks.dart`.

Closes #2073

### DIFF
--- a/pkgs/checks/CHANGELOG.md
+++ b/pkgs/checks/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.3.2-wip
 
+- Bump `test` dev dependency constraint.
 - Add `isNotA<R>()` check extension as a convenience in place of
   `not((it) => it.isA<R>())`.
 - Require Dart 3.11
--- a/pkgs/checks/pubspec.yaml
+++ b/pkgs/checks/pubspec.yaml
@@ -17,4 +17,4 @@ dependencies:
   test_api: ">=0.5.0 <0.8.0"
 
 dev_dependencies:
-  test: ^1.21.3
+  test: ^1.32.0-wip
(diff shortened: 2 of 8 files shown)


kevmoo:
Blocked on https://github.com/dart-lang/sdk/issues/64049 (MSAN crash in `dartaotruntime_msan` on `3.14.0-112.0.dev`).

natebosch:
What's your opinion on whether we should move a copy of `Never fail(String)` into `test/scaffolding.dart`?

natebosch (approved):

natebosch:
Closes #2073

kevmoo:
> What's your opinion on whether we should move a copy of `Never fail(String)` into `test/scaffolding.dart`?

We can't bring it in from matcher. If we re-create it then we'll have the opportunity for a name conflict. I figured this was the easy thing to do!

PR #2738, opened 2026-08-27
Remove redundant tag

natebosch:
This tag was added along with support for VM debugging, but it was
always redundant and likely introduced because of confusion over whether
it would impact whether this test case runs for VM tests. There is no
configuration in `dart_test.yaml` associated with this tag.

Removes a warning to make the test output be more fully grouped in the
github reporter output.


kevmoo (approved):
LGTM - but CI is sad

natebosch:
> CI is sad

We've picked up some flakes recently 😢

PR #2740, opened 2026-08-28
Fix deadlock waiting on serverSocket

natebosch:
Work around, and reduce the odds of triggering, a bug in the Dart SDK
implementation of unix domain sockets on windows.

https://github.com/dart-lang/sdk/issues/64166

On windows in Dart sockets are being inherited by subprocesses, so
spawned test processes (like the chrome browser) would inherit them, and
subsequently after receiving a connection the socket cancel would never
complete (waiting on subprocess which inherited the socket to exit) so
the `Stream.first` future would also never complete. This surfaced as a
deadlock only in the case where a browser startup happens while the
socket server is waiting for the first connection.

Drastically reduce the odds of this situation by waiting to bind the
socket server until the subprocess is about to start, instead of
listening on the socket during the compilation process as well.

Work around the situation in case it still happens with a `fastFirst`
extension which behaves like `first` except it explicitly does not await
the Future returned from `cancel`.

Add an integration test which demonstrates the conditions required to
trigger the bug. We can remove this after the SDK is fixed.


integration_tests/cross_compiler_hang/README.md
```diff
@@ -0,0 +1,96 @@
...
+```dart
+var serverSocket = await ServerSocket.bind(
+  InternetAddress(socketPath, type: InternetAddressType.unix),
+  0,
+);
+var process = await _spawnExecutable(...);
+var socket = await serverSocket.first;
+```
+
+On Windows, this interacts with two platform-level behaviors:
```
brianquinlan, line 23: Optional: Are you going to clean up this text when your SDK fix is live? Maybe you could just point to an issue?

pkgs/test_core/lib/src/runner/vm/platform.dart
```diff
@@ -618,3 +612,18 @@ Future<Set<String>> _filterCoveragePackages(
         .toSet();
   }
 }
+
+// TODO(https://github.com/dart-lang/sdk/issues/64166) remove when the min SDK
+// does not leak sockets to subproceesses.
```
brianquinlan: You could add the note to remove the test or at least remove the docs for the test here.

brianquinlan (approved):
That's a bit ugly but I don't have a better solution.

integration_tests/cross_compiler_hang/README.md
natebosch: I think it would be OK to drop the whole directory after the min SDK dependency has a fix.

pkgs/test_core/lib/src/runner/vm/platform.dart
natebosch: Added a note to remove the test.

PR #2743, opened 2026-08-30
feat(checks): add prints expectations (#2742)

### LINKED ISSUE #2742: package:checks: Add prints expectation for checking captured output
In `package:matcher`, `prints(...)` captures `print` statements during callback execution via `ZoneSpecification(print: ...)` and validates the printed output against a string or matcher:

```dart
expect(() => print("hello"), prints("hello\n"));
```

In `package:checks`, there is currently no direct replacement for `prints`. The migration document (`pkgs/checks/doc/migrating_from_matcher.md`) lists:
> `- prints`: TODO add missing expectation? Is this one worth replacing?

Migrating tests that assert on console or logging output requires either maintaining a dependency on `package:matcher` / `package:test/expect.dart` or writing custom `runZoned` capture boilerplate.

### Proposed Solution

Add a `prints` expectation extension on `Subject<void Function()>` (and async callback subjects) that executes the callback in a `Zone` intercepting `print()`, collects the output, and provides a `Subject<String>` or accepts a condition on the captured output.

Example usage:

```dart
check(() {
  print("hello");
}).prints(it()..equals("hello\n"));

// Or with string containment / pattern matching:
check(() {
  print("foo");
  print("bar");
}).prints(it()..contains("foo"));
```

### Acceptance Criteria

- [ ] Provide a `prints` expectation on synchronous and asynchronous callback `Subject`s.
- [ ] Capture all `print` invocations executed within the callback body using a `ZoneSpecification`.
- [ ] Provide a nested `Subject<String>` for the concatenated printed output to allow chaining standard `String` checks (`equals`, `contains`, `matchesPattern`, etc.).
- [ ] Support both synchronous (`void Function()`) and asynchronous (`FutureOr<void> Function()`) callbacks.
- [ ] Update `pkgs/checks/doc/migrating_from_matcher.md` to document the replacement.

kevmoo:
Add `prints()` check extensions on `Subject<void Function()>` (`FunctionPrints`) and `Subject<Future<void> Function()>` (`AsyncFunctionPrints`) to validate captured `print()` output from synchronous and asynchronous callbacks.

- Intercept `print()` via `ZoneSpecification` during function execution.
- `FunctionPrints.prints()` (on `Subject<void Function()>`) returns `Subject<String>` for synchronous checking on captured output and asserts that the callback does not return a `Future` at runtime.
- `AsyncFunctionPrints.prints()` (on `Subject<Future<void> Function()>`) awaits the returned `Future` before checking `Condition<String>` and returning `Future<Subject<String>>`.
- Add runnable `{@example}` files, update migration guide, and add tests for output capture and error rejections.

Closes #2742

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -30,7 +30,7 @@ jobs:
             os:ubuntu-latest;pub-cache-hosted
             os:ubuntu-latest
       - name: Setup Dart SDK
-        uses: dart-lang/setup-dart@65eb853c7ba17dde3be364c3d2858773e7144260
+        uses: dart-lang/setup-dart@6afc89df92d6eb3834022f73cd65adc8cdfcb92d
         with:
           sdk: stable
       - id: checkout
@@ -57,7 +57,7 @@ jobs:
             os:ubuntu-latest;pub-cache-hosted
             os:ubuntu-latest
       - name: Setup Dart SDK
-        uses: dart-lang/setup-dart@65eb853c7ba17dde3be364c3d2858773e7144260
+        uses: dart-lang/setup-dart@6afc89df92d6eb3834022f73cd65adc8cdfcb92d
         with:
           sdk: "3.11.0"
       - id: checkout
@@ -125,7 +125,7 @@ jobs:
             os:ubuntu-latest;pub-cache-hosted
             os:ubuntu-latest
       - name: Setup Dart SDK
-        uses: dart-lang/setup-dart@65eb853c7ba17dde3be364c3d2858773e7144260
+        uses: dart-lang/setup-dart@6afc89df92d6eb3834022f73cd65adc8cdfcb92d
         with:
           sdk: "3.11.0"
       - id: checkout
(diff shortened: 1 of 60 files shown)


natebosch:
I took some time this afternoon to put some thought into this and the `expectAsync` APIs. I think I'm leaning towards the idea that these things shouldn't be coupled to the testing APIs directly, we should consider separating them out into a different package. I've had gemini write a rough draft at #2755 - I think we can have some discussion about the API and decide whether or not to work towards cleaning that up or landing this instead.

natebosch (changes requested):
I think if we put one extension on `Subject<void Function()>` and another other `Subject<Future<void> Function>` then we can call them both `prints`. The one use case that design disallows is explicitly calling `printAsync` for a function reference that is statically typed as synchronous but is async in practice. I don't think that's a problem. Maybe the synchronous implementation can `assert` that it doesn't get a `Future` return value.

kevmoo:
@natebosch Done!

- Replaced `prints` / `printsAsync` on `FunctionChecks<T>` with overloaded `.prints()` extensions on `Subject<void Function()>` (`FunctionPrints`) and `Subject<Future<void> Function()>` (`AsyncFunctionPrints`).
- Added an `assert(result is! Future, ...)` in `FunctionPrints.prints` if a function statically typed as `void Function()` returns a `Future` at runtime.
- Merged `master` and added runnable `{@example}` files (`example/function/function/prints.dart` and `example/function/async_function/prints.dart`).

PTAL!

natebosch (approved):

PR #2752, opened 2026-09-05
Make zero-arg expectations getters

### LINKED ISSUE #1943: Change expectations which take no arguments to be getters - `check(foo).isTrue` instead of `check(foo).isTrue()`
I accidentally changed:
```dart
expect(foo, isNotNull);
```

to:
```dart
check(foo).isNotNull;
```

(instead of `check(foo).isNotNull();`)

This compiles and gives no warning, but of course does not actually execute the test. 

Is there a way we can detect this kind of user failure?

natebosch:
Closes #1943

### DIFF
--- a/pkgs/checks/lib/src/extensions/math.dart
+++ b/pkgs/checks/lib/src/extensions/math.dart
@@ -6,7 +6,7 @@ import '../../context.dart';
 
 extension NumChecks on Subject<num> {
   /// Expects that [num.isNaN] is true.
-  void isNaN() {
+  void get isNaN {
     context.expect(
       () => ['is not a number (NaN)'],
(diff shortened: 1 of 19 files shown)


kevmoo (approved):
YES!

pkgs/checks/lib/src/extensions/math.dart
lrhn, line 43: (Inconsisitent iwth `isNegative` which uses a conditional expression.)

lrhn (approved):

pkgs/checks/lib/src/extensions/math.dart
natebosch: Refactored for consistency.

Issue #2671, opened 2026-06-16
Find an API for making expectations that callbacks are executed


natebosch:
`packag:test` has [`expectAsync`](https://pub.dev/documentation/test/latest/test/expectAsync0.html) methods which wrap a callback and hold the test case open until it is executed. As is, these APIs suffer from the `package:matcher` problem of being top-level so they crowd the autocomplete initially and are not helped by static type information.

Should we keep this in `package:test` and continue to use it with `package:checks`, or move it to `package:master` and replace it with something that fits the API better?

What should the API be? `check(() {/*...*/).isCalled()`? `(() {/*...*/}).checkCalled()`?

FMorschel:
Why not `results` and `asyncResults`? Similar to the matchers?

Issue #994, opened 2019-03-02
`prints` should accept an optional extra param "then" or similar to validate the result of the provided function


kevmoo:

grouma:
I'm not sure I follow this request. Can you elaborate with a brief example?

kevmoo:
```dart
expect(() => process(1,2,3), prints('hello'));
```

Fine, but what if I want to check a thrown exception or the result?

```dart
expect(() => process(1,2,3), prints('wth, yo?'), then: throwsArgumentError);
```

lrhn:
This is a reasonable request since `prints` is not checking the result, but checking side-effects created during the evaluation. You can have many side effects, and still have one result afterward. Other matchers check the result (value or thrown) and there can only be one of those.

Maybe there could be a `SideEffectMatcher` with an optional `Matcher then(Matcher)` method and perhaps an `SideEffectMatcher and(SideEffectMatcher)` for chaining multiple side effects. Example:

```dart
abstract class SideEffectMatcher implements Matcher {
  SideEffectMatcher and(SideEffectMatcher nextSideEffect);
  Matcher then(dynamic result);
}
```
Then `prints` would be a `SideEffectMatcher`.

kevmoo:
@lrhn – SGTM!

natebosch:
Does this work today with `allOf`?

```dart
expect(() => process(1,2,3), allOf(prints('hello'), throwsArgumentError));
```

Would the side effects need to be ordered?

kevmoo:
Tried that. It doesn't work.

natebosch:
We could consider trying to make the `allof` version work, though taking a look at the implementation that most likely means that the function would get run twice - once to check the `prints` and once to check the `throwsArgumentError`.

Making multiple matchers cooperate and only run a `Function()` once does not look immediately feasible to me.

If it's acceptable to let the function run once per matcher I can run with that. Otherwise we'll likely need a separate approach and provide a different utility.

```dart
var printCapture = PrintCapture();
expect(printCapture.wrap(() => process(1,2,3)), throwsArgumentError);
expect(printCapture.captured, contains('hello'));
```

kevmoo:
That's not a bad idea, Nate.

PR #2755, opened 2026-09-11
Add package:spectate with PrintSpy, CallCounter, and ExitSpy

natebosch:
Towards #994, #2671, #2742

The `prints` Matcher is not quite orthogonal to other expectations that
might be checked on a callback (#994).
The `expectAsync` pattern doesn't fit cleanly into the `package:checks`
patterns.

Solve both of these problems by implementing a separate, non-test
specific package which can "specate" on executing Dart code and record
details about the interactions. The records can then be checked with
typical expectation extensions.

TAG=agy
CONV=3f8ca85a-8236-4bd5-b7a1-26d64a8fb70a

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.7.2
+# Created with package:mono_repo v6.7.3
 name: Dart CI
 on:
   push:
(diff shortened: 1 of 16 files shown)


kevmoo:
I like the idea of "Spy" - we can use 🥷 emjoi or something!

PR #2756, opened 2026-09-17
Add an example for every checks expectation, and fix the bugs they exposed

natebosch:
Adds a runnable example under `pkgs/checks/example/` for each of the 79
public members of the expectation extensions in
`pkgs/checks/lib/src/extensions/`, and links each one from its member's
dartdoc with `{@example /example/<source>/<subject>/<member>.dart}`.

Every example is a `package:test` test file with at least one check that
passes and at least one that fails, and the failure that a failing check
produces is recorded verbatim in a comment immediately following it. Members
with a single failing case use one test for both the passing and the failing
check; members with more than one failing case use a test per case, including
for the passing checks. The examples are therefore expected to fail when run
directly, which is how the failure messages stay visible. The two deprecated
members, `IterableChecks.containsInOrder` and
`IterableChecks.pairwiseComparesTo`, are skipped.

Examples are written as tests rather than plain scripts because asynchronous
expectations call `TestHandle.current` and throw `OutsideTestException`
outside a running test. Examples that would otherwise report a stack trace
construct their errors with `Future.error`, `Stream.error`, or
`Error.throwWithStackTrace(..., StackTrace.empty)` so that the recorded output
does not contain machine specific paths.

`test/example_test.dart` checks that the failures each example produces equal
the output recorded in its comments, so the documented failure messages cannot
drift from the real ones. All of the examples are run together in a single
`dart test` invocation with the JSON reporter, which associates each failure
with the suite it came from. Running them one invocation apiece instead costs
about 2.5s each, since every invocation pays the runner startup and spawns a
new `frontend_server` process initialized from
`.dart_tool/test/incremental_kernel.*`; batching them lets the runner compile
all suites incrementally against one live `frontend_server` instance and run
them concurrently, which takes the suite from 3m25s to 9s.

Recording the real output instead of the plausible output exposed five defects
in the failure messages. Each is fixed here, along with its test expectations
and the recorded output of the affected examples.

`MapChecks.isNotEmpty` described the wrong condition. Rejecting an empty map
reported `Which: is not empty`, which repeats the expectation's own wording
instead of saying what was wrong with the value. `IterableChecks.isNotEmpty`
and `StringChecks.isNotEmpty` both correctly say `is empty`. Two existing
tests asserted the incorrect string rather than catching it.

`not` emitted a trailing space. The clause was built as `'is a value that: '`,
while the matching clause for the expectation is `'is not a value that:'` with
no trailing space. Besides being inconsistent, trailing whitespace is stripped
by `dart format`, so the message could not be recorded in an example comment.

`pairwiseMatches` did not indent its nested detail. The `Actual element at
index N:` and `Which:` lines were emitted at the same level as the `does not
have an element at index N that:` line that introduces them, so they read as
unrelated top level clauses instead of as the detail of the element that
failed. `every` already indents the equivalent lines.

`equalsIgnoringWhitespace` reported offsets that did not line up with the
value it displayed. The difference is computed between the two strings after
their whitespace is collapsed, so the reported offset and the caret are
positions in the collapsed value, but the message showed the original value on
the `Actual:` line. For `check('hello   wide world')` against `'hello world'`
the message pointed at offset 7, which is a different character in the
original than in the collapsed string. The difference is now reported beneath
a line naming the collapsed value it was computed from, so the offset and
caret can be matched up with the string they refer to.

`isA`, `IterableChecks.any`, `MapChecks.containsKeyThat`, and
`MapChecks.containsValueThat` began their `Which:` clause with a capital
letter, unlike every other rejection in the package. They are now lower case.

The `a int` versus `an int` article bug in the subject label is left alone; it
is already tracked by a TODO in `lib/src/checks.dart`.

TAG=agy
CONV=adddbf0c-f10a-4c66-914a-9f0caa5eeafe


pkgs/checks/CHANGELOG.md
```diff
@@ -39,6 +39,21 @@
...
+  about the element that failed, so that it is visibly part of the clause it
+  describes.
+- Fix the failure message for `equalsIgnoringWhitespace` to report the value
+  with its whitespace collapsed. The reported offset and difference always
+  referred to the collapsed value, but the message showed the original, so the
+  two did not line up.
+- Use a consistent lower case first word for the `Which:` clause of
+  `isA`, `IterableChecks.any`, `MapChecks.containsKeyThat`, and
+  `MapChecks.containsValueThat`.
+- Remove a stray trailing space from the failure message for `not`.
```
jonasfj, line 56: Wow, this is a really good set of bug fixes just from writing examples!

pkgs/checks/example/string/string/equals_ignoring_whitespace.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:checks/checks.dart';
+import 'package:test/scaffolding.dart';
+
+void main() {
+  test('equalsIgnoringWhitespace', () {
+    // This check succeeds.
+    check('  hello   world  ').equalsIgnoringWhitespace('hello world');
+
```
jonasfj, line 9:
```suggestion

  // This check succeeds
  check('''
    newlines are whitespace
    tabs\ttoo!
  ''').equalsIgnoringWhitespace('newlines are whitespace tabs too!');
```

jonasfj (approved):
I skimmed a lot of these examples.

But with the testing in place, it's not even "load bearing" whether or not the examples are perfectly reasonable.
Because we know all the examples will run as `Expected:` 🚀

kevmoo (commented):
Nice cleanup and test setup (`test/example_test.dart`). Left two quick notes on stale inline code snippets directly above the new `{@example}` directives and one minor typo.

pkgs/checks/analysis_options.yaml
```diff
@@ -0,0 +1,6 @@
+include: ../../analysis_options.yaml
+
+analyzer:
+  errors:
+    # Not all SDK versions used on CI spport @example
```
kevmoo:
```suggestion
    # Not all SDK versions used on CI support @example
```

pkgs/checks/lib/src/extensions/async.dart
```diff
@@ -292,6 +304,8 @@ extension StreamChecks<T> on Subject<StreamQueue<T>> {
   /// state.
   /// If this expectation succeeds, consumes as many events from the source
   /// stream as are consumed by all the conditions.
+  ///
+  /// {@example /example/async/stream_queue/in_order.dart}
```
kevmoo, line 301: The inline code block a few lines above (lines 296–301) still uses the old closure-based `Condition` syntax (`(s) => s.emits((e) => e.equals(0))`), which no longer compiles. Might be worth either updating those lines to `.it()..emits(.it()..equals(0))` (so IDE hover, which strips `{@example}`, shows valid code) or removing the inline snippet in favor of `{@example}`.

pkgs/checks/lib/src/extensions/iterable.dart
```diff
@@ -200,6 +228,8 @@ extension IterableChecks<T> on Subject<Iterable<T>> {
   /// ```
   ///
   /// Values, will be compared with the equality operator.
+  ///
+  /// {@example /example/iterable/iterable/contains_equal_in_order.dart}
```
kevmoo, line 226: Similarly, the inline code block right above this (line 225) calls the deprecated `.containsInOrder([1, 2, 3])` instead of `.containsEqualInOrder([1, 2, 3])`.

pkgs/checks/example/string/string/equals_ignoring_whitespace.dart
natebosch: Added newline and tab characters in the passing example

pkgs/checks/lib/src/extensions/async.dart
natebosch: Dropped extra inline code sample from the doc, the example should be sufficient.

pkgs/checks/lib/src/extensions/iterable.dart
natebosch: Dropped extra inline code block.

kevmoo (approved):
amazing!

PR #2759, opened 2026-09-18
Use consistent *Checks naming for prints

natebosch:
Rename to `AsyncFunctionChecks` and `VoidFunctionChecks` with a comment
explaining that it exists for extension precedence.


natebosch:
I didn't notice this until just after I merged.

natebosch:
Oh, it can't be merged for the precedence to work... I think that makes sense.

kevmoo (approved):
thanks! (although check CI sadness)

PR #2760, opened 2026-09-18
Remove mentions of Travis

natebosch:
Clean up stale comment references to travis.

Remove some skips that date back to when we were using travis.


kevmoo (approved):
better late than never. Checkout out CI failures, of course.

PR #2761, opened 2026-09-18
Fix race conditions during test load

natebosch:
`Engine._addLoadSuite` awaits `LoadSuite.suite`, whose completer is only
ever completed from inside the load test's body: either when the body
finishes loading the suite, or by the `onComplete` handler the body
registers. If the load test is closed before its body runs at all,
neither happens and nothing ever completes the suite.

That stalls the task `Engine.run` added for the suite, so it never
releases its resource from `_runPool`, so `_runPool.close()` never
completes, so `Engine.close()` and in turn `Runner.close()` hang
forever.

Today this is masked in the runner: the signal handler cancels its
signal subscription immediately, so the VM exits once nothing else
holds the event loop even though `close()` never completed. Anything
that keeps the isolate alive past that point turns it into a real hang.

Hold onto the completer in `LoadSuite` and add `completeWithoutSuite`,
which completes it with a null suite if nothing else has. Call it once
the load test has finished running, from both `Engine._addLoadSuite`
and `LoadSuite.getSuite`. It is a no-op in every normal path, since the
body holds an outstanding callback until it completes the completer.

Skip running the load test when the engine is already closed. Nothing
needs to be loaded for a closed engine, and `completeWithoutSuite` still
unblocks anything waiting on the suite.


pkgs/test_core/lib/src/runner/engine.dart
```diff
@@ -529,7 +529,18 @@ class Engine {
 
     // Schedule a microtask to ensure that [onTestStarted] fires before the
     // first [LiveTest.onStateChange] event.
-    await Future.microtask(liveTest.run);
+    await null;
```
brianquinlan, line 532: `await null` schedules a microtask?

brianquinlan (approved):

pkgs/test_core/lib/src/runner/engine.dart
natebosch: Yes, any `await` waits at least a microtask. It's not a super widespread pattern because it's generally a code smell but IMO `await null;` is less ugly than `await Future.microtask(() {});` with an empty function literal which is the more explicit way to express this.

The old pattern was handy when we could use a function tear-off but it's not a simple tear-off after this change.

PR #2765, opened 2026-09-21
Refactor signal handling in executable.dart

### LINKED ISSUE #1615: Runner test for `runTest` fails with unexpected output on windows

natebosch:
Keep the signal subscription active during graceful shutdown so a second
signal can be handled explicitly in Dart before exiting with the signal
exit code, and cancel the subscription in the finally block once the
runner closes. Add logic to determine the exit code that would have been
used by the Dart VM when it was handling the second signal.
This prepares for an upcoming change which improves temporary directory
cleanup during shutdown.

The `runTests` and `completeShutdown` methods are unused externally, so
remove and inline them. Move the `signalSubscription` global variable to
a local variable in the main method. Drop the global `isShutdown`
variable. Reset `_stdinLines` in `cancelStdinLines` so it is safe to
call more than once.

Closes #1615


natebosch:
@brianquinlan friendly ping - this is a precursor to the other PR you reviewed #2763

brianquinlan (approved):
I should have looked at this in order ;-)

